PR #1, opened 2016-01-07
Show how to dispatch non-zero denominator conditions left by field in ZGaloisField

daniel-ziegler:
Don't merge this PR -- just do the necessary changes and then merge that


andres-erbsen:
This has been incorporated.

PR #3, opened 2016-06-22
Make fiat-crypto compile with both 8.4 and 8.5

JasonGross:
I submit this as a pull request, because we seem to have been avoiding `Import`s in this file up until this point, and I'm not sure what policy we want to take here.


JasonGross:
Huh, apparently I did this wrong.  (Hooray for continuous integration.)  I'll take a look at what broke in the 8.4 build tomorrow.  In the meantime, I'd still appreciate feedback on how we want to handle such name-changes.

JasonGross:
After chatting with Andres about how to do this, and fixing the CI for 8.4 and 8.5, I'm merging this.

PR #4, opened 2016-06-22
Disable Coq 8.4

JasonGross:
This rips out the support for building with Coq 8.4.  (When) should we merge this?


andres-erbsen:
I think a good reason for Coq 8.4 would be if we hit a bug that we don't know how to work around that is unambiguously fixed in 8.5. So far, most builds that fail on only some coq versions have been due to the code relying on unspecified behavior (that just happens to be nice to use on some other version), seconded by standard library changes which we already know how to handle. There are only so many Coq versions that we could drop support for to fix our builds. Responding to a version-correlated build failure by taking a step back and thinking about whether Coq should actually behave the way we just expected it to seems like a much better idea.

I'm closing this pull request now, but it may be worth keeping the branch if it is any good. I do hope this project will outlast Coq 8.4.

PR #6, opened 2016-06-23
Define the spec of Weierstrass curves

JasonGross:
This is the start of work on P256.

N.B. I've just copy/pasted the equations, and haven't actually yet
checked that addition is associative, that it forms a group, etc.

Questions that should be resolved before this is merged:

1. [x] Is there a better place for the various tactics I added than
   `src/CompleteWeierstrassCurve/Pre.v` (are we trying to minimize
   dependencies? if not, we can define `Util/Tactics.v` and put them in
   there...)
2
2. [ ] What assumptions do we need to make about `a`, `b`, characteristic,
   etc, for these curves.
3. [x] Should this be "Weierstrass", "ShortWeierstrass", or something else?
4. [x] Does my `NOT_ON_CURVE` kludge need more commenting?
5. [ ] Is my notation involving `∞` reasonable?  (Should it be exported
   somewhere?)
6. [x] Is the `match` semantics for `onCurve` readable for non-Coq-experts,
   or should we use a different syntax.
7. [x] Is the boilerplate around `add` okay, or should I try to get `Program
   `Definition` working?


src/Spec/CompleteWeierstrassCurve.v
```diff
@@ -0,0 +1,81 @@
...
+                              end }.
+    Definition coordinates (P:point) : (F*F + ∞) := proj1_sig P.
+
+    (** The following points are indeed on the curve -- see [CompleteWeierstrassCurve.Pre] for proof *)
+    Local Ltac t := intros; apply (Pre.unifiedAdd'_onCurve _ _ (proj2_sig _) (proj2_sig _)).
+
+    Program Definition zero : point := ∞.
+
+    (* If we use [Program Definition], Coq inserts too many
+       intermediate casts, and does too much destruction. *)
```
andres-erbsen: `Program Definition .. := exist _ (code) _.` has worked for me and gives nicer separation of code and proof.

src/Spec/CompleteWeierstrassCurve.v
```diff
@@ -0,0 +1,81 @@
...
+    Local Notation "'∞'" := unit : type_scope.
+    Local Notation "'∞'" := (inr tt) : core_scope.
+    Notation "2" := (1+1). Notation "3" := (1+1+1).
+    Local Notation "( x , y )" := (inl (pair x y)).
+    Local Open Scope core_scope.
+
+    Context {a b: F}.
+    Class weierstrass_params :=
+      {
+        (** TODO FIXME What do we need to say about a, b, characteristic, etc? *)
```
andres-erbsen: Here goes whatever you need to prove the group law.

andres-erbsen:
1. Only `Spec/` should avoid Import of non-spec files, Pre can import whatever it wants.
2. You probably want nonzero discriminant, maybe nonzero a b, probably 1+1<>0 and 1+1+1<>0. Associativity proof will hopefully show you goals for this.
3. Current name is fine.
4. I think it's fine, but could you just leave an underscore and prove that case inpossible in the goal that is left?
5. hmm.
6. It's fine.
7. It is acceptable, but Program Definition has worked fine for me when I tell it where my holes are.

src/Spec/CompleteWeierstrassCurve.v
JasonGross: Have you ever done this with `code` that involves explicit `match`es or `if` statements?  `Program` is breaking apart my code into multiple obligations, so I'd need to write tactic code that split apart the various cases of `Pre.unifiedAdd'_onCurve`.  Is `Local Obligation Tactic := ....  Program Definition .. := exist _ (code) _.` really all that better than `Local Ltac t := .... Definition ... refine (exist _ (code) _); t. Defined.`?

JasonGross:
4. Oh, interesting.  But then I need to pass around proofs that the results of destructing the points are actually on the curve (because it's only impossible for points on the curve), via the convoy pattern.  I think `Program` can do that automatically, but do we really want to be hiding that much smarts in the notation?  And do we really want the first projection to depend on the proofs?

andres-erbsen:
1. I'm not sure I understand all the constraints here. What makes passing proofs for Weierstrass curves harder than for Edwards curves?

JasonGross:
(Why does github replace `4.` with `1.`?)

Because my code has explicit `match` statements which are interpreted by the `Program` machinery, which the `Program` infrastructure automatically convoys.  Here's the obligation that I'm left with when I disable the obligation tactic:
```coq
   forall P1 P2 : point,
   match
     match
       coordinates P1 as anonymous'0
       return (anonymous'0 = coordinates P1 -> coordinates P2 = coordinates P2 -> F * F + ∞)
     with
     | inl p =>
         let
           (x1, y1) as p0
            return (inl p0 = coordinates P1 -> coordinates P2 = coordinates P2 -> F * F + ∞) := p in
         match
           coordinates P2 as anonymous'
           return ((x1, y1) = coordinates P1 -> anonymous' = coordinates P2 -> F * F + ∞)
         with
         | inl p0 =>
             let
               (x2, y2) as p1
                return ((x1, y1) = coordinates P1 -> inl p1 = coordinates P2 -> F * F + ∞) := p0 in
             fun (_ : (x1, y1) = coordinates P1) (_ : (x2, y2) = coordinates P2) =>
             if x1 =? x2
             then
              if y2 =? - y1
              then ∞
              else
               if y2 =? y1
               then
                ((3 * x1 ^ 2 + a) ^ 2 / (2 * y1) ^ 2 - x1 - x1,
                (2 * x1 + x1) * (3 * x1 ^ 2 + a) / (2 * y1) - (3 * x1 ^ 2 + a) ^ 3 / (2 * y1) ^ 3 -
                y1)
               else Pre.NOT_ON_CURVE
             else
              ((y2 - y1) ^ 2 / (x2 - x1) ^ 2 - x1 - x2,
              (2 * x1 + x2) * (y2 - y1) / (x2 - x1) - (y2 - y1) ^ 3 / (x2 - x1) ^ 3 - y1)
         | inr u =>
             match
               u as u0 return ((x1, y1) = coordinates P1 -> inr u0 = coordinates P2 -> F * F + ∞)
             with
             | tt => fun (_ : (x1, y1) = coordinates P1) (_ : ∞ = coordinates P2) => coordinates P1
             end
         end
     | inr u =>
         match
           u as u0 return (inr u0 = coordinates P1 -> coordinates P2 = coordinates P2 -> F * F + ∞)
         with
         | tt =>
             match
               coordinates P2 as anonymous'
               return (∞ = coordinates P1 -> anonymous' = coordinates P2 -> F * F + ∞)
             with
             | inl p => fun (_ : ∞ = coordinates P1) (_ : inl p = coordinates P2) => coordinates P2
             | inr u0 =>
                 match
                   u0 as u1 return (∞ = coordinates P1 -> inr u1 = coordinates P2 -> F * F + ∞)
                 with
                 | tt => fun (_ : ∞ = coordinates P1) (_ : ∞ = coordinates P2) => ∞
                 end
             end
         end
     end eq_refl eq_refl
   with
   | (x, y) => y ^ 2 = x ^ 3 + a * x + b
   | ∞%core => True
   end
```

andres-erbsen:
(The numerals are interpreted by markdown list canonicalizer).

Okay, I see. I'm trying to make it look reasonable and I have moderate hope that I will succeed.

JasonGross:
> I'm trying to make it look reasonable and I have moderate hope that I will succeed

You mean, you're trying a few things, and will get back to me with the results of your experiments?

andres-erbsen:
Yep, I tried a few things. I didn't manage to make anything nicer than what you did using `refine`. Is there any reason why `Program Definition` would want to rewrite all the matches even though there are no holes inside them?

My favorite indentation:

    match coordinates P1, coordinates P2 with
      | (x1, y1), (x2, y2) =>
        if x1 !=? x2 then            ((y2-y1)^2 / (x2-x1)^2 - x1 - x2,
                                       (2*x1+x2)*(y2-y1) / (x2-x1) - (y2-y1)^3 / (x2-x1)^3 - y1)
        else if y2 !=? -y1 then      ((3*x1^2+a)^2 / (2*y1)^2 - x1 - x1,
                                       (2*x1+x1)*(3*x1^2+a) / (2*y1) - (3*x1^2+a)^3/(2*y1)^3-y1)
             else                    ∞
      | ∞, ∞ =>                      ∞
      | ∞, _ =>                      coordinates P2
      | _, ∞ =>                      coordinates P1

JasonGross:
@andres-erbsen Using the negation of `eq_dec` is more annoying; I'd rather flip the cases around so I don't have to deal with it, unless you think it's important to order them that way.

JasonGross:
Updated to remove the `NOT_ON_CURVE` case, and to split off the tactic code elsewhere.

mattam82:
By default match with uses the convoy pattern but you can explicitly add match return _ with so that compilation of match goes back to the standard pattern-matching compilation algorithm and inference of the return clause. There'll be an option Program Cases in 8.6 to enable/disable this more globally.

andres-erbsen:
Thanks! Adding a `return _` seems to make the Weierstrass addition law definition work out as expected.

src/Spec/CompleteWeierstrassCurve.v
```diff
@@ -0,0 +1,78 @@
+Require Crypto.CompleteWeierstrassCurve.Pre.
+
+Module E.
+  Section WeierstrassCurves.
+    (* Short Weierstrass curves with complete addition laws. References:
```
agl: The phrase &ldquo;complete addition laws&rdquo; usually means formulae that are correct for all inputs (i.e. P+P, P+&infin;, &hellip;). I guess in the Coq sense, the formula below is complete because you handle those cases with a match, but I believe that mathematicians would consider them to be _incomplete_ formulae.

You find complete a complete (in the traditional sense) formula for addition in a [recent Microsoft paper](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/06/complete-2.pdf).

Whether _using_ the complete formula is a good idea I'm much less sure. It's defined for projective points, not affine, so it may be harder to deal at this level. At some point you'll have to switch to using projective or Jacobian representations so perhaps you planned on tying this definition to that one?

So the practical upshot may simply involve dropping the word &ldquo;complete&rdquo; here.

src/Spec/CompleteWeierstrassCurve.v
```diff
@@ -0,0 +1,78 @@
...
+    Notation "2" := (1+1). Notation "3" := (1+2).
+    Local Notation "( x , y )" := (inl (pair x y)).
+    Local Open Scope core_scope.
+
+    Context {a b: F}.
+    Class weierstrass_params :=
+      {
+        char_gt_2 : 2 <> 0;
+        char_ne_3 : 3 <> 0
+        (** TODO FIXME What do we need to say about a, b, characteristic, etc? *)
```
agl: You've already defined the field characteristic to be &ge; 3. You may want to define a = -3 if you're planning on using the formulae that depend on that. (All the NIST curves have a = -3.) If you want to prove that the set of `point`s forms a group using `add` then you'll also need to limit `a` and `b` such that the discriminant (which is -16(4a&#179; + 27b&#178;)) is non-zero I believe.

JasonGross:
@agl Thanks!  I've dropped "complete" and added the discriminant requirement.  (I'll wait to see where I need `a <> 0` or `a = -3`, and whether I need `b <> 0`)

@andres-erbsen Thoughts on `Program` with `return _` vs `refine`?

andres-erbsen:
I woul prefer `Program Definition` with `return _` because it automatically makes the proof part abstract and gives it a name like `add_obligation`. It also makes definitions involving sigma types look less different from definitions for simple types, which I think is a good thing.

I also have a slight preference for not requiring `a = -3` in specification and general proofs. We will want to use it for the optimized addition formulas (similarly to `ExtendedCoordinates.v` assuming `a=-1`), but it will not be necessary elsewhere. Flexibility to pick `a` allows translating to Weierstrass from other forms of ellitpic curves (as in Elligator).

JasonGross:
Okay, I'll look at making `Program Definition` work nicely.

@andres-erbsen @agl, proving the correctness of Algorithm 4 given in https://eprint.iacr.org/2015/1060.pdf seems to require that the order of the group of points on the curve is odd.  Is there a nice way of formulating this requirement in terms of `a` and `b`?

JasonGross:
@andres-erbsen I've updated it to use `Program Definition`

@andres-erbsen @agl I've coded up one way of forcing an odd number of points onto the curve.  Feedback welcome.

src/Spec/CompleteWeierstrassCurve.v
```diff
@@ -0,0 +1,89 @@
...
+    Local Infix "-" := Fsub. Local Infix "/" := Fdiv.
+    Local Notation "- x" := (Fopp x).
+    Local Notation "x ^ 2" := (x*x) (at level 30). Local Notation "x ^ 3" := (x*x^2) (at level 30).
+    Local Notation "'∞'" := unit : type_scope.
+    Local Notation "'∞'" := (inr tt) : core_scope.
+    Local Notation "0" := Fzero.  Local Notation "1" := Fone.
+    Local Notation "2" := (1+1). Local Notation "3" := (1+2). Local Notation "4" := (1+3).
+    Local Notation "8" := (1+(1+(1+(1+4)))). Local Notation "12" := (1+(1+(1+(1+8)))).
+    Local Notation "16" := (1+(1+(1+(1+12)))). Local Notation "20" := (1+(1+(1+(1+16)))).
+    Local Notation "24" := (1+(1+(1+(1+20)))). Local Notation "27" := (1+(1+(1+24))).
```
andres-erbsen: Defining constants as notations can get hairy. An alternative would be `Fixpoint coef (n:nat) := match n with O => Fzero | S n' => Fone + coef n' end.` Pick whichever you think is nicer overall.

andres-erbsen:
re: https://github.com/mit-plv/fiat-crypto/pull/6/files/625a1a9abaa579dc4efc03d62e403d7407440dd6#r68506335

I think it make sense to drop the word "complete" from the file and section names as well, except for the file that implements the complete addition law. The Edwards curve spec will probably keep "complete" in its name since even the spec proofs depend on the addition law being complete.

src/Spec/CompleteWeierstrassCurve.v
JasonGross: I think this way will give us fewer issues with simplification of constants, later.  When we move to, uh, 8.7, probably, we can make use of the ability to declare numerical notations in user-space (https://github.com/coq/coq/pull/156).  Until then, for this limited class of constants, I'd rather not have to deal with the headache of a function in the spec that morally shouldn't be present.  (An alternative would be, using your `coef`, `Let twenty_seven := Eval hnf in coef 27.  Notation "27" := twenty_seven.`  Do you think this would be nicer?)

JasonGross:
@andres-erbsen To prove the complete addition formula, we require `b <> 0`.  I'm not sure if this is required to prove associativity in the incomplete case.  Should I leave this out of the spec until I finish the associativity proof and see what it needs?

andres-erbsen:
https://hyperelliptic.org/EFD/g1p/auto-shortw-w12-0.html describes a family of Weierstrass curves with `b = 0`, so I think `b <> 0` is a requirement of the complete addition formulas, not of Weierstrass curves in general. As long as the proofs go through, there is no reason to require it.

src/Spec/WeierstrassCurve.v
```diff
@@ -0,0 +1,83 @@
...
+
+    (** The following points are indeed on the curve -- see [WeierstrassCurve.Pre] for proof *)
+    Local Obligation Tactic :=
+      try solve [ Program.Tactics.program_simpl
+                | intros; apply (Pre.unifiedAdd'_onCurve _ _ (proj2_sig _) (proj2_sig _)) ].
+
+    Program Definition zero : point := ∞.
+
+    (* If we use [Program Definition], Coq inserts too many
+       intermediate casts, and does too much destruction. *)
```
andres-erbsen: This comment is out of date.

src/Spec/CompleteWeierstrassCurve.v
andres-erbsen: Okay, nevermind then. I'm looking forward to Numeral Notation (and hexadecimal and binary variants of that).

JasonGross: "You mean you're not happy with `Ox~0~1~1~0` (or variants where `~` is replaced by zero-with joiners or some similar nonsense)?"

src/Spec/WeierstrassCurve.v
JasonGross: Thanks! Removed.  (I can add back one that talks about `return _` if you think it's useful.)

JasonGross:
Should we wait until we have an actual proof of associativity to merge this, or should we merge this soon and add conditions as necessary to make the proofs go thorough?

andres-erbsen:
I think we should merge this soon. It might be a good idea to add a note to spec, stating that more conditions may be added when group properties are verified. LGTM to merge.

PR #7, opened 2016-06-23
Merge Assembly Machinery into Master

varomodt:
https://github.mit.edu/plv/fiat-crypto/pull/7

@achlipala @JasonGross


_CoqProject
```diff
@@ -58,7 +58,6 @@ src/Assembly/Qhasm.v
 src/Assembly/PseudoConversion.v
 src/Assembly/AlmostConversion.v
 src/Assembly/StringConversion.v
-src/Assembly/Pipeline.v
```
JasonGross: Please don't manually modify `_CoqProject`.  It breaks the `make update-_CoqProject` target.  If you want to do something like this, instead stick your `.v` file in a separate directory (not under `src`), or just comment out all of the non-building code.

varomodt: Sorry I didn't see this before I merged: I'll make a PR to add Pipeline.v presently.

Issue #15, opened 2016-06-28
`super_nsatz` clears `H: 1 + 1 <> 0`


andres-erbsen:
In response to #10, I was trying to port `CompleteEdwardsCurve` proofs from `field_algebra` to `super_nsatz`. However, it seems that `super_nsatz` clears `H: 1 + 1 <> 0`, making a provable goal unprovable.

https://github.com/mit-plv/fiat-crypto/commit/5f522eedfece8db63ddd95360b58e14fefcdad9a

JasonGross:
Do you get it back if you run `split_field_inequalities`?

JasonGross:
> I should probably add that to the end of

On the other hand, I'm not sure it's a good idea to put much work into the kind of goals that `super_nsatz` leaves over; the intended use is "pose all the facts you need, then finish with `super_nsatz`".  If this is the spec, and left-over goals are just meant for debugging purposes, then I'm not sure whether or not it makes sense to clean up.  What do you think?

andres-erbsen:
I would definitely appreciate if `super_nsatz` won that goal right away -- I posed all facts that are necessary before calling `super_nsatz`. In particular, asserting the goal that `super_nsatz` leaves over under the original hypotheses, `assumptions` solves the goal.

andres-erbsen:
Yes, `split_field_inequalities` gives the missing assumption back.

JasonGross:
Try `repeat super_nsatz` ;)

I wrote the tactic under the assumption that `nsatz` would not leave over side-conditions.  I'll go fix this now.

JasonGross:
Try it now with 91a712295078756f162113d30412c666679c90ac ?

JasonGross:
Er, I mean the one after that commit; that one doesn't actually compile.

andres-erbsen:
Now I get the following subgoal left over:

    F : Type
    eq : F -> F -> Prop
    zero : F
    one : F
    opp : F -> F
    add : F -> F -> F
    sub : F -> F -> F
    mul : F -> F -> F
    inv : F -> F
    div : F -> F -> F
    H : field
    a : F
    sqrt_a : F
    a_nonzero : sqrt_a <> 0 /\ sqrt_a <> 0
    a_square : exists sqrt_a0 : F, sqrt_a0^2 = a
    d : F
    d_nonsquare : forall sqrt_d : F, sqrt_d^2 <> d
    x1 : F
    y1 : F
    x2 : F
    y2 : F
[...]

PR #16, opened 2016-06-29
wrap nsatz with dropRingSyntax; port CompleteEdwardsCurveTheorems to super_nsatz

andres-erbsen:
See #15.


src/CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v
```diff
@@ -96,46 +38,83 @@ Module E.
...
+      bash_step.
+      bash_step.
+      bash_step.
+      bash_step.
+      bash_step.
+      bash_step.
+      bash_step.
+      bash_step.
+      bash_step.
+      { conservative_common_denominator.
```
andres-erbsen: @JasonGross : Is `conservative_common_denominator` intended to be idempotent?

src/CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v
```diff
@@ -155,12 +134,15 @@ Module E.
...
-        destruct (eq_dec y 0); [apply nonzero_a|apply nonsquare_d with (sqrt_a/y)]; field_algebra.
+        destruct (eq_dec y 0); [apply nonzero_a|apply nonsquare_d with (sqrt_a/y)]; super_nsatz.
       Qed.
 
       Lemma solve_correct : forall x y, onCurve (x, y) <-> (x^2 = solve_for_x2 y).
       Proof.
-        unfold solve_for_x2; simpl; split; intros; field_algebra; auto using a_d_y2_nonzero.
+        unfold solve_for_x2; simpl; split; intros.
+        { conservative_common_denominator. nsatz. auto using a_d_y2_nonzero. }
+        { conservative_common_denominator_all. (* FIXME: doesn't do anything *)
```
andres-erbsen: @JasonGross : should `conservative_common_denominator_all` work here?

src/CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v
JasonGross: Yes.  It should leave the primary goal with no divisions in it, and should fail when there is no division.  Does it currently do wacky things with the side-conditions?  If so, we should guard it with a check that the goal is a field equality.

JasonGross: Oh, oops, it doesn't work with fractions in denominators yet.  I'll fix that.

src/CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v
JasonGross: I made the arbitrary decision to make it only handle hypotheses that it could solve.  I'll go fix it to make it leave over side conditions instead.

src/CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v
JasonGross: Should be fixed by b707a30 (haven't tested it on this example, though)

src/CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v
JasonGross: Should be fixed by ae61d52 (again, not tested on this example)

src/CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v
```diff
@@ -97,14 +97,17 @@ Module E.
       bash_step.
       bash_step.
       { conservative_common_denominator.
-        conservative_common_denominator.
         nsatz.
         admit.
```
andres-erbsen: The new `conservative_common_denominator` generates twice as many subgoals as running the old version twice.

src/CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v
```diff
@@ -139,10 +142,8 @@ Module E.
 
       Lemma solve_correct : forall x y, onCurve (x, y) <-> (x^2 = solve_for_x2 y).
       Proof.
-        unfold solve_for_x2; simpl; split; intros.
-        { conservative_common_denominator. nsatz. auto using a_d_y2_nonzero. }
-        { conservative_common_denominator_all. (* FIXME: doesn't do anything *)
-          common_denominator_all. nsatz. auto using a_d_y2_nonzero. }
+        unfold solve_for_x2; simpl; split; intros;
+          (conservative_common_denominator_all; [nsatz | eapply a_d_y2_nonzero; eauto]).
```
andres-erbsen: `conservative_common_denominator_all` generates a goal `False` here, probably by doing something equivalent to `unfold not; intro`. It would be nice if it left goals of the form `_ <> 0`.

src/CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v
```diff
@@ -97,14 +97,17 @@ Module E.
       bash_step.
       bash_step.
       { conservative_common_denominator.
-        conservative_common_denominator.
         nsatz.
         admit.
+        admit.
+        admit.
```
andres-erbsen: The new `conservative_common_denominator` generates four side-goals here, performing the same transformation by using the old `conservative_common_denominator_all` twice left only two side-goals.

src/CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v
JasonGross: Fixed by 9f3c224

src/CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v
JasonGross: 5a8e95d

src/Algebra.v
```diff
@@ -633,6 +633,16 @@ Module Field.
   End Homomorphism.
 End Field.
 
+(*** Tactics *)
+
+Module Import Algebra_syntax_Nsatz.
```
JasonGross: What's the reason to do it like this?  If you replace
```coq
Tactic Notation "nsatz" := nsatz 1%nat || nsatz 2%nat || nsatz 3%nat || nsatz 4%nat || nsatz 5%nat.
```
with
```coq
Ltac nsatz := nsatz 1%nat || nsatz 2%nat || nsatz 3%nat || nsatz 4%nat || nsatz 5%nat.
Tactic Notation "nsatz" := nsatz.
```
(assuming I haven't messed up my Ltac parsing; if I have, you'll need to do
```coq
Ltac nsatz_power n :=
  let nn := (eval compute in (BinNat.N.of_nat n)) in
  nsatz_sugar_power BinInt.Z0 nn.
Ltac nsatz := nsatz_power 1%nat || nsatz_power 2%nat || nsatz_power 3%nat || nsatz_power 4%nat || nsatz_power 5%nat.
Tactic Notation "nsatz" := nsatz.
Tactic Notation "nsatz" constr(n) := nsatz_power n.
```
), then you can just do `Ltac nsatz := Algebra_syntax.Nsatz.nsatz; dropRingSyntax`, which I think is slightly nicer.

JasonGross:
What's the status of this PR?

PR #18, opened 2016-07-02
Implement and prove Barrett reduction on Z

JasonGross:
This will serve as the high-level algorithm for modular reduction.

We follow Wikipedia very closely, except where we can do better (I
believe @jadep is updating Wikipedia).

Questions:
- [x] Is there a cleaner proof of `r_small` (@jadephilipoom, how does this one compare to your paper one?)  The main bottlenecks in making it shorter are that we're missing rewriting mod AC, and that we can't rewrite with `<=` lemmas.  (We could add typeclass instances for `setoid_rewrite`, but this would likely make all other rewriting in `Z` slower.  And it would be a pain.)
- [x] How do we feel about interleaving the reference and the code, and about quoting Wikipedia (@andres-erbsen, @achlipala)
- [ ] How do we feel about the unicode notations used for readability by non-Coq'ers (@andres-erbsen)

Any other comments? (@agl @mschilder123)

### DIFF
--- a/src/ModularArithmetic/BarrettReduction/Z.v
+++ b/src/ModularArithmetic/BarrettReduction/Z.v
@@ -1,0 +1,122 @@
+(*** Barrett Reduction *)
+(** This file implements Barrett Reduction on [Z].  We follow Wikipedia. *)
+Require Import Coq.ZArith.ZArith Coq.micromega.Psatz.
+Require Import Crypto.Util.ZUtil Crypto.Util.Tactics.
+
+Local Open Scope Z_scope.
+
+Section barrett.
+  Context (n a : Z)
+          (n_reasonable : n <> 0).
+  (** Quoting Wikipedia <https://en.wikipedia.org/wiki/Barrett_reduction>: *)
+  (** In modular arithmetic, Barrett reduction is a reduction
+      algorithm introduced in 1986 by P.D. Barrett. A naive way of
+      computing *)
+  (** [c = a mod n] *)
+  (** would be to use a fast division algorithm. Barrett reduction is
+      an algorithm designed to optimize this operation assuming [n] is
+      constant, and [a < n²], replacing divisions by
+      multiplications. *)
+
+  (** * General idea *)
+  Section general_idea.
+    (** Let [m = 1 / n] be the inverse of [n] as a floating point
+        number. Then *)
+    (** [a mod n = a - ⌊a m⌋ n] *)
+    (** where [⌊ x ⌋] denotes the floor function. The result is exact,
+        as long as [m] is computed with sufficient accuracy. *)
+
+    Local Notation "⌊am⌋" := (a / n) (only parsing). (* Coq automatically takes the floor *)
+
+    Theorem barrett_reduction_spec_correct
+      : a mod n = a - ⌊am⌋ * n.
+    Proof.
+      apply Zmod_eq_full; assumption.
+    Qed.
+  End general_idea.
+
+  (** * Barrett algorithm *)
+  Section barrett_algorithm.
+    (** Barrett algorithm is a fixed-point analog which expresses
+        everything in terms of integers. Let [k] be the smallest
+        integer such that [2ᵏ > n]. Think of [n] as representing the
+        fixed-point number [n 2⁻ᵏ]. We precompute [m] such that [m =
+        ⌊4ᵏ / n⌋]. Then [m] represents the fixed-point number
+        [m 2⁻ᵏ ≈ (n 2⁻ᵏ)⁻¹]. *)
+    Context (k : Z)
+            (k_good : 2 ^ (k-1) <= n < 2 ^ k)
+            (m : Z)
+            (m_good : m = 4^k / n). (* Coq automatically takes the floor *)
+    Context (a_nonneg : 0 <= a). (* Wikipedia neglects to mention this condition. *)
+
+
+    Lemma k_nonnegative : 0 <= k.
+    Proof.
+      destruct (Z_lt_le_dec k 0); try assumption.
+      rewrite !Z.pow_neg_r in * by lia; lia.
+    Qed.
+
+    (** Let *)
+    (** - [q = ⌊ma / 4ᵏ⌋] and *)
+    (** - [r = a - q n]. *)
+    Let q := (m * a) / 4^k.
+    Let r := a - q * n.
+    (** Because of the floor function, [q] is an integer and [r ≡ a
+        mod n]. *)
+    Theorem barrett_reduction_equivalent
+      : r mod n = a mod n.
+    Proof.
+      subst r q m.
+      rewrite <- !Z.add_opp_r, !Zopp_mult_distr_l, !Z_mod_plus_full by assumption.
+      reflexivity.
+    Qed.
+
+    Lemma qn_small
+      : q * n <= a.
+    Proof.
+      pose proof k_nonnegative; subst q r m.
+      assert (0 <= 2^(k-1)) by zero_bounds.
+      Zsimplify_fractions_le.
+    Qed.
+
+    (** Also, if [a < n²] then [r < 2n]. *)
+    (** N.B. It turns out that it is sufficient to assume [a < 4ᵏ]. *)
+    Context (a_small : a < 4^k).
+    Lemma r_small : r < 2 * n.
+    Proof.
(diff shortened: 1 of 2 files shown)


src/ModularArithmetic/BarrettReduction/Z.v
```diff
@@ -0,0 +1,122 @@
...
+  Section general_idea.
+    (** Let [m = 1 / n] be the inverse of [n] as a floating point
+        number. Then *)
+    (** [a mod n = a - ⌊a m⌋ n] *)
+    (** where [⌊ x ⌋] denotes the floor function. The result is exact,
+        as long as [m] is computed with sufficient accuracy. *)
+
+    Local Notation "⌊am⌋" := (a / n) (only parsing). (* Coq automatically takes the floor *)
+
+    Theorem barrett_reduction_spec_correct
```
andres-erbsen: I think the name of this theorem should include "naive" or something like that -- currently it can be quite confusing, especially in SearchAbout.

src/ModularArithmetic/BarrettReduction/Z.v
```diff
@@ -0,0 +1,122 @@
...
+
+  (** * General idea *)
+  Section general_idea.
+    (** Let [m = 1 / n] be the inverse of [n] as a floating point
+        number. Then *)
+    (** [a mod n = a - ⌊a m⌋ n] *)
+    (** where [⌊ x ⌋] denotes the floor function. The result is exact,
+        as long as [m] is computed with sufficient accuracy. *)
+
+    Local Notation "⌊am⌋" := (a / n) (only parsing). (* Coq automatically takes the floor *)
```
andres-erbsen: "Coq automatically takes the floor" might be better phrased as "/ means truncating division".

src/ModularArithmetic/BarrettReduction/Z.v
```diff
@@ -0,0 +1,122 @@
...
+  (** * Barrett algorithm *)
+  Section barrett_algorithm.
+    (** Barrett algorithm is a fixed-point analog which expresses
+        everything in terms of integers. Let [k] be the smallest
+        integer such that [2ᵏ > n]. Think of [n] as representing the
+        fixed-point number [n 2⁻ᵏ]. We precompute [m] such that [m =
+        ⌊4ᵏ / n⌋]. Then [m] represents the fixed-point number
+        [m 2⁻ᵏ ≈ (n 2⁻ᵏ)⁻¹]. *)
+    Context (k : Z)
+            (k_good : 2 ^ (k-1) <= n < 2 ^ k)
```
andres-erbsen: Do we actually use the `2 ^ (k-1) <= n` assumption anywhere? IIRC, `ed25519-amd64-51-30k` uses `n = 2^252 + 27742317777372353535851937790883648493` and `k = 256`.

src/ModularArithmetic/BarrettReduction/Z.v
```diff
@@ -0,0 +1,122 @@
...
+
+    (** In that case *)
+    (**
+<<
+          ⎧
+          ⎪ r       if r < n
+a mod n = ⎨
+          ⎪ r - n   otherwise
+          ⎩
+>> *)
```
andres-erbsen: I think this comment duplicates the definition that follows and does not add much to the code. Perhaps remove it?

src/ModularArithmetic/BarrettReduction/Z.v
```diff
@@ -0,0 +1,122 @@
...
+
+    Lemma k_nonnegative : 0 <= k.
+    Proof.
+      destruct (Z_lt_le_dec k 0); try assumption.
+      rewrite !Z.pow_neg_r in * by lia; lia.
+    Qed.
+
+    (** Let *)
+    (** - [q = ⌊ma / 4ᵏ⌋] and *)
+    (** - [r = a - q n]. *)
```
andres-erbsen: This comment duplicates the code right below it. Please remove.

andres-erbsen:
Jade's Github username is @jadephilipoom, mine is @andres-erbsen.

I generally like the way the file reads (in particular, I prefer it to the Wikipedia article being imitated).

The cases, floor and superscript symbols do not display correctly in a Arch Linux virtual terminal (with the default font nor Terminus). I if we want to keep the rational numbers analogy there, we should keep the floor symbol. The others do not add much to the coq code they are next to, and I would prefer if we removed these comments altogether.

jadephilipoom:
The proof of r_small is fiddly on paper too, so I think this version is fine. (The wikipedia article is now updated.)

achlipala:
I like the general strategy of keeping the code readable for experts, without undue distractions of long quotes, so that URL citations are good, and medium-length sequences of quoted text may also often be helpful, probably not with more than 10 lines at a time.

src/ModularArithmetic/BarrettReduction/Z.v
JasonGross: No, indeed we don't.  Here we see the power of automation; the proof fails to go through if I just remove `2 ^ (k-1) <= `, but it goes back to working if I add the assumption `0 < n` (or `0 <= n`, since we have `0 <> n` above).  I'll push the change momentarily.  Perhaps one of us should update Wikipedia with a note that `k` doesn't need to be the smallest such value?  (I'm happy to let you or @jadephilipoom do it.)

JasonGross:
Do we want to go all the way with the floor function notation, and make
```coq
Local Notation "⌊ x / y ⌋" := (x / y) (only parsing).
```
so we don't need to keep reminding non-Coq'ers that `/` means truncated division?

src/ModularArithmetic/BarrettReduction/Z.v
JasonGross: Awww, you don't like my fancy unicode braces?  Removed.  (I'm an eighth tempted to make a unicode notation for cases that prints with braces on nice lines and interprets as an if statement...)

src/ModularArithmetic/BarrettReduction/Z.v
JasonGross: Updated to "`/` is `Z.div`, which means truncated division", thanks.

src/ModularArithmetic/BarrettReduction/Z.v
JasonGross: Changed to `naive_barrett_reduction_correct`

src/ModularArithmetic/BarrettReduction/Z.v
JasonGross: Done

JasonGross:
> Github username

Oops, thanks.

> in particular, I prefer it to the Wikipedia article being imitated

Thanks!

> probably not with more than 10 lines at a time

Sounds like a good rule of thumb, in general.  I'm inclined to leave the 15 lines of introduction towards the top, to let this file be stand-alone in terms of explanation.  (My plan is also to only inline explanations like this in the highest-level specification of each algorithm I formalize, when I'm going from a nice spec.)

agl:
Looks good, but this is single-word Barrett reduction. That's rarer in cryptography than the polynomial-based, multi-word Barrett reduction (although perhaps this would be useful for proving that). The single-word form is more commonly found in crypto that deals with small fields, like code- and lattice-based crypto.

andres-erbsen:
LGTM.

@agl: I think the plan for generating code that does polynomial-based Barrett reduction is to use polynomials to represent `a` and `m`. Or is this not what polynomial-based Barrett reduction means?

agl:
It's mostly what Barrett reduction means, but the polynomial version is often presented differently. [Barrett's original paper](http://link.springer.com/chapter/10.1007%2F3-540-47721-7_24) was really all about the polynomial version but he mentions the single-word version as motivation. [More recent](http://cacr.uwaterloo.ca/hac/about/chap14.pdf) references (section 14.3.3) mostly skip that and present the multi-word version immediately. Having said that, I can imagine defining a mapping from `list Word` to `Z` and showing that the multi-word version is correct that way, in which case this is a great step.

JasonGross:
@agl My intention is to define the map from `list Word` to `Z` (or `list Z` to `Z`), and prove that if you first transform and then reduce, or if you first reduce and then transform, you get the same result.  Unless the polynomial version is fundamentally different, correctness belongs at the `Z` level of abstraction; doing things at the "wrong" level of abstraction is an easy way to make your proofs very painful in Coq.
And thanks for the references!  I'll take a look at them soon.

PR #19, opened 2016-07-02
Rename lemmas about Z in ZUtil

JasonGross:
The standard library uses `Z.*`, and `Z*` and `Z_*` are compatibility
notations, I believe.  We follow suit.

Also, eliminate a few lemmas that are duplicates of ones in the standard
library.

The first commit does the renaming; the second is purely an indentation change.

@andres-erbsen @jadephilipoom any objections?  Does this interfere with anything you're working on?


andres-erbsen:
LGTM.

PR #21, opened 2016-07-08
Make [base] and [log_cap] notations

JasonGross:
Also use [ZUtil.Z.pow2_mod].  Making `base` a notation lets us remove the dependency of
ModularBaseSystem on ModularArithmetic.PseudoMersenneBaseParamProofs.

This is a small part of reorganizing and factoring ModularBaseSystem for
use with Barrett reduction.


JasonGross:
Rebased against current master to fix things broken by bb38344557cddbc64eac0eb5b174d54c0507e08a

jadephilipoom:
Looks good to me.

PR #22, opened 2016-07-09
stop using raw nsatz, field_algebra, common_denominator

### LINKED ISSUE #10: What's the spec of field_algebra?
I added a version of `common_denominator` that is faster and cleaner; it does not perform simplification in portions of the term that it doesn't need to. (https://github.com/mit-plv/fiat-crypto/commit/2b7e45498d3a3e8856906c7eed7caec4f0053aae)

However, I can't use it in `field_algebra`, because various code depends on the exact way in which `field_algebra` calls `field_simplify_eq`.  If we can, I'd like to head off this issue early; this is the first hint of something like the design flaw in Coq's unification where the devs can't touch anything without breaking backwards compatibility, because they've extended it ad-hoc without any specification.

(@andres-erbsen @achlipala)

### LINKED ISSUE #15: `super_nsatz` clears `H: 1 + 1 <> 0`

andres-erbsen:
Supersedes #16; closes #10; closes #15.

Several replacements were made in separate commits:
- `nsatz` was redefined in `Algebra.v` as `nsatz; dropRingSyntax`, other files were changed to import that instead
- uses of `field_algebra` were replaced with `super_nsatz`, `nsatz` and `conservative_common_denominator` as appropriate.
- `common_denominator` family tactics were replaced with `conservative_common_denominator` family tactics
- `common_denominator` tactics were renamed to `field_simplify_if_div`
- `conservative_common_denominator` tactics were renamed to `common_denominator`

In some cases, `super_nsatz` is a too big hammer: when the goal contains many hypotheses of the form `?x <> 0` then `super_nsatz` takes a long time, but `intro; apply; rewrite; ring` wins quickly. This speedup was successfully applied to make the CompleteEdwardsCurveTheorems associativity proof feasible.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -28,6 +28,7 @@ src/CompleteEdwardsCurve/Pre.v
 src/Encoding/EncodingTheorems.v
 src/Encoding/ModularWordEncodingPre.v
 src/Encoding/ModularWordEncodingTheorems.v
+src/Encoding/PointEncodingPre.v
 src/Experiments/DerivationsOptionRectLetInEncoding.v
 src/Experiments/EdDSARefinement.v
 src/Experiments/GenericFieldPow.v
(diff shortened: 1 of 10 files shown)


JasonGross:
LGTM modulo travis.  If you rebase on top of current master, that should fix travis.  (Sorry for leaving the repo in a broken state while I flew to Berlin.)

PR #24, opened 2016-07-18
Move some definitions to Pow2Base

JasonGross:
These definitions don't depend on PseudoMersenneBaseParams, only on
limb_widths, and we'll want them for BarrettReduction / P256.

@jadephilipoom, does this conflict with anything you want to do?

@jadephilipoom @andres-erbsen is there any reason to prefer defining `add_to_nth` as `set_nth n (x + nth_default 0 xs n) xs` vs `update_nth n (fun y => x + y) xs`, or vice versa?

### DIFF
--- a/src/ModularArithmetic/ModularBaseSystemProofs.v
+++ b/src/ModularArithmetic/ModularBaseSystemProofs.v
@@ -10,6 +10,7 @@ Require Import Crypto.ModularArithmetic.Pow2BaseProofs.
 Require Import Crypto.ModularArithmetic.PseudoMersenneBaseParams.
 Require Import Crypto.ModularArithmetic.PseudoMersenneBaseParamProofs.
 Require Import Crypto.ModularArithmetic.ExtendedBaseVector.
+Require Import Crypto.Util.Tactics.
 Require Import Crypto.Util.Notations.
 Local Open Scope Z_scope.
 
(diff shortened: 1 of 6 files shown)


src/ModularArithmetic/ModularBaseSystemProofs.v
```diff
@@ -460,21 +375,28 @@ Section CarryProofs.
...
 
+  Lemma length_carry_and_reduce : forall i us, length (carry_and_reduce i us) = length us.
+  Proof. intros; unfold carry_and_reduce; autorewrite with distr_length; reflexivity. Qed.
+  Hint Rewrite @length_carry_and_reduce : distr_length.
+
+  Lemma length_carry : forall i us, length (carry i us) = length us.
+  Proof. intros; unfold carry; break_if; autorewrite with distr_length; reflexivity. Qed.
+  Hint Rewrite @length_carry : distr_length.
+
+  (* FIXME(jadep)(from jgross): Do we need this? *)
```
jadephilipoom: No, we don't; it's a duplicate of length_carry. I can fix this, if you don't.

src/ModularArithmetic/Pow2Base.v
```diff
@@ -39,4 +42,53 @@ Section Pow2Base.
...
+  (** ** Carrying *)
+  Section carrying.
+    (** Here we implement addition and multiplication with simple
+        carrying. *)
+    Notation log_cap i := (nth_default 0 limb_widths i).
+
+
+    Definition add_to_nth n (x:Z) xs :=
+      set_nth n (x + nth_default 0 xs n) xs.
+    (* TODO: Maybe we should use this instead? *)
```
jadephilipoom: Why is this better, though?

JasonGross: It might not be.  The abstraction seems a tiny bit cleaner, to me: we don't need to reason about `nth_default 0` anymore.  But it's not terribly important.

jadephilipoom:
Looks okay to me, if you fix the merge conflicts and satisfy Travis.

src/ModularArithmetic/Pow2Base.v
jadephilipoom: I agree, but I think we have higher priorities at the moment. You can push it onto my stack, but I can't guarantee that I'll get to it any time soon. If you want to do it, though, I'll happily merge it!

PR #25, opened 2016-07-18
Add some global settings

JasonGross:
Other ones to consider someday: `Primitive Projections` (blocking on
[bug #4946](https://coq.inria.fr/bugs/show_bug.cgi?id=4946)), `Universe
Polymorphism`, `Asymmetric Patterns` (already set in anything depending
on Bedrock/Nomega.v)

@andres-erbsen sanity check, is there anything obviously wrong with doing this?


JasonGross:
Or not.  Apparently unrecognized options are ignored, unless they are followed by arguments, in which case you get an error.

PR #26, opened 2016-07-18
Use update_nth in add_to_nth

JasonGross:
It leads to a slightly more transparent and clearer definition.

If I got everything right, nothing should depend on the judgmental
definition of [add_to_nth] anymore.

@andres-erbsen does this break any of your plans for optimized Gallina code?


andres-erbsen:
I think this is a good idea, I am guessing it won't break anything, and I think any breakage wouldn't be difficult to fix. LGTM.

PR #30, opened 2016-07-19
Add a separate non-specific target

### LINKED ISSUE #27: Separate Makefile targets for library and generated code
As the `Specific/` code becomes more developed, build times go up. It woud be nice to have Makefile target that builds everything but the `Specific` directory for use while pushing changes through the rest of the codebase. This would be especially useful when refactoring irrelevant proofs.

JasonGross:
This should fix #27.  We depend on some files in the etc/coq-scripts
submodule.

Note that you need to either run `make clean -k` or `rm -f Makefile.coq`
after pulling this to build the development.


JasonGross:
@andres-erbsen can you test this and see if it does what you want?

Makefile
```diff
@@ -46,15 +66,6 @@ install-coqprime-8.4:
...
-	$(MAKE) -f Makefile.coq clean
-	rm -f Makefile.coq
-
-cleanall: clean clean-coqprime
+cleanall:: clean clean-coqprime
 
-install: coq Makefile.coq
-	$(MAKE) install-coqprime
-	$(MAKE) -f Makefile.coq install
+install: coq install-coqprime
```
andres-erbsen: What does the install target do?

Makefile.submodule
```diff
@@ -0,0 +1,35 @@
+ifneq (,$(wildcard .git)) # if we're in a git repo
```
andres-erbsen: Why is this a Makefile, not a script? Why do we need a script at all -- running `git clone --recursive` or `git submodule update` does not seem that difficult.

Makefile
JasonGross: Since we include `Makefile.coq`, the install target first installs coqprime and builds all the files, and then installs the files as per `Makefile.coq`

Makefile.submodule
JasonGross: Because it needs to be included in `Makefile` and I don't know any good ways of running scripts unconditionally in `make`.  This is so that when I update the submodule and depend on some change that I made to coq-scripts, the build process doesn't suddenly fail for everyone.  It's unfortunate if usually you run `git pull && make`, and sometimes this fails with strange error messages and then you need to run `git pull && git submodule update && make`.

andres-erbsen: Nah. It's really not the Makefile's job to go fetch stuff from the internet and update the git repository. If hacks like this are necessary to use the submodule, using the submodule is probably not a good idea.

Is there any circumstance where Makefile.coq VOFILES differs from appending "o" to every entry of _CoqProject?

JasonGross: Is copy/pasting and maintaining the same code across multiple repositories a better idea?  I can just inline the [`sed` hacks needed to `-include Makefile.coq`](https://github.com/JasonGross/coq-scripts/blob/master/Makefile.coq.common#L85) instead.

When `_CoqProject` contains things like `-R src Crypto`, `COQPATH = "$(pwd)/coqprime"`, `-indices-matter`, etc.  See also https://github.com/ProofGeneral/PG/issues/60#issuecomment-188790660

andres-erbsen: We don't need to parse arbitrary _CoqProject files here, though. I think we could just use the same find magic to generate the list of .vo files that we use to generate the list of .v files. In general, if at all feasible, let's try to control the *inputs* to `make` and `coq_makefile` instead of trying to make sense the outputs of them.

I think duplicating code is better than breaking abstraction boundaries -- parsing a file that was never meant to be parsed by third parties is bad, as is implementing our own auto-update mechanism for git submodules given that git did not include one on purpose.

JasonGross: No, we definitely don't want to do that.  `_CoqProject` is intentionally not automatically regenerated when the list of .v files changes, because it's checked into version control.  It needs to be checked into version control so that `emacs` and `coqide` work without building it.  (We could automatically regenerate it if we wanted to, though...)  The list of `.v` files in our `Makefile` should not get out of sync with the list in `_CoqProject`.

I'll go add a commit to copy-paste the `sed` magic, though.

JasonGross:
Rebased and updated; we no longer depend on the submodule.

andres-erbsen:
LGTM. Ugh.

This totally is how Makefiles are used, https://www.gnu.org/software/make/manual/html_node/Include.html even documents the behavior of `-include` that we are relying on. This may even be less fragile than recursive make. Multiple layers of code modifying each other just makes me cringe, but fighting this isn't a goal of our project. I am guessing this solution PR will work with little enough fiddly-ness that it makes sense to go with it. Merge away.

JasonGross:
Rebased to fix merge conflicts.  Since it builds with 8.5 on travis, I'm going to merge this.

Makefiles are indeed a pain.  Most of the sed hacks can go away when we move to 8.5-only, where `coq_makefile` correctly makes `clean` a double-colon rule.

PR #36, opened 2016-07-22
Full Assembly-conversion Stack

varomodt:
@achlipala @andres-erbsen @JasonGross


JasonGross:
Not sure what you want comments on....

Travis is unhappy because `replace x with y at n by tac` isn't valid before 8.4pl4.  I can't recall if `replace x with y at n` is valid.  A workaround is:
```coq
set (tmp := x) at n; replace tmp with y by tac; clear tmp.
```

varomodt:
Thanks @JasonGross -- not looking for anything in particular :)

I've added a custom implementation of replace-at in `src/Assembly/WordizeUtil.v` that might be useful elsewhere.

JasonGross:
I'm going to push a change that will probably conflict with the PR: I'm revoking ability of `intuition` to use hints in random databases.  Otherwise there's no way I can make databases that are meant to be separate, which potentially loop when combined.  Your code seems to rely a lot on `intuition` grabbing random hints (things from `zarith` and `crelations` and `typeclass_instances`), so I've had to change a lot of your `intuition` invocations.  Additionally, I've added `Require Export Crypto.Util.FixCoqMistakes.` to the top of each file that uses `intuition`.  You're free to blow away my changes to the assembly files, as long as you keep the `Require Export Crypto.Util.FixCoqMistakes.` at the top of any file using `intuition`, and make your code compile.  The tactic I added called `debug_intuition`, which will tell you what hints went into making `intuition auto with *` successful, combined with `Print Hint *`, may prove useful for debugging.

Once I merge (in about 5 minutes), please add `Require Export Crypto.Util.FixCoqMistakes.` to any files that make use of `intuition`.

JasonGross:
In Coq <= 8.4pl2, `vm_compute` gave an anomaly on evars.  I'm Coq > 8.4pl2, it errors.  You should guard your calls to `vm_compute` in `bound_compute` with `not has_evar x` (replacing `x` with the relevant things).  Note that the `not` tactical is defined in `Crypto.Util.Tactics`.

JasonGross:
You seem to be missing Vectorize.v

andres-erbsen:
https://github.com/mit-plv/fiat-crypto/blob/c634cec9b27946e8802afb91db66174402a25a5e/src/Experiments/SpecificCurve25519.v#L437 is an example of code we want to be converting. This is the smallest unit of code on which bounds analysis can be done once and for all: if the code is correct, then there exists a list of bounds `b` such that if the inputs are (elementwise) bounded by `b++b` then the output words are bounded by `b`.

I think it should be possible to convert this to assembly as-is using 64-bit words, and with slightly more finesse we could be recognized that most values fit into 32-bit words. Later we would have a transformation that implements addition and bit operations on 64-bit words by representing them as a pair of 32-bit words, after which we could switch to 128-bit multiplication outputs and 64-bit words for other values to reduce the number of words.

PR #38, opened 2016-07-25
Convenience settings: If COQPATH is not set, set it by default

JasonGross:
This allows most users to not need to type COQPATH=... on make.

@andres-erbsen , do you think this is fine, or do you object to doing this in the `Makefile`.  (Here it really just is to eliminate typing characters.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,9 @@ OTHERFLAGS ?= -w "-deprecated-appcontext -notation-overridden"
 endif
 endif
 
+COQPATH?=$(shell pwd)/$(COQPRIME_FOLDER)
+export COQPATH
+
 coqprime:
 	$(MAKE) -C $(COQPRIME_FOLDER)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -84,5 +87,8 @@ cleanall:: clean clean-coqprime
 
 install: coq install-coqprime
 
+printenv::
+	@echo "COQPATH =        $$COQPATH"
```
andres-erbsen: What motivates the amount of spaces on this line?

andres-erbsen:
I don't object -- a Makefile making reasonable guesses about build variables is quite benign compared to relying on double-colon rules and such.

Makefile
JasonGross: Consistency with the rest of `printenv` in `Makefile.coq`

JasonGross:
What do you mean by "relying on double-colon rules"?  (Note that this runs `sed` on the output of `coq_makefile` to make `printenv` a double-colon rule so that we can append `COQPATH` to it.)

andres-erbsen:
IIRC the only specified use case for double colon rules is for for multiple targets that build the same file after changes to different input files. I may be recalling this wrong, though. Either way, I think this patch is a good idea.

JasonGross:
*head-tilt*  I don't quite see how that works.  The way I use double-colon rules is to distribute a single recipe across multiple locations; I only do this with phony targets (like `clean`).  I believe the difference is that if you give multiple recipes for single-colon rules, the latest one replaces all earlier ones, and emits a warning about shadowing; double-colon recipes are concatenated.

PR #40, opened 2016-07-28
Refactor ModularArithmetic into Zmod, expand Decidable

andres-erbsen:
ModularArithmetic now uses Algebra lemmas in various places instead of
custom manual proofs. Similarly, Util.Decidable is used to state and
prove the relevant decidability results.

Backwards-incompatible changes:

F_some_lemma -> Zmod.some_lemma
Arguments ZToField _%Z _%Z : clear implicits.
inv_spec says inv x * x = 1, not x * inv x = 1

### DIFF
--- a/src/Algebra.v
+++ b/src/Algebra.v
@@ -12,16 +12,9 @@ Module Import ModuloCoq8485.
   Infix "mod" := modulo.
 End ModuloCoq8485.
 
-Notation is_eq_dec := (DecidableRel _) (only parsing).
-Notation "@ 'is_eq_dec' T R" := (DecidableRel (R:T->T->Prop)) (only parsing).
-Notation eq_dec x y := (@dec (_ x y) _) (only parsing).
-Notation "x =? y" := (eq_dec x y) : type_scope.
-
 Section Algebra.
   Context {T:Type} {eq:T->T->Prop}.
   Local Infix "=" := eq : type_scope. Local Notation "a <> b" := (not (a = b)) : type_scope.
-
-  Local Notation is_eq_dec := (@is_eq_dec T eq).
 
   Section SingleOperation.
     Context {op:T->T->T}.
@@ -41,7 +34,7 @@ Section Algebra.
 
         monoid_op_Proper: Proper (respectful eq (respectful eq eq)) op;
         monoid_Equivalence : Equivalence eq;
-        monoid_is_eq_dec : is_eq_dec
+        monoid_is_eq_dec : DecidableRel eq
       }.
     Global Existing Instance monoid_is_associative.
     Global Existing Instance monoid_is_left_identity.
@@ -364,6 +357,23 @@ Module ScalarMult.
     Global Existing Instance scalarmult_Proper.
     Context `{is_scalarmult}.
 
+    Fixpoint scalarmult_ref (n:nat) (P:G) {struct n} :=
+      match n with
+      | O => zero
+      | S n' => add P (scalarmult_ref n' P)
+      end.
+
+    Global Instance Proper_scalarmult_ref : Proper (Logic.eq==>eq==>eq) scalarmult_ref.
+    Proof.
+      repeat intro; subst.
+      match goal with [n:nat |- _ ] => induction n; simpl @scalarmult_ref; [reflexivity|] end.
+      repeat match goal with [H:_ |- _ ] => rewrite H end; reflexivity.
+    Qed.
+
+    Lemma scalarmult_ext : forall n P, mul n P = scalarmult_ref n P.
+      induction n; simpl @scalarmult_ref; intros; rewrite <-?IHn; (apply scalarmult_0_l || apply scalarmult_S_l).
+    Qed.
+
     Lemma scalarmult_1_l : forall P, 1*P = P.
     Proof. intros. rewrite scalarmult_S_l, scalarmult_0_l, right_identity; reflexivity. Qed.
 
@@ -402,6 +412,10 @@ Module ScalarMult.
         rewrite scalarmult_add_l, scalarmult_1_l, Group.inv_op, scalarmult_S_l, Group.cancel_left; eauto. }
     Qed.
   End ScalarMultProperties.
+
+  Global Instance scalarmult_ref_is_scalarmult {G eq add zero} `{@monoid G eq add zero}
+    : @is_scalarmult G eq add zero (@scalarmult_ref G add zero).
+  Proof. split; try exact _; intros; reflexivity. Qed.
 End ScalarMult.
 
 Require Coq.nsatz.Nsatz.
@@ -588,7 +602,7 @@ Module IntegralDomain.
       : eq (mul x y) zero -> eq x zero \/ eq y zero.
     Proof.
       pose proof mul_nonzero_nonzero x y.
-      destruct (eq_dec x zero); destruct (eq_dec y zero); intuition.
+      destruct (dec (eq x zero)); destruct (dec (eq y zero)); intuition.
     Qed.
 
     Lemma mul_nonzero_nonzero_iff (x y : T)
@@ -663,7 +677,7 @@ Module Field.
 
   Lemma isomorphism_to_subfield_field
         {T EQ ZERO ONE OPP ADD SUB MUL INV DIV}
-        {Equivalence_EQ: @Equivalence T EQ} {eq_dec: forall x y, {EQ x y} + {~ EQ x y} }
+        {Equivalence_EQ: @Equivalence T EQ} {eq_dec: DecidableRel EQ}
         {Proper_OPP:Proper(EQ==>EQ)OPP}
         {Proper_ADD:Proper(EQ==>EQ==>EQ)ADD}
         {Proper_SUB:Proper(EQ==>EQ==>EQ)SUB}
@@ -673,7 +687,7 @@ Module Field.
         {R eq zero one opp add sub mul inv div} {fieldR:@field R eq zero one opp add sub mul inv div}
         {phi}
         {eq_phi_EQ: forall x y, eq (phi x) (phi y) -> EQ x y}
-        {neq_zero_one : (EQ ZERO ONE -> False)}
+        {neq_zero_one : (not (EQ ZERO ONE))}
         {phi_opp : forall a, eq (phi (OPP a)) (opp (phi a))}
         {phi_add : forall a b, eq (phi (ADD a b)) (add (phi a) (phi b))}
         {phi_sub : forall a b, eq (phi (SUB a b)) (sub (phi a) (phi b))}
@@ -753,7 +767,7 @@ Module Field.
        -> phi (INV x) = inv (phi x).
     Proof.
       intros.
-      destruct (eq_dec x ZERO); auto.
+      destruct (dec (EQ x ZERO)); auto.
       assert (phi (MUL (INV x) x) = one) as A. {
         rewrite left_multiplicative_inverse; auto using Ring.homomorphism_one.
       }
@@ -1223,8 +1237,8 @@ Section ExtraLemmas.
   Lemma only_two_square_roots'_choice x y : x * x = y * y -> x = y \/ x = opp y.
   Proof.
     intro H.
-    destruct (eq_dec x y); [ left; assumption | right ].
-    destruct (eq_dec x (opp y)); [ assumption | exfalso ].
+    destruct (dec (eq x y)); [ left; assumption | right ].
+    destruct (dec (eq x (opp y))); [ assumption | exfalso ].
     eapply only_two_square_roots'; eassumption.
   Qed.
 
--- a/src/Algebra.v
+++ b/src/Algebra.v
@@ -365,12 +365,14 @@ Module ScalarMult.
 
     Global Instance Proper_scalarmult_ref : Proper (Logic.eq==>eq==>eq) scalarmult_ref.
     Proof.
-      repeat intro; subst.
-      match goal with [n:nat |- _ ] => induction n; simpl @scalarmult_ref; [reflexivity|] end.
-      repeat match goal with [H:_ |- _ ] => rewrite H end; reflexivity.
+      repeat intro; subst; 
+        match goal with [n:nat |- _ ] =>
+                        solve [induction n; simpl scalarmult_ref; rewrite_hyp ?*; reflexivity]
+        end.
     Qed.
 
     Lemma scalarmult_ext : forall n P, mul n P = scalarmult_ref n P.
+    Proof.
       induction n; simpl @scalarmult_ref; intros; rewrite <-?IHn; (apply scalarmult_0_l || apply scalarmult_S_l).
     Qed.
 
--- a/src/Encoding/ModularWordEncodingTheorems.v
+++ b/src/Encoding/ModularWordEncodingTheorems.v
@@ -3,7 +3,7 @@ Require Import Coq.Numbers.Natural.Peano.NPeano.
 Require Import Crypto.CompleteEdwardsCurve.CompleteEdwardsCurveTheorems.
 Require Import Crypto.ModularArithmetic.PrimeFieldTheorems Crypto.ModularArithmetic.ModularArithmeticTheorems.
 Require Import Bedrock.Word.
-Require Import Crypto.Tactics.VerdiTactics.
+Require Import Crypto.Tactics.VerdiTactics Crypto.Util.Tactics.
 Require Import Crypto.Spec.Encoding.
 Require Import Crypto.Util.ZUtil.
 Require Import Crypto.Spec.ModularWordEncoding.
@@ -24,7 +24,7 @@ Section SignBit.
       assert (m < 1)%Z by (apply Z2Nat.inj_lt; try omega; assumption).
       omega.
     + assert (0 < m)%Z as m_pos by (pose proof prime_ge_2 m prime_m; omega).
-      pose proof (FieldToZ_range x m_pos).
+      pose proof (Zmod.FieldToZ_range x m_pos).
       destruct (FieldToZ x); auto.
       - destruct p; auto.
       - pose proof (Pos2Z.neg_is_neg p); omega.
@@ -35,20 +35,13 @@ Section SignBit.
     rewrite sign_bit_parity; auto.
   Qed.
 
-  Lemma sign_bit_opp : forall (x : F m), x <> 0 -> negb (@sign_bit m sz x) = @sign_bit m sz (opp x).
-  Proof.
-    intros.
-    pose proof sign_bit_zero as sign_zero.
-    rewrite !sign_bit_parity in *.
-    pose proof (F_opp_spec x) as opp_spec_x.
-    apply F_eq in opp_spec_x.
-    rewrite FieldToZ_add in opp_spec_x.
-    rewrite <-opp_spec_x, Z.odd_mod in sign_zero by (pose proof prime_ge_2 m prime_m; omega).
-    replace (Z.odd m) with true in sign_zero by (destruct (Z.prime_odd_or_2 m prime_m); auto || omega).
-    rewrite Z.odd_add, F_FieldToZ_add_opp, Z.div_same, Bool.xorb_true_r in sign_zero by assumption || omega.
-    apply Bool.xorb_eq.
-    rewrite <-Bool.negb_xorb_l.
-    assumption.
-  Qed.
+  Lemma odd_m : Z.odd m = true. Admitted.
 
+  Lemma sign_bit_opp (x : F m) (Hnz:x <> 0) : negb (@sign_bit m sz x) = @sign_bit m sz (opp x).
+  Proof.
+    pose proof Zmod.FieldToZ_nonzero_range x Hnz; specialize_by omega.
+    rewrite !sign_bit_parity, Zmod.FieldToZ_opp, Z_mod_nz_opp_full,
+      Zmod_small, Z.odd_sub, odd_m, (Bool.xorb_true_l (Z.odd x)) by
+       (omega || rewrite Zmod_small by omega; auto using Zmod.FieldToZ_nonzero); trivial.
+  Qed.
 End SignBit.
--- a/src/Encoding/ModularWordEncodingTheorems.v
+++ b/src/Encoding/ModularWordEncodingTheorems.v
@@ -6,6 +6,7 @@ Require Import Bedrock.Word.
 Require Import Crypto.Tactics.VerdiTactics Crypto.Util.Tactics.
 Require Import Crypto.Spec.Encoding.
 Require Import Crypto.Util.ZUtil.
+Require Import Crypto.Util.FixCoqMistakes.
 Require Import Crypto.Spec.ModularWordEncoding.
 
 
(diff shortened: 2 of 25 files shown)


src/Util/Decidable.v
```diff
@@ -47,9 +50,11 @@ Hint Extern 0 (Decidable (~?A)) => apply (@dec_not A) : typeclass_instances.
 Global Instance dec_eq_unit : DecidableRel (@eq unit) | 10. exact _. Defined.
 Global Instance dec_eq_bool : DecidableRel (@eq bool) | 10. exact _. Defined.
 Global Instance dec_eq_Empty_set : DecidableRel (@eq Empty_set) | 10. exact _. Defined.
-Global Instance dec_eq_nat : DecidableRel (@eq nat) | 10. exact _. Defined.
 Global Instance dec_eq_prod {A B} `{DecidableRel (@eq A), DecidableRel (@eq B)} : DecidableRel (@eq (A * B)) | 10. exact _. Defined.
 Global Instance dec_eq_sum {A B} `{DecidableRel (@eq A), DecidableRel (@eq B)} : DecidableRel (@eq (A + B)) | 10. exact _. Defined.
+Global Instance dec_eq_nat : DecidableRel (@eq nat) | 10. exact _. Defined.
```
andres-erbsen: Should these have weithts?

JasonGross: You might as well put them at `10` with all the others; if I recall correctly, the reason for the weights is so that hints in the context always get picked up before these.  (Or else because I tried not giving them weights, and everything was sad.)

src/Util/Decidable.v
```diff
@@ -60,5 +65,28 @@ Proof. solve_decidable_transparent. Defined.
...
 
+Lemma eqsig_eq {T} {U} {Udec:DecidableRel (@eq U)} (f g:T->U) (x x':T) pf pf' :
+  (exist (fun x => f x = g x) x pf) = (exist (fun x => f x = g x) x' pf') <-> (x = x').
+Proof. split; intros; [congruence | subst x'; apply f_equal, UIP_dec, Udec]. Qed.
+
+Global Instance dec_eq_eqsig
+       {T} {Tdec:DecidableRel (@eq T)}
+       {U} {Udec:DecidableRel (@eq U)}
+       (f g:T->U)
+  : DecidableRel (@eq (sig (fun x => f x = g x))).
```
JasonGross: There's a better factoring of this:
```coq
dec_eq_lift : forall {A} {Adec : DecidableRel (@eq A)} (x y : A) : DecidableRel (@eq (x = y))
dec_eq_sig : forall {A} {P : A -> Prop} {Adec : DecidableRel (@eq A)} {Pdec : forall a, DecidableRel (@eq (P a))}, DecidableRel (@eq (@sig A P))
```
The first by `UIP_dec`, the second straightforwardly.  Then this lemma should be automatically solved by typeclass resolution.

andres-erbsen: `eqsig_eq` is also used at https://github.com/mit-plv/fiat-crypto/pull/40/files#diff-4205713a9165d6688fb39119d18336deR64, https://github.com/mit-plv/fiat-crypto/pull/40/files#diff-4205713a9165d6688fb39119d18336deR43. Should I move it somewhere else, just prove it differently, or use a different lemma?

JasonGross: I've just pushed `Util.Sigma` with
```coq
path_sig_hprop_iff : forall {A} {P : A -> Prop} (P_hprop : forall (x : A) (p q : P x), p = q) (u v : @sig A P),
 u = v <-> (proj1_sig u = proj1_sig v)
```
Composing this with `UIP_dec` and `Udec` will give you `eqsig_eq` trivially (feel free to do this in this file, if you really need to).

Let me see if I can extend the automation in this file to handle `dec_eq_lift` (or maybe `dec_eq_eq`) and `dec_eq_sig`; I'd suggest dropping `dec_eq_eqsig` and letting it be resolved by typeclasses automatically whenever it's needed.

JasonGross: I've pushed enough instances to automatically resolve `dec_eq_eqsig` by typeclasses (though not in `Decidable.v`, where `Local Hint Extern 0 => solve [ solve_decidable_transparent ] : typeclass_instances.` seems to loop on this goal).

Want to rebase, drop `dec_eq_eqsig`, and re-prove `eqsig_eq` using `path_sig_hprop_iff`?

src/Experiments/SpecEd25519.v
```diff
@@ -8,44 +8,43 @@ Require Import Crypto.ModularArithmetic.PrimeFieldTheorems Crypto.ModularArithme
...
 Lemma two_lt_q : (2 < q)%Z. Proof. reflexivity. Qed.
-Lemma char_gt_2 : (1 + 1 <> (0:F q))%F. Proof. rewrite F_eq; compute; discriminate. Qed.
+Lemma char_gt_2 : (1 + 1 <> (0:F q))%F. vm_decide_no_check. Qed.
 
 Definition a : F q := opp 1%F.
-Lemma nonzero_a : a <> 0%F. Proof. rewrite F_eq; compute; discriminate. Qed.
+Lemma nonzero_a : a <> 0%F. Proof. vm_decide_no_check. Qed.
 Lemma square_a : exists b, (b*b=a)%F.
-Proof. setoid_rewrite <-F_pow_2_r. apply (minus1_is_square _); reflexivity. Qed.
+Proof. pose (@Zmod.Decidable_square q _ two_lt_q a); vm_decide_no_check. Qed.
```
andres-erbsen: @JasonGross this used to work before the `hprop`-based decidability stuff came in, and now the left hand side does not compute fully. Help me out?

JasonGross: Fixed by bcd963fb43b08f5154d4891845292524e2cad02e

jadephilipoom:
lgtm

src/Algebra.v
```diff
@@ -364,6 +357,23 @@ Module ScalarMult.
...
+      match n with
+      | O => zero
+      | S n' => add P (scalarmult_ref n' P)
+      end.
+
+    Global Instance Proper_scalarmult_ref : Proper (Logic.eq==>eq==>eq) scalarmult_ref.
+    Proof.
+      repeat intro; subst.
+      match goal with [n:nat |- _ ] => induction n; simpl @scalarmult_ref; [reflexivity|] end.
+      repeat match goal with [H:_ |- _ ] => rewrite H end; reflexivity.
```
JasonGross: An alternative from `Crypto.Util.Tactics`: `rewrite_hyp !*; reflexivity` (or `repeat rewrite_hyp *; reflexivity`)

src/Algebra.v
```diff
@@ -364,6 +357,23 @@ Module ScalarMult.
...
+    Qed.
+
+    Lemma scalarmult_ext : forall n P, mul n P = scalarmult_ref n P.
```
JasonGross: No `Proof.`?

src/Encoding/ModularWordEncodingTheorems.v
```diff
@@ -35,20 +35,13 @@ Section SignBit.
     rewrite sign_bit_parity; auto.
   Qed.
 
-  Lemma sign_bit_opp : forall (x : F m), x <> 0 -> negb (@sign_bit m sz x) = @sign_bit m sz (opp x).
+  Lemma odd_m : Z.odd m = true. Admitted.
```
JasonGross: There's no lemma that says that every prime is either 2 or odd?  (Or that if something is not odd, then it's divisible by 2, and that the only prime divisible by 2 is 2?)

src/Experiments/SpecEd25519.v
```diff
@@ -8,44 +8,43 @@ Require Import Crypto.ModularArithmetic.PrimeFieldTheorems Crypto.ModularArithme
 Require Import Crypto.Util.NatUtil Crypto.Util.ZUtil Crypto.Util.WordUtil Crypto.Util.NumTheoryUtil.
 Require Import Bedrock.Word.
 Require Import Crypto.Tactics.VerdiTactics.
-Require Import Coq.Logic.Decidable.
+Require Import Coq.Logic.Decidable Crypto.Util.Decidable.
 Require Import Coq.omega.Omega.
 
 (* TODO: move to PrimeFieldTheorems *)
-Lemma minus1_is_square {q} :  prime q -> (q mod 4)%Z = 1%Z -> isSquare (opp 1 : F q)%F.
+Lemma minus1_is_square {q} :  prime q -> (q mod 4)%Z = 1%Z -> (exists y, y*y  = opp (ZToField q 1))%F.
```
JasonGross: Nit: extra space before the `=`

JasonGross: Also after the `:`

src/Experiments/SpecEd25519.v
```diff
@@ -8,44 +8,43 @@ Require Import Crypto.ModularArithmetic.PrimeFieldTheorems Crypto.ModularArithme
...
 
 Definition a : F q := opp 1%F.
-Lemma nonzero_a : a <> 0%F. Proof. rewrite F_eq; compute; discriminate. Qed.
+Lemma nonzero_a : a <> 0%F. Proof. vm_decide_no_check. Qed.
 Lemma square_a : exists b, (b*b=a)%F.
-Proof. setoid_rewrite <-F_pow_2_r. apply (minus1_is_square _); reflexivity. Qed.
+Proof. pose (@Zmod.Decidable_square q _ two_lt_q a); vm_decide_no_check. Qed.
+Definition d : F q := (opp (ZToField _ 121665) / (ZToField _ 121666))%F.
+
+(* TODO: move to Decidable *)
```
JasonGross: Why not move it now?

src/Experiments/SpecEd25519.v
```diff
@@ -8,44 +8,43 @@ Require Import Crypto.ModularArithmetic.PrimeFieldTheorems Crypto.ModularArithme
...
+Lemma not_not P {d:Decidable P} : not (not P) <-> P.
+Proof. destruct (dec P); intuition. Qed.
```
JasonGross: Don't use `intuition` without importing `Util.FixCoqMistakes`

src/Experiments/SpecEd25519.v
```diff
@@ -8,44 +8,43 @@ Require Import Crypto.ModularArithmetic.PrimeFieldTheorems Crypto.ModularArithme
...
+  
+Global Instance dec_ex_forall_not T (P:T->Prop) {d:Decidable (exists b, P b)} : Decidable (forall b, ~ P b).
```
JasonGross: This too belongs in `Decidable.v`, right?

src/ModularArithmetic/ModularArithmeticTheorems.v
```diff
@@ -68,680 +34,326 @@ Ltac demod :=
...
-  Qed.
-
-  (* Redefine our division theory under the ring morphism *)
-  Lemma Fmorph_div_theory:
+  lazy iota beta delta [add sub mul opp FieldToZ ZToField proj1_sig] in *;
+  try apply eqsig_eq;
+  demod.
+
+(* FIXME: remove the pose proof once [monoid] no longer contains decidable equality *)
+Global Instance eq_dec {m} : DecidableRel (@eq (F m)). pose proof dec_eq_Z. exact _. Defined.
```
JasonGross: Maybe you want `eq_dec_F` or `F_dec_eq` as a name?

src/ModularArithmetic/ModularArithmeticTheorems.v
```diff
@@ -68,680 +34,326 @@ Ltac demod :=
...
+
+    Lemma square_iff (x:F m) :
+      (exists y : F m, y * y = x) <-> (exists y : Z, y * y mod m = x)%Z.
+    Proof.
+      setoid_rewrite eq_FieldToZ_iff; setoid_rewrite FieldToZ_mul; split; intro H; destruct H as [x' H].
+      - eauto.
+      - exists (ZToField _ x'); rewrite !FieldToZ_ZToField; demod; auto.
+    Qed.
+
+    (* TODO: move to ZUtil *)
```
JasonGross: Why not move now?

src/ModularArithmetic/ModularArithmeticTheorems.v
```diff
@@ -68,680 +34,326 @@ Ltac demod :=
...
+    Lemma sub_intersperse_modulus : forall x y, ((x - y) mod m = (x + (m - y)) mod m)%Z.
+    Proof.
+      intros.
+      replace (x + (m - y))%Z with (m+(x-y))%Z by omega.
```
JasonGross: Shorter proof: (if it works)
```coq
intros; push_Zmod; autorewrite with zsimplify; reflexivity
```
or else
```coq
intros; destruct (Z_zerop m); subst; push_Zmod; autorewrite with zsimplify; reflexivity.
```

Does one of these work?

src/ModularArithmetic/PrimeFieldTheorems.v
```diff
@@ -10,568 +10,172 @@ Require Import Coq.ZArith.BinInt Coq.NArith.BinNat Coq.ZArith.ZArith Coq.ZArith.
...
-    assert ((x / y) ^ 2 = x ^ 2 / y ^ 2) as square_distr_div by (field; trivial).
-    assert (y ^ 2 <> 0) as y2_nonzero by (
-      change (2%N) with (1+(1+0))%N;
-      rewrite !(proj2 (@F_pow_spec q _) _), !(proj1 (@F_pow_spec q _));
-      auto 10 using Fq_mul_nonzero_nonzero, Fq_1_neq_0).
-    rewrite (Fq_mul_eq _ z (y ^ 2)); auto.
-    rewrite <- A.
-    field; trivial.
-  Qed.
+    (* TODO: move to number thoery util *)
```
JasonGross: Why not move this now?

JasonGross:
Where's the extra argument to `ZToField` coming from?

JasonGross:
Modulo my concern with using `intuition`, I'm fine with this being merged.  Addressing the other comments I made would be nice, though.

src/ModularArithmetic/ModularArithmeticTheorems.v
andres-erbsen: Would moving it inside the module help?

andres-erbsen:
ZToField used to have its first argument (the target modulus) marked as implicit and maximally inserted. Yet it was often specified manually, and caused very slow type inference in some situations. Since now have notations for `1` and `0`, `ZToField` is used even less with an implicit argument. I could change it back if there is a good reason to keep the argument implicit.

src/Algebra.v
andres-erbsen: done.

src/Encoding/ModularWordEncodingTheorems.v
andres-erbsen: There is a lemma exactly like that. Fixed.

JasonGross:
Seems reasonable, no need to change it back.

src/ModularArithmetic/ModularArithmeticTheorems.v
andres-erbsen: This lemma was dead code, I deleted it.

src/ModularArithmetic/ModularArithmeticTheorems.v
andres-erbsen: I deleted `sub_intersperse_modulus`.

src/ModularArithmetic/PrimeFieldTheorems.v
andres-erbsen: moved.

andres-erbsen:
Should the name of the module be `Zmod` (which is also an alias for `Z.modulo`), `Znz` (which is also a standard library module), `F` (which is confusing when `m` is not prime and the structure is not a field) or something else?

src/ModularArithmetic/ModularArithmeticTheorems.v
JasonGross: Sure, if you want to stick it in a module that you generally don't import and use primarily for namespacing, that seems fine.

src/Experiments/SpecEd25519.v
andres-erbsen: fixed.

src/Experiments/SpecEd25519.v
andres-erbsen: moved.

src/Experiments/SpecEd25519.v
andres-erbsen: fixed.

src/Experiments/SpecEd25519.v
andres-erbsen: yep, moved

src/ModularArithmetic/ModularArithmeticTheorems.v
```diff
@@ -2,746 +2,346 @@ Require Import Coq.omega.Omega.
...
+  (* Fails iff the input term does some arithmetic with mod'd values. *)
+  Ltac notFancy E :=
+    match E with
+    | - (_ mod _) => idtac
+    | context[_ mod _] => fail 1
+    | _ => idtac
+    end.
+  
+  (* Remove redundant [mod] operations from the conclusion. *)
+  Ltac demod :=
```
JasonGross: This looks very similar to `pull_Zmod` in `Util.ZUtil`.  I'd suggest seeing if you can just use that.

JasonGross:
Regarding naming, I'd suggest `F`, given that everything is about applying `F` to things?  (I'd expect a namespace about `Z.modulo` to be part of a `ZUtil` hierarchy.)

src/ModularArithmetic/ModularArithmeticTheorems.v
JasonGross: Though you may have to adjust `has_no_mod` to have a case
```coq
  | -(?x' mod z) => has_no_mod x' z
```

JasonGross: Actually, better, I can just write a lemma to handle that case

src/ModularArithmetic/ModularArithmeticTheorems.v
```diff
@@ -2,746 +2,346 @@ Require Import Coq.omega.Omega.
...
+    Proof.
+      constructor; intros; unfold quotrem, id.
+
+      replace (Z.quotrem a b) with (Z.quot a b, Z.rem a b) by
+          try (unfold Z.quot, Z.rem; rewrite <- surjective_pairing; trivial).
+
+      unwrap_F; rewrite <-Z.quot_rem'; trivial.
+    Qed.
+
+    Lemma Z_mod_opp_equiv : forall x y m,  x  mod m = (-y) mod m ->
```
JasonGross: This belongs in ZUtil

src/ModularArithmetic/ModularArithmeticTheorems.v
```diff
@@ -2,746 +2,346 @@ Require Import Coq.omega.Omega.
...
+                                           (-x) mod m =   y  mod m.
+    Proof.
+      intros.
+      rewrite <-Z.sub_0_l.
+      rewrite Zminus_mod. rewrite H.
+      rewrite ?Zminus_mod_idemp_l, ?Zminus_mod_idemp_r; f_equal.
+      destruct y; auto.
+    Qed.
+
+    Lemma Z_opp_opp : forall x : Z, (-(-x)) = x.
```
JasonGross: Exists as `Z.opp_involutive`

src/ModularArithmetic/ModularArithmeticTheorems.v
```diff
@@ -2,746 +2,346 @@ Require Import Coq.omega.Omega.
...
+      destruct x; auto.
+    Qed.
+
+    Lemma Z_mod_opp : forall x m, (- x) mod m = (- (x mod m)) mod m.
```
JasonGross: As of 85dae05, this is `Z.opp_mod_mod` in `ZUtil`

JasonGross:
LGTM modulo travis.  Feel free to implement my suggestions and then merge, or merge as-is if it builds with 8.4 and 8.5.

andres-erbsen:
The naming question here is also open for `F` -- currently, the `F` type is not a field if it is given a non-prime parameter, `ModularArithmeticTheorems` does not assume `m` is a prime, `PrimeFieldTheorems` does.

src/ModularArithmetic/ModularArithmeticTheorems.v
andres-erbsen: This does not seem to be working out for me right now. Let's not do this in this PR. Later, though, I think this is a good idea.

JasonGross: Okay, maybe leave a TODO in a comment then?

src/Spec/ModularArithmetic.v
```diff
@@ -18,50 +20,45 @@ Infix "^" := BinInt.Z.pow : Z_scope.
 Infix "mod" := BinInt.Z.modulo (at level 40, no associativity) : Z_scope.
 Local Open Scope Z_scope.
 
-Definition F (modulus : BinInt.Z) := { z : BinInt.Z | z = z mod modulus }.
-Definition ZToField {m} (a:BinNums.Z) : F m := exist _ (a mod m) (Pre.Z_mod_mod a m).
-Coercion FieldToZ {m} (a:F m) : BinNums.Z := proj1_sig a.
+Module F.
+  Definition F (m : BinInt.Z) := { z : BinInt.Z | z = z mod m }.
+  Local Obligation Tactic := auto using Pre.Z_mod_mod.
```
JasonGross: In 8.6, and I think also 8.5, you need `cbv beta; auto using Pre.Z_mod_mod` here.  See https://coq.inria.fr/bugs/show_bug.cgi?id=4339

src/ModularArithmetic/ModularArithmeticTheorems.v
JasonGross: Oh, it's not working because `lia` can't prove that `m <> 0`.  I'll update the tactic I have to not need this assumption, I guess.  (Though tactics proving things about `_ mod 0` is perhaps slightly sketchy?)

JasonGross: I'll be pushing a commit momentarily that lets you replace `demod` with `pull_Zmod` and it seems to work.  (It also might cause a slight merge conflict because I had to remove a line in `ModularArithmeticTheorems` because `Fdefn` got more powerful because I added a few hints to `zarith`.

andres-erbsen: Thanks. I do think proving things about `_ mod 0` is sketchy, but I don't think adding the precondition to all lemmas about `mod` is not much better. Maybe one day `mod` and `div` can take in something like `{ z:Z | is_nonzero z = true }` as the second argument without it being a pain all over the place...

src/ModularArithmetic/ModularArithmeticTheorems.v
andres-erbsen: It was dead code, I threw it out.

src/ModularArithmetic/ModularArithmeticTheorems.v
andres-erbsen: Noted, lemma removed.

src/ModularArithmetic/ModularArithmeticTheorems.v
andres-erbsen: Thanks!

andres-erbsen:
I incorporated the latest feedback, merged manually, tested with 8.4pl6 and 8.5pl2 locally and pushed to master. If you have any further suggestions about this code, file an issue.

src/ModularArithmetic/ModularArithmeticTheorems.v
JasonGross: Much better would be for it to take  `positive` rather than `Z` or similar

PR #41, opened 2016-08-01
Add documentation: Equality, HProp, Isomorphism, Sigma

JasonGross:
As rightly requested by @andres-erbsen

### DIFF
--- a/src/Util/Equality.v
+++ b/src/Util/Equality.v
@@ -7,14 +7,17 @@
 Require Import Crypto.Util.Isomorphism.
 Require Import Crypto.Util.HProp.
 
-(** The J-rule ([eq_rect]) shows that [eq] forms a weak ∞-groupoid.
-    The first level of this groupoid is given by the identity
-    [eq_refl], the binary operation [eq_trans], and the inverse of [p
-    : x = y] given by [eq_sym p : y = x].  The second level, which we
-    provide here, says that [id ∘ p = p = p ∘ id], that [(p ∘ q)⁻¹ =
-    q⁻¹ ∘ p⁻¹], that [(p⁻¹)⁻¹ = p], and that [p ∘ p⁻¹ = p⁻¹ ∘ p = id].
-    We prove these here, as well as a few lemmas about functoriality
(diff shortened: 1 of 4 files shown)


src/Util/Isomorphism.v
```diff
@@ -1,3 +1,32 @@
...
+  forall x : A, r (s x) = x.
+
+Class IsEquiv {A B : Type} (f : A -> B) := BuildIsEquiv {
+  equiv_inv : B -> A ;
+  eisretr : Sect equiv_inv f;
+  eissect : Sect f equiv_inv;
+  eisadj : forall x : A, eisretr (f x) = ap f (eissect x)
+}.
+>> *)
+(** - Contratible fibers: [∀ y : B, Contr { x : A | f x = y }] where
```
andres-erbsen: Did you mean "contractible"?

src/Util/Equality.v
```diff
@@ -1,7 +1,20 @@
...
+(** * The Structure of the [eq] Type *)
+(** As described by the project of homotopy type theory, there is a
+    lot of structure inherent in the type of propositional equality,
+    [eq].  We build up enough lemmas about this structure to deal
+    nicely with proofs of equality that come up in practice in this
+    project. *)
 Require Import Crypto.Util.Isomorphism.
 Require Import Crypto.Util.HProp.
 
+(** The J-rule ([eq_rect]) shows that [eq] forms a weak ∞-groupoid.
```
andres-erbsen: This first line here is daunting because it has many fancy words in it. I can understand the paragraph, though.

andres-erbsen:
Much appreciated. LGTM as is, but please see whether you can do something about the commented things before you merge.

src/Util/Isomorphism.v
JasonGross: Yes, thanks

src/Util/Equality.v
JasonGross: Reworded to introduce known things first, and then use them to talk about unknown things, rather than the other way around.

PR #42, opened 2016-08-01
Define Montgomery reduction / multiplication on Z

JasonGross:
This is partly done for my own benefit, to internalize how Montgomery
multiplication works, and partly done as a template for word-based
Montgomery multiplication when the carrying does not take advantage of
the fact that we are using a pseudomersenne prime.

(@agl @mschilder123 @andres-erbsen @jadephilipoom) I'm looking for feedback on style, as well as correctness of transcription for ECC.

### DIFF
--- a/src/ModularArithmetic/Montgomery/Z.v
+++ b/src/ModularArithmetic/Montgomery/Z.v
@@ -1,0 +1,275 @@
+(*** Montgomery Multiplication *)
+(** This file implements Montgomery Form, Montgomery Reduction, and
+    Montgomery Multiplication on [Z].  We follow Wikipedia. *)
+Require Import Coq.ZArith.ZArith Coq.micromega.Psatz Coq.Structures.Equalities.
+Require Import Crypto.Util.ZUtil Crypto.Util.Tactics.
+Require Import Crypto.Util.Notations.
+
+Declare Module Nop : Nop.
+Module Import ImportEquivModuloInstances := Z.EquivModuloInstances Nop.
+
+Local Existing Instance eq_Reflexive. (* speed up setoid_rewrite as per https://coq.inria.fr/bugs/show_bug.cgi?id=4978 *)
+
+Local Open Scope Z_scope.
+Delimit Scope montgomery_naive_scope with montgomery_naive.
+Delimit Scope montgomery_scope with montgomery.
+Definition montgomeryZ := Z.
+Bind Scope montgomery_scope with montgomeryZ.
+
+Section montgomery.
+  Context (N : Z)
+          (N_reasonable : N <> 0)
+          (R : Z)
+          (R_good : Z.gcd N R = 1).
+  Local Notation "x ≡ y" := (Z.equiv_modulo N x y) : type_scope.
+  Local Notation "x ≡ᵣ y" := (Z.equiv_modulo R x y) : type_scope.
+  Context (R' : Z)
+          (R'_good : R * R' ≡ 1).
+  (** Quoting Wikipedia <https://en.wikipedia.org/wiki/Montgomery_modular_multiplication>: *)
+  (** In modular arithmetic computation, Montgomery modular
+      multiplication, more commonly referred to as Montgomery
+      multiplication, is a method for performing fast modular
+      multiplication, introduced in 1985 by the American mathematician
+      Peter L. Montgomery. *)
+  (** Given two integers [a] and [b], the classical modular
+      multiplication algorithm computes [ab mod N]. Montgomery
+      multiplication works by transforming [a] and [b] into a special
+      representation known as Montgomery form. For a modulus [N], the
+      Montgomery form of [a] is defined to be [aR mod N] for some
+      constant [R] depending only on [N] and the underlying computer
+      architecture. If [aR mod N] and [bR mod N] are the Montgomery
+      forms of [a] and [b], then their Montgomery product is [abR mod
+      N]. Montgomery multiplication is a fast algorithm to compute the
+      Montgomery product. Transforming the result out of Montgomery
+      form yields the classical modular product [ab mod N]. *)
+
+  Lemma R'_good' : R' * R ≡ 1.
+  Proof. rewrite <- R'_good; apply f_equal2; lia. Qed.
+
+  Definition to_montgomery_naive (x : Z) : montgomeryZ := x * R.
+  Definition from_montgomery_naive (x : montgomeryZ) : Z := x * R'.
+  Lemma to_from_montgomery_naive x : to_montgomery_naive (from_montgomery_naive x) ≡ x.
+  Proof.
+    unfold to_montgomery_naive, from_montgomery_naive.
+    rewrite <- Z.mul_assoc, R'_good'.
+    autorewrite with zsimplify; reflexivity.
+  Qed.
+  Lemma from_to_montgomery_naive x : from_montgomery_naive (to_montgomery_naive x) ≡ x.
+  Proof.
+    unfold to_montgomery_naive, from_montgomery_naive.
+    rewrite <- Z.mul_assoc, R'_good.
+    autorewrite with zsimplify; reflexivity.
+  Qed.
+
+  (** * Modular arithmetic and Montgomery form *)
+  Section general.
+    (** Addition and subtraction in Montgomery form are the same as
+        ordinary modular addition and subtraction because of the
+        distributive law: *)
+    (** [aR + bR = (a+b)R], *)
+    (** [aR - bR = (a-b)R]. *)
+    (** This is a consequence of the fact that, because [gcd(R, N) =
+        1], multiplication by [R] is an isomorphism on the additive
+        group [ℤ/Nℤ]. *)
+
+    Definition add : montgomeryZ -> montgomeryZ -> montgomeryZ := fun aR bR => aR + bR.
+    Definition sub : montgomeryZ -> montgomeryZ -> montgomeryZ := fun aR bR => aR - bR.
+    Local Infix "+" := add : montgomery_scope.
+    Local Infix "-" := sub : montgomery_scope.
+
+    Lemma add_correct_naive x y : from_montgomery_naive (x + y) = from_montgomery_naive x + from_montgomery_naive y.
+    Proof. unfold from_montgomery_naive, add; lia. Qed.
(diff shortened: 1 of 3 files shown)


src/ModularArithmetic/Montgomery/Z.v
```diff
@@ -0,0 +1,275 @@
...
+      rewrite <- !Z.mul_assoc, R'_good.
+      autorewrite with zsimplify; apply (f_equal2 Z.modulo); lia.
+    Qed.
+  End general.
+
+  (** * The REDC algorithm *)
+  Section redc.
+    (** While the above algorithm is correct, it is slower than
+        multiplication in the standard representation because of the
+        need to multiply by [R′] and divide by [N]. Montgomery
```
andres-erbsen: divide? Are you assuming `(fun x => x mod m)` is implemented using division?

src/ModularArithmetic/Montgomery/Z.v
```diff
@@ -0,0 +1,275 @@
...
+    Section redc.
+      Context (T : Z)
+              (T_in_range : 0 <= T < R * N).
+
+      Lemma N'_good' : N' * N ≡ᵣ -1.
+      Proof. rewrite <- N'_good; apply f_equal2; lia. Qed.
+
+      Let m := ((T mod R) * N') mod R.
+      Let t := (T + m * N) / R.
+      Definition reduce : montgomeryZ
```
andres-erbsen: It is quite hard to tell that reduce will have only one argument. Not sure whether we care.

src/ModularArithmetic/Montgomery/Z.v
JasonGross: Is there a faster way for `m` not a power of two and `x` large?  Maybe I should insert more instances of "I'm [quoting Wikipedia](https://en.wikipedia.org/wiki/Montgomery_modular_multiplication#The_REDC_algorithm)"...

andres-erbsen:
I haven't implemented Montgomery reduction before, so I can't judge how closely this lines up with the code you would want to have when using a polynomial representation of Z.  I think I would skim through https://golang.org/src/math/big/nat.go#L199 and https://golang.org/src/math/big/nat.go#L1056 (which have helpful comments) and maybe https://github.com/libtom/libtommath/blob/master/bn_mp_montgomery_reduce.c if I wanted to line this PR up with a polynomial-representation implementation, but I won't do this now.

Does this PR mean that you will not be implementing polynomial-based Barrett reduction? The motivation on my part here is that I was hoping I could reuse the generic code for polynomial-representation Barrett reduction for the Ed25519 exponent scalars. In an ideal word we could have a somewhat generic recipe for changing between representations of Z, so it might be worth thinking about how different the `Barrett reduction on Z -> Barrett reduction on polynomials in Z` and `Montgomery reduction reduction on Z -> Montgomery reduction on polynomials in Z` transformations need to look like anyway.

src/ModularArithmetic/Montgomery/Z.v
andres-erbsen: Nah, division is a very reasonable implementation. The edit I would have suggested would be "divide by [N]" -> "reduce modulo [N]", but reading this the third time now, I don't find it confusing anymore, so maybe nevermind.

JasonGross:
> so I can't judge how closely this lines up with the code you would want to have when using a polynomial representation of Z.

I plan to look at that next; I think it's better to have this version, and look at the polynomial version after, rather than trying to do them simultaneously.  [Wikipedia also has a section on the polynomial version](https://en.wikipedia.org/wiki/Montgomery_modular_multiplication#Montgomery_arithmetic_on_multiprecision_integers)

> Does this PR mean that you will not be implementing polynomial-based Barrett reduction? 

It turns out that some of the optimizations necessary for doing Barrett-reduction efficiently (namely, [computing `q = (m * (a >> (k-1))) >> (k+1)` rather than `q = (m * a) >> k`]( https://github.com/JasonGross/fiat-crypto/blob/wip-barrett-modular/src/ModularArithmetic/BarrettReduction/ZFuzzyPaper.v#L72) and only works for 25519 (`k = 255`) by luck, and may or may not work for P256; this optimization only guarantees that `a - q * n < 3 * n`, and a brute-force search of around 700 cases shows that `a - q * n < 2 * n` for 25519.  (P256 requires just over `2^256 - 2^254` cases using the method that works for 25519, so I'm still not sure if it works.)  So I probably won't be implementing polynomial-based Barrett reduction, but I'm hopeful that I'll be able to port the proofs from integer-based Montgomery to polynomial-based Montgomery without much code, if not derive the algorithm itself, and I'm hopeful that the transformation will be similar for (unoptimized) Barrett reduction.  I might even include this as a test-case that the machinery is modular enough.

andres-erbsen:
I assume your experiments were for the Curve25519 underlying field `F (2^255-19)`, not the exponent field `F (2^252 +  27742317777372353535851937790883648493)` ? I would want code for the latter. However, the corresponding Barrett reduction in `ed25519-ref` at https://github.com/floodyberry/supercop/blob/master/crypto_sign/ed25519/ref/sc25519.c#L88 performs two conditional subtractions, as HAC 14.44(ii) says, and so does https://github.com/floodyberry/supercop/blob/master/crypto_sign/ed25519/amd64-51-30k/sc25519_barrett.s#L1029.

src/ModularArithmetic/Montgomery/Z.v
andres-erbsen: No. I think there is value in keeping top-level specification as small as possible, but Montgomery reduction isn't that. I can't currently think of a case where I would want the definitions but not the proofs.

JasonGross:
> `F (2^255-19)`

Yes, that's correct.

> performs two conditional subtractions

Yes, this is safe regardless of the modulus, as long as it's sufficiently within bounds.  I haven't pushed through / cleaned up the code that shows this (most of the content is there, excepting the part that shows that we're not negative), but I could probably do so (for `Z`) in a few hours to a day of work (unless it turns out that you can get something negative, and that something special has to be done in this case).  Would that be useful to you?

andres-erbsen:
Yes.

JasonGross:
Okay, let me finish cleaning this up (I'm going to separate out the proofs and add a few alternate bounds), and then I'll add the `(m * (a >> (k-1))) >> (k+1)` version of Barrett.  (Do you have a good name for it?  I've been calling it `ZPaper.v` because it's what's in [Barrett's original paper](http://download.springer.com/static/pdf/905/chp%253A10.1007%252F3-540-47721-7_24.pdf?originUrl=http%3A%2F%2Flink.springer.com%2Fchapter%2F10.1007%2F3-540-47721-7_24&token2=exp=1470169156~acl=%2Fstatic%2Fpdf%2F905%2Fchp%25253A10.1007%25252F3-540-47721-7_24.pdf%3ForiginUrl%3Dhttp%253A%252F%252Flink.springer.com%252Fchapter%252F10.1007%252F3-540-47721-7_24*~hmac=60cfce7831df48d48917f136b334cf6c56c28ecbc5de27325a2de739c1a48775).)

src/ModularArithmetic/Montgomery/Z.v
JasonGross: I think I want to separate out the proofs; @mschilder123 suggested a bunch of alternate bounds one might want, and the file is getting a bit unwieldy to read at this point.

JasonGross:
@andres-erbsen I've moved the proofs to `ZProofs.v` and added a number of alternate bounds.

JasonGross:
> `F (2^252 + 27742317777372353535851937790883648493)`

Do you take `k = 253` here?  If not, I think you can be off by a lot more than 2.  In particular, I need that `a mod 2^(k-1) < n` for `a` the input and `n` the modulus.

> https://github.com/floodyberry/supercop/blob/master/crypto_sign/ed25519/ref/sc25519.c#L88
```c
  /* XXX: Can it really happen that r<0?, See HAC, Alg 14.42, Step 3 
   * If so: Handle  it here!
   */
```

Assuming they're implementing the same algorithm we are, the answer here is "no".

andres-erbsen:
Let's move the Barrett reduction discussion to https://github.com/mit-plv/fiat-crypto/issues/43 and leave this ticket for Montgomery form stuff.

src/ModularArithmetic/Montgomery/Z.v
```diff
@@ -0,0 +1,158 @@
...
+    (** Computing products in Montgomery form requires removing the
+        extra factor of [R]. While division by [R] is cheap, the
+        intermediate product [(aR mod N)(bR mod N)] is not divisible
+        by [R] because the modulo operation has destroyed that
+        property. *)
+    (** Removing the extra factor of R can be done by multiplying by
+        an integer [R′] such that [RR' ≡ 1 (mod N)], that is, by an
+        [R′] whose residue class is the modular inverse of [R] mod
+        [N]. Then, working modulo [N], *)
+    (** [(aR mod N)(bR mod N)R' ≡ (aR)(bR)R⁻¹ ≡ (ab)R  (mod N)]. *)
```
jadephilipoom: I think the R⁻¹ is intended to be R' here. If not, it should be explained somehow.

src/ModularArithmetic/Montgomery/Z.v
```diff
@@ -0,0 +1,158 @@
...
+    m ← ((T mod R)N′) mod R
+    t ← (T + mN) / R
+    if t ≥ N then
+        return t − N
+    else
+        return t
+    end if
+end function
+>> *)
+    Context (N' : Z). (* N' is N⁻¹ mod R *)
```
jadephilipoom: Above on line 89 it says "Integer `N′` in `[0, R − 1]` such that `NN′ ≡ −1 mod R`", so shouldn't `N'` be `-N⁻¹ mod R`?

src/ModularArithmetic/Montgomery/Z.v
JasonGross: This is literally a quotation from Wikipedia.  I think they mean `(aR)(bR) / R`, but just wanted to write that without the fraction.

src/ModularArithmetic/Montgomery/Z.v
JasonGross: Good catch.  Fixed.

jadephilipoom:
Looks good.

Issue #43, opened 2016-08-03
Implement barrett reduction for ed25519 exponent field


andres-erbsen:
modulus `m = 2^252 +  27742317777372353535851937790883648493`, length `k = 256`, (precomputed `mu = 1852673427797059126777135760139006525645217721299241702126143248052143860224795`,) input `a < 2^512`. As in https://github.com/floodyberry/supercop/blob/master/crypto_sign/ed25519/ref/sc25519.c#L41.

andres-erbsen:
Or maybe we are mistaken to assume base `b=2`. Following HAC 14.42 as the code suggests, `b=256`, `k=32` would make sense as well, and satisfy `2^(k-1) < n < 2^k`.

JasonGross:
Oh, hrm, interesting.  I don't think I have anything that relies on `b = 2`, so it should be easy to generalize to any `b > 1`.  The computation is `q = (mu * (a >> (256 - 32)) >> (256 + 32)`?

andres-erbsen:
The guess I have with `b=256=2^8` would give `256^(32-1) = 2^(256-8)` and `256^(32+1) = 2^(256+8)`, so `q = (mu * (a >> (256 - 8)) >> (256 + 8)`.

JasonGross:
Ugh, why are there so many variations on Barrett Reduction.  Not only does this version split apart the shifts, it also does reduction modulo `b^(k+1)` early (which might be where the possibility of negative values comes from).

JasonGross:
@andres-erbsen Is it acceptable to replace

> (3) If `r < 0` then `r ← r + bᵏ⁺¹`

with

(3) `r ← r mod bᵏ⁺¹`

?  I think this would make the proofs much easier, and it seems like, by choosing `k` and `b` carefully with respect to the machine word size, you get this for free.

JasonGross:
What needs to be done to complete this, on top of #69?

JasonGross:
Closed by #77.

PR #44, opened 2016-08-03
Add a generalized version of Barrett Reduction

JasonGross:
In this version, we split up the integer division so that we are less
likely to overflow in intermediate computations.

This is still not the version in HAC 14.42; that version also does early
reduction modulo b^(k+1).

This is work towards #43

### DIFF
--- a/src/ModularArithmetic/BarrettReduction/ZGeneralized.v
+++ b/src/ModularArithmetic/BarrettReduction/ZGeneralized.v
@@ -1,5 +1,13 @@
 (*** Barrett Reduction *)
-(** This file implements a slightly-generalized version of Barrett Reduction on [Z]. *)
+(** This file implements a slightly-generalized version of Barrett
+    Reduction on [Z].  This version follows a middle path between the
+    Handbook of Applied Cryptography (Algorithm 14.42) and Wikipedia.
+    We split up the shifting and the multiplication so that we don't
+    need to store numbers that are quite so large, but we don't do
+    early reduction modulo [b^(k+offset)] (we generalize from HAC's [k
+    ± 1] to [k ± offset]).  This leads to weaker conditions on the
+    base ([b]), exponent ([k]), and the [offset] than those given in
+    the HAC. *)
 Require Import Coq.ZArith.ZArith Coq.micromega.Psatz.
 Require Import Crypto.Util.ZUtil Crypto.Util.Tactics Crypto.Algebra.
 
@@ -62,7 +70,11 @@ Section barrett.
     Context (n_pos : 0 < n) (* or just [0 <= n], since we have [n <> 0] above *)
             (a_nonneg : 0 <= a).
 
-    Context (k_big_enough : offset <= k).
+    Context (k_big_enough : offset <= k)
+            (a_small : a < b^(2*k))
+            (** We also need that [n] is large enough; [n] larger than
+                [bᵏ⁻¹] works, but we ask for something more precise. *)
+            (n_large : a mod b^(k-offset) <= n).
 
     (** Now *)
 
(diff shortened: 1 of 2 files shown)


src/ModularArithmetic/BarrettReduction/ZGeneralized.v
```diff
@@ -0,0 +1,136 @@
...
+      break_match; auto with lia.
+    Qed.
+
+    (** In that case, we have *)
+    Theorem barrett_reduction_small
+      : a mod n = if r <? n
+                  then r
+                  else if r <? 2 * n
+                       then r - n
+                       else r - 2 * n.
```
andres-erbsen: The following seems slightly more natural to me (and it would match implementations that use conditional subtraction):

    let r := if r<?n then r else r-n in
    let r := if r<?n then r else r-n in
    r

(the current form is fine as well)

src/ModularArithmetic/BarrettReduction/ZGeneralized.v
```diff
@@ -0,0 +1,136 @@
+(*** Barrett Reduction *)
```
andres-erbsen: A comment explaining the differences of the two Barrett reduction variants would be appreciated.

src/ModularArithmetic/BarrettReduction/ZGeneralized.v
```diff
@@ -0,0 +1,136 @@
...
+      Z.simplify_fractions_le.
+      autorewrite with pull_Zpow pull_Zdiv zsimplify; reflexivity.
+    Qed.
+
+    (** Also, if [a < n²] then [r < 2n]. *)
+    (** N.B. It turns out that it is sufficient to assume [a < b²ᵏ]. *)
+    Context (a_small : a < b^(2*k)).
+    (** We also need that [n] is large enough; [n] larger than bᵏ⁻¹
+        works, but we ask for something more precise. *)
+    Context (n_large : a mod b^(k-offset) <= n).
```
andres-erbsen: Is there a reason we don't declare these requirements next to the other ones?

JasonGross: They're not needed to prove that `q * n` is small, nor to prove that the reduction is correct modulo `n`, only that it is in the relevant range.  If you think it's better, I can move these up to the rest of the hypotheses, or I can move each one that's up there down to just above the first theorem that needs it.  What do you think?

andres-erbsen: I think collecting all preconditions in the same place is easier on the reader. I don't foresee using this code in a way that does not depend on the output being small.

andres-erbsen:
LGTM as is, see suggestions.

src/ModularArithmetic/BarrettReduction/ZGeneralized.v
JasonGross: Changed (I added a few spaces, though)

src/ModularArithmetic/BarrettReduction/ZGeneralized.v
andres-erbsen: If you really think a precondition can be annoying to satisfy, and only some lemma depends on it, add it to that lemma -- not context.

src/ModularArithmetic/BarrettReduction/ZGeneralized.v
JasonGross: Done

src/ModularArithmetic/BarrettReduction/ZGeneralized.v
JasonGross: Moved

JasonGross:
Builds locally in 8.4pl1 and 8.6, so I'm going to merge it.

PR #45, opened 2016-08-04
Implement Barrett Reduction following HAC 14.42

JasonGross:
From http://cacr.uwaterloo.ca/hac/about/chap14.pdf

This should take care of most of #43, at least the specification on Z
part of it.

### DIFF
--- a/src/ModularArithmetic/BarrettReduction/ZHandbook.v
+++ b/src/ModularArithmetic/BarrettReduction/ZHandbook.v
@@ -128,14 +128,12 @@ Section barrett.
     Qed.
 
     Theorem barrett_reduction_small
-      : x mod m = if r <? m
-                  then r
-                  else if r <? 2 * m
-                       then r - m
-                       else r - 2 * m.
+      : x mod m = let r := if r <? m then r else r-m in
+                  let r := if r <? m then r else r-m in
+                  r.
     Proof.
-      pose proof r_small.
-      destruct (r <? m) eqn:?, (r <? 2 * m) eqn:?; Z.ltb_to_lt; rewrite !r_eq_alt' in *; try lia.
+      pose proof r_small. cbv zeta.
+      destruct (r <? m) eqn:Hr, (r-m <? m) eqn:?; rewrite !r_eq_alt', ?Hr in *; Z.ltb_to_lt; try lia.
       { symmetry; eapply (Zmod_unique x m q3); lia. }
       { symmetry; eapply (Zmod_unique x m (q3 + 1)); lia. }
       { symmetry; eapply (Zmod_unique x m (q3 + 2)); lia. }
--- a/src/ModularArithmetic/BarrettReduction/ZHandbook.v
+++ b/src/ModularArithmetic/BarrettReduction/ZHandbook.v
@@ -71,7 +71,7 @@ Section barrett.
       assert (0 < b^(k-offset)) by zero_bounds.
       assert (x / b^(k-offset) <= b^(2*k) / b^(k-offset)) by auto with zarith lia.
       assert (x / b^(k-offset) <= b^(k+offset)) by (autorewrite with pull_Zpow zsimplify in *; assumption).
-      subst q3 q2 q1 μ Q.
+      subst q1 q2 q3 Q r r1 r2 R μ.
       rewrite (Z.div_mul_diff_exact' (b^(2*k)) m (x/b^(k-offset))) by auto with lia zero_bounds.
       rewrite (Z_div_mod_eq (_ * b^(2*k) / m) (b^(k+offset))) by lia.
       autorewrite with push_Zmul push_Zopp zsimplify zstrip_div zdiv_to_mod.
--- a/src/ModularArithmetic/BarrettReduction/ZHandbook.v
+++ b/src/ModularArithmetic/BarrettReduction/ZHandbook.v
@@ -48,17 +48,23 @@ Section barrett.
 
     Let q1 := x / b^(k-offset). Let q2 := q1 * μ. Let q3 := q2 / b^(k+offset).
     Let r1 := x mod b^(k+offset). Let r2 := (q3 * m) mod b^(k+offset).
-    Let r := let r := r1 - r2 in
-             if r <? 0 then r + b^(k+offset) else r.
+    (** At this point, the HAC says "If [r < 0] then [r ← r + bᵏ⁺¹]".
+        This is equivalent to reduction modulo [b^(k+offset)], as we
+        prove below.  The version involving modular reduction has the
+        benefit of being cheaper to implement, and making the proofs
+        simpler, so we primarily use that version. *)
+    Let r_mod_3m      := (r1 - r2) mod b^(k+offset).
+    Let r_mod_3m_orig := let r := r1 - r2 in
+                         if r <? 0 then r + b^(k+offset) else r.
 
-    Lemma r_eq_alt : r = (r1 - r2) mod b^(k+offset).
+    Lemma r_mod_3m_eq_orig : r_mod_3m = r_mod_3m_orig.
     Proof.
       assert (0 <= r1 < b^(k+offset)) by (subst r1; auto with zarith).
       assert (0 <= r2 < b^(k+offset)) by (subst r2; auto with zarith).
-      subst r; cbv zeta.
+      subst r_mod_3m r_mod_3m_orig; cbv zeta.
       break_match; Z.ltb_to_lt.
-      { apply (Zmod_unique (r1 - r2) _ (-1)); lia. }
-      { apply (Zmod_unique (r1 - r2) _ 0); lia. }
+      { symmetry; apply (Zmod_unique (r1 - r2) _ (-1)); lia. }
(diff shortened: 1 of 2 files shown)


src/ModularArithmetic/BarrettReduction/ZHandbook.v
```diff
@@ -0,0 +1,144 @@
+(*** Barrett Reduction *)
```
andres-erbsen: A comment explaining the differences of the two Barrett reduction variants would be appreciated.

src/ModularArithmetic/BarrettReduction/ZHandbook.v
```diff
@@ -0,0 +1,144 @@
...
+            (m_pos : 0 < m)
+            (base_pos : 0 < b)
+            (k_good : m < b^k)
+            (μ_good : μ = b^(2*k) / m) (* [/] is [Z.div], which is truncated *)
+            (x_nonneg : 0 <= x)
+            (offset_nonneg : 0 <= offset)
+            (k_big_enough : offset <= k).
+    (** N.B. We generalize to [k ± offset] from [k ± 1]. *)
+    (** INPUT: positive integers [x = (x₂ₖ₋₁···x₁x₀)_b], [m =
+        (mₖ₋₁···m₁m₀)_b] (with [mₖ₋₁ ≠ 0]), and [µ = ⌊b²ᵏ/m⌋. *)
```
andres-erbsen: Since this file does not deal with representations of integers, the "INPUT" line is currently confusing. Since the meaning of `x` and `m` is already explained above, I think this comment can be removed.

src/ModularArithmetic/BarrettReduction/ZHandbook.v
```diff
@@ -0,0 +1,144 @@
...
+    (** OUTPUT: [r = x mod m] *)
```
andres-erbsen: The "OUTPUT" line is redundant as well

src/ModularArithmetic/BarrettReduction/ZHandbook.v
```diff
@@ -0,0 +1,144 @@
...
+    (** 1. [q₁ ← ⌊x/bᵏ⁻¹⌋], [q₂ ← q₁ · µ], [q₃ ← ⌊q₂/bᵏ⁺¹⌋].
+        2. [r₁ ← x mod bᵏ⁺¹], [r₂ ← q₃ · m mod bᵏ⁺¹], [r ← r₁ - r₂].
+        3. If [r < 0] then [r ← r + bᵏ⁺¹].
+        4. While [r ≥ m] do: [r ← r - m].
+        5. Return([r]). *)
```
andres-erbsen: This comment duplicates the code right below it, having the potential failure mode of causing infinite confusion if they get out of sync. IMHO, the Coq code below is clear enough and the comment can be removed.

In fact, even right now we have a comment that contains 5 steps above a paragraph of code that performs only the first 4. If you really want to match HAC, you should include step 5 in the same code block. If you need a creative name for what is currently called `r` (but is not equal to `x mod m` as claimed in the introductory paragraph), I would use `r_mod_3m`.

src/ModularArithmetic/BarrettReduction/ZHandbook.v
```diff
@@ -0,0 +1,144 @@
...
+
+    Let q1 := x / b^(k-offset). Let q2 := q1 * μ. Let q3 := q2 / b^(k+offset).
+    Let r1 := x mod b^(k+offset). Let r2 := (q3 * m) mod b^(k+offset).
+    Let r := let r := r1 - r2 in
+             if r <? 0 then r + b^(k+offset) else r.
+
+    Lemma r_eq_alt : r = (r1 - r2) mod b^(k+offset).
```
andres-erbsen: I would use the RHS here as the definition because it is both what you need to make the proof work AND it is likely to yield faster code, but if you do not want to deviate from HAC, that's fine.

src/ModularArithmetic/BarrettReduction/ZHandbook.v
```diff
@@ -0,0 +1,144 @@
...
+      break_match; Z.ltb_to_lt.
+      { apply (Zmod_unique (r1 - r2) _ (-1)); lia. }
+      { apply (Zmod_unique (r1 - r2) _ 0); lia. }
+    Qed.
+
+    Context (x_small : x < b^(2*k)).
+    (** We also need that [m] is large enough; [m] larger than bᵏ⁻¹
+        works, but we ask for something more precise. *)
+    Context (m_large : x mod b^(k-offset) <= m)
+            (m_small : 3 * m <= b^(k+offset)).
```
andres-erbsen: Is there a reason we don't declare these requirements above next to all other context variables?

src/ModularArithmetic/BarrettReduction/ZHandbook.v
```diff
@@ -0,0 +1,144 @@
...
+      rewrite r_eq_alt'.
+      autorewrite with zsimplify push_Zmod; reflexivity.
+    Qed.
+
+    (** (ii) In step 2, observe that [-bᵏ⁺¹ < r₁ - r₂ < bᵏ⁺¹], [r₁ -
+             r₂ ≡ (Q - q₃)m + R (mod bᵏ⁺¹)], and [0 ≤ (Q - q₃)m + R <
+             3m < bᵏ⁺¹] since [m < bᵏ] and [3 < b]. If [r₁ - r₂ ≥ 0], then [r₁
+             - r₂ = (Q - q₃)m + R]. If [r₁ - r₂ < 0], then [r₁ - r₂ + bᵏ⁺¹
+             = (Q - q₃)m + R]. In either case, step 4 is repeated at
+             most twice since [0 ≤ r < 3 m] *)
```
andres-erbsen: This comment references assumption `3 < b` that is not declared in this context.

Referencing step 4 is also confusing because our code does not do a while loop (and it is a good idea not to).

All in all, I am not sure whether it adds much to the discussion. If you feel like we need to explain more, a comment that actually talks about the code Coq knows about would be much preferred.

src/ModularArithmetic/BarrettReduction/ZHandbook.v
```diff
@@ -0,0 +1,144 @@
...
+      rewrite r_eq_alt; subst Q R r1 r2.
+      autorewrite with pull_Zmod zsimplify; lia.
+    Qed.
+
+    Theorem barrett_reduction_small
+      : x mod m = if r <? m
+                  then r
+                  else if r <? 2 * m
+                       then r - m
+                       else r - 2 * m.
```
andres-erbsen: The following seems slightly more natural to me (and it would match implementations that use conditional subtraction):

    let r := if r<?n then r else r-n in
    let r := if r<?n then r else r-n in
    r

(the current form is fine as well)

andres-erbsen:
I am excited to have this code in the repository.

While I think following an established reference is a good idea, I do not see repeating HAC to the letter as a worthy goal. HAC is 20 years old now, thus, for example, it does not worry about timing attacks like we do today. It also expresses algorithms in a purely imperative setting, which does not fit that well with Coq and can lead to particularly confusing translations. Let's look out for that.

src/ModularArithmetic/BarrettReduction/ZHandbook.v
JasonGross: Done.  (I'll add the comment in the other file soon.)

src/ModularArithmetic/BarrettReduction/ZHandbook.v
JasonGross: I've updated the code to use primarily the RHS, but also include theorems for the LHS version.

src/ModularArithmetic/BarrettReduction/ZHandbook.v
JasonGross: Done

src/ModularArithmetic/BarrettReduction/ZHandbook.v
JasonGross: Moved

src/ModularArithmetic/BarrettReduction/ZHandbook.v
JasonGross: Comment removed

src/ModularArithmetic/BarrettReduction/ZHandbook.v
JasonGross: Changed

src/ModularArithmetic/BarrettReduction/ZHandbook.v
JasonGross: Comment removed

src/ModularArithmetic/BarrettReduction/ZHandbook.v
JasonGross: Removed

andres-erbsen:
LGTM assuming the build passes.

JasonGross:
Merging, because this builds in 8.6 and 8.4pl1 locally.

PR #46, opened 2016-08-04
Test fewer Coq versions on Travis

JasonGross:
Each version of 8.4 takes around 17 minutes to complete.  I've yet to see anything that behaves differently among 8.4{,pl1,pl2,pl3,pl4}, or among 8.4pl{5,6}.  So if we keep 8.4pl3 (the version of Coq distributed in debian trusty), and 8.4pl6, we should capture the range of behaviors, and save about an hour-and-a-quarter per test.


andres-erbsen:
I'm all for it. Do we expect that the commented lines will stop working due to repository-rot or will they still be functional whenever we decide to re-enable some of them?

JasonGross:
I expect they should keep working (the ppa is mine), at least until we move to 8.5-only.

PR #47, opened 2016-08-07
Remove unnecessary pseudo notations

varomodt:
@JasonGross sorry it took me so long to fix this


JasonGross:
Want to remove the relevant `Reserved Notation` bits in Util/Notations.v at the same time?

Other than that, LGTM

varomodt:
Done. Waiting on Jenkins

Le dim. 7 août 2016 à 15:03, Jason Gross <notifications@github.com> a
écrit :

PR #49, opened 2016-08-09
Specify a type of bounded integers for modular arithmetic

JasonGross:
Also use it to implement Montgomery reduction and Barrett reduction.

The `ZLikeOps` type is a collection of the operations needed for Montgomery and Barrett in (roughly) the machine code I'm targeting.  It's probable there's a better specification for the base language here.  Feedback on this point welcome.

Note that, once the initial bounds are established, if the algorithm is in the right form (does modular reduction in all the right places to make it obvious that nothing goes out of bounds), then it's very simple to convert from operations on `Z` to operations on this type.

Next up is instantiating this type with digits / a `limb_width`-based representation.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -50,10 +50,13 @@ src/ModularArithmetic/PrimeFieldTheorems.v
 src/ModularArithmetic/PseudoMersenneBaseParamProofs.v
 src/ModularArithmetic/PseudoMersenneBaseParams.v
 src/ModularArithmetic/Tutorial.v
+src/ModularArithmetic/ZBounded.v
 src/ModularArithmetic/BarrettReduction/Z.v
+src/ModularArithmetic/BarrettReduction/ZBounded.v
 src/ModularArithmetic/BarrettReduction/ZGeneralized.v
 src/ModularArithmetic/BarrettReduction/ZHandbook.v
 src/ModularArithmetic/Montgomery/Z.v
+src/ModularArithmetic/Montgomery/ZBounded.v
 src/ModularArithmetic/Montgomery/ZProofs.v
 src/Spec/CompleteEdwardsCurve.v
 src/Spec/EdDSA.v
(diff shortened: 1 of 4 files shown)


JasonGross:
@jadephilipoom How does this compare to the abstraction you're implementing with `BaseSystem` / `ModularBaseSystem`?

andres-erbsen:
I am not sure whether it is worth trying to share this code between Montgomery and Barrett reduction. The `large`, `medium`, `small` classification seems quite arbitrary and naively generalizing this would explode the number of classes proportionally to how many different sequences of operations we can perform until the representation becomes invalid. I wonder whether "almost montgomery reduction" (from https://eprint.iacr.org/2011/239.pdf) and "non-reduced montgomery reduction" (code https://rt.openssl.org/Ticket/Display.html?id=3054&user=guest&pass=guest, paywalled paper "Software Implementation of Modular Exponentiation, Using Advanced Vector Instructions Architectures") fit this paradigm, but I would guess not because the only redundant representation we have (for pseudo-Mersenne) does not. I don't know what the correct abstraction would be though, the `Zlike` implementations really are designed to the specification "behaves like Z if you use this in the algorithm it was designed for"...

JasonGross:
> paywalled paper

Do you know about the .libproxy.mit.edu "trick"?  http://link.springer.com.libproxy.mit.edu/chapter/10.1007%2F978-3-642-31662-3_9#page-1 gives access for anyone with MIT credentials.

> "almost montgomery reduction"

Oh, hey, that's the alternate version Marius told me about ([`partial_reduce`](https://github.com/mit-plv/fiat-crypto/blob/master/src/ModularArithmetic/Montgomery/Z.v#L110) and [`partial_reduce_alt`](https://github.com/mit-plv/fiat-crypto/blob/master/src/ModularArithmetic/Montgomery/Z.v#L116) that I actually synthesize code for here.  The standard version is actually trickier to get code for, because you have to somehow deal with the case where the addition overflows, and there's no way to do that written into the code.  (On the other hand, the version that first does almost-montgomery-reduction and then does a final reduce should be straightforward; it just gets an extra conditional subtract at the end.)

> "non-reduced montgomery reduction"

Not sure what's going on here with vector instructions, but the algorithm on `Z` seems to be basically the same as that of almost montgomery reduction.

>  naively generalizing this would explode the number of classes

We could just index the types over a `nat` standing in for roughly `length limb_widths`...  I'm inclined to hold off on this for now, though.

andres-erbsen:
I'm not sure of what you mean by "We could just index the types over a nat standing in for roughly length limb_widths", but I agree that it is not blocking for this pull request. I guess the main takeaway here is that representations whose use cases are restricted in complicated ways are just inherently harder to work with.

PR #52, opened 2016-08-11
Initial work on an architecture interface for ℤ/nℤ

JasonGross:
This provides a cleaner interface for the bottom level implementation,
as well as an implementation of multiplying 128x128 -> 256.

This is very rough, right now, but I want to solicit feedback on it before I start work on the proofs.

In particular:
- [x] Is the interface at the bottom reasonable?  It's missing conditional assignment, but that's kind-of separate, anyway.
- [x] There are almost certainly better names.  Any suggestions?
- [ ] Should this go in its own folder, or in ModularArithmetic
- [x] The code for ShiftLeft and ShiftRight are complicated.  Is there a standard way to do this?
- [ ] I lifted the code for `DoubleArchitectureBoundedFullMulOpsOfHalfWidthMulOps` from @mschilder123's code; does it generalize to ripple-carry multiplication?
- [x] Should I move all the code to proper `Definition`s, rather than sticking it inline in the `Instance`s?

Any other thoughts / comments?


src/BoundedArithmetic/DoubleBounded.v
```diff
@@ -0,0 +1,114 @@
...
+Require Import Crypto.Util.Notations.
+
+Import ListNotations.
+Local Open Scope list_scope.
+Local Open Scope nat_scope.
+Local Open Scope Z_scope.
+Local Open Scope type_scope.
+
+Local Coercion Z.of_nat : nat >-> Z.
+Local Notation eta x := (fst x, snd x).
```
andres-erbsen: Why is this a good idea?

src/BoundedArithmetic/DoubleBounded.v
```diff
@@ -0,0 +1,114 @@
...
+
+Section with_single_ops.
+  Section single.
+    Context (n : size) {base_ops : ArchitectureBoundedOps n}.
```
andres-erbsen: I don't think `base_ops` should be bundled here, it is perfectly reasonable for half of the constructions given half of the fields from the bundle.

src/BoundedArithmetic/DoubleBounded.v
```diff
@@ -0,0 +1,114 @@
...
+      := match x return match x with
+                        | (a, [b0; b1]) => A * (B * B)
+                        | _ => True
+                        end
+         with
+         | (a, [b0; b1]) => (a, (b0, b1))
+         | _ => I
+         end.
+
+    Local Instance DoubleArchitectureBoundedOps : ArchitectureBoundedOps (2 * n)%nat
```
andres-erbsen: Again, I would like to (also?) have a non-bundled version of these definitions.

src/BoundedArithmetic/Interface.v
```diff
@@ -0,0 +1,131 @@
...
+Local Coercion Z.of_nat : nat >-> Z.
+
+Class ArchitectureBoundedOps (n : size) :=
+  { BoundedType : Type (* [n]-bit word *);
+    decode : BoundedType -> Z;
+    encode : Z -> BoundedType;
+    ShiftRight : forall a : size, BoundedType * BoundedType -> BoundedType;
+    (** given [(high, low)], constructs [(high << (n - a)) + (low >>
+          a)], i.e., shifts [high * 2ⁿ + low] down by [a] bits *)
+    ShiftLeft : forall a : size, BoundedType -> BoundedType * BoundedType;
```
andres-erbsen: Huh, this instruction is actually slightly different from what mainstream processors have: http://x86.renejeschke.de/html/file_module_x86_id_291.html

src/BoundedArithmetic/Interface.v
```diff
@@ -0,0 +1,131 @@
...
+    CarryAdd : forall (carry : bool) (x y : BoundedType), bool * BoundedType;
+    (** Ouputs [(x + y + if carry then 1 else 0) mod 2ⁿ], together
+          with a boolean that's [true] if the sum is ≥ 2ⁿ, and [false]
+          if there is no carry *)
+    CarrySub : forall (carry : bool) (x y : BoundedType), bool * BoundedType;
+  (** Ouputs [(x - y - if carry then 1 else 0) mod 2ⁿ], together
+          with a boolean that's [true] if the sum is negative, and [false]
+          if there is no borrow *) }.
+
+Inductive BoundedHalfType {n} {ops : ArchitectureBoundedOps n} :=
```
andres-erbsen: What is the motivation for this factoring of `decode_half`?

src/BoundedArithmetic/Interface.v
```diff
@@ -0,0 +1,131 @@
...
+Definition decode_half {n_over_two : size} {ops : ArchitectureBoundedOps (2 * n_over_two)%nat} (x : BoundedHalfType) : Z
+  := match x with
+     | UpperHalf v => decode v / 2^n_over_two
+     | LowerHalf v => (decode v) mod 2^n_over_two
+     end.
+
+Class ArchitectureBoundedFullMulOps n {ops : ArchitectureBoundedOps n} :=
+  { Mul : BoundedType -> BoundedType -> BoundedType * BoundedType
+  (** Outputs [(high, low)] *) }.
+Class ArchitectureBoundedHalfWidthMulOps n {ops : ArchitectureBoundedOps n} :=
```
andres-erbsen: Requiring multiplication on all pairs of halves is quite strict.

src/BoundedArithmetic/DoubleBounded.v
JasonGross: The notation is just a convenience.  Doing eta-expansion means we don't have to destruct anything to do rewriting or other manipulation.

src/BoundedArithmetic/Interface.v
```diff
@@ -0,0 +1,131 @@
...
+  { HalfWidthMul : BoundedHalfType -> BoundedHalfType -> BoundedType }.
+
+Class ArchitectureBoundedProperties {n} (ops : ArchitectureBoundedOps n) :=
```
andres-erbsen: Again, I don't think these should be bundled. add-with-carry is add-with-carry, regardless of what instruction set it comes from, and our definitions should apply even if the other operations do not. I.e., I think this file should be aiming for the second and third portability goal of programming languages as by https://cr.yp.to/qhasm/20050129-portable.txt .

src/BoundedArithmetic/Interface.v
JasonGross: I guess I should use the standard one then.  I only care about left-shifting by 128 in the standard way.  I guess I can implement the double width shifter by using both left and right shifts.

src/BoundedArithmetic/Interface.v
JasonGross: It's a factoring of HalfWidthMul.  I didn't want four copies, and an inductive type is nicer than a Boolean since "true" doesn't naturally mean lower or higher.  Everything else is just to allow a nicer phrasing of the lemmas around HalfWidthMul

src/BoundedArithmetic/Interface.v
JasonGross: Do some platforms only support some of them?  Are we targeting any such platform?  The Go code for the machine I'm targeting factors the function in this way.

src/BoundedArithmetic/Interface.v
JasonGross: You want a separate class for each instruction?

I'm not sure what you mean, even after reading that file.  Would it be quicker for you to reorganize the file yourself than to try to explain it to me?  I don't have any sense of how this file is to be used beyond my specific use case, so I'm kind-of lost in figuring out how to design it.

src/BoundedArithmetic/Interface.v
andres-erbsen: I think it is fine for you to use the fancy instruction as long as it makes engineering-sense. I would take it as a sign that you *really* shouldn't bundle the arguments, though. I would also prefer if the `SiftRight` and `ShiftLeft` did not have matching names if they don't have matching type signatures. Maybe let's call the fancy one `SpreadLeft` or `SplitLeft`? (and if I ever need to do something with non-fancy shifts, I think it will not be too much trouble proving the things I need).

src/BoundedArithmetic/Interface.v
andres-erbsen: On it.

src/BoundedArithmetic/Interface.v
andres-erbsen: TIL: some ARM processors have all 4: https://developer.arm.com/docs/dui0646/latest/3-the-cortex-m7-instruction-set/36-multiply-and-divide-instructions/363-smlawb-smlawt-smlabb-smlabt-smlatb-and-smlatt PPC has only one (http://www.ds.ewi.tudelft.nl/vakken/in1006/instruction-set/mulhwu.html), I don't think x86 has any.

src/BoundedArithmetic/Interface.v
andres-erbsen: Essentially, I wouldn't mind having a specific instruction set arithmetic specification in this repository. Yet the constructions you are making would be reusable if they only took as input the instructions they need, not the entire instruction set for your machine. The same way as I would not mind having an arithmetic specification for your processor, it would also make sense to have a special instantiation of these constructions for your ZLike, one-by-one. I guess if I squinted really hard, I could shove this case under the "definitional equality is very strict" complaint -- writing lemmas that only apply when the low-level function is a projection from your record is limiting on how they can be used.

andres-erbsen: I applied the conventions that I had in mind for the instruction set specifications at https://github.com/mit-plv/fiat-crypto/commit/ae6af4896acb7efcf757c0cb8bcddf84292e073c, read file-by-file, bottom-to-top. The file that contains machine and application specific ZLike construction does not build yet because I did not port it. Poke me when you have formed an opinion about whether this approach seems good to you.

JasonGross: The problem with ae6af48 is that it makes automation and rewrite databases hard / slow.  There's no easy way to recognize the instructions, so doing an `autorewrite` will try every single function of the right arity.

andres-erbsen: Yeah, that makes sense. "try every single function of the right arity" really shouldn't be slow (one typeclass table lookup) but I guess that would be too much to ask. Can we work around this by manually resolving the typeclasses instances while adding hints to the hint database?

JasonGross:
@andres-erbsen I've cherry-picked your commit and reworked things a bit to be based on records.  I'm about to head off to lunch, but I figured I'd push what I had so far, so you could comment.  Thoughts?  (I also managed to find [two](https://coq.inria.fr/bugs/show_bug.cgi?id=5011) [bugs](https://coq.inria.fr/bugs/show_bug.cgi?id=5012) in Coq in the process.)

andres-erbsen:
Mmmh. I really don't like this idea of forcing all definitions that use a function to project it out of a record just to make autorewrite faster. But if we need this to have our proofs run in reasonable time, well, I guess we have to. Did you try making a hint database of lemmas like `fst_sub_with_carry (fancy_machine.subc myFancyMachine)`?

Other than that, I am quite pleased with where this is going.

src/BoundedArithmetic/Interface.v
JasonGross: Yes, you could add hints with the instances pre-resolved.  We could even do this mostly-automatically with a module functor, probably.  But I don't know any design patterns that work this way with (open-universe) classes to be confident that I can get it right easily, without good test-cases.

JasonGross: The reason that it is slow is that the way `(e)rewrite` works is that it first does subterm matching on the goal, and then does typeclass resolution.  It's not just looking at identifiers, either; if you have something like `H : forall f : Z -> Z, P f -> forall x, f x = g x`, and you `rewrite f`, it's going to try things like `Z.succ` (if that appears anywhere), `Z.add 5` (if you have `5 + x`), etc.  If you have `H : forall T (f : T -> T), P f -> forall x, f x = g x`, it's going to try basically every single application node in the goal.  I can't guarantee that it'll be slow, but it's a very easy way to accidentally be slow.

JasonGross:
>  Did you try making a hint database of lemmas like

No, I didn't.  I'm doing this preventatively, based on prior experience with similar things.  I'm going to try to get it working like this first, and I'll leave a `TODO` comment to try dropping the record wrappers later if it gets annoying.

andres-erbsen:
Are you currently expecting feedback on this?

JasonGross:
No.  I'll close it for now, and reopen and mention you and update it when I am.

JasonGross:
@andres-erbsen rebased and updated with a single commit on top of what was previously there.  The `admit`s are all gone, and the proofs are separated from the implementations.  I have ripple-carry-add defined and proven for arbitrary tuples, and my version of ripple-carry-multiplication only for 2-tuples.

src/BoundedArithmetic/Interface.v
```diff
@@ -0,0 +1,339 @@
...
+  Record spread_left_immediate := { sprl :> W -> imm -> W * W (* [(low, high)] *) }.
+
+  Class is_spread_left_immediate (sprl : spread_left_immediate) :=
+    {
+      decode_fst_spread_left_immediate : forall r count,
+          0 <= count < n
+          -> decode (fst (sprl r count)) = (decode r << count) mod 2^n;
+      decode_snd_spread_left_immediate : forall r count,
+        0 <= count < n
+        -> decode (snd (sprl r count)) = (decode r << count) >> n;
```
andres-erbsen: 8.4 does not like this semicolon.

src/BoundedArithmetic/DoubleBounded.v
```diff
@@ -0,0 +1,126 @@
...
+       end.
+
+  Section ripple_carry_adc.
+    Context {n W} {decode : decoder n W} (adc : add_with_carry W).
+
+    Global Instance ripple_carry_adc {k} : add_with_carry (tuple W k)
+      := {| Interface.adc := ripple_carry_tuple adc k |}.
+  End ripple_carry_adc.
+
+  (* TODO: Would it made sense to make generic-width shift operations here? *)
```
andres-erbsen: If we end up needing wide shifts, this seems like the right place to put them.

JasonGross:
@andres-erbsen I've updated things based on your suggestions, and reworked the rewriting strategy so that we no longer rely on things being projections.  (I think the ability to infer which operation we're talking about when writing code separate from proofs still relies on projections, though.)

I think this is ready for the next round of code review.

src/BoundedArithmetic/DoubleBounded.v
```diff
@@ -0,0 +1,113 @@
...
+
+(** The list is low to high; the tuple is low to high *)
+Definition tuple_decoder {n W} {decode : decoder n W} {k : nat} : decoder (k * n) (tuple W k)
+  := {| decode w := BaseSystem.decode (base_from_limb_widths (repeat n k))
+                                      (List.map decode (List.rev (Tuple.to_list _ w))) |}.
+Global Arguments tuple_decoder : simpl never.
+Hint Extern 3 (decoder _ (tuple ?W ?k)) => let kv := (eval simpl in (Z.of_nat k)) in apply (fun n decode => (@tuple_decoder n W decode k : decoder (kv * n) (tuple W k))) : typeclass_instances.
+
+Section ripple_carry_definitions.
+  Definition ripple_carry {T} (f : T -> T -> bool -> bool * T)
```
andres-erbsen: Is this definition ever used? Given that it is here, would it be a good idea to define `ripple_carry_tuple` using `ripple_carry` and `from_list` and `to_list` (or perhaps a more flexible variant of `on_list`)?

src/BoundedArithmetic/Interface.v
```diff
@@ -0,0 +1,306 @@
...
+Global Arguments decode {n W _} _.
+
+Class is_decode {n W} (decode : decoder n W) :=
+  decode_range : forall x, 0 <= decode x < 2^n.
+
+Class rewrite_eq {A} (x y : A)
+  := by_rewrite : x = y.
+Arguments by_rewrite {A} _ _ {_}.
+
+Class rewrite_right_to_left_eq {A} (x y : A)
```
andres-erbsen: These classes deserve a comment.

src/BoundedArithmetic/DoubleBoundedProofs.v
```diff
@@ -0,0 +1,471 @@
...
+Local Open Scope list_scope.
+Local Open Scope nat_scope.
+Local Open Scope Z_scope.
+Local Open Scope type_scope.
+
+Local Coercion Z.of_nat : nat >-> Z.
+Local Coercion Pos.to_nat : positive >-> nat.
+Local Notation eta x := (fst x, snd x).
+
+Local Infix "==" := rewrite_eq.
```
andres-erbsen: Perhaps `<~=~>`>

andres-erbsen:
Are you planning to remove ZLike? Are you planning to do so before merging this? 

It should be possible to restore "the ability to infer which operation we're talking about when writing code" without projections by using named context variables like `Context {ldi} {is_ldi:is_load_immediate ldi}`. I think this would be cumbersome when dealing with `fancy_machine` because coq does not have Agda's `open` for records, but otherwise quite reasonable.

src/BoundedArithmetic/ArchitectureToZLikeProofs.v
```diff
@@ -0,0 +1,111 @@
...
+    { abstract t. }
+    { abstract t. }
+    { abstract t. }
+    { abstract t. }
+    { abstract t. }
+    { abstract t. }
+    { abstract t. }
+    { abstract t. }
+    { abstract t. }
+Hint Resolve Z.div_pos : zarith.
```
andres-erbsen: Is this hint in the middle of a proposed `par:t` intentional? Should this be removed when we switch to `par`, or moved up?

src/BoundedArithmetic/Interface.v
```diff
@@ -0,0 +1,306 @@
...
+      decode_fst_spread_left_immediate :> forall r count,
+          0 <= count < n
+          -> decode (fst (sprl r count)) =~> (decode r << count) mod 2^n;
+      decode_snd_spread_left_immediate :> forall r count,
+        0 <= count < n
+        -> decode (snd (sprl r count)) =~> (decode r << count) >> n
+
+    }.
+
+  Definition Build_is_spread_left_immediate' (sprl : spread_left_immediate)
```
andres-erbsen: Could these alternative `Build_` definitions be in the proofs file?

src/BoundedArithmetic/ArchitectureToZLikeProofs.v
JasonGross: Good catch.  It's already been moved to `ZUtil`, but I forgot to remove it here after rebuilding / merging the change to master.

src/BoundedArithmetic/DoubleBounded.v
JasonGross: No, it's not.  I've removed it.  I don't really want to redo the proofs with a different variant of it.

src/BoundedArithmetic/Interface.v
JasonGross: Moved to a different file with some comments.  Let me know if there needs to be more explanation.

src/BoundedArithmetic/DoubleBoundedProofs.v
JasonGross: Done.

src/BoundedArithmetic/Interface.v
JasonGross: Moved

JasonGross:
> Are you planning to remove ZLike? Are you planning to do so before merging this?

I could probably do that, if you think it'd be useful to you.  Otherwise we could wait until we revisit these things for whatever machine you're targeting.  In either case, I think I'd prefer to get this merged first, though.

> It should be possible to restore ...

I disagree.  If you look at `ZLikeOps_of_ArchitectureBoundedOps`, there is:
```coq
      decode_large := decode;
      decode_small := decode;
```
When you print this (thanks to the printing of coercions), you get
```coq
decode_large := tuple_decoder;
decode_small := fancy_machine.decode;
```
Similarly,
```coq
      Mul x y := muldw x y;
      CarryAdd x y := adc x y false;
```
work by instance resolution, the first picking up `mul_double_multiply` and the second picking up `ripple_carry_adc fancy_machine.adc`.

andres-erbsen:
I agree that `open` would not work in `ZLikeOps_of_ArchitectureBoundedOps`, but IMHO, making the arguments explicit would make that file easier to read. In particular, if I were to ever hit an error that typeclass resolution failed to resolve the `adc` in `CarryAdd x y := adc x y false`, I don't think I would be able to do anything smarter than `git bisect` to figure out what caused this. Generalizing this, I think that code and other definitions that will later be the object of study (in Coq proofs or in front of human eyes) should not use fancy implicit argument resolution. I am yet to see a case where local type inference that either finds the only solution or errors is not sufficient for creating concise definitions.

Anyway, I think you should merge this and go on with your life regardless of whether you agree that we should remove the projections or see a way of doing it. I will probably get to dig through this code much more thoroughly in a month or two (i.e. whenever I get to making a constant-time implementation of curve25519 exponent field arithmetic).

JasonGross:
I will probably not get a chance to remove `ZLike`, given that I only have 2-3 weeks left in my internship, and I want to get to something approximating straight-line assembly

Issue #53, opened 2016-08-12
RFC: We need bundling


andres-erbsen:
Parametrizing every definition in a module individually leads to explosion in the number of parameters. See https://gist.github.com/andres-erbsen/3c4989ad05673f6807728b9ecccd0557 for an example. It would be nice if we figured out to how to modularize our code in a way that does not cause that.

In particular, I am thinking of the following:

Record field := { F:Type; add : F->F->F; ...; is_field F add ... }

```
Section Edwards.
  Context {f:field}.
  (* "from f import *" *)
  Let add := F.(add)
  Let mul := F.(mul)
  ...

 Lemma edwards_curve_is_group : is group ... .
 Qed.

  (* only this would be used from outside *)
 Definition edwards_curve_group : group := {
   op := point_addition;
   ...
   is_group := edwards_curve_is_group
 }
```

The module system could do this automatically for us, but would not let us `Check` module instantiations, determine module parameters from Ltac, or parametrize modules by something other than modules.

I am not sure how mechanically we could convert the existing codebase, but let's figure out what to do first.

Another question is how liberally we should use this construction. I definitely wouldn't want generic lemmas (e.g., about properties of fields) to talk about projections from bundles -- it is important that a lemma can be applied even if the goal is not in bundled form.

JasonGross:
I will argue against using the module system this way.  It enforces the choice on all downstream code, modules and module functors are some of the hairiest parts of the kernel, as I understand it.

andres-erbsen:
Yeah, I am quite strongly against using the module system as well. This RFC is primarily for figuring out how to get the behavior we want without the module system, presumably using records.

andres-erbsen:
I think this has been de-facto decided in favor of this proposal.

PR #54, opened 2016-08-17
Speed up src/Specific/GF25519.v

JasonGross:
```
After    | File Name                      | Before   || Change
-----------------------------------------------------------------
0m13.37s | Total                          | 0m40.29s || -0m26.91s
-----------------------------------------------------------------
0m10.09s | Specific/GF25519               | 0m37.00s || -0m26.91s
0m03.29s | Experiments/SpecificCurve25519 | 0m03.30s || -0m00.00s
```

Making this a PR because I don't want to step on anyone's toes.  (Also, I didn't bother to change it here because I didn't notice it until just now, but `repeat progress` and `repeat` have the same behavior.)

### DIFF
--- a/src/Specific/GF25519.v
+++ b/src/Specific/GF25519.v
@@ -110,7 +110,7 @@ Proof.
   reflexivity.
 Qed.
 
-Definition appify2 {T} (op : fe25519 -> fe25519 -> T) (f g : fe25519) := 
+Definition appify2 {T} (op : fe25519 -> fe25519 -> T) (f g : fe25519) :=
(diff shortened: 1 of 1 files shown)


andres-erbsen:
We will probably merge this as is tomorrow.

After we figure out #53, the plan is to rewrite all specific code, converting from many `Definition`s and one record constructor to a tree of tactics that builds something like `{ field_ops & is_field field_ops /\ isomorphic field_ops fq_field_ops }`. I was brainstorming tactic design for this over chat a couple of days ago, and the approach seems viable (so for requiring a max length on `let'` binders is the main annoyance).

andres-erbsen:
LGTM

PR #57, opened 2016-08-25
Integrating ed25519 components

andres-erbsen:
This puts together several key components of the ed25519, with proofs wherever easy and admits (for now) if not. The main achievement is an elliptic curve scalar multiplication routine that uses the extended representation of the underlying Edwards curve and a non-uniform base system with lazy carrying for the field elements. I have extracted this code and executed it using `GHC`, passing the test that multiplication by group order returns identity. A lot of work is to be done, still -- the correctness proof of the described implementation has several non-trivial admits and the limbs of the field elements are stored as arbitrary-precision integers, not machine words.

The main hurdle in this patch is `Shrink Obligations`, which seems to have different behavior in 8.6 and older versions regardless of the state of the flag. @JasonGross please investigate and help -- I think this PR as-is does not build in 8.6.

The merge is also likely to be somewhat nasty, but I haven't looked into it yet.

@jadephilipoom please review the EdDSA and Ed25519 specs, cross-referencing with "eddsa for more curves" and the ed25519 RFC wherever relevant.

@JasonGross please see whether the `setoid_rewrite`s at https://github.com/mit-plv/fiat-crypto/compare/fiat-crypto-integration?expand=1#diff-5366caedd04016399124d528353b4a5cR134 and the next derivation in that file can be made less fragile.

Other crap that leaked in:

- `F.of_nat` and `F.to_nat`
- `Require`-s moved out of sections
- fix reversed carry chain in GF25519


JasonGross:
If
```coq
Global Unset Shrink Abstract.
Global Unset Shrink Obligations.
```
does not allow the code to build in both 8.5 and 8.6, that's a bug in Coq.  I'll look into it tomorrow.

GitHub claims there are no merge conflicts.

src/Algebra.v
```diff
@@ -618,6 +627,7 @@ Module IntegralDomain.
 End IntegralDomain.
 
 Module Field.
+  Require Coq.setoid_ring.Field_theory.
```
JasonGross: This is unsupported.  Coq only fully supports requires at top level.  (Symptoms of unsupportedness include anomalies, inconsistent states, and sad bug minimizers.)

src/CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v
```diff
@@ -8,6 +9,7 @@ Require Import Coq.Classes.Morphisms.
 Require Import Coq.Relations.Relation_Definitions.
 Require Import Crypto.Util.Tuple Crypto.Util.Notations Crypto.Util.Tactics.
 Require Export Crypto.Util.FixCoqMistakes.
+Require Import Ring_tac Field_tac.
```
JasonGross: Style nit: fully qualify requires

src/CompleteEdwardsCurve/ExtendedCoordinates.v
```diff
@@ -125,6 +129,7 @@ Module Extended.
         pose proof (add_coordinates_correct P Q) as Hrep.
         destruct P as [ [ [ [ ] ? ] ? ] [ HP [ ] ] ]; destruct Q as [ [ [ [ ] ? ] ? ] [ HQ [ ] ] ].
         autounfold with bash in *; simpl in *.
+        repeat rewrite field_div_definition; repeat rewrite field_div_definition in Hrep.
```
JasonGross: Shorter: `rewrite ?field_div_definition in Hrep |- *.`  which you prefer is up to you

src/Experiments/EdDSARefinement.v
```diff
@@ -1,130 +1,246 @@
...
 
-Section EdDSA.
-  Context `{prm:EdDSA}.
-  Context {eq_dec:DecidableRel Eeq}.
-  Local Infix "==" := Eeq.
-  Local Notation valid := (@valid E Eeq Eadd EscalarMult b H l B Eenc Senc).
-  Local Infix "*" := EscalarMult. Local Infix "+" := Eadd. Local Infix "++" := combine.
-  Local Notation "P - Q" := (P + Eopp Q).
-  Local Arguments H {_} _.
+(* TODO: move lemmas to better places *)
```
JasonGross: I'd suggest making `Util.Logic` or, better, `Util.Relations`

src/Experiments/EdDSARefinement.v
```diff
@@ -1,130 +1,246 @@
...
-  Context {Proper_Eenc : Proper (Eeq==>Logic.eq) Eenc}.
-  Context {Proper_Eopp : Proper (Eeq==>Eeq) Eopp}.
-  Context {Proper_Eadd : Proper (Eeq==>Eeq==>Eeq) Eadd}.
+Lemma combine_eq_iff {a b} (A:word a) (B:word b) C :
+  combine A B = C <-> A = split1 a b C /\ B = split2 a b C.
+Proof. intuition; subst; auto using split1_combine, split2_combine, combine_split. Qed.
 
-  Context {decE:word b-> option E}.
-  Context {decS:word b-> option nat}.
+Local Ltac logic :=
```
JasonGross: Is this stronger than `intuition eauto`?

src/Experiments/EdDSARefinement.v
```diff
@@ -1,130 +1,246 @@
...
+  repeat match goal with
+         | |- _ => intro
+         | H:exists _, _ |- _ => destruct H
+         | H:_ /\ _ |- _ => destruct H
+         | |- _ => solve [eauto]
+         | |- _ => split
+         end.
+Lemma exists_and_indep_l {A B} P Q :
```
JasonGross: Belongs on something like `Util.Logic`

src/Experiments/EdDSARefinement.v
```diff
@@ -1,130 +1,246 @@
...
+ weqb
+   (ERepEnc
+      (ErepAdd (SRepERepMul (s0) (EToRep B))
+         (SRepERepMul
+            (SRepDecModL (H _ (split1 b b sig ++ pk ++ message)))
+            (ErepOpp _)))) (split1 b b sig))
+
+       false).
+
+        etransitivity. Focus 2. {
```
JasonGross: Do you know about `2: tac.`?

src/Util/GlobalSettings.v
```diff
@@ -22,3 +22,6 @@ Global Set Asymmetric Patterns.
 
 (** Better-behaved [simpl] *)
 (** Set SimplIsCbn. *)
+
+(* [Program Definition] obligations only use the context variables they need *)
```
JasonGross: Your comment is backward.  Or you don't make it clear that you're talking about the 8.6 default and that we're changing that.

src/Algebra.v
andres-erbsen: done.

src/Experiments/EdDSARefinement.v
andres-erbsen: yeah, but these cases are nested (and the indentation is misleading)

andres-erbsen:
I added `Global Unset Shrink Abstract` and moved the `Require` out of modules, but the build still fails in my slightly dated 8.6.

JasonGross:
You seem to have broken the 8.4 (and 8.5) build(s) with the latest commits.

JasonGross:
> the build still fails in my slightly dated 8.6

[Bug #5044](https://coq.inria.fr/bugs/show_bug.cgi?id=5044)  Your `bash` tactic is the culprit.  Try this (don't `pose proof E.char_gt_2` if you don't need to) instead:
```coq
    Local Ltac bash' :=
      repeat match goal with
             | |- Proper _ _ => intro
             | _ => progress intros
             | [ H: _ /\ _ |- _ ] => destruct H
             | [ p:E.point |- _ ] => destruct p as [ [??] ? ]
             | [ p:point |- _ ] => destruct p as [ [ [ [??] ? ] ? ] ? ]
             | _ => progress autounfold with bash in *
             | |- _ /\ _ => split
             | _ => solve [neq01]
             | _ => solve [eauto]
             | _ => solve [intuition eauto]
             | _ => solve [etransitivity; eauto]
             | |- _*_ <> 0 => apply Ring.zero_product_iff_zero_factor
             | [H: _ |- _ ] => solve [intro; apply H; super_nsatz]
             | _ => progress uninv
             | |- Feq _ _ => super_nsatz
             end.
    Local Ltac bash := bash'; pose proof E.char_gt_2; bash'.
```

JasonGross:
What's up with your changes to `ExtendedCoordinates`?  It's 2x slower now than it was (2m on travis rather than 1m).  Is this expected?

src/Experiments/EdDSARefinement.v
```diff
@@ -1,130 +1,246 @@
...
-      rewrite weqb_true_iff.
-      f_equiv. exact Hvalid.
-    }
-  Qed.
+Lemma exists_and_equiv_r {A} {RA} {Equivalence_RA:Equivalence RA}
+      P {Proper_P: Proper (RA==>iff) P} (ref:A)
+  : (exists a, P a /\ RA a ref) <-> P ref.
+Proof.
+  logic; try match goal with [H:_|-_] => (rewrite H || rewrite <-H); assumption end.
+  repeat (assumption||reflexivity||econstructor); assumption. (* WHY the last [assumption]?*)
```
JasonGross: You could instead do `repeat repeat (assumption||reflexivity||econstructor).`  The answer is goal ordering and evar information.

JasonGross:
> see whether the setoid_rewrites at https://github.com/mit-plv/fiat-crypto/compare/fiat-crypto-integration?expand=1#diff-5366caedd04016399124d528353b4a5cR134 and the next derivation in that file can be made less fragile.

You're not going to get `setoid_rewrite` to do this transformation in a nice way.  You can't rely on higher-order unification to find non-unique solutions for you in the way you want.  I can probably get an approach working where first you "tag" a term with the proof that you want to rewrite with, and then you push the tags in until either you hit bottom (whence you drop them), or until you find the pattern you're supposed to be rewriting.  Except `setoid_rewrite` is modulo delta, so it might unfold too many things.  If Ltac supported manipulation of open terms, you could just use `pattern`, but, alas, it does not (yet).  If you can figure out a way to break apart your transformation into multiple steps, none of which require higher-order unification to infer non-unique solutions (for example, if you have an inverse function to `Eenc`, that is probably sufficient), then you might be able to make things work nicely.

src/CompleteEdwardsCurve/ExtendedCoordinates.v
```diff
@@ -50,16 +53,17 @@ Module Extended.
...
 
-    Obligation Tactic := bash.
+    Local Obligation Tactic := bash.
 
     Program Definition from_twisted (P:Epoint) : point := exist _
       (let (x,y) := E.coordinates P in (x, y, 1, x*y)) _.
 
     Program Definition to_twisted (P:point) : Epoint := exist _
-      (let '(X,Y,Z,T) := coordinates P in ((X/Z), (Y/Z))) _.
+      (let '(X,Y,Z,T) := coordinates P in let iZ := Finv Z in ((X*iZ), (Y*iZ))) _.
```
andres-erbsen: @JasonGross I think this change and the tactics to undo it in proofs made this file slower. I did this to make the generated code not duplicate the inversion. Not sure how to best fix it.

JasonGross: It's possible that moving `uninv` to the top of `bash` will be enough to fix it.  You can also do something like:
```coq
Let to_twisted_final (P:point) := (let '(X,Y,Z,T) := coordinates P in let iZ := Finv Z in ((X*iZ), (Y*iZ))).
Let to_twisted_fast (P:point) := (let '(X,Y,Z,T) := coordinates P in ((X/Z), (Y/Z))).
Lemma to_twisted_speed (P : point) : to_twisted_final P = to_twisted_fast P. ... Qed.
```
and then rewrite with `to_twisted_speed` wherever you need to.

andres-erbsen:
The parts of this that did not have open issues were merged as commits
1ea69cd53ff8472bb23c338d0e3fcac0a1f9ada5 through e51fb9ecd03fb1ce14870b40312d7259da0c4776

PR #60, opened 2016-09-03
Common PHOAS syntax

### LINKED ISSUE #58: Uniform Reflective Language
We now have something like eight different variants of reflected syntax: `Input.expr` and `Output.expr` in https://github.com/mit-plv/fiat-crypto/blob/jgross-phoas/src/Experiments/FancyMachineMontgomery256.v (mine) , https://github.com/mit-plv/fiat-crypto/blob/rsloan-phoas/src/Experiments/SpecificCurve25519.v (@varomodt 's), and https://github.com/mit-plv/fiat-crypto/blob/phoas/src/Experiments/SpecificCurve25519.v (@andres-erbsen 's), my untyped symbolic (`SymbolicExpr`) in https://github.com/mit-plv/fiat-crypto/blob/jgross-phoas/src/Experiments/FancyMachineMontgomery256.v (I also have `Syntax`, which I use only for notations), and possibly another one that is the input to @varomodt 's assembly pipeline.  We should probably unify them, to some extend, so that we're not duplicating work and copy-pasting.

I propose that we settle on how to design the inductive syntax here.  There are some high-level questions:
- How much do we want to enforce by construction?
  + Should the input and output languages for linearization be the same?
  + Should the output language enforce things like that you don't have wrongly-nested `let`s, or that you don't have trees of operations?
  + Should we enforce that we only have variables of type `bool` and of type word?

There are some questions of what constructions we should support.  I need the following features:
- Kinds of operators:
  + 1-to-1 (ldi)
  + 2-to-1 (e.g., mulhwll)
  + 3-to-1 (e.g., selc)
  + 3-to-2 (e.g., adc, subc)
- types:
  + booleans (for the carry bit)
  + words (or whatever type is carrying bounded integers)
  + literal `bool` constants (for `adc`)
  + literal `Z` constants (for `ldi`, `shl`, `shr`)
  + abstract `Z` constants (things whose identity should be fixed at reification time, which are of type `Z`, but which `Z_beq` will block on)
- generic operations:
  + `let ... in ...` (and a variant for handling *n*-to-2 operators)
  + possibly: function abstraction, for things like parameterizing multiplication over its word arguments

The things I want to be able to do with the language are:
- reification & interpretation
- linearization of lets and of function application
- common-subexpression-elimination (including combining a `fst` of an `adc` and a `snd` of the same `adc` into a single "carry-let-binding")
  + note that cse for PHOAS seems to require a variant of the syntax with decidable equality, to be used for keys / indexing of variables
- notations that let it look like assembly (since I'm too lazy to write stringification at the moment)

What do we need elsewhere?

There are also some design questions:
- How should we represent abstract constants?
- Should the `let ... in ...` be different for different numbers of arguments?
- Should we fuse operator application and `let ... in ...`?

JasonGross:
We also have linearization of function application / lets,
constant-inlining, and reification.

This closes #58.

@andres-erbsen @varomodt can you take a look and let me know if this fits your needs?  I've tried to stick fairly close to the syntax of https://github.com/mit-plv/fiat-crypto/tree/rsloan-phoas.  If you want, I can probably submit PRs (on Tuesday, hopefully) against https://github.com/mit-plv/fiat-crypto/tree/rsloan-phoas and/or https://github.com/mit-plv/fiat-crypto/tree/phoas that replace the reflective stuff with definitions based on this.

It might also be pretty straightforward to prove a very general bounds-checking-like lemma about this syntax: if interpreting any operation in two different ways gives pointwise-related functions, then the interpretation of any expression (in those two different ways) gives related values.  (This isn't quite right; I need to figure out how to deal with constants, but I shall think about this, and may find some way to make @varomodt 's job easier.  Or it might not work out.)

There are examples in `TestCase.v`.  The reification code is long (about 400 lines total for two different ways of reifying things), and I'm not super-happy with it, but I'm pretty happy with the expression language itself.


andres-erbsen:
Seeing `InputSyntax` reminded me of something that might be useful to know: I have no managed to prove `Wf` for the reified term on the phoas branch. It may be because I am messing up (but what I tried worked for smaller terms), or it may be because Ltac is slow and hopeless. The last-resort solution would be to reify into a language that does not need `Wf` to compile out of (probably a language that uses deBrujin indices) and compile from that to PHOAS.

JasonGross:
@andres-erbsen That's probably because the well-foundedness proof is exponential in the size of the term you're proving well-founded (every constructor records the entirety of the term that it's proving well-founded), and, well, you're building it in Ltac, so you're probably retypchecking the syntax tree of the expression *at least* once for each constructor.

Let me know if 4524fe5 fixes your problem; it's a reflective approach to proving well-foundedness (I got nerd-sniped for the past 6 hours or so), and should admit a linear proof term rather than an exponential one (and the linear constant is generally less than 1).  There's an example in at the bottom of `TestCases.v` with an inline script that's parametric in the expression.  (It doesn't yet work for terms that have free variables across the entire expression (i.e., top-level lambdas), but it should be straightforward to add that to the reflective lemma.)  It's basically equivalent to using DeBruijn indices, but doesn't require more reification code.

Come to think of it, @achlipala might have some way of doing this by compiling PHOAS to DeBruijn and back to PHOAS again, together with a proof that the DeBruijn -> PHOAS transformation always gives well-founded expressions; that's essentially what I'm doing here, but more messily.

src/Reflection/Syntax.v
```diff
@@ -0,0 +1,198 @@
...
+      | Op {t1 tR} : op t1 tR -> exprf t1 -> exprf tR
+      | Let : forall {tx}, exprf tx -> forall {tC}, (interp_flat_type_gen var tx -> exprf tC) -> exprf tC
+      | Pair : forall {t1}, exprf t1 -> forall {t2}, exprf t2 -> exprf (Prod t1 t2).
+      Bind Scope expr_scope with exprf.
+      Inductive expr : type -> Type :=
+      | Return {t} : exprf t -> expr t
+      | Abs {src dst} : (var src -> expr dst) -> expr (src -> dst).
+      Bind Scope expr_scope with expr.
+      Global Coercion Return : exprf >-> expr.
+      Fixpoint smart_interp_flat_map {f g}
```
andres-erbsen: I don't understand the code from here to the next section start.

src/Reflection/Syntax.v
```diff
@@ -0,0 +1,198 @@
...
+    Local Notation interp_flat_type := (interp_flat_type_gen interp_base_type).
+    Section expr.
+      Context {var : base_type_code -> Type}.
+
+      (** N.B. [Let] destructures pairs *)
+      Inductive exprf : flat_type -> Type :=
+      | Const {t : flat_type} : interp_type t -> exprf t
+      | Var {t} : var t -> exprf t
+      | Op {t1 tR} : op t1 tR -> exprf t1 -> exprf tR
+      | Let : forall {tx}, exprf tx -> forall {tC}, (interp_flat_type_gen var tx -> exprf tC) -> exprf tC
```
andres-erbsen: If I understand correctly, this `Let` binds multiple values if the let-clause expression is of tuple type?

src/Reflection/ReifyDirect.v
```diff
@@ -0,0 +1,212 @@
+(** * Direct reification of PHOAS Representation of Gallina *)
```
andres-erbsen: Are we going to be using this file?

src/Reflection/Syntax.v
JasonGross: Would it be better if I added a comment that says: Sometimes, we want to deal with partially-interpreted expressions, things like `prod (exprf A) (exprf B)` rather than `exprf (Prod A B)`, or like `prod (var A) (var B)` when we start with the type `Prod A B`.  These convenience functions let us recurse on the type in only one place, and replace one kind of pairing operator (be it `pair` or `Pair` or anything else) with another kind, and simultaneously mapping a function over the base values (e.g., `Var` (for turning `var` into `exprf`) or `Const` (for turning `interp_base_type` into `exprf`))."

src/Reflection/Syntax.v
JasonGross: That's correct.  That's the "N.B." comment at the top of the `Inductive`; if you have a suggestion of a better comment, I'm happy to replace it.

Doing this makes a lot of things nice, e.g., it completely eliminates the need to deal with MatchPair, while, combined with `SmartVar`, it basically lets us ignore the fact that we have more than value whenever we don't care.

src/Reflection/ReifyDirect.v
JasonGross: Probably not; I wrote it before `ReifyExact` before realizing that it would make some things painful and slow, and before realizing that I could mostly hide the fact that we're dealing with two languages from the user, and then just left it in.  Shall I remove it?

src/Reflection/WfReflective.v
```diff
@@ -0,0 +1,306 @@
+Require Import Coq.Arith.Arith Coq.Logic.Eqdep_dec.
```
andres-erbsen: This file is still black magic to me. I probably don't need to understand all of it, but some description of high-level design choices would be very nice.

src/Reflection/WfReflective.v
```diff
@@ -0,0 +1,306 @@
...
+Require Import Crypto.Reflection.Syntax.
+Require Import Crypto.Util.Notations Crypto.Util.Tactics Crypto.Util.Option Crypto.Util.Sigma Crypto.Util.Prod.
+
+Inductive pointed_Prop := trivial | inject (_ : Prop).
```
andres-erbsen: For example, why is this type a good idea? Why choose this over a boolean decider?

src/Reflection/WfReflective.v
```diff
@@ -0,0 +1,306 @@
...
+    | [ |- wff _ _ _ _ _ _ ] => constructor
+    | _ => progress unfold and_pointed_Prop in *
+    end.
+  Local Ltac t := repeat t_step.
+  Fixpoint reflect_wff (G : list (sigT (fun t => var1 t * var2 t)%type))
+           {t1 t2 : flat_type}
+           (e1 : @exprf (fun t => nat * var1 t)%type t1) (e2 : @exprf (fun t => nat * var2 t)%type t2)
+           {struct e1}
+    : match reflect_wffT (duplicate_type G) e1 e2 with
+      | Some p => to_prop p
```
andres-erbsen: There is probably some magic here with the false case. how does it work?

src/Reflection/ReifyDirect.v
andres-erbsen: Either remove it or move it to `Expermients/` if you want to keep it building.

src/Reflection/WfReflective.v
JasonGross: Because equality on `var t` is not decidable.

src/Reflection/WfReflective.v
JasonGross: If the reflective procedure fails (returns `None`), then we can't hope to prove `wff` of the expression.  We could just as well have done
```coq
match reflect_wffT (duplicate_type G) e1 e2, flat_type_eq_semidec_transparent t1 t2 with
| Some p, Some q => to_prop p -> @wff base_type_code interp_base_type op var1 var2 G t2 (eq_rect _ exprf (unnatize_exprf (List.length G) e1) _ q) (unnatize_exprf (List.length G) e2)
| _, _ => True
end
```
Would that be easier to understand?

JasonGross:
@andres-erbsen I've added a bunch of documentation to `WfReflective`, generalized it to handle free variables (`expr`, not just `exprf`), and stuck a single tactic at the bottom which is the interface for proving `Wf` goals.  (I've also factored some common things out into other files which I've pushed to master, e.g., making `inversion_sigma` and `inversion_prod` stronger, adding a `PointedProp` file).  Note that the time difference between the reflective `Wf` proofs and the non-reflective `Wf` proofs is already noticeable on very small examples (the example in the test case file has a factor of 3x-8x speedup when using the reflective version).

Also, I miscalculated when I claimed the non-reflective proof term is exponential in size; it's actually only quadratic in size.

src/Reflection/Syntax.v
andres-erbsen: How about "`Let` binds the components of a pair to separate variables, and does so recursively"?

src/Reflection/WfReflective.v
andres-erbsen: Thanks, this helps a lot!

andres-erbsen:
Generally LGTM. I will be excited to see how this integrates into existing stuff.

@jadephilipoom: I wonder if we could use this to make a tactic that reifies, flattens, reflects, and leaves the proof abstract so that it is lightweight enough to be used in Specific and such.

src/Reflection/Syntax.v
JasonGross: Done

JasonGross:
Given that travis was happy, and all I've done since then is updated the comment and rebased, I'm going to merge this.

PR #61, opened 2016-09-06
Small example of bounds-calculation with dependent types

JasonGross:
This might properly belong in Experiments rather than TestCase.  It demos the
ability to transform from one kind of constants to another kind, and to plug
bounds calculation functions into [var] to get bounds.  There's not currently
any sort of correctness theorem, and I'm not entirely sure what one would look
like.  Probably something that says that if you map a function over the syntax
tree and then interpret, you could instead have first interpreted and then
applied the function.

I'm hopeful that this will provide a template for integrating this version of
the syntax with rsloan-phoas.

I probably won't have any more time to play with this aspect of the reflection before the end of my internship; I'll be porting the CSE code, and the assembly for montgomery.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -63,6 +63,7 @@ src/ModularArithmetic/Montgomery/Z.v
 src/ModularArithmetic/Montgomery/ZBounded.v
 src/ModularArithmetic/Montgomery/ZProofs.v
 src/Reflection/CommonSubexpressionElimination.v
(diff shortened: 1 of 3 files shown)


JasonGross:
I'm going to merge this so I can stop rebasing it on top of master.  It's a small example, and having the code in the repo shouldn't do any harm.

PR #63, opened 2016-09-07
Add Barrett and Montgomery for the 256-bit machine

JasonGross:
This pushes Montgomery and Barrett all the way down to assembly, and I think marks the end of the work I'm doing for my internship.  I'm making this a PR mostly for visibility and as a sanity check about adding this to master, and that Experiments is the right place to leave this.  (Is it?)

Other interested parties might include @agl and @mschilder123


src/Experiments/FancyMachine256.v
```diff
@@ -0,0 +1,359 @@
...
+       | OPmulhwll => SOPmulhwll
+       | OPmulhwhl => SOPmulhwhl
+       | OPmulhwhh => SOPmulhwhh
+       | OPselc => SOPselc
+       | OPaddm => SOPaddm
+       end.
+
+  Definition CSE {t} e := @CSE base_type SConstT op_code base_type_beq SConstT_beq op_code_beq internal_base_type_dec_bl interp_base_type op symbolicify_const symbolicify_op t e (fun _ => nil).
+End reflection.
+Section instances.
```
andres-erbsen: Is it necessary to have both typeclass instances and tactic overrides for operations? Could we do with just tactics (because that is faster)?

src/Experiments/FancyMachineBarrett256.v
```diff
@@ -0,0 +1,129 @@
+Require Import Crypto.Experiments.FancyMachine256.
+Require Import Crypto.ModularArithmetic.BarrettReduction.ZBounded.
+Require Import Crypto.ModularArithmetic.BarrettReduction.ZHandbook.
+
+Section expression.
+  Let b : Z := 2.
+  Let k : Z := 253.
+  Let offset : Z := 3.
+  Context (ops : fancy_machine.instructions (2 * 128)) (props : fancy_machine.arithmetic ops).
+  (** 25519 dsa *)
```
andres-erbsen: Please change or remove this comment. dsa != ecdsa !=  eddsa. I assume you mean to indicate that this file implements arithmetic in the field of integers modulo the order of the curve25519 base point?

src/Experiments/FancyMachineBarrett256.v
```diff
@@ -0,0 +1,129 @@
...
+  Let props' := ZLikeProperties_of_ArchitectureBoundedOps ops m H 250 H' H''.
+  Let ops' := (ZLikeOps_of_ArchitectureBoundedOps ops m 250).
+  Local Existing Instances props' ops'.
+  Local Notation fst' := (@fst fancy_machine.W fancy_machine.W).
+  Local Notation snd' := (@snd fancy_machine.W fancy_machine.W).
+  Local Notation SmallT := (@ZBounded.SmallT (2 ^ 256) (2 ^ 250) m
+                                  (@ZLikeOps_of_ArchitectureBoundedOps 128 ops m _)).
+  Definition ldi' : load_immediate SmallT := _.
+  Let isldi : is_load_immediate ldi' := _.
+  Axiom (μ_range : 0 <= b ^ (2 * k) / m < b ^ (k + offset)).
```
andres-erbsen: Should this axiom be a context variable instead?

src/Experiments/FancyMachine256.v
```diff
@@ -0,0 +1,359 @@
...
+    (at level 200, format "'c.RightShifted' { x ,  v }").
+Notation "'c.RightShifted' { x , v }" :=
+  (cRightShifted (cVar x) v)
+    (at level 200, format "'c.RightShifted' { x ,  v }").
+Notation "'λ'  x .. y , t" := (cAbs (fun x => .. (cAbs (fun y => t)) ..))
+  (at level 200, x binder, y binder, right associativity).
+
+Definition Syntax := forall var, @syntax var.
+
+Section assemble.
```
andres-erbsen: Perhaps a comment that this section turns PHOAS expressions with fancy machine operations into specialized fancy machine syntax trees?

andres-erbsen:
I am leaning towards putting this in Specifc instead of Experiments -- it seems like complete result on its own, these files are unlikely to be reused for something else, and I don't think we are expecting to need to remove it any time soon. Currently, `Specific` code does not quantify over anything (no context variables), but this seems secondary. We *could* make versions of this with the curve constants plugged in, though.

LGTM (but see suggestions).

Overall, I am quite impressed with how all of this turned out, and I expect that everything that is even slightly upstream from here will get reused in some form. Currently, path of least resistance to an implementation of p256 base field arithmetic and curve25519 exponent field arithmetic on consumer devices would be to fill in for the wide instructions here using the double-width constructions you have already verified. I will probably punt doing this until our curve25519 group operations are fast enough to be interesting.

src/Experiments/FancyMachineBarrett256.v
JasonGross: I was basing this on Marius' comment in his source code:
```go
// Mulmod where 4**253/mod fits in 256 bits.
// src1 and src2 < 2**253.
// Useful for 25519 dsa operations.
```
I think this is for the 25519 exponent modulus field?  Do you have a suggested replacement comment?

src/Experiments/FancyMachineBarrett256.v
JasonGross: Good catch; thanks!

src/Experiments/FancyMachine256.v
JasonGross: Added

src/Experiments/FancyMachine256.v
JasonGross: Removed

JasonGross:
Thanks!  I'm pretty happy with how it turned out too.

PR #66, opened 2016-09-19
Add a non-higher-order syntax, and register assignment

JasonGross:
The register assigner will return [None] if you give it an invalid
assignment (too short, attempting to eliminate live code, attempting to
merge registers that can't be merged).

No correctness proofs yet, nor any sort of automatic register
assignment.  I'm planning to write a dead-code-eliminator register
assigner next.  It should also be moderately straight-forward to write a
default register allocator, or perhaps a register post-allocator, that
takes a manual register allocation, and renumbers variables in a greedy
way.  (This would let you pre-specify that some registers should be
merged, but leave the rest of the assignment up to the algorithm.)

Main code review questions: Should I stick in more examples?  Is the interface reasonable?

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -78,7 +78,10 @@ src/Reflection/TestCase.v
 src/Reflection/WfProofs.v
 src/Reflection/WfReflective.v
 src/Reflection/WfRel.v
+src/Reflection/Named/Compile.v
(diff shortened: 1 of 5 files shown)


JasonGross:
Superseded by #69

PR #68, opened 2016-09-22
Dead code elimination and register assignment

JasonGross:
This is on top of #66 

Still no proofs, but there's working dead code elimination and register assignment.  The interface is that you provide a list of names, and these names are used, in order, after dead code is dropped.  (You can also skip dead code elimination when compiling.)  If you try to reuse a register that still alive when you overwrite it, the compilation returns `None`.  This integrates well with the carry bit; as long as all operations that set/clear the carry bit bind a carry value in the syntax tree, you can use the same "register" for all of the carry bits, and you'll be informed if you're trying to use a clobbered carry bit.

The fancy machine assembly is now checked and interpreted (rather than there being a pretty-printing-only syntax tree).

Warts:
- `vm_compute` would take upwards of 2 hours (estimated); we use `lazy` instead.  (See [bug #5096](https://coq.inria.fr/bugs/show_bug.cgi?id=5096).)
- This is nearly enough to do a nice version of Barrett / Montgomery without common-subexpression elimination; what's left is either doing correctness proofs, or figuring out how to do sanity-checking without eta-expansion, or moving to 8.5 and using primitive projections.

Things I want feedback on:
- the interface for compilation / dead code elimination / register assignment (is it reasonable and useful?)
- dropping support for `vm_compute` (using `lazy` here, but also supporting `native_compute`)
- anything else you feel like commenting on

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -79,7 +79,9 @@ src/Reflection/WfProofs.v
 src/Reflection/WfReflective.v
 src/Reflection/WfRel.v
 src/Reflection/Named/Compile.v
+src/Reflection/Named/ContextOn.v
 src/Reflection/Named/DeadCodeElimination.v
+src/Reflection/Named/EstablishLiveness.v
 src/Reflection/Named/NameUtil.v
 src/Reflection/Named/RegisterAssign.v
 src/Reflection/Named/Syntax.v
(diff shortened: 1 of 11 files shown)


JasonGross:
Closing this; I managed to finish scalable barrett

PR #69, opened 2016-09-22
Scalable Scalars, Dead Code Removal, Register Assignment

JasonGross:
This is on top of #66, and replaces #68.

Barrett and Montgomery no longer need CSE.

Still no proofs, but there's working dead code elimination and register assignment.  The interface is that you provide a list of names, and these names are used, in order, after dead code is dropped.  (You can also skip dead code elimination when compiling.)  If you try to reuse a register that still alive when you overwrite it, the compilation returns `None`.  This integrates well with the carry bit; as long as all operations that set/clear the carry bit bind a carry value in the syntax tree, you can use the same "register" for all of the carry bits, and you'll be informed if you're trying to use a clobbered carry bit.

The fancy machine assembly is now checked and interpreted (rather than there being a pretty-printing-only syntax tree).

Wart:
- `vm_compute` would take upwards of 2 hours (estimated); we use `lazy` instead.  (See [bug #5096](https://coq.inria.fr/bugs/show_bug.cgi?id=5096).)

Things I want feedback on:
- the interface for compilation / dead code elimination / register assignment (is it reasonable and useful?)
- dropping support for `vm_compute` (using `lazy` here, but also supporting `native_compute`)
- anything else you feel like commenting on

### DIFF
--- a/src/BoundedArithmetic/DoubleBoundedProofs.v
+++ b/src/BoundedArithmetic/DoubleBoundedProofs.v
@@ -12,7 +12,7 @@ Require Import Crypto.Util.ZUtil.
 Require Import Crypto.Util.ListUtil.
 Require Import Crypto.Util.Tactics.
 Require Import Crypto.Util.Notations.
-Require Import Crypto.Util.LockedLet.
+Require Import Crypto.Util.LetIn.
 
 Import ListNotations.
 Local Open Scope list_scope.
@@ -239,20 +239,20 @@ Global Instance decode_mul_double
 
 Lemma ripple_carry_tuple_SS' {T} f k xss yss carry
   : @ripple_carry_tuple T f (S (S k)) xss yss carry
-    = llet xss := xss in
-      llet yss := yss in
+    = dlet xss := xss in
+      dlet yss := yss in
       let '(xs, x) := eta xss in
       let '(ys, y) := eta yss in
-      llet addv := (@ripple_carry_tuple _ f (S k) xs ys carry) in
+      dlet addv := (@ripple_carry_tuple _ f (S k) xs ys carry) in
       let '(carry, zs) := eta addv in
-      llet fxy := (f x y carry) in
+      dlet fxy := (f x y carry) in
       let '(carry, z) := eta fxy in
       (carry, (zs, z)).
 Proof. reflexivity. Qed.
 
 Local Ltac eta_expand :=
   repeat match goal with
-         | _ => rewrite !unlock_let
+         | _ => progress unfold Let_In
          | [ |- context[let '(x, y) := ?e in _] ]
            => rewrite (surjective_pairing e)
          | _ => rewrite <- !surjective_pairing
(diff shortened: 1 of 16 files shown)


JasonGross:
Update: Fancy machine no longer uses CSE

JasonGross:
Okay, I'll merge this as-is (with a comment.)

1. Yes.  Register names are per-component-of-tuple, not per-instruction.  I handle carries by a single special carry-bit pseudo-register.  This automatically ensures that instructions using the carry-bit are correctly ordered, as long as every instruction that clobbers the carry bit is correctly noted as doing so in its type.  (So if you want to really make sure of this, you'll need a multiplication instruction that correctly returns relevant flags.)
2. 20-50 lines of generic code, I'd guess?  (You'd actually want to take in a map `forall src dst, op src dst -> interp_flat_type_gen (fun _ => Register) src -> bool`.)
3. The main issue I see is figuring out the right way to design this phase.  It should be pretty straightforward to, e.g., take in a map of locations to register rebindings to insert, together with whatever the "move" operation is, and just check that nothing bad happens.  But it's unclear to me exactly how the input should look.

> A simple register-assignment interface example would be nice indeed.

https://github.com/mit-plv/fiat-crypto/pull/69/files#diff-fc1d32cc87c131349763d5ad44a1b4dbR159

> I think the only way to determine whether this is going to be useful is to meet up in person and talk through ourselves all the remaining pipeline stages. I am optimistic, though.

Sure, I'll be back at MIT tomorrow (though possibly not at the group meeting, since it seems to conflict with the poster session?)

> Would it be possible to define a ltac that is `native_compute` in versions where that is available and `lazy` otheriwse?

We'd need to stick it somewhere that we can import different files based on the version of Coq.  For example, we could make a `Compat84` folder and a `Compat85` folder, stick a `Util.v` file in each, bind whichever one matches with the version we're using to `VersionCompat`, and do `Import VersionCompat.Util`, and stick `Ltac native_compute_or_lazy := native_compute.` in `Compat84/Util.v` and `Ltac native_compute_or_lazy := lazy.` in `Compat85/Util.v`.  However, note that `vm_compute` seems to have been fixed in 8.6; see [bug #5096](https://coq.inria.fr/bugs/show_bug.cgi?id=5096), and, more importantly, note that `native_compute` falls back to `vm_compute` when the native compiler is disabled (the default on Windows), and that `lazy` is actually a bit faster than `native_compute` here.

src/BoundedArithmetic/DoubleBoundedProofs.v
```diff
@@ -235,14 +236,40 @@ Global Instance decode_mul_double
...
+      let '(xs, x) := eta xss in
+      let '(ys, y) := eta yss in
+      dlet addv := (@ripple_carry_tuple _ f (S k) xs ys carry) in
+      let '(carry, zs) := eta addv in
+      dlet fxy := (f x y carry) in
+      let '(carry, z) := eta fxy in
+      (carry, (zs, z)).
+Proof. reflexivity. Qed.
+
+Local Ltac eta_expand :=
```
JasonGross: Done

PR #70, opened 2016-09-26
montgomery-curve

andres-erbsen:
Unsuccessful attempt at formalizing Montgomery coordinates for elliptic curves. Unsuccessful, because I ran out of steam before I managed to establish isomorphism to weierstrass. It would be kinda nice if this kept building, so that it is not that bad to pick it up later.

### DIFF
--- a/src/Experiments/MontgomeryCurve.v
+++ b/src/Experiments/MontgomeryCurve.v
@@ -1,0 +1,287 @@
+Require Crypto.Algebra.
+Require Crypto.Util.GlobalSettings.
+
+Require Crypto.Spec.WeierstrassCurve.
+Module W := Crypto.Spec.WeierstrassCurve.E. (* TODO: rename the module? *)
+
+Module M.
+  Section MontgomeryCurve.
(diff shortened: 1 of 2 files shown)


src/Experiments/MontgomeryCurve.v
```diff
@@ -0,0 +1,287 @@
...
+        assert (H:LHS = LH);
+          [|rewrite H; assumption]
+      end;
+      try field; repeat split; try trivial.
+
+    Global Instance WeierstrassParamsOfMontgomery : @W.weierstrass_params
+                                                      F Feq Fzero Fone Fopp Fadd Fmul
+                                                      WeierstrassA WeierstrassB.
+    Proof.
+      (* I am not sure [common_denominator] is behaving correctly here *)
```
JasonGross: We probably actually want to replace `common_denominator` with a tactic that that replaces `n / d` with a new variable `q` and an equation `d <> 0 -> q * d = n`, specialize by using `nsatz_nonzero` or something, and then run `nsatz`.  I think this version will behave better.

JasonGross:
Seems fine by me; feel free to merge it.

PR #72, opened 2016-10-03
Add spec for x86

JasonGross:
I haven't yet written code to make sure this is everything I need, but I think it suffices.

Feedback I want:
1. Do all these instructions exist.
2. Which of them do things with the carry flag?

Also, I'm planning to implement reduce-modulo as
```
let '(CF, _) := sub ret modulus in
let '(CF, v) := subc 0 0 in
let maybe_modulus := and v modulus in
let '(CF, ret) := sub ret maybe_modulus in
Return ret
```

### DIFF
--- a/src/BoundedArithmetic/Interface.v
+++ b/src/BoundedArithmetic/Interface.v
@@ -288,3 +288,34 @@ Module fancy_machine.
       add_modulo :> is_add_modulo addm
     }.
 End fancy_machine.
+
+Module x86.
+  Local Notation imm := Z (only parsing).
+
+  Class instructions (n : Z) :=
+    {
+      W : Type (* [n]-bit word *);
+      decode :> decoder n W;
+      ldi :> load_immediate W;
+      shrd :> shift_right_doubleword_immediate W;
+      shl :> shift_left_immediate W;
+      shr :> shift_right_immediate W;
+      adc :> add_with_carry W;
+      subc :> sub_with_carry W;
+      muldw :> multiply_double W;
+      and :> bitwise_and W
+    }.
+
+  Class arithmetic {n} (ops:instructions n) :=
+    {
+      decode_range :> is_decode decode;
+      load_immediate :> is_load_immediate ldi;
+      shift_right_doubleword_immediate :> is_shift_right_doubleword_immediate shrd;
+      shift_left_immediate :> is_shift_left_immediate shl;
+      shift_right_immediate :> is_shift_right_immediate shr;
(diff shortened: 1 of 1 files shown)


andres-erbsen:
I looked up some instructions: destination is the leftmost argument, `r` means register, `m` means memory location, `imm` means immediate. I'm referencing the 64-bit versions, 32-bit variants do exist. We probably only care about the carry flag CF and not the other flags because I think the instructions we use never read them.

- selc: `CMOVC r64, r/m64`, no flags -- Move if carry (CF=1) http://www.felixcloutier.com/x86/CMOVcc.html
- ldi: `MOV r64, imm64`, no flags -- http://www.felixcloutier.com/x86/MOV.html
- `SHRD r/m64, r64, imm8`, tramples CF SF ZF PF OF AF -- "Shift r/m64 to right imm8 places while shifting bits from r64 in from the left." IIRC this means that the first register is modified in-place. http://www.felixcloutier.com/x86/SHRD.html
- `SHL r/m64, imm8`, tramples CF SF ZF PF OF AF -- http://www.felixcloutier.com/x86/SAL:SAR:SHL:SHR.html
- `SHR r/m64, imm8`, tramples CF SF ZF PF OF AF -- http://www.felixcloutier.com/x86/SAL:SAR:SHL:SHR.html
- `ADC r64, r/m64`, reads CF, writes CF, tramples SF ZF PF OF AF -- http://www.felixcloutier.com/x86/ADC.html
- subc: `SBB r64, r/m64`, reads CF, writes CF, tramples SF ZF PF OF AF -- http://www.felixcloutier.com/x86/SBB.html computes CF+a-b
- `MUL r/m64`, writes OF CF, tramples SF ZF PF AF -- "RDX:RAX ← RAX ∗ r/m64" http://www.felixcloutier.com/x86/MUL.html
- `AND r64, r/m64`, tramples CF SF ZF PF OF AF -- http://www.felixcloutier.com/x86/AND.html

So looks good to me. You may want to also use a subtraction that ignores the carry flag: `SUB r64, r/m64` http://www.felixcloutier.com/x86/SUB.html that writes all 6 flags.

Your implementation of conditional subtraction looks reasonable, I think https://github.com/floodyberry/supercop/blob/master/crypto_sign/ed25519/amd64-64-24k/sc25519_barrett.s#L1029 correspons to the following:

        let '(CF, _) := sub ret modulus in
        let maybe_modulus := ldi 0 in
        let maybe_modulus := cmovc maybe_modulus modulus in
        let '(CF, ret) := sub ret maybe_modulus in
        Return ret

PR #74, opened 2016-10-11
word manipulation functions for secret key formatting

andres-erbsen:
These need to be constant time with respect to the words being manipulated, so converting to N and operating there is no good.

@JasonGross @achlipala: I introduced some machinery for working with words whose sizes are equal by `Logic.eq` but not convertible, and I am not sure whether what I did is a good way to go about it.  The core of it is `cast_word {n m} : forall {pf:wordsize_eq n m}, word n -> word m`, an identity function that only uses the proof argument in impossible cases, so the proof can be opaque. I also added hooks for proving word size equalities using hints (`omega` for now), and `Qed`-wrapped the goal type so that other tactic/typeclass machinery does not mess with it.

No correctness proofs yet, but should be rather straightforward, so I am punting until we actually decide we want to do this this way.

### DIFF
--- a/src/Util/WordUtil.v
+++ b/src/Util/WordUtil.v
@@ -1,6 +1,7 @@
 Require Import Coq.Numbers.Natural.Peano.NPeano.
 Require Import Coq.ZArith.ZArith.
 Require Import Crypto.Util.NatUtil.
+Require Import Crypto.Util.Tactics.
 Require Import Bedrock.Word.
 Require Import RelationClasses.
 
@@ -77,6 +78,7 @@ Proof. exact (proj1 ((proj2_sig wordsize_eq_sig) a b)). Qed.
 Ltac wordsize_eq_to_eq :=
   repeat match goal with
          | [H: wordsize_eq _ _ |- _] => apply wordsize_eq_eq in H
+         | [H: wordsize_eq _ _ |- _] => unique pose proof (wordsize_eq_eq _ _ H)
          | |- wordsize_eq _ _ => apply eq_wordsize_eq
          end.
 
@@ -105,3 +107,15 @@ Definition clearlow {n b:nat} {H:n <= b} (w:word b) : word b :=
 
 Definition setbit {n b:nat} {H:n < b} (w:word b) : word b :=
   wor (cast_word( wzero (b-n-1)  ++ wones 1 ++ wzero n )) w.
+
+Lemma wordToNat_cast_word : forall {n} (w:word n) m pf,
+  wordToNat (@cast_word n m pf w) = wordToNat w.
+Proof.
+  induction w; destruct m eqn:Heqm;
+    simpl; intros; wordsize_eq_to_eq;
+      rewrite ?IHw; solve [trivial | discriminate].
+Qed.
+
+Lemma wordToN_cast_word {n} (w:word n) m pf :
+  wordToN (@cast_word n m pf w) = wordToN w.
+Proof. rewrite !wordToN_nat, wordToNat_cast_word; reflexivity. Qed.
--- a/src/Util/WordUtil.v
+++ b/src/Util/WordUtil.v
@@ -2,6 +2,7 @@ Require Import Coq.Numbers.Natural.Peano.NPeano.
 Require Import Coq.ZArith.ZArith.
 Require Import Crypto.Util.NatUtil.
 Require Import Bedrock.Word.
+Require Import RelationClasses.
 
 Local Open Scope nat_scope.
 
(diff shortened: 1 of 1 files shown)


JasonGross:
Is there any reason to assume that `Class wordsize_eq (x y : nat) := wordsize_eq_to_eq : x = y.` won't work?  If you add a typeclass hint that says `hnf in *`, you're shooting yourself in the foot, and I don't think we need to protect against that.  Which makes `Qed`ing the type of the class overkill.  (Your way also breaks if you use a `wordsize_eq` hypothesis in the type of another hypothesis.)

andres-erbsen:
Hmm. I did the Qed because it was necessary in a similar case when we were trying to get `field` to work on `F q` (which we later have up on), it is not necessary in the cases where it is currently used. If you think we are better off without the Qed, I will remove it. Is the basis of your objection that it would be infeasible to extend the `wordsize_eq_to_eq` conversion tactic to cover all cases we will encounter?

JasonGross:
> trying to get `field` to work on `F q`

That was [bug #4810](https://coq.inria.fr/bugs/show_bug.cgi?id=4810), that `field` ended up being linear in the number of record fields with a fairly large constant, no?  In this case, there's no record, and no field.  
I'm arguing against the `Qed` here because it seems like it'll make things a bit more complicated and a bit slower (and possibly much more complicated in a few cases) and I don't see any benefit to it.

src/Util/WordUtil.v
```diff
@@ -63,3 +65,42 @@ Definition wfirstn n {m} (w : Word.word m) {H : n <= m} : Word.word n.
...
+Class wordsize_eq (x y : nat) := wordsize_eq_to_eq : x = y.
+Ltac wordsize_eq_tac := cbv beta delta [wordsize_eq] in *; omega.
+Ltac gt84_abstract t := t. (* TODO: when we drop Coq 8.4, use [abstract] here *)
+Hint Extern 100 (wordsize_eq _ _) => gt84_abstract wordsize_eq_tac : typeclass_instances.
+
+Program Fixpoint cast_word {n m} : forall {pf:wordsize_eq n m}, word n -> word m :=
+  match n, m return wordsize_eq n m -> word n -> word m with
+  | O, O => fun _ _ => WO
+  | S n', S m' => fun _ w => WS (whd w) (@cast_word _ _ _ (wtl w))
+  | _, _ => _ (* impossible *)
```
JasonGross: `Program` has a nifty notation where you can use `!` to indicate that something is impossible (notation for `False_rect`)

src/Util/WordUtil.v
```diff
@@ -1,7 +1,9 @@
 Require Import Coq.Numbers.Natural.Peano.NPeano.
 Require Import Coq.ZArith.ZArith.
 Require Import Crypto.Util.NatUtil.
+Require Import Crypto.Util.Tactics.
 Require Import Bedrock.Word.
+Require Import RelationClasses.
```
JasonGross: I'd prefer this to be fully qualified.

src/Util/WordUtil.v
```diff
@@ -63,3 +65,42 @@ Definition wfirstn n {m} (w : Word.word m) {H : n <= m} : Word.word n.
...
+  wand (cast_word( wzero n ++ wones (b-n) )) w.
+
+Definition setbit n {b} {H:n < b} (w:word b) : word b :=
+  wor (cast_word( wzero n ++ wones 1 ++ wzero (b-n-1) )) w.
+
+Definition clearbit n {b} {H:n < b} (w:word b) : word b :=
+  wand (cast_word( wones n ++ wzero 1 ++ wones (b-n-1) )) w.
+
+Lemma wordToNat_cast_word : forall {n} (w:word n) m pf,
+    wordToNat (@cast_word n m pf w) = wordToNat w.
```
JasonGross: Want to factor this into a theorem that says `forall n w, @cast_word n n eq_refl w = w`?  Then this and the next just become `destruct pf; apply cast_word_refl.`

JasonGross: Er, I mean, `destruct pf; now rewrite cast_word_refl`.

src/Util/WordUtil.v
andres-erbsen: Done.

src/Util/WordUtil.v
andres-erbsen: Done.

src/Util/WordUtil.v
```diff
@@ -63,3 +66,49 @@ Definition wfirstn n {m} (w : Word.word m) {H : n <= m} : Word.word n.
...
+  wand (cast_word( wones n ++ wzero 1 ++ wones (b-n-1) )) w.
+
+Lemma cast_word_refl {n} (w:word n) : @cast_word n n eq_refl w = w.
+Proof.
+  induction w;
+    repeat match goal with
+           | _ => solve [trivial]
+           | _ => progress (simpl @cast_word;f_equal)
+           | |- context [@cast_word ?n ?n ?pf _ ] =>
+             pattern pf; rewrite (Eqdep_dec.UIP_refl_nat n pf)
```
JasonGross: Oh, ew, is this case necessary because `Program Fixpoint` does something more complicated than `f_equal pred pf` to fill in the recursive proof argument?

PR #76, opened 2016-10-17
Ix86nitial work on exponent field as Z

andres-erbsen:


src/Specific/X86/Exponent25519.v
```diff
@@ -0,0 +1,124 @@
...
+Import NPeano.
+
+Local Notation modulusv := (2^252 + 27742317777372353535851937790883648493)%Z.
+Local Coercion Z.of_nat : nat >-> Z.
+Local Notation eta x := (fst x, snd x).
+Local Notation eta3 x := (eta (fst x), snd x).
+Local Notation eta4 x := (eta3 (fst x), snd x).
+Local Notation eta4' x := (eta (fst x), eta (snd x)).
+
+Section x86.
```
andres-erbsen: This section name is no longer relevant, right?

src/Specific/X86/Exponent25519.v
```diff
@@ -0,0 +1,124 @@
...
+    end.
+  Definition SRepDecModL : Word.word (256 + 256) -> SRep.
+  Proof.
+    let v' := (eval cbv [SRepDecModL'] in SRepDecModL') in
+    let v' := (eval cbv beta iota delta [Let_In] in v') in
+    let v := (*change_values*) v' in
+    unify v v';
+      exact v.
+  Defined.
+  (* TODO(jadep):what's S2Rep? *)
```
andres-erbsen: S2Rep is F.to_Z here, I think.

JasonGross: Thanks, updated.

src/Specific/X86/Exponent25519.v
JasonGross: Changed to `Z`

PR #77, opened 2016-10-17
SC25519

JasonGross:


jadephilipoom:
Looks good--we can merge it once Travis is satisfied.

JasonGross:
Travis is happy with what I added in this commit, so I'll merge it.

PR #78, opened 2016-10-17
Add opt versions of add, sub, opp

JasonGross:
Implementing @jadephilipoom's idea to carry on these to make the word versions into a field.

@jadephilipoom, there are two `Admitted` proofs here.  Can I assign you to fill them in?

### DIFF
--- a/src/ModularArithmetic/ModularBaseSystem.v
+++ b/src/ModularArithmetic/ModularBaseSystem.v
@@ -60,9 +60,22 @@ Section ModularBaseSystem.
   (* Placeholder *)
   Definition div (x y : digits) : digits := encode (F.div (decode x) (decode y)).
 
+  Definition carry_ (carry_chain : list nat) (us : digits) : digits :=
+    from_list (carry_sequence carry_chain [[us]]) (length_carry_sequence length_to_list).
+
+  Definition carry_add (carry_chain : list nat) (us vs : digits) : digits :=
+    carry_ carry_chain (add us vs).
   Definition carry_mul (carry_chain : list nat) (us vs : digits) : digits :=
-    from_list (carry_sequence carry_chain [[mul us vs]]) (length_carry_sequence length_to_list).
-  
+    carry_ carry_chain (mul us vs).
+  Definition carry_sub (carry_chain : list nat) (modulus_multiple: digits)
+             (modulus_multiple_correct : decode modulus_multiple = 0%F)
+             (us vs : digits) : digits :=
+    carry_ carry_chain (sub modulus_multiple modulus_multiple_correct us vs).
+  Definition carry_opp (carry_chain : list nat) (modulus_multiple : digits)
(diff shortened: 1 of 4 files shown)


src/ModularArithmetic/ModularBaseSystem.v
```diff
@@ -60,9 +60,30 @@ Section ModularBaseSystem.
...
-  
+    carry_ carry_chain (mul us vs).
+  Definition carry_sub (carry_chain : list nat) (modulus_multiple: digits)
+             (modulus_multiple_correct : decode modulus_multiple = 0%F)
+             (us vs : digits) : digits :=
+    carry_ carry_chain (sub modulus_multiple modulus_multiple_correct us vs).
+  Definition carry_opp (carry_chain : list nat) (modulus_multiple : digits)
+             (modulus_multiple_correct : decode modulus_multiple = 0%F)
+             (x : digits) : digits :=
+    carry_ carry_chain (opp modulus_multiple modulus_multiple_correct x).
```
jadephilipoom: Why are we carrying after [opp]? It should just call [sub], and we don't want to carry twice.

src/Specific/GF25519.v
```diff
@@ -345,6 +405,14 @@ Proof.
   + intros; rewrite opp_correct, opp_opt_correct; reflexivity.
 Qed.
 
+
+(** TODO(jadep from jgross): Fill me in *)
+Lemma carry_field25519 : @field fe25519 eq zero one carry_opp carry_add carry_sub mul inv div.
```
jadephilipoom: Do we want the field defined at this level of abstraction? We really only need the extra carries for fields with bounded types; this has unbounded Z. It makes more sense to me to have a different type with bounds built-in, and define all new operations for that type. We can prove them equivalent to the unbounded operations and then prove that the new type and operations form a field.

src/ModularArithmetic/ModularBaseSystem.v
JasonGross: @andres-erbsen told me to leave the non-carrying versions in there.  I made the design decision to duplicate the very boilerplate `carry_ carry_chain (f args)`, rather than duplicating the definition of `opp`, etc.  Do you think it would be better to copy/paste the definition of `opp`, and use `carry_sub` instead?  (I see that `carry_{pow,inv,div}` are unnecessary, though; I'll remove them.)

jadephilipoom: I think using `carry_sub` makes more sense.

JasonGross: Done

jadephilipoom:
Looks good to me. Once it builds, go ahead and merge it. Then I'll prove the admits--should be pretty straightforward.

JasonGross:
Builds with 8.5pl2; the other build failures are common to master (which should now be fixed), so I'm merging it.

PR #81, opened 2016-10-20
Start instantiating boundedness things

JasonGross:
Rob, can you merge this with your branch, assuming it doesn't conflict with anything?  It's a merge of master in rsloan-phoas, together with three commits at the end that implement what I could of the boundedness integration.  (Note that I still need to replace `Z` with `word`.)


JasonGross:
Superseded by #83

PR #86, opened 2016-10-24
Use Bounded Z in GF25519Bounded and fill Ed25519 with bounded things

JasonGross:
This plugs admitted lemmas for the boundedness proofs into Experiments/Ed25519.

This should not change the code that's being run (though I haven't looked at the extracted code at all; e.g., we might need more extraction directives).

Currently, `feDec` is unsound; I was under the impression that `unpack` requires 25/26 bounds on its input to produce sane output.  But `feDec` gives only the guarantee that the elements are bounded by the wordsize.  So there's something fishy going on here. (@andres-erbsen @jadephilipoom do you know what's going on?)

Reverting the change from bounded words to bounded Z should take no more than 2.5 hours (an overestimate for how long it took me to make the bounded-Z changes for PR), if we decide to go that way.

@andres-erbsen I'd like to assign you to inspecting the extracted code and deciding what, if anything, needs further unfolding / extraction directives.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -129,6 +129,7 @@ src/Specific/GF1305.v
 src/Specific/GF25519.v
 src/Specific/GF25519Bounded.v
 src/Specific/GF25519BoundedCommon.v
+src/Specific/GF25519BoundedCommonWord.v
 src/Specific/SC25519.v
 src/Specific/FancyMachine256/Barrett.v
 src/Specific/FancyMachine256/Core.v
--- a/src/Assembly/GF25519BoundedInstantiation.v
+++ b/src/Assembly/GF25519BoundedInstantiation.v
@@ -5,7 +5,7 @@ Require Import Crypto.Assembly.Compile.
 Require Import Crypto.Assembly.LL.
 Require Import Crypto.Assembly.GF25519.
 Require Import Crypto.Specific.GF25519.
-Require Import Crypto.Specific.GF25519BoundedCommon.
+Require Import Crypto.Specific.GF25519BoundedCommonWord.
 Require Import Crypto.Util.Tactics.
 Require Import Crypto.Util.LetIn.
 Require Import Crypto.Util.Tuple.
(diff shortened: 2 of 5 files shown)


jadephilipoom:
As far as I can tell, [GF25519.unpack], which is used in feDec, outputs digits with the correct 25/26 bounds. When I try:

`
Eval compute in option_map (Tuple.map (n := 10) Z.log2) (feDec (feEnc (ModularBaseSystem.encode (F.of_Z GF25519.modulus (GF25519.modulus - 2))))).
`
I get
`Some
         (24%Z, 25%Z, 24%Z, 25%Z, 24%Z, 25%Z, 24%Z, 25%Z, 24%Z,
         25%Z)
`

Do you have a counterexample?

JasonGross:
Oh, okay, I guess I made a mistake in the spec then.  What about `pack`?  Does that give anything other than wordsize bounds on the output?

jadephilipoom:
`pack` requires that its input be bounded with the 25/26 bounds (in particular, iinput should have been run through `freeze` first). Its output will be "wire format"--meaning totally full digits, all bits used. So yes, nothing other than wordsize bounds on the output.

JasonGross:
Ewww, so this means that `pack` does not return the same type that `unpack` takes in :-(

JasonGross:
Updated to not require anything other than wordsize bounds on `pack`/`unpack`, and remove the corresponding `admit`.

jadephilipoom:
Looks correct now.

andres-erbsen:
I am failing to get extraction output to compile after this. In particular, it seems that having `Z` judgementally equal to `Word64` is necessary for `fe25519WToZ`, `wire_digitsWToZ`, and `exist_fe25519` to typecheck. Here are the extraction directives I am using:

    Extract Inlined Constant GF25519BoundedCommon.word64 => "Data.Word.Word64".
    Extract Inlined Constant GF25519BoundedCommon.word64ToZ => "Prelude.fromIntegral".
    Extract Inlined Constant GF25519BoundedCommon.ZToWord64 => "Prelude.fromIntegral".
    Extract Inlined Constant GF25519BoundedCommon.w64eqb => "(Prelude.==)".
    Extraction Inline GF25519BoundedCommon.Build_bounded_word.
    Redirect "/tmp/sign.hs" Recursive Extraction sign.

Here is the error message:

    Sign.hs:1164:3: error:
        • Couldn't match type ‘Data.Word.Word64’ with ‘Prelude.Integer’
          Expected type: Fe25519
            Actual type: Fe25519W
        • In the expression: x
          In an equation for ‘fe25519WToZ’: fe25519WToZ x = x
    
    Sign.hs:1168:3: error:
        • Couldn't match type ‘Data.Word.Word64’ with ‘Prelude.Integer’
          Expected type: Wire_digits
            Actual type: Wire_digitsW
        • In the expression: x
          In an equation for ‘wire_digitsWToZ’: wire_digitsWToZ x = x
    
    Sign.hs:1449:33: error:
        • Couldn't match type ‘Prelude.Integer’ with ‘Data.Word.Word64’
          Expected type: Fe0
            Actual type: (((((((((Prelude.Integer, Prelude.Integer),
                                 Prelude.Integer),
                                Prelude.Integer),
                               Prelude.Integer),
                              Prelude.Integer),
                             Prelude.Integer),
                            Prelude.Integer),
                           Prelude.Integer),
                          Prelude.Integer)
        • In the expression:
            (,)
              ((,)
                 ((,) ((,) ((,) ((,) ((,) ((,) ((,) x0 x1) x2) x3) x4) x5) x6) x7)
                 x8)
              x9
          In a case alternative:
              (,) x0 x1
                -> (,)
                     ((,)
                        ((,) ((,) ((,) ((,) ((,) ((,) ((,) x0 x1) x2) x3) x4) x5) x6) x7)
                        x8)
                     x9
          In the expression:
            case p6 of {
              (,) x0 x1
                -> (,)
                     ((,)
                        ((,) ((,) ((,) ((,) ((,) ((,) ((,) x0 x1) x2) x3) x4) x5) x6) x7)
                        x8)
                     x9 }

JasonGross:
That's probably because I was relying on coercion insertion, and now it's just unfolding things.  I'll take a stab at it tomorrow, or, if you want to try tonight, inserting the conversion calls into the various functions that currently just do type annotations should work.  (Alternatively, a `Local Opaque word64` might also be enough.)

JasonGross:
> I am failing to get extraction output to compile after this.

I've pushed a change opacifying `word64`.  When I do extraction and run `ghc`, I get a spew of messages like
```
/tmp/sign.hs:260:20: Not in scope: ‘Data.Bits.testBit’

...

/tmp/sign.hs:1188:3:
    Not in scope: type constructor or class ‘Data.Word.Word64’

...
```

andres-erbsen:
Yep, that is a separate issue. I am not aware of any way to make extraction add import statements to the beginning of the file.

JasonGross:
8.4 build is broken, but it was already broken on master, so I'm merging.

JasonGross:
> I am not aware of any way to make extraction add import statements to the beginning of the file.

How about this truly horrible kludge?

```coq
Extraction Language Haskell.

Extract Constant nat_rec => "nat_rect                                                      {--}
                                                                                           import foo".

Definition foo := nat_rec.

Recursive Extraction foo.
```
gives
```haskell
module Main where

import qualified Prelude

data Nat =
   O
 | S Nat

nat_rec :: a1 -> (Nat -> a1 -> a1) -> Nat -> a1
nat_rec = nat_rect                                                      {--}                                                                                            import foo

foo :: a1 -> (Nat -> a1 -> a1) -> Nat -> a1
foo f f0 n =
  nat_rec f f0 n


```

Alternatively @zeldovich might know a way; I believe he's worked on Haskell extraction a bit.

JasonGross:
Oh, ew, I didn't realize emacs was just line-wrapping my spaces there.... I wonder if I can get it to insert the line-breaks necessary to get an `import` to show up at the beginning of a line...

zeldovich:
We use a slightly less fragile kludge: we pass `-F -pgmF ./fiximports.py` to GHC when compiling Coq-extracted Haskell code, where `fiximports.py` is:

  https://github.com/mit-pdos/fscq-impl/blob/master/src/fiximports.py

PR #87, opened 2016-10-24
Generalize add_coordinates

JasonGross:
This should give us the option to remove carries between add and mul
when we do bounded things.

Submitted as a PR for the purposes of sanity-checking.


andres-erbsen:
LGTM assuming all goals in the `Admitted` lemma are for nonzero denominators.

JasonGross:
I'll look into whether or not that's the case, and also look at how much slower this change makes things, and see if I can find a formulation of `admit` that makes it harder to accidentally admit things other than nonzero denominator obligations.

andres-erbsen:
The nonzero admits are the the second item on my todo list anyway, so don't put a lot of effort into there. It will take me at least a couple of days to get there, though.

PR #91, opened 2016-11-01
Fix JGross' Word Admits

varomodt:
@JasonGross Almost done -- I'm using a PR so you can review it easily


src/Reflection/Z/Interpretations.v
```diff
@@ -342,50 +399,204 @@ Module BoundedWord64.
                  | progress autorewrite with push_word64ToZ
                  | progress repeat apply conj ].
 
-  Tactic Notation "admit" := abstract case proof_admitted.
+  Ltac ktrans k := do k (etransitivity; [|eassumption]); assumption.
+  Ltac trans' := first [ assumption | ktrans ltac:1 | ktrans ltac:2 ].
```
JasonGross: This is not valid in 8.6.  Try `ltac:(1)` and `ltac:(2)`?

src/Reflection/Z/Interpretations.v
```diff
@@ -342,50 +399,204 @@ Module BoundedWord64.
...
+      unfold Word64.word64ToZ; rewrite wordToN_wminus;
+      apply Z.le_add_le_sub_l in H; abstract first
+    [ sub_mono
+    | transitivity (lower1 - lower0)%Z; [assumption|]; sub_mono
+    | eapply Z.le_lt_trans; [|eassumption]; apply Z.log2_le_mono; sub_mono ].
   Defined.
 
   Definition mul : t -> t -> t.
   Proof.
+    Ltac mul_mono :=
```
JasonGross: Please don't put tactics inside proofs.  Do `Local Ltac ...` right above the theorem statement.

src/Reflection/Z/Interpretations.v
```diff
@@ -342,50 +399,204 @@ Module BoundedWord64.
...
+    [ mul_mono
+    | transitivity (lower1 * lower0)%Z; [assumption|]; mul_mono
+    | eapply Z.le_lt_trans; [|eassumption]; apply Z.log2_le_mono; mul_mono ].
   Defined.
 
   Definition shl : t -> t -> t.
   Proof.
-    build_binop Word64.w64shl ZBounds.shl; t_start;
-      admit.
+    Ltac shl_mono := etransitivity;
```
JasonGross: Same here: move the `Ltac` up, make it local

src/Reflection/Z/Interpretations.v
```diff
@@ -342,50 +399,204 @@ Module BoundedWord64.
...
+      | _ => progress trans'
+      | _ => progress omega
+      end).
   Defined.
 
   Definition shr : t -> t -> t.
   Proof.
-    build_binop Word64.w64shr ZBounds.shr; t_start;
-      admit.
+    Ltac shr_mono := etransitivity;
```
JasonGross: And here: move the `Ltac`

src/Reflection/Z/Interpretations.v
```diff
@@ -164,10 +221,11 @@ Module ZBounds.
   Definition shr : t -> t -> t
     := t_map2 (fun lx ly ux uy => {| lower := lx >> uy ; upper := ux >> ly |}).
   Definition land : t -> t -> t
-    := t_map2 (fun lx ly ux uy => {| lower := 0 ; upper := Z.min ux uy |}).
+    := t_map2 (fun lx ly ux uy => {| lower := 0 ;
+                                     upper := 2^(Z.succ (Z.min (Z.log2 ux) (Z.log2 uy))) |}).
```
JasonGross: This is unnecessarily tight.  There should be a proof already in `ZUtil` or the standard library that `Z.land x y <= x` and `Z.land x y <= y` for nonnegative things.

src/Reflection/Z/Interpretations.v
```diff
@@ -164,10 +221,11 @@ Module ZBounds.
...
   Definition lor : t -> t -> t
     := t_map2 (fun lx ly ux uy => {| lower := Z.max lx ly;
-                                     upper := 2^(Z.max (Z.log2 (ux+1)) (Z.log2 (uy+1))) - 1 |}).
+                                     upper := 2^(Z.succ (Z.max (Z.log2 ux) (Z.log2 uy))) |}).
```
JasonGross: Isn't there a tighter bound?  It would be nice if `lor (2^k-1) (2^k-1)` said that the upper bound is `2^k-1`, not `2^k`.  (That is, it would be nice if it's always possible to saturate the upper bounds, and if it's the case that repeatedly applying an operation that doesn't change the value won't make the upper bound grow without bound.)

src/Util/WordUtil.v
```diff
@@ -271,43 +271,96 @@ Proof.
...
-  := (forall {sz} (x y : word sz),
-         (0 <= Zop (Z.of_N (wordToN x)) (Z.of_N (wordToN y))
-          -> Z.log2 (Zop (Z.of_N (wordToN x)) (Z.of_N (wordToN y))) < Z.of_nat sz
-          -> Z.of_N (wordToN (wop x y)) = (Zop (Z.of_N (wordToN x)) (Z.of_N (wordToN y))))%Z).
+Local Notation bounds_2statement wop Zop := (forall {sz} (x y : word sz),
+     (0 <= Zop (Z.of_N (wordToN x)) (Z.of_N (wordToN y))
+  -> (Z.log2 (Zop (Z.of_N (wordToN x)) (Z.of_N (wordToN y))) < Z.of_nat sz)
+  -> (Z.of_N (wordToN (wop x y)) = (Zop (Z.of_N (wordToN x)) (Z.of_N (wordToN y)))))%Z).
+
+Require Import Crypto.Assembly.WordizeUtil.
```
JasonGross: Please move these to the top of the file.

PR #97, opened 2016-11-10
Freeze stubs

JasonGross:
@andres-erbsen Here are some stubs in src/Specific/GF25519BoundedCommon.v

I leave it to you to decide when to merge this; it currently breaks compilation of extracted code by having an `admit` in computational code.  I haven't removed the things I put in to handle reifying conditional subtract yet, but it should be pretty easy to remove them once this is merged.

(We should have talked about this earlier; I probably could have gotten the spurious carries dropped in time for the paper if we had taken this path from the beginning; I wouldn't have had any of the nasty n-tuple proofs that I spent so long struggling with.


src/Specific/GF25519BoundedCommon.v
```diff
@@ -689,3 +689,51 @@ Notation iunop_WireToFE_correct_and_bounded irop op
...
+dlet a25 := (Z.shiftr a24 25 + Z.land a15 67108863)%Z in
+dlet a26 := (Z.shiftr a25 26 + Z.land a16 33554431)%Z in
+dlet a27 := (Z.shiftr a26 25 + Z.land a17 67108863)%Z in
+dlet a28 := (Z.shiftr a27 26 + Z.land a18 33554431)%Z in
+
+((Z.land a28 33554431 )%Z, (Z.land a27 67108863 )%Z,
+(Z.land a26 33554431 )%Z, (Z.land a25 67108863 )%Z,
+(Z.land a24 33554431 )%Z, (Z.land a23 67108863 )%Z,
+(Z.land a22 33554431 )%Z, (Z.land a21 67108863 )%Z,
+(Z.land a20 33554431 )%Z, ( Z.land a19 67108863 )%Z).
```
andres-erbsen: This seems to be just three carry loops. Could `GF25519.ge_modulus` also go through the bounds checking pipeline and get a proof that it returns the same thing as it would in Z? I realize it is not really great design to have it in terms of Z in the first place, but eh, we have it there now and going this way might be easy.

andres-erbsen:
Why would we want to reify conditional subtract?

src/Specific/GF25519BoundedCommon.v
JasonGross: Already there, as `ge_modulusW`

JasonGross:
Who are you asking?

andres-erbsen:
Oh, nevermind, I misread. "haven't removed the things I put in to handle reifying conditional subtract" indeed means they *will* go away.

JasonGross:
>  they *will* go away.

Yes.  I meant to write "haven't yet", but misplaced my "yet".  Oops.

jadephilipoom:
I've written the word version of conditional subtract and am working on filling in the stubs. I may leave a few admits if the proofs take too long.

JasonGross:
Merged via #104

PR #98, opened 2016-11-11
Remove special conditional subtract

JasonGross:
This is on top of #97.  Here's the current list of `Print Assumptions` in 8.6:
```coq
Axioms:
WordUtil.wlast_combine : forall (sz : nat) (a : Word.word sz) (b : bool),
                         WordUtil.wlast (Word.combine a (Word.WS b Word.WO)) =
                         b
WordUtil.wlast : forall sz : nat, Word.word (sz + 1) -> bool
WordUtil.winit_combine : forall (sz : nat) (a : Word.word sz)
                           (b : Word.word 1),
                         WordUtil.winit (Word.combine a b) = a
WordUtil.winit : forall sz : nat, Word.word (sz + 1) -> Word.word sz
shiftr_range : forall a n m : Z,
               (0 <= a < 2 ^ (n + m))%Z -> (0 <= Z.shiftr a n < 2 ^ m)%Z
GF25519BoundedCommon.proof_admitted : False
 used in postfreezeW_subproof to prove: GF25519BoundedCommon.fe25519W ->
                                        GF25519BoundedCommon.fe25519W
prime_q : Znumtheory.prime (2 ^ 255 - 19)
prime_l : Znumtheory.prime (2 ^ 252 + 27742317777372353535851937790883648493)
pow2_mod_range : forall a n m : Z,
                 (n <= m)%Z -> (0 <= ZUtil.Z.pow2_mod a n < 2 ^ m)%Z
GF25519BoundedCommon.postfreezeW_correct_and_bounded :
forall x : GF25519BoundedCommon.fe25519W,
GF25519BoundedCommon.is_bounded (GF25519BoundedCommon.fe25519WToZ x) = true ->
GF25519BoundedCommon.fe25519WToZ (GF25519BoundedCommon.postfreezeW x) =
GF25519BoundedCommon.postfreeze (GF25519BoundedCommon.fe25519WToZ x) /\
GF25519BoundedCommon.is_bounded
  (GF25519BoundedCommon.fe25519WToZ (GF25519BoundedCommon.postfreezeW x)) =
true
GF25519BoundedCommon.postfreeze : GF25519.fe25519 -> GF25519.fe25519
Tuple.map_S : forall (n : nat) (A B : Type) (f : A -> B)
                (xs : Tuple.tuple' A n) (x : A),
              Tuple.map f (xs, x) = (Tuple.map f xs, f x)
Tuple.map2_S : forall (n : nat) (A B C : Type) (f : A -> B -> C)
[...]

PR #100, opened 2016-11-11
Make [neg] a unary operation in reflection

JasonGross:
It now takes a meta-level [Z] argument which is the bit width.  I haven't modified any of the extraction directives, so I don't know if extraction still works nicely.

On top of #97 and #98


JasonGross:
Merged via #104

PR #102, opened 2016-11-11
New Word/Z Util Lemmas

varomodt:
Most of the admits in Ed25519.v are already proved. Keeping this PR open 'til I prove the rest.

I'll refactor the proofs I imported into WordUtil later: I know they're messy.

@JasonGross


JasonGross:
Thanks!  I've manually checked the build; it works with my changes on 8.4pl3 and 8.6, so I'm merging it.

PR #107, opened 2016-11-16
Generalize add_coordinates

JasonGross:
This is in preparation for dropping extra carries.  What remains to be
done, after this and #106, is to finish packaging up the reified
[add_coordinates] so that it can operate on [Tuple.tuple
GF25519BoundedCommon.fe25519 4], and then to prove

```coq
forall twice_d P1 P2, Tuple.fieldwise GF25519BoundedCommon.eq
  (<reflected add_coordinates> twice_d P1 P2)
  (@ExtendedCoordinates.Extended.add_coordinates GF25519BoundedCommon.fe25519
        GF25519Bounded.add GF25519Bounded.sub GF25519Bounded.mul twice_d
   P1 P2)
```

I'm not sure how to do this, or even what the right structure for the
proof is.

### DIFF
--- a/src/CompleteEdwardsCurve/ExtendedCoordinates.v
+++ b/src/CompleteEdwardsCurve/ExtendedCoordinates.v
@@ -92,6 +92,35 @@ Module Extended.
     Global Instance Proper_to_twisted : Proper (eq==>E.eq) to_twisted. Proof. safe_bash. Qed.
     Lemma to_twisted_from_twisted P : E.eq (to_twisted (from_twisted P)) P. Proof. unsafe_bash. Qed.
 
+    Section Proper.
+      Global Instance point_Proper : Proper (fieldwise (n:=4) Feq ==> iff)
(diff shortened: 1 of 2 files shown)


andres-erbsen:
Perhaps something like https://github.com/mit-plv/fiat-crypto/blob/master/src/Specific/GF25519.v#L762 ?

JasonGross:
That looks like a good intermediate!  Does a similar proof work with the `carry_*` versions?

andres-erbsen:
I think it should, but I am not sure which version you will want.

JasonGross:
If you give me
```coq
Tuple.fieldwise eq
    (@ExtendedCoordinates.Extended.add_coordinates _ add sub mul td P Q)
    (@ExtendedCoordinates.Extended.add_coordinates _ carry_add carry_sub mul td P Q)
```
I think I should be able to make something work from there.

Issue #14, opened 2016-06-28
A Big Rename


JasonGross:
Is there any reason that we use `CompleteEdwardsCurve/CompleteEdwardsCurveTheorems.v` rather than `CompleteEdwardsCurve/Theorems.v`?  From my experience with the category theory library in HoTT, managing the complete namespace in each file independently of the directory structure becomes a headache, and can easily lead to inconsistent styling.

andres-erbsen:
I agree that repeating the directory name in each file name is redundant. The current directory structure outside `Spec` has not received much thought. Perhaps better names would be `Curves/Edwards/Affine.v` and `Curves/Edwards/Extended.v` -- "Theorems" isn't particularly informative in a Coq development either. Generalizing from this a little bit, I think we should mirror EFD's `shape/coordinates/formula`  structure for elliptic curves because (1) the properties we verify mostly match the properties they verify (2) I don't see a problem with EFD's directory structure.
It's still unclear how expressive a name should be: `CompleteTwistedEdwards` vs `CompleteEdwards` vs `TwistedEdwards` vs `Edwards` -- for some reason i included the completeness requirement in the name but omitted the specific curve shape. Including both seems like a slippery slope...

andres-erbsen:
I think we should do a cleanup/rename pass on the repository, perhaps the next time I feel like doing low-brain work. Here is what I propose:

1. Keep `Spec/` as-is because it is fine
2. Remove `BaseSystem` and `ModularBaseSystem` and their users
3. Remove `Assembly` and all of its users
4. Remove `Spec/Encoding`
5. Remove `PointEncoding`
6. See whether we still use `VerdiTactics`, remove it unless we use it heavily
7. I wouldn't touch `Reflection` because I don't understand it well enough
8. Consolidate elliptic curve files to paths like `Curves/Weierstrass/Projective.v`, `Curves/Edwards/Twisted`. Spec still stays in spec.
9. Remove most of `Specific`, only keep FancyMachine stuff?
10. Remove `SpecificGen`
12. Create a new directory for "BaseSystem" kind of stuff, perhaps "Arithmetic"?
13. Remove `coqprime-8.4`? I can't remember the project built on 8.4, so meh, we probably are not going to fix it.

jadephilipoom:
++

6. IIRC, `break_if` requires `VerdiTactics`, and is used in many places.
[...]

PR #111, opened 2017-02-13
Rework Algebra library and tactics, prove Weierstrass curve group law

andres-erbsen:
This branch introduces multiple significant improvements to the high-level algebra library.

- `fsatz`, a tactic that implications between equalities and inequalities of rational functions. It is intended to be complete in algebraically complete fields, and has a [test suite](https://github.com/mit-plv/fiat-crypto/compare/fsatz?expand=1#diff-c51c67c9209b2248e600fec991708a8aR338). As of right now, I don't have a goal that Sage or Magma can prove but `fsatz` cannot (it is a couple of times slower, though).
    * `IntegralDomain.prove_constant_nonzero` proves foals of the form `e <> 0` in integral domains where `e` can contain `1`, addition and multiplication. It is intended to be complete, but currently is only complete if `e` is syntactically a product of prime factors.
    * `inverses_to_equations_by field solver_tac` replaces `common_denominator_in_all`. Unlike its predecessor, it does not change anything other than inverses, and does not cause the size of the goal to blow up
- A proof that Weierstrass curve points form a group (a ~15-minute compile on my machine)
- A proof that Montgomery curve addition corresponds to Weierstrass curve addition
- Significant cleanup of CompleteEdwardsCurve (especially CompleteEdwardsCurve.Pre)
    * instead of re-proving nonzero-denominator goals using ad-hoc tactic in 5 places, now we prove them when we define point addition
- theory of lower bounds on ring characteristics
- homomorphism from `Z` to any ring `R` (by Jade)
- split `Algebra.v` into multiple files with backwards-compatible names

This branch builds up to (but excluding) `src/Specific/GF25519ExtendedAddCoordinates.v` and `src/Experiments/Ed25519.v`. Both Jason and Jade also have high-value feature branches that build except for Ed25519, and a rewrite of `Experiments/Ed25519` is long overdue anyway. Let's discuss the logistics of re-integrating work in a way less fragile than current Ed25519 at the Friday meeting.

Smaller TODOs:

- [x] should the now-unused `only_two_square_roots` tactics be kept around or deleted?
- [ ] clean up Spec/MontgomeryCurve, prove it isomorphic to WeierstrassCurve
- [ ] rename all files for curves to something saner (see #14)
- [x] change the strategy for nonzero-denominator goals in ExtendedCoordinates to match that of CompleteEdwardsCurve
- [x] prove `1 <= N.pos x`
- [ ] prove `F p` has characteristic greater than `p-1`
- [ ] make a weakening lemma for "characteristic greater than"
- [x] write a fast unverified integer factorization function and use it in `IntegralDomain.solve_constant_nonzero`, then remove `char_huge` precondition from WeierstrassCurveTheorems


src/CompleteEdwardsCurve/ExtendedCoordinates.v
```diff
@@ -231,8 +206,8 @@ Module Extended.
                  | [ |- and _ _ ] => split
                  | [ H : ?x = ?y |- context[?x] ] => is_var x; rewrite H
                  | _ => reflexivity
-                 end. }
-      Qed.
+                 end. admit. }
+      Admitted. (* TODO: FIXME *)
```
JasonGross: Do you have a sense of what's left to do here?

src/Util/Tactics.v
```diff
@@ -27,6 +27,16 @@ Ltac contains search_for in_term :=
   | appcontext[search_for] => idtac
   end.
 
+Ltac debuglevel := constr:(0%nat).
+
+Ltac solve_debugfail tac :=
```
JasonGross: This tactic is closer to `try solve`, right?  Why are we succeeding when it fails to solve the goal, rather than emitting an error message?

src/WeierstrassCurve/WeierstrassCurveTheorems.v
```diff
@@ -0,0 +1,182 @@
...
+           _
+           (match W.coordinates P return _ with
+            | inl (x1, y1) => inl (x1, Fopp y1)
+            | _ => P
+            end)
+           _.
+    Next Obligation. t. Qed.
+
+    Global Instance commutative_group : abelian_group(eq:=W.eq)(op:=W.add)(id:=W.zero)(inv:=inv).
+    Proof. Time t. Time Qed.
```
JasonGross: Do you want to split apart this proof script, so that first we split up the goal, and then we run `t` on all the subgoals?  This way, in 8.5, we can do `par: t.`

src/WeierstrassCurve/WeierstrassCurveTheorems.v
```diff
@@ -0,0 +1,182 @@
...
+          (xA yA : F)
+          (A : yA ^ 2 = xA ^ 3 + a * xA + b)
+          (xB yB : F)
+          (B : yB ^ 2 = xB ^ 3 + a * xB + b)
+          (Hx: xA = xB)
+          (Hy:yB <> Fopp yA)
+      : yB = yA.
+    Proof. fsatz. Qed.
+
+    Definition is_redundant {T} (x:T) := x.
```
JasonGross: Maybe this should get its own small util file?

andres-erbsen:
I am thinking of putting this on master soon, without fixing the remaining todos, ExtendedCoordinates importers or PointeEncoding.

andres-erbsen:
@JasonGross do you have any objections to merging this and removing ./src/Specific/GF25519ExtendedAddCoordinates.v and ./src/Specific/GF25519Reflective/Reified/AddCoordinates.v plus whatever depends on them? I would like to start merging the work on this branch with Jade's.

JasonGross:
I have no objection to removing `./src/Specific/GF25519ExtendedAddCoordinates.v`

I'm a bit concerned about removing `./src/Specific/GF25519Reflective/Reified/AddCoordinates.v`, since it'll be a bit harder to appropriately generalize from just LadderStep.  How much work is it to make `./src/Specific/GF25519Reflective/Reified/AddCoordinates.v` work?

JasonGross:
Also, I'd like to make it so that the code builds on Travis, even if that means disabling files from the Makefile

andres-erbsen:
@JasonGross again, I am not understanding why this build is failing. Travis doesn't show me a log, it says "There was an error while loading data.".

andres-erbsen:
heads-up to @JasonGross , @jadephilipoom : I am planning to merge this as soon as the build passes.

andres-erbsen:
Merge `fsatz`, split `Algebra`, `F` has `positive` modulus, change curve files, remove PointEncoding

PR #112, opened 2017-02-21
Merge new base system

jadephilipoom:
This should serve as a good connecting point for Jason's reflective stuff. I defined the `add` and `mul` operations for 32-bit GF25519 at the bottom of NewBaseSystem.v --Jason, feel free to move that to a different file or something if it suits you better. To get `add` and `mul` for different bases, you should be able to just change the three `Let` statements found right above the GF25519 versions.

(note : I deleted a fair amount of saturated-arithmetic code from previous versions of NewBaseSystem, because further brainstorming sessions have elicited better strategies and that code is probably getting completely thrown out in the long run.)


src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+              {mul_correct : forall x y, fst (mul x y) + word_max * snd (mul x y) = x * y}
+              {end_wt:Z} {end_wt_pos : 0 < end_wt}
+      .
+
+      Definition sat_multerm (t t' : limb) {T} (f:list limb->T) :=
+        dlet tt' := mul (snd t) (snd t') in
+              f ((fst t*fst t', runtime_fst tt') :: (fst t*fst t'*word_max, runtime_snd tt') :: nil)%list.
+      Definition sat_mul (p q : list limb) {T} (f:list limb->T) := 
+        flat_map_cps (fun t => @flat_map_cps _ _ (sat_multerm t) q) p f.
+      (* TODO (jgross): kind of an interesting behavior--it infers the type arguments like this but fails to check if I leave them implicit *)
```
JasonGross: Can you elaborate on this?  In particular, what's the term that checks, and what's the term that doesn't check?

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
+Require Import Crypto.Util.Tactics Crypto.Util.Decidable Crypto.Util.LetIn. 
+Require Import ZArith Nsatz Psatz Coq.omega.Omega.
```
JasonGross: Please qualify these `Require`s

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+
+Require Import Coq.ZArith.BinIntDef. Local Open Scope Z_scope.
+Require Import Crypto.Util.ZUtil Crypto.Util.ListUtil.
+Require Import Crypto.Util.CPSUtil.
+
+Require Import Coq.Lists.List. Import ListNotations.
+Require Crypto.Util.Tuple. Local Notation tuple := Tuple.tuple.
+Require Import Recdef.
```
JasonGross: Please qualify this `Require`

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+
+(* TODO: move *)
+Lemma fst_pair {A B} (a:A) (b:B) : fst (a,b) = a. reflexivity. Qed.
```
JasonGross: This belongs in `src/Util/Prod.v`; please move it there and end it with `Defined` rather than `Qed` (because it's simple enough that I think it should be transparent by default, in case we ever need to cast across it).

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+Lemma snd_pair {A B} (a:A) (b:B) : snd (a,b) = b. reflexivity. Qed.
+Create HintDb cancel_pair discriminated. Hint Rewrite @fst_pair @snd_pair : cancel_pair.
+
+Lemma push_id {A} (a:A) : id a = a. reflexivity. Qed.
```
JasonGross: We should find a better place for this.  Maybe a new file `src/Util/Functions.v`?

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+Create HintDb push_id discriminated. Hint Rewrite @push_id : push_id.
+
+Ltac prove_id :=
```
JasonGross: This should be `Local Ltac`, unless you're exporting it.

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+         | _ => break_if
+         | _ => break_match
+         | _ => contradiction
+         | _ => reflexivity
+         | _ => nsatz
+         | _ => solve [auto]
+         end.
+
+Create HintDb push_eval discriminated.
+Ltac prove_eval :=
```
JasonGross: This should be `Local Ltac`, unless you're exporting it

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+         | _ => split 
+         | H : _ /\ _ |- _ => destruct H
+         | H : Some _ = Some _ |- _ => progress (inversion H; subst)
+         | _ => discriminate
+         | _ => reflexivity
+         | _ => nsatz
+         end.
+
+Delimit Scope runtime_scope with RT.
+Definition runtime_mul := Z.mul. Global Infix "*" := runtime_mul : runtime_scope.
```
JasonGross: You probably want
```coq
Global Notation "a * b" := (runtime_mul a%RT b%RT) : runtime_scope.
```
and similar, or else writing `(a * (b * c))%RT` will parse as `runtime_mul a (Z.mul b c)`, which is probably not what you want?

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+Definition runtime_add := Z.add. Global Infix "+" := runtime_add : runtime_scope. 
+Definition runtime_div := Z.div. Global Infix "/" := runtime_div : runtime_scope. 
+Definition runtime_modulo := Z.modulo. Global Infix "mod" := runtime_modulo : runtime_scope.
+Definition runtime_fst {A B} := @fst A B.
+Definition runtime_snd {A B} := @snd A B.
+
+Module B.
+  Let limb := (Z*Z)%type. (* position coefficient and run-time value *)
```
JasonGross: Uhhhhh, why are you using `Let` at top-level in a module?  You should probably be using `Definition` or else `Local Notation`, or possibly `Local Definition`, since `Let` is equivalent to `Local Definition` but emits a warning.

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+               end.
+    Qed. Hint Rewrite split_id : uncps.
+    Lemma eval_split_noncps s p (s_nonzero:s<>0):
+      eval (fst (split_noncps s p)) + s*eval (snd (split_noncps s p))  = eval p.
+    Proof.
+      cbv [split_noncps];  induction p; prove_eval.
+        match goal with H:_ |- _ =>
+                        unique pose proof (Z_div_exact_full_2 _ _ s_nonzero H)
+        end; nsatz.
+    Qed. Hint Rewrite @eval_split_noncps using auto : push_eval.
```
JasonGross: Do you mean `using solve [ auto ]`?  This will leave over side-conditions that can't be solved with `auto`; I'd suggest either adding a comment that this is deliberate, or wrapping `auto` with `solve`.

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+        nsatz.
+      Qed. Hint Rewrite eval_carryterm_noncps using auto : push_eval.
+      Lemma carry_id w fw p {T} f:
+        @carry w fw p T f = f (carry_noncps w fw p).
+      Proof. cbv [carry carry_noncps]; prove_id. Qed.
+      Hint Rewrite carry_id : uncps.
+      Lemma eval_carry_noncps w fw p (fw_nonzero:fw<>0):
+        eval (carry_noncps w fw p) = eval p.
+      Proof. cbv [carry carry_noncps]; induction p; prove_eval. Qed.
+      Hint Rewrite eval_carry_noncps using auto : push_eval.
```
JasonGross: Same comment as above about `using auto` vs `using solve [ auto ]`.

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+        autorewrite with uncps push_id.
+        rewrite ListUtil.combine_update_nth_r at 1.
+        rewrite <-(update_nth_id i (List.combine _ _)) at 2.
+        rewrite <-!(ListUtil.splice_nth_equiv_update_nth_update _ _ (weight 0, 0)); cbv [ListUtil.splice_nth id];
+          repeat match goal with
+                 | _ => progress (apply Zminus_eq; ring_simplify)
+                 | _ => progress autorewrite with push_eval cancel_pair distr_length
+                 | _ => progress rewrite <-?ListUtil.map_nth_default_always, ?map_fst_combine, ?List.firstn_all2, ?ListUtil.map_nth_default_always, ?nth_default_seq_inbouns, ?plus_O_n
+                 end; trivial; lia.
+        Unshelve.
```
JasonGross: Is there a way to do this proof without `Unshelve`?  I think it'd be nice if we can, but not essential.

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+    apply f_equal2; nsatz. Qed.
+End Karatsuba.
+
+Local Coercion Z.of_nat : nat >-> Z.
+Import Coq.Lists.List.ListNotations. Local Open Scope list_scope.
+Import B.
+Require Import Crypto.Algebra.
+
+(* TODO : move *)
+Definition lift_tuple2 {R S T n} f (g:R->S)
```
JasonGross: This belongs in `src/Util/Tuple.v`, right?

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+Proof.
+  induction n; intros; cbv [chained_carries_noncps]; simpl chained_carries.
+  { apply Positional.eval_carry_noncps; auto. }
+  { rewrite chained_carries_id.
+    etransitivity;[|apply IHn; solve[auto]].
+    apply Positional.eval_carry_noncps; auto. }
+Qed. Hint Rewrite eval_chained_carries_noncps : push_eval.
+  
+
+Ltac assert_preconditions :=
```
JasonGross: `Local Ltac`?

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+           unique assert (wt 0%nat = 1) by (cbv; congruence)
+         | |- context [Positional.from_associational ?wt ?n] =>
+           unique assert (forall i, wt i <> 0) by (intros; apply Z.pow_nonzero; try (cbv; congruence); solve [zero_bounds])
+         | |- context [Positional.from_associational ?wt ?n] =>
+           unique assert (n <> 0%nat) by (cbv; congruence)
+         | |- context [Positional.carry ?wt ?i] =>
+           unique assert (wt (S i) / wt i <> 0) by (cbv; congruence)
+         end.
+
+Ltac op_simplify :=
```
JasonGross: `Local Ltac`?

src/NewBaseSystem.v
```diff
@@ -0,0 +1,551 @@
...
+         | |- context [Positional.carry ?wt ?i] =>
+           unique assert (wt (S i) / wt i <> 0) by (cbv; congruence)
+         end.
+
+Ltac op_simplify := 
+  cbv - [runtime_add runtime_mul Let_In];
+  cbv [runtime_add runtime_mul];
+  repeat progress rewrite ?Z.mul_1_l, ?Z.mul_1_r, ?Z.add_0_l, ?Z.add_0_r.
+
+Ltac prove_op sz x :=
```
JasonGross: `Local Ltac`?

src/NewBaseSystem.v
jadephilipoom: Done.

src/NewBaseSystem.v
jadephilipoom: Done.

src/NewBaseSystem.v
jadephilipoom: Done.

src/NewBaseSystem.v
jadephilipoom: I put it in CPSUtil. We can maybe move it somewhere else later, but I think that makes sense for now.

src/NewBaseSystem.v
jadephilipoom: Done.

src/NewBaseSystem.v
jadephilipoom: Done.

src/NewBaseSystem.v
jadephilipoom: Done.

src/NewBaseSystem.v
jadephilipoom: Done.

src/NewBaseSystem.v
jadephilipoom: No, this is intentional. I know it's against convention, but it seems fine in this case. The side condition that would be left over is a simple nonzero-ness proof, and calling this function makes no sense if that input is zero. Sometimes this is trivial to solve and sometimes it's not, so it's better left over.

src/NewBaseSystem.v
jadephilipoom: Same response as above.

src/NewBaseSystem.v
jadephilipoom: I tried a bit and was unsuccessful (at least without tradeoffs making the proof net uglier).

src/NewBaseSystem.v
jadephilipoom: No, it shouldn't actually have had anything to do with tuples in the first place. I correctly generalized it and moved it to Util/Sigma.v.

src/NewBaseSystem.v
jadephilipoom: No, this is exported.

src/NewBaseSystem.v
jadephilipoom: No, exported.

src/NewBaseSystem.v
jadephilipoom: No, this is exported.

src/NewBaseSystem.v
jadephilipoom: If the inner term is `(fun t => flat_map_cps (sat_multerm t) q)`, it does not check. If it's `(fun t => @flat_map_cps _ _ (sat_multerm t) q)`, it does check. I think this is because `flat_map_cps` has a third implicit argument right after the first two explicit ones, and Coq tries to infer that prematurely.

PR #117, opened 2017-02-27
Add a linear-time sigT-returning reflective identity function

JasonGross:
By parametrizing over a let-in function, we can have a constant time* `vm_compute`able function on syntax trees that returns a sigma type.

Note that this is currently more general than it needs to be, because I was playing with monads before I realized that simply parameterizing on `let_in` would work.  If you prefer, I can strip out the monadic parameterization, and make the code a bit simpler.  (If we're going to merge this at all, that is.)

*I assert constant-time based on this plot of the times on my machine for the `pexist_idf` case:
image

If we fuse that with the `ppush_let_in_on_types` operation, as is done in the second commit, we get about a 50x speedup (still about 2x slower than just walking the tree):
image

Alas, I don't see a way to get out of needing two separate reduction steps; if we fuse the two remaining reductions, it blows up to exponential time.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -106,6 +106,7 @@ src/Reflection/BoundByCastWf.v
 src/Reflection/CommonSubexpressionElimination.v
 src/Reflection/Conversion.v
 src/Reflection/CountLets.v
+src/Reflection/DependentIdentitySigmaByEq.v
 src/Reflection/Equality.v
 src/Reflection/Eta.v
 src/Reflection/EtaInterp.v
--- a/src/Reflection/DependentIdentitySigmaByEq.v
+++ b/src/Reflection/DependentIdentitySigmaByEq.v
@@ -46,40 +46,93 @@ Section demo_by_eq.
 
   Section parametrized.
     Context (M : Type -> Type)
-            (let_in : forall A B, A -> (A -> M B) -> M B)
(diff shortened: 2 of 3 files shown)


andres-erbsen:
Cool, I wasn't sure whether this could be done. The performance seems "good enough".

Instead of "constant time", do you mean linear time?

The two reductions can be chained in Ltac, right?

I think we should merge something like this if we actually decide to try to write a bounds inference algorithm in this style. Whether this is a good idea depends on (1) how nice a sigT-returning bounds pass would have been in the vaporware-world I proposed and (2) how much additional difficulty is introduced by the two recursive calls instead of one. I feel quite optimistic about (1), and have no idea about (2) yet. Thoughts?

JasonGross:
Yes, I mean linear.

Yes, they can be chained in Ltac.

> how nice a sigT-returning bounds pass would have been in the vaporware-world I proposed and 

I'm not sure what you mean by this.  What's the vaporware-world you proposed?

> how much additional difficulty is introduced by the two recursive calls instead of one. 

What's the second recursive call here?  The second call to vm_compute just instantiates the `let_in` argument held abstract in the first call with `fun _ _ x f => f x`.

andres-erbsen:
I guess we are not doing this one either. However, it is a nice illustration of the difficulties we had with PHOAS.

PR #119, opened 2017-03-01
Switch to fully uncurried form for reflection

JasonGross:
This will eventually make a number of reflective proofs easier, and I got tired of coding for the curried form of things.

Unfortunately, the correctness lemmas for AddCoordinates and LadderStep
no longer work (because of different arities), and there's a proof in
Experiments/Ed25519 that I've admitted.

The correctness lemmas will be easy to re-add when we have a more
general version that handle arbitrary type shapes.

How do you all feel about merging this presently (given that it breaks a bit of Experiments/Ed25519, as well as the things following Reified/AddCoordinates and Reified/LadderStep)?  I've commented out or admitted the non-building bits of code, so Travis should accept this.


andres-erbsen:
Honestly, I am not sure whether there is any value in keeping Ed25519.v building at all -- we mostly just want to eliminate it. Sure, we will want to look at it to know what pitfalls to avoid, but it doesn't need to be up-to-date for that. The fsatz branch breaks large chunks of Ed25519.v as well, in ways that are much more difficult to fix. And the final version we actually want would depend on the new bounds inference code. So perhaps we should remove Ed25519.v altogether to decrease friction on new code. @jadephilipoom , you looked into Ed25519 -- is there anything there that is actually conceptually difficult and would be worth preserving?

Does this PR change any pre-reflection code? If not, then I think it is fine to break Ed25519.v to the extent of this PR regardless of whether we keep the rest of it around.

JasonGross:
> Does this PR change any pre-reflection code?

No.  Literally the only file this PR touches outside the reflective bit of the pipeline is `src/Experiments/Ed25519.v`.

andres-erbsen:
Feel free to merge then

PR #120, opened 2017-03-01
Add a version of word-size-selection based on sigT

JasonGross:
This is on top of #119

Note that we're not doing anything with this, yet.


andres-erbsen:
Cool. I guess the next thing to discuss would be how to prove this correct?

andres-erbsen:
I remember from meetings that we decided to use a first-order term represesentation to implement this pass instead.

PR #123, opened 2017-03-02
Add very simple Named version of MapCast

JasonGross:
I went with the version that cast the operation transformer inserting
casts, because that's easier to do without operation codes.


src/Reflection/NamedBase/MapCast.v
```diff
@@ -0,0 +1,46 @@
...
+                        & exprf base_type_code Name (pick_typeb _ bounds) }
+    := match e in exprf _ _ t return option { bounds : interp_base_type_bounds t
+                                                       & exprf base_type_code Name (pick_typeb _ bounds) } with
+       | Var t x
+         => option_map
+              (fun bounds => existT _ bounds (Var x))
+              (lookupb (t:=t) ctx x)
+       | LetIn tx n ex tC eC
+         => match @mapf_cast ctx _ ex with
+            | Some (existT x_bounds ex')
```
andres-erbsen: Unused variable `ex'`. Shouldn't this case be returning a  `LetIn`?

src/Reflection/NamedBase/Syntax.v
```diff
@@ -0,0 +1,100 @@
...
+          Context (output_interp_base_type : base_type_code -> Type)
+                  (interp_var : forall t, var t -> output_interp_base_type t)
+                  (interp_op : forall src1 src2 dst, output_interp_base_type src1 -> output_interp_base_type src2 -> output_interp_base_type dst)
+                  (interp_let : forall tx (ex : output_interp_base_type tx)
+                                       tC (eC : var tx -> output_interp_base_type tC),
+                      output_interp_base_type tC).
+
+          Fixpoint interp_genf (ctx : Context) {t} (e : exprf t)
+            : prop_of_option (wff ctx e) -> output_interp_base_type t
+            := match e in exprf t return prop_of_option (wff ctx e) -> output_interp_base_type t with
```
andres-erbsen: Are you sure it is a good idea to use well-formedness here? Couldn't we just return `option`?

src/Reflection/NamedBase/Syntax.v
```diff
@@ -0,0 +1,100 @@
...
+                        end
+           | BinOp _ _ _ arg1 arg2
+             => @wff ctx _ arg1 /\ @wff ctx _ arg2
+           | LetIn tx n ex _ eC
+             => @wff ctx _ ex
+                /\ inject (forall v, prop_of_option (@wff (extendb (t:=tx) ctx n v) _ eC))
+           end%option_pointed_prop.
+
+      Section interp_gen.
+          Context (output_interp_base_type : base_type_code -> Type)
```
andres-erbsen: Why generalize over these variables? We use only one instantiation of this, right?

src/Reflection/NamedBase/Syntax.v
```diff
@@ -0,0 +1,100 @@
...
+Delimit Scope nexpr_scope with nexpr.
+Module Export Named.
+  Section language.
+    Context (base_type_code : Type)
+            (interp_base_type : base_type_code -> Type)
+            (Name : Type).
+
+    Inductive exprf : base_type_code -> Type :=
+    | Var {t} : Name -> exprf t
+    | BinOp {t1 t2 tR} : exprf t1 -> exprf t2 -> exprf tR
```
andres-erbsen: This is just `Add`, right?

src/Reflection/NamedBase/Syntax.v
```diff
@@ -0,0 +1,100 @@
+(** * Named Representation of Gallina *)
+Require Import Crypto.Util.PointedProp.
+Require Import Crypto.Util.Notations.
+
+Class Context {base_type_code} (Name : Type) (var : base_type_code -> Type) :=
```
andres-erbsen: This typeclass thing is weird; could we use separate functions?

src/Reflection/NamedBase/MapCast.v
```diff
@@ -0,0 +1,46 @@
...
+              => @mapf_cast (extendb (t:=tx) ctx n x_bounds) _ eC
+            | None => None
+            end
+       | BinOp t1 t2 tR arg1 arg2
+         => match @mapf_cast ctx _ arg1, @mapf_cast ctx _ arg2 with
+            | Some (existT arg1_bounds arg1'),
+              Some (existT arg2_bounds arg2')
+              => Some (existT _ _ (cast_op _ _ _ _ _ arg1' arg2'))
```
andres-erbsen: Please fill in bounds arguments to all functions, even if they can be inferred.

src/Reflection/NamedBase/MapCast.v
JasonGross: Oops, good catch.  Yes.

src/Reflection/NamedBase/MapCast.v
JasonGross: I don't anticipate getting a chance to do this before Monday; feel free to use the "maintainers can edit" feature to update the PR.  (There's a small chance I'll find time, though, and I'll fix it if I get a chance.)

src/Reflection/NamedBase/Syntax.v
JasonGross: I didn't notice it was a class when copying code.  It should work fine as a record.  Separate functions would also work, but would be a bit more of a pain.

src/Reflection/NamedBase/Syntax.v
JasonGross: Yes.  Or mul, or whatever single operation you want to pick.

src/Reflection/NamedBase/Syntax.v
JasonGross: I believe there's an alternate instantiation that will compile this code to PHOAS, and this allows for code reuse.  Maybe that's a bad thing though?

src/Reflection/NamedBase/Syntax.v
JasonGross: It's a bad idea to use well-formedness here unless we need it; I think my past self made the wrong decision, and returning option is the way to go if we can do it.  Does the let-in case break if we return option?

JasonGross: Actually, I think the let-in case does break if we use `option`.  Do you see a way to make it go through without asking for a fail-value?

src/Reflection/NamedBase/Syntax.v
JasonGross: Fixed to be `Record`

src/Reflection/NamedBase/MapCast.v
JasonGross: I found time; fixed.

src/Reflection/NamedBase/MapCast.v
JasonGross: Fixed

src/Reflection/NamedBase/Syntax.v
JasonGross: I've added a file that interprets to PHOAS using this definition.

JasonGross:
@andres-erbsen I tried adding `sub` and lower bounds (not just upper bounds) (see the latest commit on this PR), and have not succeeded at making the arithmetic side-conditions go through.  I'm also not sure that I got the selection logic right.  Can you help?

JasonGross:
We no longer need this, now that we have a fully-proven admit-free full-fledged version at https://github.com/mit-plv/fiat-crypto/commit/f2814cf2ba24fbddc3cb091a4fddd19a8c807b89

PR #124, opened 2017-03-03
Use Redirect to display less spew on Travis

### LINKED ISSUE #121: The build produces more than 10000 lines of output
The build produces more than 10000 lines of output. This makes the Travis web interface slow, and one needs to click on "raw log" to see the remaining lines of output.

I think we don't need 10000 lines of output. Most of the spew seems to come from `src/Specific/GF25519Reflective/Reified/.*Display.v`. Could we just disable this, or redirect it into a file?

JasonGross:
This closes #121


Makefile
```diff
@@ -97,14 +97,6 @@ Makefile.coq: Makefile _CoqProject
 	$(SHOW)'COQ_MAKEFILE -f _CoqProject > $@'
 	$(HIDE)$(COQBIN)coq_makefile -f _CoqProject | sed s'|^\(-include.*\)$$|ifneq ($$(filter-out $(FAST_TARGETS),$$(MAKECMDGOALS)),)~\1~else~ifeq ($$(MAKECMDGOALS),)~\1~endif~endif|g' | tr '~' '\n' | sed s'/^clean:$$/clean::/g' | sed s'/^Makefile: /Makefile-old: /g' | sed s'/^printenv:$$/printenv::/g' > $@
 
-$(DISPLAY_NON_JAVA_VO:.vo=.log) : %Display.log : %.vo %Display.v src/Reflection/Z/CNotations.vo
```
andres-erbsen: What was wrong with the previous version of this code? Do redirects not work on travis?

JasonGross: The previous version had separate targets for `*Display.vo` and `*Display.log`.  The default target built the .vo files.  The alternative to this PR is to just filter out those vo files from the default target

andres-erbsen: Yes, we should filter them out instead. I stand by my coq bug report that asks for Redirect to be removed.

PR #128, opened 2017-03-14
Add a Named version of MapCast

JasonGross:
Based on @andres-erbsen's work towards #123.

I haven't ported the example from #123, though I kept the statement of the correctness lemma basically the same (modulo changing the form of the equality-decidability we require for names and base-type-codes).

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -161,6 +161,8 @@ src/Reflection/Named/FMapContext.v
 src/Reflection/Named/InterpretToPHOAS.v
 src/Reflection/Named/InterpretToPHOASInterp.v
 src/Reflection/Named/InterpretToPHOASWf.v
+src/Reflection/Named/MapCast.v
+src/Reflection/Named/MapCastInterp.v
 src/Reflection/Named/NameUtil.v
 src/Reflection/Named/PositiveContext.v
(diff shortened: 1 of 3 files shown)


andres-erbsen:
This looks pretty much like what I wrote, yes. Nothing sticks out to me as odd. Are you interested on detailed feedback on some parts here?

Perhaps the next interesting thing to do would be to instantiate this as a word size selector parametrized on a `list nat` that contains the allowed finite word sizes?

JasonGross:
>  Are you interested on detailed feedback on some parts here?

If anything seems unreadable or unreasonable, I want to know that.  Otherwise, shall we merge?

> Perhaps the next interesting thing to do would be to instantiate this as a word size selector parametrized on a list nat that contains the allowed finite word sizes?

Do you want to see this before or after getting code for things like `add` and `mul` that we can feed into this?  And do you want me to do it, or do you want to do it?  (I've been spending some time today working on connecting this up to the reification bits and PHOAS and getting proofs that we can go back and forth working.  I could instead switch to instantiating, but there are bits that I don't know how to do (e.g., truncation), and I think you have a clearer picture of it than I do, right now, so if we should probably discuss design at least a bit if you want me to implement this.)

PR #131, opened 2017-03-17
Have cast_op return option exprf instead of op

JasonGross:
This is, unfortunately, a bit of a pain to wrap up into a PHOAS transformation,
because we have to compile the arguments to PHOAS, then apply the
transformation, then compile back.  Hopefully it won't be too tricky to
prove that the composition is the identity.

### DIFF
--- a/src/Reflection/MapCastByDeBruijn.v
+++ b/src/Reflection/MapCastByDeBruijn.v
@@ -1,8 +1,10 @@
+Require Import Coq.Numbers.BinNums.
 Require Import Crypto.Reflection.SmartMap.
 Require Import Crypto.Reflection.Named.Syntax.
 Require Import Crypto.Reflection.Named.MapCast.
 Require Import Crypto.Reflection.Named.InterpretToPHOAS.
 Require Import Crypto.Reflection.Named.Compile.
+Require Import Crypto.Reflection.Named.IdContext.
 Require Import Crypto.Reflection.Named.PositiveContext.
 Require Import Crypto.Reflection.Named.PositiveContext.Defaults.
 Require Import Crypto.Reflection.Linearize.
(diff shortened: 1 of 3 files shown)


JasonGross:
I'm thinking this may be the wrong approach.  In particular, it makes for either non-uniform or complicated cast logic.  For example, the strategy you implement [here](https://github.com/mit-plv/fiat-crypto/pull/123/files#diff-e53a99ce3199e10cdb4d60dd9b3f029dR207) fails for `shr`, because `((uint32_t)x) >> 32` is definitely not the same as `(uint32_t)(x >> 32)` for a 64-bit `x` (so you need to first operate and then cast to the output size), while `(uint64_t)(x * y)` is not the same as `((uint64_t)x) * ((uint64_t)y)` for 32-bit `x` and `y`.  So, if we stick to uni-type operations and casts, it looks like we need to cast all arguments up to the maximum of the types, and then cast the output down, and reason about all the complexity entailed by these extra casts.  @andres-erbsen Do you think this is worth it?

andres-erbsen:
Your description of the situation matches my understanding. I'm not sure what would be the best thing to do either.

- I believe the cast-operate-cast pattern would work. For targeting C, the middle type would need to be a appropriately-sized word; conceptually, doing the operation in Z makes slightly more sense.
- The machine architectures are have operations that cast on their own, for example `mul : word 32 -> word  32 -> word 64` as well as `mul' : word 32 -> word 32 -> word 32`. Pretty-printing this style as C code would not be that difficult, I think going the other way would require another pass that does some for of rewriting through let binders.

Do whichever you think we will have an overall easier time with.

JasonGross:
> Do whichever you think we will have an overall easier time with.

I think we'll have a significantly easier time with the version where we parameterize operations on the types they input and return separately.

> I believe the cast-operate-cast pattern would work. For targeting C, the middle type would need to be a appropriately-sized word; conceptually, doing the operation in Z makes slightly more sense.

Yes, I think I agree with you here; we can even do this uniformly over all operations and type possibilities: https://github.com/JasonGross/fiat-crypto/blob/wip-map-cast-by-debruijn/src/Reflection/Z/Syntax.v#L56-L85  This should allow for more generic proofs, too.

andres-erbsen:
I read we decided not to do this.

PR #138, opened 2017-04-01
Work around bug #5434

JasonGross:
This works around [bug #5434](https://coq.inria.fr/bugs/show_bug.cgi?id=5434) that `vm_compute` breaks an invariant asserted on [line 115 of `pretyping/constr_matching.ml`](https://github.com/coq/coq/blob/V8.6/pretyping/constr_matching.ml#L115).  This was triggering whenever we tried to reify one of the operations.

Does this seem fine to you two, @jadephilipoom and @andres-erbsen ?

N.B. This commit cannot work with Coq 8.4, unless I do some typeclass hackery that I didn't want to bother with.

### DIFF
--- a/src/NewBaseSystem.v
+++ b/src/NewBaseSystem.v
@@ -903,6 +903,34 @@ Positional.to_associational_cps Positional.to_associational Positional.eval Posi
                   (@runtime_mul));
   [replace_with_vm_compute t1; clear t1|reflexivity].
 
+(** This block of tactic code works around bug #5434
+    (https://coq.inria.fr/bugs/show_bug.cgi?id=5434), that
+    [vm_compute] breaks an invariant in pretyping/constr_matching.ml.
+    So we refresh all of the names in match statements in the goal by
+    crawling it. *)
+Ltac replace_match_with_destructuring_match T :=
+  match T with
+  | ?F ?X
+    => let F' := replace_match_with_destructuring_match F in
+       let X' := replace_match_with_destructuring_match X in
+       constr:(F' X')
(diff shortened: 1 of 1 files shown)


andres-erbsen:
I don't understand how this tactic works around the linked bug. More comment would be appreciated.

JasonGross:
How about this?
```coq
(** This block of tactic code works around bug #5434
    (https://coq.inria.fr/bugs/show_bug.cgi?id=5434), that
    [vm_compute] breaks an invariant in pretyping/constr_matching.ml.
    So we refresh all of the names in match statements in the goal by
    crawling it.

    In particular, [replace_with_vm_compute] creates a [vm_compute]d
    term which has anonymous binders where pretyping expects there
    to be named binders.  This shows up when you try to match on the
    function (the branch statement of the match) with an Ltac pattern like
    [(fun x : ?T => ?C)] rather than [(fun x : ?T => @?C x)]; we use the
    former in reification to save the cost of many extra invocations of
    [cbv beta].  Luckily, patterns like [(fun x : ?T => @?C x)] don't trigger
    this anomaly, so we can walk the term, fixing all match statements
    whose branches are functions whose binder names were eaten by
    [vm_compute] (note that in a match, every branch where the
    corresponding constructor takes arguments is represented internally
    as a function (lambda term)).  We fix the match statements by
    pulling out the branch with the [@?] pattern that doesn't trigger the
    anomaly, and then recreating the match with a destructuring [let]
    that hasn't been through [vm_compute], and therefore has name
    information that constr_matching is happy with. *)
```

andres-erbsen:
Yep, that does the trick even without looking up the bug. thanks. Merge as desired.

PR #141, opened 2017-04-02
Reflective Pipeline

JasonGross:
This is not yet ready for merge, but it builds and I think I'm approximately ready for feedback.  Things still to be done:

- [x] document the pipeline files (the only one well-documented right now is `Pipeline/Definition.v` (which is the only one that should be changed in normal, non-debugging usage))
- [ ] Prove bounds lemmas for negatives on `<<`, `>>`, `&`, `|`
- [ ] Prove the remaining admit about equality of cast-op (hopefully @achlipala can help me with the second one, and give me something comparable to the 8x improvement he gave me for the first admit)
- [x] Figure out why `Defined` in `IntegrationTest` is taking forever (where forever means, currently, > 20 hours and > 10 GB of RAM)
- [x] maybe give nicer error messages when bits of the reflective machinery fail
- [x] Decide whether or not we want to get something comparable to `Experiments/Ed25519.v` (which this PR removes, because its dependencies no longer build) before merging ( @andres-erbsen ?)
- [ ] Teach reification about context variables so that it can hopefully handle `add_coordinates` and `ladderstep`.
- [ ] Make use of the `CNotations` to print some things
- [x] remove obsolete makefile targets

Sorry it's so big.  If we merge #138 and #139, I'll rebase this on top of those, and the diff will get a bit smaller.


andres-erbsen:
I will read the diff after it is rebased. I will probably have some amount suggestions for renaming and moving code.

I think `Experiments/Ed25519.v` is broken for many reasons (good and bad), and can be removed for now. We will try and make something comparable once we have `add_coordinates` building the way we want it to.

Yes, please remove the Makefile targets that depend on `Experiments/Ed25519.v`, including all targets that operate on extracted code.

JasonGross:
Rebased.  I've also separated it out into two commits that remove things, and one commit that adds things, so it'll be easier to read the relevant diff without it being cluttered by 14,000 lines of removed code.

src/Reflection/Z/Syntax/Util.v
```diff
@@ -6,18 +6,16 @@ Require Import Crypto.Reflection.TypeInversion.
 Require Import Crypto.Reflection.Z.Syntax.
 Require Import Crypto.Util.FixedWordSizesEquality.
 Require Import Crypto.Util.NatUtil.
+Require Import Crypto.Util.HProp.
```
andres-erbsen: I don't understand what is going on in this file. Maybe it is not something that should be fixed right now, though.

src/Reflection/Z/Bounds/Pipeline/Definition.v
```diff
@@ -0,0 +1,125 @@
...
+  := Build_ProcessedReflectivePackage_from_option_sigma
+       e input_bounds
+       (let e := Linearize e in
+        let e := InlineConst e in
+        let e := MapCast e input_bounds in
+        option_map
+          (fun v
+           => let 'existT b e := v in
+              existT _ b (ExprEta (InlineConst e)))
+          e).
```
andres-erbsen: Why is this just not `option_map InlineConst` or something simple like that?

src/Reflection/Z/Bounds/Pipeline/Glue.v
```diff
@@ -0,0 +1,153 @@
+(** * Reflective Pipeline: Glue Code *)
```
andres-erbsen: I think this file is a good example of minimal but sufficient comments.

src/Reflection/Z/Bounds/Pipeline/PreDefinitions.v
```diff
@@ -0,0 +1,52 @@
+(** * Definitions which are used in the pipeline, but shouldn't be changed *)
```
andres-erbsen: I think this file defines the output type of the reflective pipeline. Consider explaining and renaming (but it's okay as-is as well).

src/Reflection/Z/Syntax/Util.v
JasonGross: It's possible that I don't need `HProp`, and can remove it, if that's what you're asking about.  More generally, this file has two kinds of changes:
1. adapting to the change in the definition of `interp_base_type` (which now can return `wordT` and is not just `fun _ => Z`) and the changes in the definition of `op`, and
2. adding various useful definitions and lemmas that operate on the syntax that I didn't have a better place for (like `cast_back_flat_const` and `Zinterp_op_genericize_op`

src/Reflection/Z/Bounds/Pipeline/Definition.v
JasonGross: Because word-size-selection returns a `sigT` of bounds and an expression, and I don't have `projT2_map` (if I did, it would be `option_map (projT2_map (fun e => EtaExpr (InlineConst e))) e`).

src/Reflection/Z/Syntax/Util.v
andres-erbsen: I think my main issue is that I didn't understand the file before either. I think lemmas about casting and range analysis of Z should have their own file.

src/Reflection/Z/Bounds/Pipeline/PreDefinitions.v
JasonGross: `OutputTypeDefinition.v`?

src/Reflection/Z/Bounds/Pipeline/Definition.v
andres-erbsen: ah, okay. That form looks much nicer, especially when the variable name `e` is used only once.

src/Reflection/Z/Bounds/Pipeline/PreDefinitions.v
andres-erbsen: yes, or maybe `OutputType.v`

src/Reflection/Z/Bounds/Pipeline/Glue.v
JasonGross: I've just pushed a commit that added more comments to this file to explain some of the most arcane or fragile bits of Ltac hackery I do.  (Also a check to give nice error messages (hopefully) when things don't satisfy my preconditions.)

src/Reflection/Z/Syntax/Util.v
JasonGross: There's very little to do with casting and range analysis of `Z`; most of what appears to be about casting and range analysis of `Z` is actually about how casting and range analysis of `Z` lift over reified-type-products, and how they play with lifting operators.

The exceptions are `cast_const_id` and `cast_const_idempotent`, which are not currently used (though I think they might be used in current master), and talk about how casting plays with base-type comparisons.  Should I remove them?  (I'm not sure if I'll need them again in the future; I may need them in proving the bounds interpretation word-size-selection-side-conditions.)

src/Reflection/Z/Bounds/Pipeline/PreDefinitions.v
JasonGross: Done

src/Reflection/Z/Syntax/Util.v
andres-erbsen: I don't think I understand enough to propose next steps. Let's punt this to whenever we try to clean up the tuple-structure code.

src/Reflection/Z/Bounds/Pipeline/Definition.v
andres-erbsen: Sure.

andres-erbsen:
I don't have any further comments right now.

JasonGross:
@andres-erbsen I've rewritten the reflective-only part of the pipeline; it's now mostly in Gallina, and there's very little interesting going on in Ltac.  Also, the thing making `Defined` hang happens in the glue code...

JasonGross:
And, of course, the slowness in `Defined` is from the single line `refine (let pf := _ in ((proj2 pf : B) : let f' := proj1 pf in B))`

src/Reflection/Z/Bounds/Pipeline/Definition.v
JasonGross: Updated to
```coq
  := Build_ProcessedReflectivePackage_from_option_sigma
       e input_bounds
       (let e := Linearize e in
        let e := InlineConst e in
        let e := MapCast e input_bounds in
        option_map
          (projT2_map
             (fun b e'
              => let e' := InlineConst e' in
                 let e' := ExprEta e' in
                 e'
          e).
```

Does that look good to you?

andres-erbsen:
> Updated to
> ```coq
>   := Build_ProcessedReflectivePackage_from_option_sigma
>        e input_bounds
>        (let e := Linearize e in
>         let e := InlineConst e in
>         let e := MapCast e input_bounds in
>         option_map
>           (projT2_map
>              (fun b e'
>               => let e' := InlineConst e' in
>                  let e' := ExprEta e' in
>                  e'
>           e).
> ```
>
> Does that look good to you?

Yes.

JasonGross:
I've made some questionable modifications to Jade's side of the pipeline in order to get `Defined` to run in reasonable time (about 10s).  But it finishes now!  And I've documented pretty much all of the pipeline, so I'm ready for more comments @andres-erbsen .

We should meet at some point to get a goal for `add_coordinates` or `ladderstep` up, so I can work on getting reflection to handle that.  And we should meet to talk about if my changes to @jadephilipoom 's side of the pipeline are okay.

src/Specific/IntegrationTest.v
```diff
@@ -43,17 +72,66 @@ Section BoundedField25p5.
...
-    cbn beta iota delta [fst snd] in carry_addZ.
-    apply f_equal.
+    let carry_addZ' := (eval cbv beta iota delta [carry_addZ proj1_sig add_sig carry_sig fst snd runtime_add runtime_and runtime_mul runtime_opp runtime_shr sz] in carry_addZ) in
+    let carry_addZ'' := fresh carry_addZ in
+    rename carry_addZ into carry_addZ'';
+      pose carry_addZ' as carry_addZ;
+      replace carry_addZ'' with carry_addZ by abstract (cbv beta iota delta [carry_addZ'' proj1_sig add_sig carry_sig fst snd runtime_add runtime_and runtime_mul runtime_opp runtime_shr sz]; reflexivity);
+      clear carry_addZ''.
+    all:save_state_and_back_to_sig.
+    apply sig_f_equal.
```
andres-erbsen: This doesn't look terrifying, but I still don't understand what is going on in the last 10 lines. You seem to have removed a `change` in favor of a corresponding `replace`, and then there are some tactics that I don't get. For the replace, could this be written in the etransitivity-evaluate-reflexivity style, or is there a reason why a straight replace is nicer?

src/Specific/IntegrationTest.v
```diff
@@ -34,6 +37,32 @@ Section BoundedField25p5.
...
+    | lazymatch goal with
+      | [ |- { c | _ = ?f ?x } ]
+        => let rexpr := head x in
+           let rexprZ := fresh "rexprZ" in
+           set (rexprZ := rexpr)
+      end ].
+  (** TODO: Move me *)
+  Definition sig_f_equal {T A B} (f : A -> B) {x y : T -> A}
+             (p : { t : T | x t = y t })
+    : { t : T | f (x t) = f (y t) }
```
andres-erbsen: I guess ideally we would factor out `fun a b => f a = b a` into a variable, but I guess unification wouldn't be happy with this?

src/Specific/IntegrationTest.v
```diff
@@ -34,6 +37,32 @@ Section BoundedField25p5.
...
+
+  (** TODO MOVE ME *)
+  Ltac save_state_and_back_to_sig :=
+    [> reflexivity
+    | lazymatch goal with
+      | [ |- { c | _ = ?f ?x } ]
+        => let rexpr := head x in
+           let rexprZ := fresh "rexprZ" in
+           set (rexprZ := rexpr)
+      end ].
```
andres-erbsen: The last two tactics should have documentation

src/Reflection/Z/Bounds/Pipeline/ReflectiveTactics.v
```diff
@@ -0,0 +1,244 @@
...
+   (** ** computed output bounds are not looser than the given output bounds *)
+   (** we do subst and we don't [vm_compute] first because we want to
+       get an error message that displays the bounds *)
+   subst_let; clear; abstract vm_cast_no_check (eq_refl true) |
+   (** ** removal of a cast across the equality proof above *)
+   abstract_cbv_rhs |
+   (** ** unfolding of [interp] constants *)
+   abstract_cbv_interp_rhs |
+   (** ** boundedness of inputs *)
+   abstract handle_bounds_from_hyps ].
```
andres-erbsen: How does one maintain this list when the order of pipeline passes changes? I think this should be documented.

andres-erbsen: Perhaps it would make sense to put `src/Reflection/Z/Bounds/Pipeline.v`, `src/Reflection/Z/Bounds/Pipeline/Composition.v` and `src/Reflection/Z/Bounds/Pipeline.v` into the same file as this tactic to indicate that they need to be edited in unison.  OTOH, I think tactics whose specs don't reference the pipeline could be factored out.

src/Specific/IntegrationTest.v
JasonGross: Using `replace` lets you fully separate out the conversion work with `abstract` and decouples it from evar instantiation (in a way that I'm still just a bit fuzzy on the details of), and also lets you work locally.  The slow behavior of `Defined`, I believe, was rooted in the fact that `cbv in carry_addZ` and `change` don't insert `Cast` nodes in the proof tree, and thus cause `Defined` to sometimes try different things.  I noticed this when I discovered that if I remove the `cbv` from inside the `abstract`, the `abstract reflexivity` itself takes a really long time (I didn't wait for it to finish).  I suspect that `revert carry_addZ; cbv beta iota delta [stuff]; intro carry_addZ` would also work (since I think `cbv`, unlike `cbv in H`, does insert a `Cast` node), but there's no way to separate off the work that's done there with `abstract`.  This is in general true of the etransitivity-evaluate-reflexivity style, which is unfortunate when it comes to debugging.  One idea that I'm considering is that I'd like to have is a set of tactics that simulates the etransitivity-evaluate-reflexivity style in a way the keeps the sigma types around and doesn't have floating evars, which would allow wrapping all of the transformations in `abstract`.  But I'm not sure, and I think it might be good to bring @achlipala into our discussion here on the issues debugging slowness in the etransitivity-evaluate-reflexivity proof-by-refinement style.

src/Specific/IntegrationTest.v
JasonGross: Good idea, let me try it

src/Reflection/Z/Bounds/Pipeline/ReflectiveTactics.v
JasonGross: I'll fuse `Composition.v` and `ReflectiveTactics.v`, and pick more descriptive names for the unify-abstract-reflexivity tactics and stuff them in a `Util/Tactics/*` file.  Note that `Composition` should not need to be changed when you change the transformations in the pipeline.  I expect we'll need to change this when we need a fancier bounds relaxation lemma, or if we radically change the correctness condition of bounds analysis, or if we need new side-conditions in other passes that don't follow from well-foundedness.  The only transformations so far that need side-conditions are bounds analysis, bounds relaxation, and binder renaming (this one just because of issues with `vm_compute`).

JasonGross: Fused

src/Specific/IntegrationTest.v
JasonGross: ahahahaha, the unifier is fine with it, but somehow it triggers
```
Anomaly: File "pretyping/constr_matching.ml", line 115, characters 15-21: Assertion failed. Please
report at http://coq.inria.fr/bugs/.
```
in doing reflection.

JasonGross: Actually, nvm, I found a way around that.  I'm sure you'll love the name I gave to the binder that must not be an `_` and must not be the letter `a` (for different reasons) ;)

src/Specific/IntegrationTest.v
JasonGross: Documented

JasonGross:
I'm going to merge this, having talked with Andres.  Travis hasn't finished yet, but it builds fully on my machine, and I want to modify master without having to rebuild this.  I'll fix anything that breaks on travis.

PR #145, opened 2017-04-04
make elliptic curve proofs faster and split them into files

### LINKED ISSUE #144: Split MontgomeryX into two files
It's annoying when I have to wait 15 minutes for WeierstrassCurveTheorems to build just to use the definition of `xzladderstep`, which (I think) doesn't depend on the theorems.

andres-erbsen:
closes #144

### DIFF
--- a/src/Algebra.v
+++ b/src/Algebra.v
@@ -140,8 +140,15 @@ Section Algebra.
     Global Existing Instance field_div_Proper.
   End AddMul.
 
-  Definition char_ge {T} (eq:T->T->Prop) (zero:T) (inj:BinPos.positive->T) C := forall p, BinPos.Pos.lt p C -> not (eq (inj p) zero).
+  Definition char_ge (zero:T) (inj:BinPos.positive->T) C := forall p, BinPos.Pos.lt p C -> not (eq (inj p) zero).
   Existing Class char_ge.
+  Lemma char_ge_weaken id of_pos C
+        (HC:@char_ge id of_pos C) c (Hc:BinPos.Pos.le c C)
+    : @char_ge id of_pos c.
+  Proof. intros ??; eauto using BinPos.Pos.lt_le_trans. Qed.
+  (* TODO: make a typeclass instance that applies this lemma
+  automatically using [vm_decide] for the inequality if both
+  characteristics are literal constants. *)
 End Algebra.
 
 Section ZeroNeqOne.
--- a/src/MontgomeryXProofs.v
+++ b/src/MontgomeryXProofs.v
@@ -1,0 +1,57 @@
+Require Import Crypto.Algebra Crypto.Algebra.Field.
+Require Import Crypto.Util.Sum Crypto.Util.Prod Crypto.Util.LetIn.
+Require Import Crypto.Spec.MontgomeryCurve Crypto.MontgomeryCurve.
+Require Import Crypto.MontgomeryX BinPos.
+
+Module M.
+  Section MontgomeryCurve.
+    Context {F Feq Fzero Fone Fopp Fadd Fsub Fmul Finv Fdiv}
+            {field:@Algebra.field F Feq Fzero Fone Fopp Fadd Fsub Fmul Finv Fdiv}
+            {Feq_dec:Decidable.DecidableRel Feq}
+            {char_ge_5:@Ring.char_ge F Feq Fzero Fone Fopp Fadd Fsub Fmul 5}.
+    Local Infix "=" := Feq : type_scope. Local Notation "a <> b" := (not (a = b)) : type_scope.
+    Local Infix "+" := Fadd. Local Infix "*" := Fmul.
+    Local Infix "-" := Fsub. Local Infix "/" := Fdiv.
+    Local Notation "0" := Fzero.  Local Notation "1" := Fone.
+    Local Notation "'∞'" := (inr tt) : core_scope.
+    Local Notation "( x , y )" := (inl (pair x y)).
+
+    Let char_ge_3:@Ring.char_ge F Feq Fzero Fone Fopp Fadd Fsub Fmul (BinNat.N.succ_pos (BinNat.N.two)).
+    Proof. eapply char_ge_weaken; eauto; vm_decide. Qed.
+
+    Context {a b: F} {b_nonzero:b <> 0}.
+    Context {a24:F} {a24_correct:(1+1+1+1)*a24 = a-(1+1)}.
+    Local Notation add := (M.add(a:=a)(b_nonzero:=b_nonzero)(char_ge_3:=char_ge_3)).
+    Local Notation opp := (M.opp(a:=a)(b_nonzero:=b_nonzero)).
+    Local Notation point := (@M.point F Feq Fadd Fmul a b).
+    Local Notation xzladderstep := (M.xzladderstep(a24:=a24)(Fadd:=Fadd)(Fsub:=Fsub)(Fmul:=Fmul)).
+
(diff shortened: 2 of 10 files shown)


src/Algebra.v
```diff
@@ -140,8 +140,15 @@ Section Algebra.
...
   End AddMul.
 
-  Definition char_ge {T} (eq:T->T->Prop) (zero:T) (inj:BinPos.positive->T) C := forall p, BinPos.Pos.lt p C -> not (eq (inj p) zero).
+  Definition char_ge (zero:T) (inj:BinPos.positive->T) C := forall p, BinPos.Pos.lt p C -> not (eq (inj p) zero).
   Existing Class char_ge.
+  Lemma char_ge_weaken id of_pos C
+        (HC:@char_ge id of_pos C) c (Hc:BinPos.Pos.le c C)
+    : @char_ge id of_pos c.
+  Proof. intros ??; eauto using BinPos.Pos.lt_le_trans. Qed.
+  (* TODO: make a typeclass instance that applies this lemma
```
JasonGross:
```coq
Class Pos_le_class x y := make_pos_le : BinPos.Pos.le x y.
Ltac check_literal_pos_constant x :=
  lazymatch x with
  | BinNat.N.succ_pos ?x => check_literal_pos_constant
  | BinNat.N.two => idtac
  ....
  end.
Hint Extern 1 (Pos_le_class ?x ?y) => check_literal_pos_constant x; check_literal_pos_constant y; vm_decide : typeclass instances.
Existing Instance char_gt_weaken.
```

src/Util/Decidable.v
```diff
@@ -72,6 +71,15 @@ Ltac solve_decidable_transparent_with tac :=
 Ltac solve_decidable_transparent := solve_decidable_transparent_with firstorder.
 
 Local Hint Extern 0 => solve [ solve_decidable_transparent ] : typeclass_instances.
+(*
+Goal forall x y, Decidable (Pos.compare_cont Gt x y = Gt).
+  intros. exact _.
+(* Toplevel input, characters 6-7: *)
+(* > exact _. *)
+(* >       ^ *)
+(* Anomaly: Uncaught exception Invalid_argument("List.fold_right2"). Please report at *)
```
JasonGross: Did you report it?  Or minimize it with the [bug minimizer](https://github.com/JasonGross/coq-tools)?

src/Algebra.v
andres-erbsen: Ugh, later then.

src/Util/Decidable.v
andres-erbsen: Minimized manually, haven't reported or minimized using code. I would appreciate if you did that, if not I will look into this tomorrow.

src/Algebra.v
JasonGross: Sure.  You had a todo, and I I knew how to solve it, so I figured I'd sketch the solution for you to use if you want to.  Feel free to not use it if it's more trouble than it's worth.

andres-erbsen:
I think I am done with this for now (worksforme); going to sleep.

JasonGross:
Thanks!  How much faster is this?

Feel free to merge if travis passes.

src/Util/Decidable.v
JasonGross: Minimal test case (minimized manually):
```coq
Require Import Coq.PArith.BinPos.
Goal forall x y, {Pos.compare_cont Gt x y = Gt} + {Pos.compare_cont Gt x y <> Gt}.
  intros.
  decide equality.
```
Reported as [bug #5449](https://coq.inria.fr/bugs/show_bug.cgi?id=5449).

src/MontgomeryXProofs.v
```diff
@@ -0,0 +1,57 @@
...
+
+    Lemma xzladderstep_correct
+          (Q Q':point) x z x' z' x1 x2 z2 x3 z3
+          (Hl:Logic.eq (pair(pair x2 z2)(pair x3 z3)) (xzladderstep x1 (pair x z) (pair x' z')))
+          (H:match M.coordinates Q with∞=>z=0/\x<>0|(xQ,y)=>xQ=x/z/\z<>0  (* TODO *) /\ y <> 0 (* TODO: prove this from non-squareness of a^2 - 4 *) end)
+          (H':match M.coordinates Q' with∞=>z'=0/\x'<>0|(xQ',_)=>xQ'=x'/z'/\z'<>0 end)
+          (H1:match M.coordinates (add Q (opp Q')) with∞=>False|(x,y)=>x=x1/\x<>0 end):
+      match M.coordinates (add Q Q) with∞=>z2=0/\x2<>0|(xQQ,_)=>xQQ=x2/z2/\z2<>0 end /\
+      match M.coordinates (add Q Q') with∞=>z3=0/\x3<>0|(xQQ',_)=>xQQ'=x3/z3/\z3<>0 end.
+    Proof. t; abstract fsatz. Qed.
```
JasonGross: Note that `abstract` makes `Qed`'d proofs slower, in most cases, because the `Qed` will retypecheck the subproof.

src/WeierstrassCurve/WeierstrassCurveTheorems.v
```diff
@@ -40,10 +33,158 @@ Module W.
...
+        move b at bottom;
+        move a at bottom;
+        repeat match goal with [H: ?x = Fopp ?y |- _] => is_var x; is_var y; revert H end; intros;
+        repeat match goal with [H: ?x = ?y |- _] => is_var x; is_var y; revert H end; intros;
+        try split.
+      (* The [discriminant_nonzero] hypothesis makes [fsatz] slow but
+      is necessary in some cases. Thus, we wrap it in [id] by detault
+      to hide it from [nsatz] but unfold it when normal nsatz fails *)
+      (* TODO: why does par not work here? *)
+      Ltac s := abstract (
```
JasonGross: This sketches me out.  Why are we separating out every call to `s`?  Also, note that `abstract` makes `Qed`'d proofs slower, in most cases, because the `Qed` will retypecheck the subproof.

Re `par` not working: one of your subtactics must use `abstract`; see the newly reported [bug #5452](https://coq.inria.fr/bugs/show_bug.cgi?id=5452)

src/Util/Decidable.v
JasonGross: We should probably strip the comment now that this has been reported.

src/MontgomeryXProofs.v
andres-erbsen: Okay... I guess I don't understand the performance situation here well enough to improve it. My advertised speedup was on the non-Qed part.

However, I think the last time I tried doing this without abstract, my computer just ran out of memory. I pushed a new version with timing annotaitons on everything except the Qed, which might be taking half of the time.

andres-erbsen:
I think I am done with this. I can't figure out what is up with performance, though -- it is faster interactively, but I think the Qed is just correspondigly slower or something. Merge away, or debug yourself if you care.

JasonGross:
I'll plan to look into performance later today or tomorrow.

andres-erbsen:
@JasonGross please rebase and merge this whenever convenient for you -- you seem to be changing the files this PR touches.

JasonGross:
Will do.  I expect I'll do this tomorrow or Thursday.  (The only things I'm changing are the `Proof using` statements, so rebasing shouldn't be hard.)

JasonGross:
Merged as https://github.com/mit-plv/fiat-crypto/commit/f578406dd028f3d8d2afb16fb5d8e7d15b9a022c with a timing diff.

PR #147, opened 2017-04-04
Add [Proof using] to most proofs

### LINKED ISSUE #146: We should add [Proof using] statements to our theorems
It will make `make quick` faster.  I wrote a [script here](https://github.com/JasonGross/coq-tools#proof-using-helper) that we can use.

JasonGross:
This closes #146 and makes `make quick` faster.

The changes were generated by adding [Global Set Suggest Proof Using.]
to GlobalSettings.v, and then following [the instructions for a script I
wrote](https://github.com/JasonGross/coq-tools#proof-using-helper).

I'm relying on Travis to make sure that I didn't break anything.


JasonGross:
Builds on my machine, so I'm going to merge

PR #148, opened 2017-04-05
Add non-building Opp reification

JasonGross:
I'd like to reify `opp`.  Currently, `opp_sig` uses `Z.opp`, which is not a supported operation.  @jadephilipoom can you fix that, and then make sure that this builds on top of your fix, and then merge this (or let me know and I can merge it)?

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -233,6 +233,7 @@ src/Spec/WeierstrassCurve.v
 src/Specific/GF1305.v
 src/Specific/GF25519.v
 src/Specific/IntegrationTest.v
+src/Specific/IntegrationTestOpp.v
 src/Specific/NewBaseSystemTest.v
 src/Specific/SC25519.v
 src/Specific/FancyMachine256/Barrett.v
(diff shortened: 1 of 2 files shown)


andres-erbsen:
Why does the synthesis output contain `opp` anyway. Should it?

JasonGross:
It contains expressions like `x + 1 * (1 * (1 * -y + 0))`  @jadephilipoom can say more

JasonGross:
Also, this is fixed with #149, where I add support for `Z.opp` in the reifier and the expression simplifier which can (hopefully) make it go away.

andres-erbsen:
Hmm. I think it is fine to leave negation-elimination to the reflective tactics, but doing so shouldn't introduce a dependency on old BaseSystem.

andres-erbsen:
Looking at Z.syntax, I think `Neg` should go away and we should use `Opp` instead.

JasonGross:
Sure, we can remove `Neg`; maybe save it for the cleaning when we decide what to do with cmovle and cmovne?

JasonGross:
This PR is no longer needed, since we have a test with sub

PR #150, opened 2017-04-06
Add IntegrationTestLadderstep.v

JasonGross:
I believe this fully refines the ladderstep goal.  It takes forever to run (I haven't seen it finish yet, but it seems to be making slow by steady progress).

The tactic invocations on my side of the pipeline are expanded so that we can see timing data; the sequence is identical to an invocation of `refine_reflectively`.


JasonGross:
Oops, I forgot to make `mul` carry.  Also, the reflective side-condition 2 fails for reasons I'm having trouble tracking down, because it takes Coq multiple minutes to print the goal >.<  I'm not sure I can debug this any further, at least not so long as we're still on a 32 bit version.

PR #151, opened 2017-04-06
Add IntegrationTest for Sub (doesn't work)

JasonGross:
Currently it seems that `sub` doesn't bounds-check; it fails with
```
The term "eq_refl" has type "true = true" while it is expected to have type
 "Interpretation.Bounds.is_tighter_thanb
    (r[0 ~> 2251799813685247], r[-18014398509481984 ~> 18014398509481984], r[-18014398509481993 ~> 18014398509481992],
    r[-72057594037927936 ~> 72057594037927936], r[-72057594037927936 ~> 72057594037927936])
    (r[0 ~> 2533274790395904], r[0 ~> 2533274790395904], r[0 ~> 2533274790395904], r[0 ~> 2533274790395904],
    r[0 ~> 2533274790395904]) = true".
```

@jadephilipoom do you know what's up with this?  You seem to be using 2^52 in the first limb and 2^51 in all the others?

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -236,6 +236,7 @@ src/Spec/WeierstrassCurve.v
 src/Specific/GF1305.v
 src/Specific/GF25519.v
 src/Specific/IntegrationTestMul.v
+src/Specific/IntegrationTestSub.v
 src/Specific/NewBaseSystemTest.v
 src/Specific/SC25519.v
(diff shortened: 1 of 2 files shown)


JasonGross:
Now it builds.  Thanks @jadephilipoom !

PR #153, opened 2017-04-13
Ladderstep less sigma

JasonGross:
This rewrites the ladderstep goal.

It's now more amenable to use in the montgomery loop, and the display will only have one copy of the code.

Unfortunately, it's much more ad-hoc and kludgy now.


JasonGross:
Merging, because this is the best we currently have, after chatting with @andres-erbsen a bit on signal.  Let's put this on the agenda for the group meeting tomorrow @jadephilipoom @achlipala

PR #154, opened 2017-04-14
Use the for-loop notation in Montgomery.XZ

JasonGross:
We also need to stuff the local notations in a scope so that we can
still access the notations in core_scope

### DIFF
--- a/src/Curves/Montgomery/XZ.v
+++ b/src/Curves/Montgomery/XZ.v
@@ -64,11 +64,11 @@ Module M.
       (* Ideally, we would verify that this corresponds to x coordinate
          multiplication *)
       Local Open Scope core_scope.
-      Definition montladder (bound : positive) (testbit:nat->bool) (u:F) :=
+      Definition montladder (bound : positive) (testbit:Z->bool) (u:F) :=
         let '(P1, P2, swap) :=
-            for (int i = BinInt.Z.pos bound; i >= 0%Z; i--)
+            for (int i = BinInt.Z.pos bound; i >= 0; i--)
                 updating ('(P1, P2, swap) = ((1%F, 0%F), (u, 1%F), false)) {{
-              dlet s_i := testbit (BinInt.Z.to_nat i) in
+              dlet s_i := testbit i in
               dlet swap := xor swap s_i in
               let '(P1, P2) := cswap swap P1 P2 in
               dlet swap := s_i in
--- a/src/Curves/Montgomery/XZ.v
+++ b/src/Curves/Montgomery/XZ.v
@@ -2,6 +2,7 @@ Require Import Crypto.Algebra.Field.
 Require Import Crypto.Util.GlobalSettings Crypto.Util.Notations.
 Require Import Crypto.Util.Sum Crypto.Util.Prod Crypto.Util.LetIn.
 Require Import Crypto.Util.Decidable.
+Require Import Crypto.Util.ForLoop.
 Require Import Crypto.Spec.MontgomeryCurve Crypto.Curves.Montgomery.Affine.
 
 Module M.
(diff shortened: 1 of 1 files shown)


src/Curves/Montgomery/XZ.v
```diff
@@ -11,13 +12,15 @@ Module M.
...
-    Local Infix "-" := Fsub. Local Infix "/" := Fdiv.
-    Local Notation "x ^ 2" := (x*x).
-    Local Notation "0" := Fzero.  Local Notation "1" := Fone.
+    Local Infix "+" := Fadd : F_scope. Local Infix "*" := Fmul : F_scope.
+    Local Infix "-" := Fsub : F_scope. Local Infix "/" := Fdiv : F_scope.
+    Local Notation "x ^ 2" := (x*x) : F_scope.
+    Local Notation "0" := Fzero : F_scope.  Local Notation "1" := Fone : F_scope.
     Local Notation "'∞'" := (inr tt) : core_scope.
-    Local Notation "( x , y )" := (inl (pair x y)).
+    Local Notation "( x , y , .. , z )" := (inl (pair .. (pair x y) .. z)) : F_scope.
```
andres-erbsen: Is this notation ever used?

src/Curves/Montgomery/XZ.v
```diff
@@ -51,7 +54,29 @@ Module M.
...
+      Context {cswap:bool->F*F->F*F->(F*F)*(F*F)}.
+
+      Local Notation xor := Coq.Init.Datatypes.xorb.
+
+      (* Ideally, we would verify that this corresponds to x coordinate
+         multiplication *)
+      Local Open Scope core_scope.
+      Definition montladder (bound : positive) (testbit:nat->bool) (u:F) :=
+        let '(P1, P2, swap) :=
+            for (int i = BinInt.Z.pos bound; i >= 0%Z; i--)
```
andres-erbsen: Could the for-loop constants be parsed in `Z_scope`?

src/Curves/Montgomery/XZ.v
```diff
@@ -51,7 +54,29 @@ Module M.
...
       end.
+
+      Context {cswap:bool->F*F->F*F->(F*F)*(F*F)}.
+
+      Local Notation xor := Coq.Init.Datatypes.xorb.
+
+      (* Ideally, we would verify that this corresponds to x coordinate
+         multiplication *)
+      Local Open Scope core_scope.
+      Definition montladder (bound : positive) (testbit:nat->bool) (u:F) :=
```
andres-erbsen: `testbit : Z -> bool` seems to fit better here.

src/Curves/Montgomery/XZ.v
JasonGross: Yes, it's used in the match statement, at least

src/Curves/Montgomery/XZ.v
JasonGross: I'll fix this when I'm next at a computer.

src/Curves/Montgomery/XZ.v
JasonGross: Argh, this should happen automatically, because the arguments to `Z.geb ` are bound to `Z_scope`, but, if I recall correctly, I have a long-standing open bug that Coq seems to have two separate scope stacks that occasionally play poorly together.  I'll fix this when I'm next at a computer.

JasonGross: Turns out we already don't need the `%Z`.  The bug I was thinking of was [#3990](https://coq.inria.fr/bugs/show_bug.cgi?id=3990), but I think it doesn't apply here.

src/Curves/Montgomery/XZ.v
JasonGross: Fixed

PR #156, opened 2017-04-15
Inline a24

JasonGross:
@jadephilipoom Can you ack the change I made involving `solve_constant_sig` in the `ArithmeticSynthesisTest` files?


src/Specific/ArithmeticSynthesisTest130.v
```diff
@@ -25,6 +25,7 @@ Section Ops51.
   Definition coef_div_modulus : nat := 2. (* add 2*modulus before subtracting *)
   Definition carry_chain1 := Eval vm_compute in (seq 0 (pred sz)).
   Definition carry_chain2 := ([0;1])%nat.
+  Definition a24 := 121665%Z. (* XXX TODO(andreser) FIXME?  Is this right for this curve? *)
```
andres-erbsen: `xzladderstep` on 2^130-5 is cryptographically useless, I don't think anybody has ever searched for parameters for it.

PR #162, opened 2017-04-23
`xzladderstep`, `montladder` WIP

andres-erbsen:
I don't know how to make these proofs less ugly. OTOH this would be good validation for the `montladder` which we use as a spec for the code in our "1.0" milestone, so maybe it would be worth figuring this out.

- [x] automate `to_xz_add` (@JasonGross , would you like to take a try at this?). Perhaps there is a better way to formulate `projective`, I think that might fix some of the issues.
- ~~extend `to_xz_add'` to `montladder` (need loop invariant framework)~~
- ~~prove that `to_x (montladder (to_xz ...) ...)` is `to_x (to_xz (M.mul ...) ...)`~~
- [x] connect to `group` proofs from `AffineProofs` (need commutativity, need to allow for all characteristic proofs)

The long proof is transcribed from https://cr.yp.to/ecdh/curve25519-20060209.pdf appendix B.2.


src/Curves/Montgomery/XZProofs.v
```diff
@@ -39,46 +44,299 @@ Module M.
...
+        abstract fsatz.
+      }
+      { rename f into xQ. rename f0 into yQ.
+        intro HH.
+        apply H5.
+        assert ((x' - z') * (x + z) - (x' + z') * (x - z) =0 /\ ((x' - z') * (x + z) + (x' + z') * (x - z))=0) by (
+        repeat match goal with
+               | H : _ |- _ => apply Hierarchy.zero_product_zero_factor in H; destruct H as [H|H]
+               end; split; try contradiction; try assumption); destruct_head and; clear HH H8.
+        clear Hsq2 Hsq1. (* TODO: removing this clear gives stack overflow *)
```
JasonGross: What's the stack overflow in? `nsatz_compute`?

src/Curves/Montgomery/XZProofs.v
```diff
@@ -39,46 +44,299 @@ Module M.
...
+        assert (x'*x = z'*z) by fsatz.
```
JasonGross: Why do you need this assert?  I thought `fsatz ` was supposed to be assert-invariant?

JasonGross:
> need loop invariant framework

Do you want me to give you this?  I don't expect it to take more than a day.

> automate to_xz_add (@JasonGross , would you like to take a try at this?).

Sure, I can take a look tomorrow or Tuesday, but I'd like to know what your assessment is of what's the main block to automation; is it that fsatz is too slow?  Or has a bug that makes it stack overflow?

src/Curves/Montgomery/XZProofs.v
```diff
@@ -30,6 +33,8 @@ Module M.
     Local Notation xzladderstep := (M.xzladderstep(a24:=a24)(Fadd:=Fadd)(Fsub:=Fsub)(Fmul:=Fmul)).
     Local Notation to_xz := (M.to_xz(Fzero:=Fzero)(Fone:=Fone)(Feq:=Feq)(Fadd:=Fadd)(Fmul:=Fmul)(a:=a)(b:=b)).
 
+    Definition projective (P:F*F) :=
+      (fst P = 0 -> snd P <> 0) /\ (snd P = 0 -> fst P <> 0).
```
JasonGross: You don't need both of these; both are curry-equivalent to `fst P = 0 -> snd P = 0 -> False `.  Do you need it in the form you have for some particular reason?

src/Curves/Montgomery/XZProofs.v
```diff
@@ -39,46 +44,299 @@ Module M.
...
+      /\ eq (to_xz (Madd Q Q')) (snd (xzladderstep x1 xz x'z'))
+      /\ projective (snd (xzladderstep x1 xz x'z')).
+    Proof.
+      pose proof (nonsquare_a24 (fst xz/snd xz  - fst xz/snd xz)) as Hsq1.
+      pose proof (y_nonzero Q) as Hsq2.
+      destruct xz as [x z];
+      destruct x'z' as [x' z'];
+      destruct Q as [Q pfQ]; destruct Q as [[xQ yQ]|[]];
+      destruct Q' as [Q' pfQ']; destruct Q' as [[xQ' yQ']|[]];
+      t.
```
JasonGross: Is there a reason to not case on `dec (_ = 0)` for x, x', z, z' up here, before `t`?  I would expect the rest of the proof to follow by `fsatz `, modulo stack overflows.

src/Curves/Montgomery/XZProofs.v
andres-erbsen: I was trying to help automation, but I think what is there right now is not that useful. Perhaps `if dec (snd P = 0) then fst P <> 0 else True`?

src/Curves/Montgomery/XZProofs.v
andres-erbsen: Hmm, I guess it's worth a try. I didn't do any blind exploration because everything is slow.

src/Curves/Montgomery/XZProofs.v
andres-erbsen: I would guess it is unnecessary. I developed the proofs by moving asserts forward, I probably just forgot that one.

src/Curves/Montgomery/XZProofs.v
andres-erbsen: I would guess so, but I don't know. Something in `fsatz`.

andres-erbsen:
Here is a (probably incomplete) list of blockers to automation:

- making use of `nonsquare_a24` (tell it what a square root is, get `False`)
- `nsatz` doesn't prove `x*x = 0 -> x = 0`.
-  implications in `projective` are currently specialized semi-manually
-  it is hard to tell whether a `fsatz` call would eventually succeed or not
- unclear whether there is a common pattern of case analysis that would mitigate these issues

src/Curves/Montgomery/XZProofs.v
JasonGross: Sure.  Or even stick both `dec `s in if statements

JasonGross:
> nsatz doesn't prove x*x = 0 -> x = 0.

Isn't this the point of power 2?  Is this just because we specialize our nsatz to power 1?

JasonGross:
> > nsatz doesn't prove x*x = 0 -> x = 0.
>
> Isn't this the point of power 2?  Is this just because we specialize our nsatz to power 1?

Indeed, this is the case; we have
```coq
Ltac nsatz := nsatz_power 1%nat.
```
and `fsatz` just calls `nsatz`.  If you make a version of `fsatz` that takes a power, and then use `fsatz 1 || fsatz 2`, that should handle your goal.  Using `nsatz` from the standard library, this works:
```coq
Require Import Coq.nsatz.Nsatz.

Goal forall x y : Z, x * x = 0 -> x = 0.
  intros.
  nsatz.
Qed.
```

andres-erbsen:
Yes, power 2 fixes the specific case of x*x=0 -> x=0, but it didn't seem
to fix the cases where that was just an intermediate step towards
another goal.

JasonGross:
> but it didn't seem to fix the cases where that was just an intermediate step towards another goal.

Can you get a small example of this?  It seems like a potential bug in the implementation of `nsatz` that we could report and poke the devs to fix, if it's not assert-invariant in this situation

andres-erbsen:
Maybe -- I didn't keep any. I doubt that it would be a bug, though -- I wouldn't be surprised if an even higher power was inherently necessary to do the assert-free proof.

JasonGross:
> automate to_xz_add (@JasonGross , would you like to take a try at this?). Perhaps there is a better way to formulate projective, I think that might fix some of the issues.

Done in #165, with reformulation of `projective` that you suggested.

andres-erbsen:
Since the remaining TODO-s are about loops, I think we should merge this now and revisit once we have decided on how to deal with loops. @JasonGross ?

andres-erbsen:
because I needed to look it up

Issue #157, opened 2017-04-15
Word-by-Word Montgomery Reduction


andres-erbsen:
https://eprint.iacr.org/2011/239.pdf Lemma 1
https://eprint.iacr.org/2013/816.pdf Algorithm 1

High-level behavior: divides `T` by `R` modulo `N` (with not necessarily minimal output). Makes use of the fact that `R = w i` for some `i` by repeatedly dividing by the bounds of each digit up to `i`. Each such division is implemented by first adjusting `T` such that it is `0 mod (bound i)` and discarding its lowest limb. The adjustment is
`T += N * ((((T  mod (bound i)) * ((-N^-1) mod (bound i))) mod (bound i))`, which obviously preserves the value `mod N`. Looking `mod (bound i)` this does `T += -T`, so the result must be divisible by `bound i` and if the last limb has divmod relationship with the remaining limbs, discarding it executes the division. I believe the proof in https://github.com/mit-plv/fiat-crypto/blob/master/src/Arithmetic/MontgomeryReduction/Proofs.v covers each single step here, but operating on multiple limbs needs to be argued by canonical saturated basesystem logic. Yet another reason to have nice support for reasoning about canonicity properties of `compact` output.

andres-erbsen:
More resources:

https://cseweb.ucsd.edu/classes/fa06/cse246/montgomery.pdf section 2  
http://colinandmargaret.co.uk/Research/CDW_ELL_99.pdf section 2  
https://github.com/openssl/openssl/blob/master/crypto/ec/asm/ecp_nistz256-x86_64.pl#L607  
Out of these, the I think the code is the most likely to be correct.

I figured out how to express the algorithm without depending on the definitional structure of basesystem.

Given `A, B < R`, we want to compute `A * B / R mod N`. `R = bound 0 * ... * bound (n-1)`  
Precompute `k i := (-1 mod N) mod (bound i)`  
Recurse for a as many iterations as `A` has limbs, varying `A := A, S := 0, bound : nat -> Z`:

1. `let r := bound 0 in`
2. `let '(A, a) := divmod A r in` -- if A is saturated, take limb. otherwise trickier, do delayed carry.
3. `let S := S + a * B in` -- word-bignum multiplication, row-wise addition, I think no carry out
4. `let s := S mod r in` -- take lowest limb
5. `let q := s * k i mod r` -- word-by-word multiplication, single word out
[...]

PR #163, opened 2017-04-24
Add loop invariant framework for for-loops

JasonGross:
Also fix bugs in for-loop definition to make the proofs go through

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -248,7 +248,10 @@ src/Util/Unit.v
 src/Util/WordUtil.v
 src/Util/ZRange.v
 src/Util/ZUtil.v
+src/Util/ForLoop/Instances.v
+src/Util/ForLoop/InvariantFramework.v
 src/Util/ForLoop/Tests.v
+src/Util/ForLoop/Unrolling.v
 src/Util/Logic/ImplAnd.v
 src/Util/Sigma/Associativity.v
 src/Util/Sigma/Lift.v
--- a/src/Util/ForLoop.v
+++ b/src/Util/ForLoop.v
@@ -9,35 +9,44 @@ Section with_body.
   Fixpoint repeat_function (count : nat) (st : stateT) : stateT
     := match count with
        | O => st
-       | S count' => repeat_function count' (body count' st)
+       | S count' => repeat_function count' (body count st)
        end.
 End with_body.
 
 Local Open Scope bool_scope.
 Local Open Scope Z_scope.
 
-Definition for_loop (i0 finish : Z) (step : Z) {stateT} (initial : stateT) (body : Z -> stateT -> stateT)
+Definition for_loop {stateT} (i0 finish : Z) (step : Z) (initial : stateT) (body : Z -> stateT -> stateT)
   : stateT
-  := let signed_step := (finish - i0) / step in
-     let count := Z.to_nat ((finish - i0) / signed_step) in
-     repeat_function (fun c => body (i0 + signed_step * Z.of_nat (count - c))) count initial.
+  := let count := Z.to_nat (Z.quot (finish - i0 + step - Z.sgn step) step) in
+     repeat_function (fun c => body (i0 + step * Z.of_nat (count - c))) count initial.
 
 
 Notation "'for' i (:= i0 ; += step ; < finish ) 'updating' ( state := initial ) {{ body }}"
   := (for_loop i0 finish step initial (fun i state => body))
      : core_scope.
 
+Module Import ForNotationConstants.
+  Definition eq := @eq Z.
+  Module Z.
+    Definition ltb := Z.ltb.
+    Definition ltb' := Z.ltb.
+    Definition gtb := Z.gtb.
+    Definition gtb' := Z.gtb.
+  End Z.
+End ForNotationConstants.
+
 Delimit Scope for_notation_scope with for_notation.
-Notation "x += y" := (x = Z.pos y) : for_notation_scope.
-Notation "x -= y" := (x = Z.neg y) : for_notation_scope.
+Notation "x += y" := (eq x (Z.pos y)) : for_notation_scope.
+Notation "x -= y" := (eq x (Z.neg y)) : for_notation_scope.
 Notation "++ x" := (x += 1)%for_notation : for_notation_scope.
 Notation "-- x" := (x -= 1)%for_notation : for_notation_scope.
 Notation "x ++" := (x += 1)%for_notation : for_notation_scope.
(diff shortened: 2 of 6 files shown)


andres-erbsen:
If I understand correctly, this only provides adaquate support for loops with `abs(step)=1`? That's fine, but please document.

A high-level example file with the minimal required imports and a simple proof would be nice.

JasonGross:
>  this only provides adaquate support for loops with `abs(step)=1`?

I don't know what you mean by "adequate".  There is special support for cases with `abs(step)=1`, because in these cases, we know for sure that the step divides `finish - i0`.  In other cases, the final value of the loop counter needs to be adjusted away from `finish` for cases where the step does not divide `finish - i0`.  In these cases, you can use `for_loop_ind_`{`lt`,`gt`,`le`,`ge`}.  (Note that these are slightly more general than anything you can write with the notation; should I specialize them to the notation (requiring `step` to be judgmentally `Z.pos _` or `Z.neg _`, requiring the helper proofs to be judgmentally `eq_refl`)?)

> A high-level example file with the minimal required imports and a simple proof would be nice.

Will do

andres-erbsen:
Sorry about the confustion. My comment was about `Pbody : forall c st, i0 <= c < finish \/ finish < c <= i0 -> P c st -> P (c + step) (body (c + step) st)`, which I think loses information about `c - start mod step`. Something within `iff` of `Pbody : forall c st, exists n, c = i0+n*step -> i0 <= c < finish \/ finish < c <= i0 -> P c st -> P (c + step) (body (c + step) st)` would probably allow more uses of this lemma.

andres-erbsen:
Regarding specialization of lemmas: specialize if not doing so would make using them more difficult.

JasonGross:
Finally, I think I have it right this time.  I've added lemmas that prove loop-unrolling, integrated your suggestion about `Pbody`, fixed the definition of for-loops yet again, added a bunch of test cases to ensure that everything goes through correctly, with steps of both `1` and `2`, in both directions, with `<`, `<=`, `>`, `>=`.  I think everything should work correctly now.  Unfortunately, it's a bit longer than it was before.  I don't really feel like trying to automate the arithmetic proofs more, having spent the day struggling with them.

andres-erbsen:
I'm sorry that this turned out to be a rabbithole.

`for (int i = 0; i < 0; i++) updating (broken = false) {{ true }}` should be allowed, and it should output `false`. This contradicts the current unrolling lemmas. More generally, `for (int i = 0; i < x; i++) updating (executed = false) {{ true }}` with symbolic `x` should output `0 <? x`. I am guessing that whatever keeps these loops from working is causing you sadness in the proofs.

Meh. Actually, only `<; +=` is what I can quickly provide confident answers about. I think one criteria for niceness of invariant lemmas is an invariant quantifying over all indices in the already iterated range can be proven to hold while only doing proofs involving the same comparison operator as the loop syntax, but I doubt this uniquely specifies the correct behavior of all loop variants. I agree that the `<=; +=` case is weird, the only way I have used this with `P := forall x, start <= x < i, P' x state` which needs `P start`, `P i state -> P (i+1) (step i state)` and gives `forall x, start <= x <= finish, P' x state` -- notice that the `<` only appears in the invariant specification, not in the steps that need to be proven. `=high-1; >= 0; -=` is quite common, other variants of `>=; -=` I don't know about.

Perhaps we should focus on having nice support for particular looping patterns instead of eh-okay good support for all linearly terminating `for` loops in this ad-hoc manner. I am not currently confident that we got all edge cases right here.

If we were to take another stab at loops, I would try to first define `for (state = init; test; afterbody) decreasing (measure) {{ body}}` with `init : T`, `test : T -> bool`, `measure: T -> nat` (or `N`?), `body afterbody : T -> T`. The decreasingness proof would be resolved by typeclasses, failing that given in a `Program Definition` obligation. Under the hood, the actual recursion would be on `measure init : nat`, returning `option T` and deriving `False` in the `None` case using the termination proof. Then we would make nice notations for the common `T := Z * T'` case, including `+=` and etc in the third clause, and make sure typeclasses always prove termination in these cases. The main upside of this approach would be that the all derived loops could be validated by (1) reading the notations defining them instead of the general loop and (2) reading the invariant-framework lemma of the general loop (for which I guess we could prove the other direction to show that it really captures everything). I think we should be able to reify the specific cases as well, denoting them using the very same notations.

I don't know what the best thing to do right now is. Perhaps to work on something else and see which way we think about this after a couple of days?

JasonGross:
> `for (int i = 0; i < 0; i++) updating (broken = false) {{ true }}` should be allowed

I disagree.  At the very least, I think it's coherently fine to have empty loops fail to typeclass-resolve, which is what currently happens with this loop.  This framework only handles loops that pass through the body at least once.  (This does make the unrolling lemmas a bit awkward, because you have to include the empty-loop-check in for the next iteration of the loop, rather than the current one.)
Note that the direction is derived from the bounds, not from the sign of the step.

> is causing you sadness in the proofs

What's causing me sadness in the proofs is that it is tricky to write down (and prove) a general equation that says how many times a loop body is executed.  The bulk of the work is in shoving around division and absolute values / `Z.sgn` to account for positives, negatives, and the various edge cases when the step does / does not divide `x - start`.

> `for (state = init; test; afterbody) decreasing (measure) {{ body }}`

Oh, interesting.  I think this may point at a place where the approach I used is causing pain: I fused the termination measure with the loop state, and so have to convert back and forth between `nat` and `Z` in a few places, with some extra subtraction.  The approach you're suggesting, I think, would separate out the count from the state.  I suspect that eliminating the `option` to get `False` based on the termination proof might be quite painful (it is not, in fact, even clear to me how to phrase termination in a way that lets you do this at all, unless your termination obligation is just that the loop does not return `None`).  Separately, you need to handle decreasingness, unless you just throw that information away?

> I don't know what the best thing to do right now is. Perhaps to work on something else and see which way we think about this after a couple of days?

Certainly, we should merge some of this; the existing code is definitely wrong, and this code (the things in `ForLoop.v` and `Tests.v`) is better.  I think this code does the right thing on all non-empty loops, and I think the invariant framework I gave you, though ugly under the hood, should allow you to prove what you want to prove.  I think, together with the unrolling lemmas, it uniquely specifies the behavior.

> Perhaps to work on something else and see which way we think about this after a couple of days?

Yes, I'll plan to take a look at #157

andres-erbsen:
I agree that it is fine to merge this, but perhaps include a link to this issue in the source so we don't forget and assume this is something it isn't.

JasonGross:
I'll make an issue and include a link to it in the source.

PR #165, opened 2017-04-25
Better automation for xzladder

JasonGross:
This merges master into `xzladder` for improvements to `specialize_by_assumption` and for `destruct_head'_sum` (rebasing would be better, but I didn't want to force-push to your branch, @andres-erbsen ).

On top of that, I've reworked the `t` tactic in `src/Curves/Montgomery/XZProofs.v` to be about twice as fast, moved a bit of the work of `fsatz` earlier so that each call to `fsatz` only takes 1-5 seconds (`Qed` takes around 26 seconds), and automated pretty much all of `projective_fst_xzladderstep`.  I left in a bunch of `Time` and `Profile` directives; feel free to strip them out or whatever.

### DIFF
--- a/src/Curves/Montgomery/XZProofs.v
+++ b/src/Curves/Montgomery/XZProofs.v
@@ -54,6 +54,7 @@ Module M.
       | _ => progress subst
       | _ => progress specialize_by_assumption
       | [ H : ?x = ?x |- _ ] => clear H
+      | [ H : ?x <> ?x |- _ ] => specialize (H (reflexivity _))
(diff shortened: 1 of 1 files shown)


JasonGross:
Note that this is a PR against the `xzladder` branch, and not against master

JasonGross:
Argh, apparently `nsatz` is better at the tip of `v8.6` than it is on released 8.6.... I'll separate out the improvements to `t` from the rest, and see what I can do about this

JasonGross:
Now fixed

andres-erbsen:
Merged into #162

PR #166, opened 2017-04-28
Initial stab at word-by-word montgomery

JasonGross:
I think I got the loop itself wrong, though:
1. I'm worried that it'll overflow off the end of the positional
   representation, since I'm not actually dividing by 2^s
2. I'm not carrying/reducing anywhere, so the getting the nth value
   might be wrong
3. I'm not sure if I got indexing right.

But I want to submit this early version to get comments from @andres-erbsen, before I put
more effort into it.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -22,6 +22,7 @@ src/Arithmetic/BarrettReduction/HAC.v
 src/Arithmetic/BarrettReduction/Wikipedia.v
 src/Arithmetic/MontgomeryReduction/Definition.v
 src/Arithmetic/MontgomeryReduction/Proofs.v
+src/Arithmetic/MontgomeryReduction/WordByWord/Definition.v
 src/Compilers/CommonSubexpressionElimination.v
 src/Compilers/CommonSubexpressionEliminationDenote.v
 src/Compilers/CommonSubexpressionEliminationInterp.v
--- a/src/Arithmetic/MontgomeryReduction/WordByWord/Definition.v
+++ b/src/Arithmetic/MontgomeryReduction/WordByWord/Definition.v
@@ -5,6 +5,7 @@
     https://eprint.iacr.org/2013/816.pdf. *)
 Require Import Coq.ZArith.ZArith.
 Require Import Crypto.Arithmetic.Core.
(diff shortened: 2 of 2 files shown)


andres-erbsen:
1. I am guessing it actually won't overflow, but I do think it would make your life easier if you divided by `2^s` (see 3.).
2. Yes, you want to be using `Arithmetic.Saturated` (in particular, `Columns`) instead of `Arithmetic.Core`. You will probably need to poke @jadephilipoom to figure out how to assemble high-level primitives like multiplication out of that.
3. Hmm, me too. I also think indexing is not essential to the algorithm here. Each loop iteration adjusts for and then chops off the lowest limb of `T`, so I think it may be possible to avoid indexing and carrying around the known-to-be-zero limbs altogether.

andres-erbsen:
In particular, each mod and div would be implemented as "take the lowest remaining limb" or "discard the lowest remaining limb"

jadephilipoom:
> Yes, you want to be using `Arithmetic.Saturated` (in particular, `Columns`) instead of `Arithmetic.Core`.

Jason and I talked this morning about whether to use `Columns` or `Positional`; we ended up deciding on `Positional` because there's only one term per weight to start with and only one piece of state to carry between weights, so `Columns` seemed unnecessary. Why do you think we need it?

> In particular, each mod and div would be implemented as "take the lowest remaining limb" or "discard the lowest remaining limb"

That's only true if you're in a saturated basesystem--is word-by-word Montgomery saturated-only?

Jason, in the first step, you want to do `T{{i}} mod (weight (S i) / weight i)`, not `T{{i}} mod weight i`. For example, if your weight function is `(fun i => 2^(64*i)`, then you want to take the nth limb mod `2^64`, not mod `2^(64*n)`.

JasonGross:
> Jason, in the first step, you want to do

Done.

> Jason, maybe we can find a time Monday to talk about how to use that?

Yes, let's. 1pm-2pm?

JasonGross:
@andres-erbsen Updated with help from @jadephilipoom.  It now uses saturated / columns internally, requires a uniform power-of-2 base.  Feedback requested.

PR #168, opened 2017-05-03
projective Weierstrass: (2P = 2Q -> P = Q) -> not exceptional

andres-erbsen:
Prove correctness of single-case addition procedure for Weierstrass curve points `P` and `Q` such that `(P+P = Q+Q) -> P = Q`. We probably want to wrap these definitions for the simple case of `forall P, P <> zero -> P + P <> zero` where `P-Q = zero <-> (P-Q)+(P-Q) = zero`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,17 +24,11 @@ matrix:
     - env: COQ_VERSION="v8.7"      TARGETS="no-curves-proofs-non-specific"               COQ_PACKAGE="coq"           PPA="ppa:jgross-h/coq-8.7-daily"
     - env: COQ_VERSION="v8.7"      TARGETS="curves-proofs"                               COQ_PACKAGE="coq"           PPA="ppa:jgross-h/coq-8.7-daily"
     - env: COQ_VERSION="v8.7"      TARGETS="selected-specific selected-specific-display" COQ_PACKAGE="coq"           PPA="ppa:jgross-h/coq-8.7-daily"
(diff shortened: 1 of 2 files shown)


andres-erbsen:
I don't know how to make this build on anything other than trunk and I don't really care right now.

JasonGross:
@andres-erbsen Is the goal of this PR to get rid of the
```coq
(y_PmQ_nz: match W.coordinates (W.add (to_affine P) (to_affine (opp Q))) return Prop with
                       | inr _ => True
                       | inl (_, y) => y <> 0
                       end)
```
assumption on `add`?  (Is that what "odd" has to do with?)  And do you mind if I rebase and squash this (without merging the PR into master), so that I don't have to keep merging master into this PR?

andres-erbsen:
This PR proves `not_exceptional` and `not_exceptional_y` equivalent. `not_exceptional_y` is useful when analyzing elliptic curve points (and operations) coordinate-by-coordinate. However, to actually prove that a pair of points is not exceptional in the context of an application that treats elliptic curve points as elements of a mostly abstract group, `not_exceptional` is much more convenient. The next step would be to show that if a group has odd order, all pairs of points are `not_exceptional` (this works for NISTP256).
Given a definition of order of a group, I expect the proof would be simple. Separately, using this addition formula for Curve25519 (or any other even-order curve) would require the higher-level algorithm to prove `not_exceptional` for all pairs of points in it.

Please rebase or squash as necessary.

andres-erbsen:
http://adam.chlipala.net/theses/andreser.pdf starting last paragraph of page 78

PR #169, opened 2017-05-12
Remove dead code in ZUtil (shiftl_by)

JasonGross:
Is there any reason to keep this?


jadephilipoom:
Nothing in our project depends on it now, but we did agree on a general principle of keeping Util lemmas whether or not they had become obsolete, with the idea that they could be more generally useful.

That being said, this was just a convenience definition because I was writing shifts a lot and wanted the arguments to be in the other order. So its general usefulness is dubious, and I'm happy with deleting it.

PR #170, opened 2017-05-12
Add definitions for zselect and add_get_carry

JasonGross:
In ZUtil/Definitions.v

@jadephilipoom Are these definitions that you can work with?  (I changed them a bit from the definitions you're using, and am having them work only for powers of 2, in part so that the reified code doesn't need to carry around large values on every add, and so that I don't need to hard-code the expanded value of 2¹²⁸ into the notations); does this seem reasonable?)

@andres-erbsen Is this general enough?  Is restricting ourselves to powers of two okay?


jadephilipoom:
Looks reasonable to me.

andres-erbsen:
LGTM.

Issue #171, opened 2017-05-12
Compiler optimization : add with carry


jadephilipoom:
@JasonGross Here's the longer-form example I promised you.

Suppose we are adding `[a0, a1, a2]` to `[b0, b1, b2]`. We have an operation `ADD` of
type `Z -> Z -> Z * Z` and an operation `ADC` of type `Z -> Z -> Z -> Z * Z`. The
code produced initially (by the basesystem side of things) will be equivalent to:

```
s0, c0 := ADD a0 b0
s1', c1' := ADD c0 a1
s1, c1'' := ADD s1' b1
c1 := c1' + c1''
s2', c2' := ADD c1 a2
s2, c2'' := ADD s2' b2
c2 := c2' + c2''
...
```

The compiler optimization should convert this to:

```
s0, c0 := ADD a0 b0
s1, c1 := ADC c0 a1 b1
s2, c2 := ADC c1 a1 b2
...
```

Referencing Jason's earlier objection in the meeting today--It's okay to use one carry instead of
two because the bounds work out. For example, in the first code snippet, it is
not possible that c1' and c1'' are both 1; since a1 and b1 are both < 2^n ,
their sum must be < 2^{n+1} - 1.  Therefore, even when c0 is added in, the sum
is < 2^{n+1} and the carry c1 must be 1 or 0.

For reference, the exact Gallina form of the initial code will be

```
let sum_carry0 := add_get_carry a0 b0 in
let carry0 := snd sum_carry0 in
let sum_carry1 := add_get_carry c0 a1 in
let carry1 := snd sum_carry1 in
let sum_carry2 := add_get_carry (fst sum_carry1) b1 in
let carry2 := carry1 + snd sum_carry2 in
let sum_carry3 := add_get_carry carry2 a2 in
let carry3 := snd sum_carry3 in
let sum_carry4 := add_get_carry carry3 b2 in
let carry4 := carry3 + snd sum_carry4 in
...
```

JasonGross:
Ewww, this is horrendous.  I don't currently support rewriting let-expressions, so I'll need to make a dedicated pass for this particular optimization.  It'll be a pass that is hard-coded to exactly this pattern (up to syntactic equality; no inlining of lets or anything), or whatever pattern actually shows up in the reified form of freeze at the appropriate location.  This also needs to come after bounds analysis, right?

JasonGross:
Turns out it doesn't need to come after bounds analysis.

[...]

PR #174, opened 2017-05-13
Add-carry optimization proof

JasonGross:
This is on top of #170 

It proves the transformation described in #171 correct on `Z` (the transformation is not yet implemented reflectively)

### DIFF
--- a/src/Util/ZUtil/AddGetCarry.v
+++ b/src/Util/ZUtil/AddGetCarry.v
@@ -1,0 +1,52 @@
+Require Import Coq.ZArith.ZArith Coq.micromega.Lia.
+Require Import Crypto.Util.ZUtil.Definitions.
+Require Import Crypto.Util.ZUtil.Hints.ZArith.
+Require Import Crypto.Util.ZUtil.Tactics.PullPush.Modulo.
+Require Import Crypto.Util.ZUtil.Tactics.DivModToQuotRem.
+Require Import Crypto.Util.LetIn.
+Local Open Scope Z_scope.
+
+Local Notation eta x := (fst x, snd x).
+
+Module Z.
+  Section with_bitwidth.
+    Context (bitwidth : Z).
+
+    Local Notation ADD := (Z.add_get_carry bitwidth).
+    Local Notation ADC := (Z.add_with_get_carry bitwidth).
+
+    Local Arguments Z.pow : simpl never.
+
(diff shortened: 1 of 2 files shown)


src/Util/ZUtil/AddGetCarry.v
```diff
@@ -0,0 +1,52 @@
...
+      { rewrite Z.pow_neg_r by assumption.
+        autorewrite with zsimplify.
+        reflexivity. }
+      { f_equal.
+        { push_Zmod; reflexivity. }
+        { Z.div_mod_to_quot_rem; nia. } }
+      { subst; autorewrite with zsimplify; reflexivity. }
+    Qed.
+
+    Lemma add_get_carry_to_add_with_get_carry_cps {T} c a b (P : Z -> Z -> T)
```
andres-erbsen: @jadephilipoom is this transformation sufficient to turn two-row columnwise compact into a row compact?

JasonGross: This should be sufficient to handle the transformation requested in #171.  If it turns out to be not sufficient for that, I'll figure out how to fix it.  Shall we merge this and continue the discussion of sufficiency on #171?

andres-erbsen: sgtm

src/Util/ZUtil/AddGetCarry.v
```diff
@@ -0,0 +1,52 @@
...
+    Context (bitwidth : Z).
+
+    Local Notation ADD := (Z.add_get_carry bitwidth).
+    Local Notation ADC := (Z.add_with_get_carry bitwidth).
+
+    Local Arguments Z.pow : simpl never.
+
+    Lemma add_get_carry_to_add_with_get_carry_cps_gen {T} c a b (P : Z -> Z -> Z -> Z -> Z -> T)
+      : (let '(a', c1) := ADD c a in
+         let '(s, c2) := ADD a' b in
```
jadephilipoom: In the current code, I don't use destructing `let`s, I use `fst` and `snd`. The lemma should probably be stated in this form if it's going to be used directly.

JasonGross: The lemma will not be used directly, but will be used to prove the reflective transformation.  The reflective machinery converts all `let`s to destructuring `lets` (and correspondingly unfolds `fst` and `snd`).  As long as you let-bind the results rather than duplicating them, this will be fine.

jadephilipoom: Okay. In that case, everything looks good to me.

Issue #158, opened 2017-04-15
CSE/AC the squaring function used in `xzladderstep`


andres-erbsen:
We want to eliminate common subexpressions when doing field element multiplications `x*x`. These common subexpressions need to be recognized modulo associativity and commutativity. The carrying code would expand in size exponentially if `let`-binders were inlined, so that shouldn't happen during CSE (as a workaround, we could CSE a non-carrying squaring function and then plug it in, but meh).

JasonGross:
Basically implemented (without proofs) at https://github.com/JasonGross/fiat-crypto/tree/zzz-cse-square

But didn't we find that this made things slower?

andres-erbsen:
It did make the first benchmark I ran marginally slower, but I would suspect it was compiler noise, not a real effect. The effectiveness of CSE depends a lot on what the input code looks like, I would not be surprised if it did end up being useful. So let's say it is not a priority to prove it for now, but we should include a version that uses it in any benchmarks we run and prove it if we deem the performance to be worth the work. So if there is anything that could be merged with an admit for correctness, perhaps we should do so.

JasonGross:
@andres-erbsen Do you want to have CSE'd versions built by default, alongside the non-CSE'd versions?  The simplest way to do this will make both versions' correctness proofs depend on the CSE admits; is that acceptable?

briansmith:
If the CSE version isn't 20% faster then doesn't that mean it isn't working the way it is expected to work? When we replaced the non-CSE code with hand-written CSE'd code, it was 20% faster.

andres-erbsen:
The CSE experiments we did at HACS were on curve25519-donna, not fiat-crypto output. This was because curve25519 donna does the inlining in the C compiler but fiat-crypto produces already inlined C source, making later experimentation inconvenient. The relevant timings are as follows:

    curve25519-donna with fsquare=fmul (approximating no source-level CSE): 70us
    curve25519-donna: 61us
    curve25519-fiat51: 75us
    curve25519-fiat51+CSE: 75us

[...]

PR #175, opened 2017-05-13
CSE modulo arithmetic without inlining

JasonGross:
This takes care of most of #158.  The remaining bits are reworking the Wf and interpretation lemmas to actually work.  (The former needs a only bit of rethinking and rephrasing to handle the fact that sometimes we change the stored symbolic expression from a complicated one to a fresh variable, while the latter needs major surgery, which Adam tells me is easy.)


andres-erbsen:
LGTM NT.

JasonGross:
NT?

andres-erbsen:
I didn't run the tests (but now travis has passed anyway).

JasonGross:
So it means "not tested" or "no tests"?

PR #177, opened 2017-05-14
Use specialized square in ladderstep

JasonGross:


JasonGross:
Got ack from Andres on signal, and it builds fine on my machine, so merging.

PR #172, opened 2017-05-13
Reification machinery for adc, zselect

JasonGross:
This is on top of #170.

This adds reflective support for adc and zselect.  I haven't yet done pretty-printing for them; I figure I can do that when we actually have code coming out, so I can check my pretty-printer against actual code, rather than printing blindly.

@jadephilipoom `freeze_sig`, unlike the other `_sig` lemmas, takes a hypothesis for its correctness condition.  I don't know what to do with this when talking about field elements.  (Maybe we don't want to talk about field elements at all?)  I left you two `FIXME`s in `src/Specific/IntegrationTestFreeze.v`; can you fix them (and you might have to adapt the place where I use `freeze` before the reflective goal, too), and point me at a commit that I can cherry-pick onto this branch?  It should compile up to (but not including) the `Fail` line in that file (the `Fail` line takes a very, very long time to fail, because typeclasses and backtracking).

@jadephilipoom Can you update `freeze` to be in terms of the `zselect` and `add_get_carry` defined in `src/Util/ZUtil/Definitions.v`?

@andres-erbsen I left you with two goals which are all in terms of `Z` at the bottom of `src/Compilers/Z/Bounds/InterpretationLemmas/IsBoundedBy.v`.  I'm not sure that they are true; bounds analysis on add-get-carry might have to change.  Can you handle this?

PR #181, opened 2017-05-17
Add compiler optimization for add-with-carry 

### LINKED ISSUE #171: Compiler optimization : add with carry

JasonGross:
This closes #171, and is on top of #172.

It's unfortunately a bit fragile, and takes a really long time (about 60s) to prove correct, because Coq is bad at deep dependent pattern matching.  (Help in speeding it up welcome.)

We enable a-normal form for the freeze test, because the rewriter only works when the arguments to adc are var or const.  Unfortunately, this isn't enough in this case, because anf makes the masking interleaved with the adc.  We'd need either cse (currently not proven correct) to eliminate this, or we'd need to make the rewriter know that it's allowed to skip a statement (which might end up ~doubling the proof time >.>).


JasonGross:
N.B. This breaks the build on trunk.  I'll look into it shortly.

JasonGross:
Trunk should now be fixed (issue was that `rewrite` got stronger, probably after econstr was merged)

PR #188, opened 2017-06-02
Don't rely on autogenerated names

JasonGross:
This fixes all of the private-names warnings emitted by compiling fiat-crypto with https://github.com/coq/coq/pull/268 (minus the ones in coqprime, which I didn't touch).

@ejgallego got me curious enough to try compiling fiat-crypto with that branch, and this was the result.

### DIFF
--- a/Bedrock/Nomega.v
+++ b/Bedrock/Nomega.v
@@ -27,12 +27,12 @@ Theorem Nneq_out : forall n m,
   n <> m
   -> nat_of_N n <> nat_of_N m.
   intuition.
-  apply nat_of_N_eq in H0; tauto.
+  match goal with H0 : _ |- _ => apply nat_of_N_eq in H0; tauto end.
 Qed.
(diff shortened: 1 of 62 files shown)


JasonGross:
I'd like to have this merged sooner rather than later, if it's to be merged at all; it touches a lot of code, and I'd prefer not to have to rebase it.

andres-erbsen:
Is there any reason not to merge it? Explicit names seem like a universally good idea...

JasonGross:
There are a few minor infelicities to work around false positives in the warning, such as [`let pf := pf in abstract ...`](https://github.com/mit-plv/fiat-crypto/pull/188/files#diff-954cf1f4fc1aff1368d5d1ffb9d84fb6R26) to work around the warning being confused about names inside of `abstract`.  That's the only thing I can think of, and I'd lean towards merging it anyway.

PR #189, opened 2017-06-08
Update word-by-word montgomery with ctx vars

JasonGross:
As per
https://github.com/mit-plv/fiat-crypto/issues/157#issuecomment-307212425


andres-erbsen:
I like the algorithm I wrote in https://gist.github.com/andres-erbsen/1240b792f43b33bc0d2b5600734bc2e1 more than the one here, although the differences are probably not that much work to reconcile. I don't know what I should review this PR for. Jade can probably tell whether the context variables are reasonable better than I can.

JasonGross:
> I like the algorithm I wrote in https://gist.github.com/andres-erbsen/1240b792f43b33bc0d2b5600734bc2e1 more than the one here

What do you like about it?  As far as I can tell, the only differences are:
1. inlining
2. `dlet` vs `let`
3. unfoldedness of `fst` and `snd`
4. you have some extra intermediate steps (and therefore slightly different variable names)

1 is the hardest to reconcile, because I think it's nice to split the definition from the proofs, so we'd need to use something like `Local Definition`.  I personally like my way of doing 3 better, though if you prefer `fst` and `snd` over `let '(x, y) := ...`, I can change that.  Having mine match yours on 2 and 4 is pretty easy (the only exception being that I have to use `S'` with `let '`, because otherwise it thinks it's `nat` successor).

> I don't know what I should review this PR for

For how much you like the way I wrote the algorithm. :-P

Issue #194, opened 2017-06-13
ArithmeticSynthesisTest for Montgomery multiplication


JasonGross:
How far away are we from having something like `ArithmeticSynthesisTest` that uses montgomery multiplication (leaving in admits)?  If we can get that, then I can start looking into adding loops to the reflective machinery.

jadephilipoom:
The things on my radar are 1) stronger `small`, which we will presumably need somewhere (kind of holding off on this until I figure out exactly where) , 2) actually defining a `Z` version of two-output mul (this should be easy and doesn't have to be done by me), and 3) the admits left in the API section of `Arithmetic/Saturated.v`. Loops are also a todo, although I'm mostly considering that not-my-problem. Beyond that, I can't think of anything else we need, but maybe Andres has more.

JasonGross:
I did 2 (https://github.com/mit-plv/fiat-crypto/commit/80687ca21d9f46fd4dd98bda163f096a43dd76b7).  If you want me to take care of the numlimbs admits, I can do that.  But I think we should hold off on the admits for now; I'd like to get all of the computational parts working, and then come back and fill in the admits after.

Does this mean that you could basically write an ArithmeticSynthesisTest-like file using wbw-montgomery now?

andres-erbsen:
We need to rewrite WW_MM in CPS.

jadephilipoom:
I propose eliminating the context variables and directly depending on the operations in `Arithmetic/Saturated`.

JasonGross:
@andres-erbsen Like this? https://github.com/mit-plv/fiat-crypto/pull/200

jadephilipoom:
Unfortunately, it's not that simple. You have to CPSify *every* operation inside `redc` that isn't getting compiled away--that means, if you leave in the context variables, you have to take in not just e.g. `divmod`, but `divmod_cps` and `divmod_id` and `divmod`, and then use `divmod_cps` in the code and `divmod` in the proofs. This explodes when you are using context variables (I went through this with Karatsuba once already) and that's why I'm saying we should get rid of them. 

Giving you a CPS version of the `Saturated` API will not be hard, since the current version is just wrapping things that are already in CPS. I should have that soon.

JasonGross:
*nods*  What's the thing that we can't do with non-cps'd versions that we can do with cps versions?  (I'm asking in part because the reification machinery doesn't care; can cps things for us.)

jadephilipoom:
[...]

PR #200, opened 2017-06-15
CPSify word-by-word montgomery algorithm

JasonGross:
As per
https://github.com/mit-plv/fiat-crypto/issues/194#issuecomment-308756307


andres-erbsen:
Shouldn't we be calling CPS vesrions of `Saturated` operations? I'm not sure what use there is in CPS-ing just the loop.

JasonGross:
Superseded by 8079cc4

PR #203, opened 2017-06-16
C Array Notations

JasonGross:


andres-erbsen:
Did you try it out? Would it be easy for you to give me `x25519_c64.c` with this enabled?

JasonGross:
Still working on it; had to track down [bug 5608](https://coq.inria.fr/bugs/show_bug.cgi?id=5608) because printing wasn't working.  Is there another valid notation for arrays that doesn't use bare curly braces?  e.g., `int x[1] = ({ a });` or `int x[1] = {( a )};`?

JasonGross:
> Would it be easy for you to give me x25519_c64.c with this enabled?

Almost.  There's an issue: I can't distinguish between input variables and other variables, and so I think some mangling needs to be done in your script to translate input variables into arrays, or something like that; is this a thing you can do?

andres-erbsen:
donna_c64.c 165944
	fiat51da.c 174392
	fiat51d.c 174832
	fiat51dui.c 176152
	fiat51du.c 176596
	fiat51u.c 187760
	fiat51.c 189488
	
	broadwell
	Thread(s) per core:    1
	Model name:            Intel(R) Core(TM) i7-5600U CPU @ 2.60GHz
	Model:                 61
	CPU min MHz:           500.0000
	CPU MHz:               2499.865
	CPU max MHz:           3200.0000
	/sys/devices/system/cpu/intel_pstate/no_turbo: 1
	sh measure.sh  6.84s user 0.23s system 100% cpu 7.059 total
	tee measure.txt  0.00s user 0.00s system 0% cpu 7.058 total

I would say this speedup is not worth the hackiness of this trick.

JasonGross:
Thoughts on merging the notation files without enabling it (or adding separate targets for it) vs closing this without merging it?

andres-erbsen:
I would say no merge. If for some reason we want to dig it up, may the version control history help us.

JasonGross:
> may the version control history help us.

That would be the point of merging, or merging and then immediately reverting; I plan to delete this branch once this PR is closed, so discovering it would require relying on GitHub to preserve the branch because it's a PR, and it wouldn't actually be stored in the git repo.

andres-erbsen:
Jason Gross <notifications@github.com> writes:

>> may the version control history help us.
>
> That would be the point of merging, or merging and then immediately reverting; I plan to delete this branch once this PR is closed, so discovering it would require relying on GitHub to preserve the branch because it's a PR, and it wouldn't actually be stored in the git repo.

Okay, go merge and revert.

PR #204, opened 2017-06-16
Switch to using tuples for word-by-word montgomery

JasonGross:
The new parameterized definitions and proofs are in
WordByWord/Abstract/Dependent/*; the old ones are untouched (and unused)
in WordByWord/Abstract/*.  I replaced definitions I didn't know how to
write in the Saturated API with the use of an axiom.

### DIFF
--- a/src/Arithmetic/Saturated.v
+++ b/src/Arithmetic/Saturated.v
@@ -720,18 +720,16 @@ Section API.
 
   Definition zero {n:nat} : T n := B.Positional.zeros n.
 
-  Axiom JADE_HOW_DO_I_IMPLEMENT_THIS : forall {T}, T.
-
-  Definition join0_cps {n:nat} (p : T n) {R} (f:T (S n) -> R) : R
-    := JADE_HOW_DO_I_IMPLEMENT_THIS.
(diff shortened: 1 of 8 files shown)


src/Arithmetic/Saturated.v
```diff
@@ -711,83 +711,87 @@ End UniformWeight.
...
+
+  Definition zero {n:nat} : T n := B.Positional.zeros n.
 
-  Definition zero (n:nat) : T := to_list _ (B.Positional.zeros n).
+  Axiom JADE_HOW_DO_I_IMPLEMENT_THIS : forall {T}, T.
 
-  Definition divmod_cps (p : T) {R} (f:T * Z->R)
-    := f (List.tl p, List.hd 0 p).
-  Definition divmod p : T * Z := divmod_cps p id.
+  Definition join0_cps {n:nat} (p : T n) {R} (f:T (S n) -> R) : R
```
jadephilipoom: This is new, but from reading the usage it seems like it adds 0 to the high end--tell me if that's wrong.

JasonGross: That's correct (that should also be basically the only thing that satisfies both the type signature and `eval_join0`).

PR #210, opened 2017-06-19
Weaken preconditions on small_add

JasonGross:
You could probably write a better proof than I could, since you have more knowledge of and familiarity with the underlying building blocks, but I got nerdsniped.  @jadephilipoom I'll leave this for you to decide what to do with.


jadephilipoom:
Looks good to me.

PR #224, opened 2017-06-26
Synthesize Nonzero

JasonGross:
@jadephilipoom Can you sanity check / ack the changes to `Saturated.v`?

@andres-erbsen Does the spec seem fine to you?

### DIFF
--- a/src/Arithmetic/Saturated.v
+++ b/src/Arithmetic/Saturated.v
@@ -681,6 +681,7 @@ Section API.
 
   Definition zero {n:nat} : T n := B.Positional.zeros n.
 
+  (** Returns 0 iff all limbs are 0 *)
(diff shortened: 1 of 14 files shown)


src/Arithmetic/Saturated.v
```diff
@@ -681,6 +681,11 @@ Section API.
 
   Definition zero {n:nat} : T n := B.Positional.zeros n.
 
+  Definition nonzero_cps {n} (p : T n) {cpsT} (f : Z -> cpsT) : cpsT
```
jadephilipoom: Could you put a comment here with the spec? Just "return 0 iff all limbs are 0" is good enough.

JasonGross: Done

Issue #222, opened 2017-06-25
add-with-carry optimization anti-patterns


JasonGross:
I found a major issue with my implementation of #171.  @jadephilipoom requested optimizing
```
s0, c0 := ADD a0 b0
s1', c1' := ADD c0 a1
s1, c1'' := ADD s1' b1
c1 := c1' + c1''
```
to
```
s0, c0 := ADD a0 b0
s1, c1 := ADC c0 a1 b1
s2, c2 := ADC c1 a1 b2
...
```
I believe (though I'm only like 80% confident) that this isn't actually powerful enough; I think the assignment of `c0` doesn't always come on the line immediately preceding `ADD c0 a1`, in particular when we put the code in a-normal form (which is required for doing this rewrite in some cases, in part due to the limitations of variable equality testing and the named representation).

Therefore, the optimization I actually implemented is the strictly more powerful:
```
```
s1', c1' := ADD c0 a1
s1, c1'' := ADD s1' b1
c1 := c1' + c1''
```
to
```
s1, c1 := ADC c0 a1 b1
s2, c2 := ADC c1 a1 b2
...
```

However, this breaks Montgomery `opp`, which has code like (128-bit version):
```coq
    (fun var : Syntax.base_type -> Type =>
     λ '(H, x1)%core,
     ℤ H0, ℤ x3 = addcarryx_u128ℤ(0x0, 0x0, Op (Syntax.Opp Syntax.TZ Syntax.TZ) (Return x1));
     ℤ H1, ℤ x5 = addcarryx_u128ℤ(0x0, 0x0, H0);
     ℤ x6 = x3 +ℤ x5;
     ℤ H2, ℤ x8 = addcarryx_u128ℤ(0x0, 0x0, Op (Syntax.Opp Syntax.TZ Syntax.TZ) (Return H));
     ℤ H3, ℤ x10 = addcarryx_u128ℤ(0x0, x6, H2);
     ℤ x11 = x8 +ℤ x10;
     ℤ x12 = x11 == 0 ?ℤ 0x0 : 0xffffffffffffffffffffffffffffffffL;
     ℤ H4, ℤ x14 = addcarryx_u128ℤ(0x0, H1, x12 &ℤ 0xffffffffffffffffffffffffL);
     ℤ H5, ℤ x16 = addcarryx_u128ℤ(0x0, 0x0, H4);
     ℤ x17 = x14 +ℤ x16;
[...]

PR #226, opened 2017-06-29
Smarter addition/subtraction chains

jadephilipoom:
As per #222. Putting this up early so we can try to make code from it, but it's unpolished--there are currently several admitted proofs and context variables don't line up.

- [x] Fix the context variables in the Montgomery code to match the new arguments that the add and subtract functions take. @JasonGross, can you do this?
- [ ] Fix the proofs that are currently admitted.
- [ ] Maybe a separate larger TODO for me, clean up the code in `Saturated`; it should be split into several files and organized before I drive myself insane.

### DIFF
--- a/src/Arithmetic/Saturated.v
+++ b/src/Arithmetic/Saturated.v
@@ -171,93 +171,6 @@ Hint Opaque Associational.mul Associational.multerm : uncps.
 Hint Rewrite @Associational.mul_id @Associational.multerm_id : uncps.
 Hint Rewrite @Associational.eval_mul @Associational.eval_map_multerm using (omega || assumption) : push_basesystem_eval.
 
-Module Positional.
-  Section Positional.
-    Context (weight : nat->Z) {s:Z}. (* s is number at which to split *)
-    Section GenericOp.
-      Context {op : Z -> Z -> Z}
-              {op_get_carry : Z -> Z -> Z * Z} (* no carry in, carry out *)
-              {op_with_carry : Z -> Z -> Z -> Z * Z}. (* carry in, carry out  *)
-
-      Fixpoint chain_op'_cps {n}:
-        option Z->Z^n->Z^n->forall T, (Z*Z^n->T)->T :=
-        match n with
-        | O => fun c p _ _ f =>
-                 let carry := match c with | None => 0 | Some x => x end in
-                 f (carry,p)
-        | S n' =>
-          fun c p q _ f =>
-            (* for the first call, use op_get_carry, then op_with_carry *)
-            let op' := match c with
-                       | None => op_get_carry
-                       | Some x => op_with_carry x end in
-            dlet carry_result := op' (hd p) (hd q) in
-              chain_op'_cps (Some (fst carry_result)) (tl p) (tl q) _
-                            (fun carry_pq =>
-                               f (fst carry_pq,
-                                  append (snd carry_result) (snd carry_pq)))
-        end.
-      Definition chain_op' {n} c p q := @chain_op'_cps n c p q _ id.
-      Definition chain_op_cps {n} p q {T} f := @chain_op'_cps n None p q T f.
-      Definition chain_op {n} p q : Z * Z^n := chain_op_cps p q id.
-
-      Lemma chain_op'_id {n} : forall c p q T f,
-        @chain_op'_cps n c p q T f = f (chain_op' c p q).
-      Proof.
-        cbv [chain_op']; induction n; intros; destruct c;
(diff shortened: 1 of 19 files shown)


JasonGross:
> Fix the context variables in the Montgomery code to match the new arguments that the add and subtract functions take. @JasonGross, can you do this?

I have to also modify `Saturated` to do this, because I need `add : T n -> T n -> T (S n)` that doesn't drop the final carry, and the versions you wrote seem to do the wrong thing.

src/Arithmetic/Saturated.v
```diff
@@ -747,36 +847,49 @@ Section API.
...
 
-  Definition add_cps {n m pred_nm} (p : T n) (q : T m) {R} (f:T (S pred_nm)->R) :=
-    Columns.add_cps (uweight bound) p q
-      (fun carry_result => Tuple.left_append_cps (fst carry_result) (snd carry_result)
-      f).
-  Definition add {n m pred_nm} p q : T (S pred_nm) := @add_cps n m pred_nm p q _ id.
-
-  Definition sub_then_maybe_add_cps {n} mask (p q r : T n) {R} (f:T n -> R) :=
-    Columns.unbalanced_sub_cps (n3:=n) (uweight bound) p q
+  Definition add_cps {n} (p q: T n) {R} (f:T n->R) :=
```
JasonGross: This is wrong; it should be `T n -> T n -> T (S n)`

src/Arithmetic/Saturated.v
```diff
@@ -917,47 +1039,50 @@ Section API.
       pose proof Z.mul_split_div; pose proof Z.mul_split_mod;
       pose proof div_correct; pose proof modulo_correct.
 
-    Lemma eval_add_nz n m pred_nm p q :
-      pred_nm <> 0%nat ->
-      eval (@add n m pred_nm p q) = eval p + eval q.
+    Lemma eval_add_nz n p q :
+      n <> 0%nat -> (0 <= eval p + eval q < uweight bound n) ->
```
JasonGross: This is not a precondition that should be needed.

src/Arithmetic/Saturated.v
```diff
@@ -917,47 +1039,50 @@ Section API.
...
+    Lemma eval_add_z n p q :
       n = 0%nat ->
-      m = 0%nat ->
-      eval (@add n m pred_nm p q) = eval p + eval q.
+      eval (@add n p q) = eval p + eval q.
     Proof. intros; subst; reflexivity. Qed.
 
-    Lemma eval_add n m pred_nm p q (Hpred_nm : pred_nm = 0%nat -> n = 0%nat /\ m = 0%nat)
-      :  eval (@add n m pred_nm p q) = eval p + eval q.
+    Lemma eval_add n p q (H:0 <= eval p + eval q < uweight bound n)
```
JasonGross: This is not a precondition that should be needed.

src/Arithmetic/Saturated.v
```diff
@@ -747,36 +847,49 @@ Section API.
...
-  Definition sub_then_maybe_add_cps {n} mask (p q r : T n) {R} (f:T n -> R) :=
-    Columns.unbalanced_sub_cps (n3:=n) (uweight bound) p q
+  Definition add_cps {n} (p q: T n) {R} (f:T n->R) :=
+    Positional.sat_add_cps (s:=bound) p q _
+      (* drops last carry, this relies on bounds *)
+      (fun carry_result => f (snd carry_result)).
+  Definition add {n} p q : T n := @add_cps n p q _ id.
+
+  (* Wrappers for additions with slightly uneven limb counts *)
+  Definition add_S1_cps {n} (p: T (S n)) (q: T n) {R} (f:T (S n)->R) :=
```
JasonGross: This is wrong; the return type should be `T (S (S n))`

src/Arithmetic/Saturated.v
```diff
@@ -747,36 +847,49 @@ Section API.
...
+    join0_cps q (fun Q => add_cps p Q f).
+  Definition add_S1 {n} p q := @add_S1_cps n p q _ id. 
+  Definition add_S2_cps {n} (p: T n) (q: T (S n)) {R} (f:T (S n)->R) :=
```
JasonGross: This is wrong; the return type should be `T (S (S n))`

andres-erbsen:
I think it might be a good idea to merge this as-is, and deal with the admits as issues. @jadephilipoom ?

jadephilipoom:
And yes, I agree that admits can be dealt with as issues.

PR #237, opened 2017-07-03
Make gcc_scope and icc_scope

JasonGross:
This allows us, in principle, to have different code output for gcc vs
icc by issuing an `Open Scope icc_scope` or `Open Scope gcc_scope`
before the display output line.  By default, currently, `icc_scope` is
open.  Opening `gcc_scope` will swap the argument order for `subborrowx`.


andres-erbsen:
Is the idea that we would duplicate all p256 display files for gcc and non-gcc compilers? More clutter seems like not ideal. I will try and look into whether I can use C preprocessor macros to detect gcc and redefine addcarryx there, so we can compile the same C code with different compilers.

andres-erbsen:
I think I got the preprocessor macro right but still not passing tests on gcc -O3 (but -Og passes)... will investigate after dinner

andres-erbsen:
The issue seems to be something really nasty with link-time optimization, inline assembly constraints, and sbb. Not that unlikely that it is another gcc bug. Do you know gcc inline assembly constraint notation?

andres-erbsen:
All hail gcc. This took "only" ~6 hours. Result: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81300

PR #241, opened 2017-09-15
Update etc scripts to include governor

JasonGross:
It needs to be in performance, not powersave, to work well on my
machine.

While we're at it, also have the scripts print usage if you pass no
arguments, rather than giving an error message about $1 being unset.

### DIFF
--- a/etc/governor.sh
+++ b/etc/governor.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+set -eu
+
+usage() {
+  generators="$1"
+  echo "USAGE: $0 <$(echo "$generators" | tr ' ' '|')>"
+  exit 111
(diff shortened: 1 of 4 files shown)


etc/governor.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+set -eu
+
+usage() {
+  generators="$1"
+  echo "USAGE: $0 <$(echo "$generators" | tr ' ' '|')>"
```
andres-erbsen: Please add a comment with a concrete example of the usage line.

JasonGross: Added `USAGE: ./governor.sh <performance|powersave>`

PR #243, opened 2017-09-15
Update Makefile by deduplicating some file lists

JasonGross:
@andres-erbsen Are you fine with using more makefile variables and less copy-paste?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,70 +163,70 @@ $(DISPLAY_JAVA_VO:.vo=.log) : %JavaDisplay.log : %.vo %JavaDisplay.v src/Compile
 	$(HIDE)$(COQC) $(COQDEBUG) $(COQFLAGS) $*JavaDisplay.v | sed s'/\r\n/\n/g' > $@.tmp && mv -f $@.tmp $@
 
 src/Specific/X25519/C64/test: src/Specific/X25519/C64/compiler.sh src/Specific/X25519/x25519_test.c $(DISPLAY_X25519_C64_VO:Display.vo=.c) $(DISPLAY_X25519_C64_VO:Display.vo=.h) src/Specific/X25519/C64/scalarmult.c
-	src/Specific/X25519/C64/compiler.sh -o src/Specific/X25519/C64/test -I liblow -I src/Specific/X25519/C64/ src/Specific/X25519/x25519_test.c $(DISPLAY_X25519_C64_VO:Display.vo=.c) src/Specific/X25519/C64/scalarmult.c
+	src/Specific/X25519/C64/compiler.sh -o $@ -I liblow -I src/Specific/X25519/C64/ $(filter %.c %.s,$^)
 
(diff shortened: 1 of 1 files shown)


andres-erbsen:
I am in general interested in cleaning up the low-level compilation in the Makefile, but perhaps a slightly bigger but more principled change would be better? If this seems like the best we can do right now, what is the pattern for when you use `%` and when not?

JasonGross:
Ideally, we could have rules like this:

```makefile
TEST_FOLDERS := src/Specific/X25519/C64 third_party/openssl-curve25519 third_party/openssl-nistz256-amd64 ...
TEST_BINARIES := $(addsuffix /test,$(TEST_FOLDERS))
MEASURE_BINARIES := $(addsuffix /measure,$(TEST_FOLDERS))
MEASUREMENTS := $(addsuffix /measurements.txt,$(TEST_FOLDERS))

$(TEST_BINARIES) : %/test : %/compiler.sh ????? # how do we list the dependencies here?
	$*/compiler.sh -o $@ -I liblow -I $*/ $(filter %.c %.s,$^)

$(MEASURE_BINARIES) : %/measure : %/compiler.sh measure.c ????? # how do we list the dependencies here?
	$*/compiler.sh -o $@ -I liblow -I $* $(filter %.c %.s,$^) -D UUT=??? # how do we get the UUT here?

$(MEASUREMENTS) : %/measurements.txt : %/measure capture.sh etc/machine.sh etc/cpufreq etc/tscfreq
	./capture.sh $* ??? # how do we get the number of iterations here?
```
For each of these, we could either drop a file in the relevant directory listing the dependencies/UUT/iteration count, or we can have makefile variables for dependencies.  Or, perhaps better, we can separate out the declaration of dependencies, i.e.,
```makefile
src/Specific/X25519/C64/test src/Specific/X25519/C64/measure: $(DISPLAY_X25519_C64_VO:Display.vo=.c) $(DISPLAY_X25519_C64_VO:Display.vo=.h) src/Specific/X25519/C64/scalarmult.c
src/Specific/X25519/C64/test: src/Specific/X25519/x25519_test.c
src/Specific/X25519/C64/measure: UUT=crypto_scalarmult_bench
src/Specific/X25519/C64/measurements.txt: COUNT=2047
...

$(TEST_BINARIES) : %/test : %/compiler.sh
	$*/compiler.sh -o $@ -I liblow -I $*/ $(filter %.c %.s,$^)

$(MEASURE_BINARIES) : %/measure : %/compiler.sh measure.c
	$*/compiler.sh -o $@ -I liblow -I $* $(filter %.c %.s,$^) -D UUT=$(UUT)

$(MEASUREMENTS) : %/measurements.txt : %/measure capture.sh etc/machine.sh etc/cpufreq etc/tscfreq
	./capture.sh $* $(COUNT)
```

JasonGross:
Updated to factor more things

JasonGross:
I'm going to merge this, since the issues are just travis timing out, and I believe I've tested it locally

PR #248, opened 2017-10-06
Curves from JSON

JasonGross:
The X25519 curves are now generated from `.json` files.  This code only works in >= 8.7, because it makes use of the recently-merged-from-fiat `transparent_abstract` tactic to allow defining things in tactics without massive slowdown.  The structure is as follows:
0. The module types and tactic definitions that set up the infrastructure live in `src/Specific/Framework/`
1. There are `.json` files in `src/Specific/CurveParameters/` that specify curve characteristics.  These configuration files don't yet handle Montgomery/P256, nor Karatsuba.  @andres-erbsen @jadephilipoom do you have thoughts on unifying the synthesis in all these different cases?

A simple example is `x2555_130.json`, which is:
```json
{
    "modulus"          : "2^255-5",
    "base"             : "130",
    "a24"              : "121665 (* XXX TODO(andreser) FIXME?  Is this right for this curve? *)",
    "sz"               : "3",
    "bitwidth"         : "128",
    "carry_chain1"     : "default",
    "carry_chain2"     : ["0", "1"],
    "coef_div_modulus" : "2",
    "operations"       : ["ladderstep"]
}
```
A more complicated example is `x25519_c64.json`:
```json
{
    "modulus"          : "2^255-19",
    "base"             : "51",
    "a24"              : "121665",
    "sz"               : "5",
    "bitwidth"         : "64",
    "carry_chain1"     : "default",
    "carry_chain2"     : ["0", "1"],
    "coef_div_modulus" : "2",
    "operations"       : ["femul", "fesquare", "freeze", "ladderstep"],
    "extra_files"      : ["X25519_C64/scalarmult.c"],
    "compiler"         : "gcc -march=native -mtune=native -std=gnu11 -O3 -flto -fomit-frame-pointer -fwrapv -Wno-attributes",
    "mul_header"       : "(* Micro-optimized form from curve25519-donna-c64 by Adam Langley (Google) and Daniel Bernstein. See <https://github.com/agl/curve25519-donna/blob/master/LICENSE.md>. *)",
    "mul_code"
    :
    "
  uint128_t t[5];
  limb r0,r1,r2,r3,r4,s0,s1,s2,s3,s4,c;

  r0 = in[0];
  r1 = in[1];
  r2 = in[2];
  r3 = in[3];
  r4 = in[4];

  s0 = in2[0];
  s1 = in2[1];
  s2 = in2[2];
  s3 = in2[3];
  s4 = in2[4];

  t[0]  =  ((uint128_t) r0) * s0;
  t[1]  =  ((uint128_t) r0) * s1 + ((uint128_t) r1) * s0;
  t[2]  =  ((uint128_t) r0) * s2 + ((uint128_t) r2) * s0 + ((uint128_t) r1) * s1;
  t[3]  =  ((uint128_t) r0) * s3 + ((uint128_t) r3) * s0 + ((uint128_t) r1) * s2 + ((uint128_t) r2) * s1;
  t[4]  =  ((uint128_t) r0) * s4 + ((uint128_t) r4) * s0 + ((uint128_t) r3) * s1 + ((uint128_t) r1) * s3 + ((uint128_t) r2) * s2;

  r4 *= 19;
  r1 *= 19;
  r2 *= 19;
  r3 *= 19;

  t[0] += ((uint128_t) r4) * s1 + ((uint128_t) r1) * s4 + ((uint128_t) r2) * s3 + ((uint128_t) r3) * s2;
  t[1] += ((uint128_t) r4) * s2 + ((uint128_t) r2) * s4 + ((uint128_t) r3) * s3;
  t[2] += ((uint128_t) r4) * s3 + ((uint128_t) r3) * s4;
  t[3] += ((uint128_t) r4) * s4;
",
    "square_header"    : "(* Micro-optimized form from curve25519-donna-c64 by Adam Langley (Google) and Daniel Bernstein. See <https://github.com/agl/curve25519-donna/blob/master/LICENSE.md>. *)",
    "square_code"
    :
    "
  uint128_t t[5];
  limb r0,r1,r2,r3,r4,c;
  limb d0,d1,d2,d4,d419;

  r0 = in[0];
  r1 = in[1];
  r2 = in[2];
  r3 = in[3];
  r4 = in[4];

  do {
    d0 = r0 * 2;
    d1 = r1 * 2;
    d2 = r2 * 2 * 19;
    d419 = r4 * 19;
    d4 = d419 * 2;

    t[0] = ((uint128_t) r0) * r0 + ((uint128_t) d4) * r1 + (((uint128_t) d2) * (r3     ));
    t[1] = ((uint128_t) d0) * r1 + ((uint128_t) d4) * r2 + (((uint128_t) r3) * (r3 * 19));
    t[2] = ((uint128_t) d0) * r2 + ((uint128_t) r1) * r1 + (((uint128_t) d4) * (r3     ));
    t[3] = ((uint128_t) d0) * r3 + ((uint128_t) d1) * r2 + (((uint128_t) r4) * (d419   ));
    t[4] = ((uint128_t) d0) * r4 + ((uint128_t) d1) * r3 + (((uint128_t) r2) * (r2     ));
"
}
```
3. The `src/Specific/CurveParameters/remake_curves.sh` script holds a list of curves to be made, what directories they should end up living in, and it invokes `src/Specific/Framework/make_curve.py` to transform these files into outputs.  The Python script fills in a few defaults (such as computing `s` and `c` from the modulus, if you don't pass them explicitly), and does a lot of processing on the C code that is pasted verbatim from donna to get it to be in the right format for Coq.

This Python script creates the files:
- `CurveParameters.v` (the Coq-ified version of the json file, which instantiates an appropriate module type);
- `Synthesis.v`, which instantiates a `MakeSynthesisTactics` with the curve parameter modules, invokes a tactic from the applied module functor to synthesize all of the relevant non-reflective bits (basically, what used to live in @jadephilipoom 's `ArithmeticSynthesisTest.v`), and then instantiates another module functor `PackageSynthesis` which defines notations via tactics in terms to access the names of the various fields defined by the synthesis tactic; 
- any other files you ask it for, such as `compiler.sh`, `femul.v`, `femulDisplay.v`.  All of the `*Display.v` files are simple, and all the the operation synthesis files have a single `Definition` (with the appropriate type), and solve the definition by invoking a single tactic defined in `PackageSynthesis`, e.g., `synthesize_mul` or `synthesize_ladderstep`.


src/Specific/Framework/make_curve.py
```diff
@@ -0,0 +1,333 @@
...
+
+def repeat_until_unchanged(update, arg):
+    changed = True
+    while changed:
+        last = repr(arg)
+        arg = update(arg)
+        changed = (repr(arg) != last)
+    return arg
+
+def format_c_code(header, code, numargs, sz, indent='      ', closing_indent='            '):
```
andres-erbsen: What does this function do? What is the input format, what is the output format, how are they related?

src/Specific/Framework/ArithmeticSynthesisFramework.v
```diff
@@ -215,6 +229,99 @@ Module MakeArithmeticSynthesisTestTactics (Curve : CurveParameters.CurveParamete
...
+  Ltac pose_half_sz sz half_sz :=
+    let v := (eval compute in (sz / 2)%nat) in
+    cache_term v half_sz.
+  Ltac pose_half_sz_nonzero half_sz half_sz_nonzero :=
+    cache_proof_with_type_by
+      (half_sz <> 0%nat)
+      ltac:(cbv; congruence)
+             half_sz_nonzero.
+
+  Ltac basesystem_partial_evaluation_RHS :=
```
andres-erbsen: I am pretty sure there is another copy of this in basesystem... should probably deduplicate sometime.

src/Specific/Framework/ArithmeticSynthesisFramework.v
```diff
@@ -215,6 +229,99 @@ Module MakeArithmeticSynthesisTestTactics (Curve : CurveParameters.CurveParamete
...
+              | ?t _ => t
+              end) in
+    let t1 := fresh "t1" in
+    pose (t1 := t);
+     transitivity
+      (t1 (@id_with_alt) (@Let_In) (@runtime_and) (@runtime_shr) (@runtime_opp) (@runtime_add)
+         (@runtime_mul));
+     [ replace_with_vm_compute t1; clear t1 | reflexivity ].
+
+  (** XXX TODO(jadep) FIXME: Is sqrt(s) the right thing to pass to goldilocks_mul_cps (the original code hard-coded 2^224 *)
```
andres-erbsen: The goldilocks rule applies if p = 2^2k - 2^k - 1. However you extract the middle term should not really matter.

andres-erbsen:
Okay, I think I have a vague idea of what is going on here. The main gripe I have is that I think we have four different metaprogramming mechanisms now (partial evaluation, other tactics, module functors, python code-spew), and of course it would be nice to have fewer. I think the current state of this PR is jolly good for paper purposes, but as we want to integrate this into Chrome CI, I would want to at least justify the need for each layer.

Strawman 1: write a gallina definition that takes in a Coq record encoding whatever is in that JSON file and generates a [record -> record] such that the first record is a list of preconditions that are solvable by a tactic and the second record contains all the operations we care about.

src/Specific/Framework/make_curve.py
JasonGross: The input format is donna C code snippets, the output format is Gallina.  Assignments are turned into `dlet`, casts are stripped, arrays are exploded (so `output[0]` becomes `output0`, etc).  More ad-hoc, the inputs are assumed to live in the arrays `in`, `in2`, etc., and the outputs are assumed to live in the array `outputs` if it is named, otherwise whatever the first declared array is in the code.  Array ordering of limbs is assumed to be reversed from tuple ordering.

src/Specific/Framework/ArithmeticSynthesisFramework.v
JasonGross: Sounds good

JasonGross:
Question: is moving to requiring 8.7 fine?

Note: I don't quite recall what the setup was in this PR (I've done some reorganization in the one that adds Montgomery support, and like that one much better; maybe I should include the reorganization here, and just remove the Montgomery-specific bits for this PR).

> Strawman 1: write a gallina definition that takes in a Coq record encoding whatever is in that JSON file and generates a [record -> record] such that the first record is a list of preconditions that are solvable by a tactic and the second record contains all the operations we care about.

There are three issues that I have:
1. The second record will either need to duplicate most of the fields of the first record, or will need to be parameterized on it; types line `feBW` are generated from some of the synthesis targets/side conditions, and need to have a particular structure of identifiers that show up on unfolding both so that the reflective pipeline glue works, and so that things are fast (that is, syntactic equality, and not just judgmental equality, matters here).  I think the solution here is to figure out what our final synthesis output actually is (field * ladderstep?), and how to unify the synthesis targets between Montgomery (where the bound is p) and non-Montgomery (solinas?), where there's a per-word bound.  If I reorganize the curve parameters into a record, would you be able to write me the record type of synthesis targets?

2. This doesn't actually save us much boilerplate, at the moment.  Perhaps the issue is that I don't know what @jadephilipoom 's `solve_F_op` tactic does, nor how much of `ArithmeticSynthesisTest` can be moved to Gallina, so I'm not sure how to break apart those `sig` proofs like I did in Montgomery.  In any case, it seems like the record would have basically one tactic per precondition, and we'd use all the same tactics, but instead of combining the results into tuples (as is done here) or associative lists of identifiers (as is done in Montgomery), we'd be hardcoding the types (which seems less modularly extensible, at least at a first glance).  Perhaps one issue here is that I don't have a good sense of the parameter space, so my design targets maximal extensibility.  Another is that there's currently no uniformity in how the partial unfolding and solving is done.  Some things go by compute, others by vm_compute, others by a cbv -list, others by a cbv list, others by a "run vm_compute in this part and then abstract in this other part", and all of them need to hide the computed terms behind identifiers so that we don't incur a ~10x slowdown (and this is what we need 8.7 for).

3. I'm not sure how much of the preglue and glue transformations I can do in Gallina.  Hopefully clamping the types at both ends will let me do useful things, since I'll know the nesting levels of tuples, etc.

andres-erbsen:
So, let me make sure I understand the 8.7 case correctly: we need 8.7 to get transparent_abstract, we need transparent_abstract to make synthesized code hide behind a delta instead of a zeta, and we need to do that to speed up typechecking in goals with lots of tactic-generated code. I assume fiat people have already investigated alternatives to this, and nothing reasonable came up? (It is *such* a pity that so many things in Coq are sensitive to delta instead of beta+zeta...).

andres-erbsen:
I think is nothing magical in the `solve_` tactics, and just reading them should give a good view: https://github.com/mit-plv/fiat-crypto/blob/master/src/Arithmetic/Core.v#L1086

andres-erbsen:
> types line feBW are generated from some of the synthesis targets/side conditions

The code for picking word bounds should be easily translated from Ltac to gallina, if it isn't Gallina already. I also don't understand what the glue does, but just wrapping additional layers around it instead of detangling some sounds miserable.

JasonGross:
> to make synthesized code hide behind a delta instead of a zeta, and we need to do that to speed up typechecking in goals with lots of tactic-generated code

Ha.  Hahaha.  No.  If only it were that simple, then we could use `pose` as long as we stayed in the same proof context.  That is part of it, but there is also a large part that we need to make the content of the definitions disappear entirely from the proof context, so that tactics don't see them, and (when relevant) printing doesn't see them, and Ltac `match goal` doesn't see them, and typeclass resolution doesn't see them (and some tactics are sensitive to the different between proof-context-definitions and actual definitions, i.e., some tactics treat proof-context-definitions as `zeta`-reducible, and other tactics (such as `cbv`) treat them as `delta`-reducible).

JasonGross:
> The code for picking word bounds should be easily translated from Ltac to gallina, if it isn't Gallina already.

`Eval vm_compute`, `Eval cbv`, etc, are not part of Gallina as such.  All of the code for this is already `Eval <reduction-strategy> in <some fixed constr>`, or some `Ltac` equivalent.

JasonGross:
> I also don't understand what the glue does

The (pre)glue code does three things: (1) reassociate sigmas, arrows, products, etc; (2) `rewrite <- sig_from_previous_layer_of_synthesis by make_use_of_preconditions_tac ()`; (3) `change <some computed expression based on reflection primitives and with [dlet] in the right places`.

> but just wrapping additional layers around it instead of detangling some sounds miserable.

Agreed.  I'm going to try to do some detangling on top of #250.

andres-erbsen:
Okay, I did indeed miss the copious `Eval .. in ..` inside Gallina code. Thinking what to do about it. I still don't understand the complications that you are referring to wrt transparent abstract.

JasonGross:
> I still don't understand the complications that you are referring to wrt transparent abstract.

I haven't fully investigated the cause of the slowness, but it has something to do with where the terms show up, and how much you can control unfolding, I think.  If you want to play around with it, [the tip](https://github.com/JasonGross/fiat-crypto/commit/36f7c9407192a625d91b4656bfea9e4ebf9c9320) of [my zzz-auto-instantiate-curves-factor-tactics](https://github.com/JasonGross/fiat-crypto/tree/zzz-auto-instantiate-curves-factor-tactics) branch has an 8.6-compatible version of using tactics and records, rather than modules and notations, which shows a 6.5x slowdown in femul, an 11x slowdown in freeze, and a 3.5x slowdown in ArithmeticSynthesisTest.  I got to the point where I was trying random things to make up the performance difference with existing code, and so I gave up and instead went back and redesigned my approach so that I'd produce terms that were syntactically equal to the original ones up to renaming.  This led to much more predictable performance preservation.

andres-erbsen:
Okay. I guess if Coq performance was understandable we wouldn't be having this discussion anyway.

src/Specific/Framework/ArithmeticSynthesisFramework.v
JasonGross: This one also includes `id_tuple_with_alt id_tuple'_with_alt` in the `cbv delta` list.  @jadephilipoom (or @andres-erbsen ) is it fine to add those to the one in `src/Arithmetic/Core.v`?

src/Specific/Framework/ArithmeticSynthesisFramework.v
JasonGross: *nods*  I've updated the Gallina-from-JSON Python script to know about this default.  (I could do it in Gallina instead of Python, if that's preferred.)

andres-erbsen:
I don't know how to do better on the overall design. So I guess it is going to be this one. Some smaller things that might be worth looking into

- maybe report a clear coq performance issue for it being so much slower with records? For example, would synthesizing a record instead of a module (using the strategy here) be a small enough diff to report its performance impact as ridiculous?
- have clear rules for what should be in python and what should be in tactics? For example, carry chain inference is not fundamentally different from deciding whether to do goldilocks or not.
- figure out the easiest way to get Ocaml 4.02.1 (for coq8.7) on Debian/Ubuntu...

JasonGross:
> I don't know how to do better on the overall design. So I guess it is going to be this one.

I think it's worth moving more things into Gallina, and removing the modules.  I will move the curve parameters from a module into a record, but the framework is still going to use modules internally.  In particular, the modules are there so that we can refer to things with names in the type of, e.g., our final `add` synthesis goal.  Do you have a sense of how to write the top-level synthesis goal (`field` or `ring` or whatever, + `ladderstep` + ...?)  Preferably in a way that supports both saturated and unsaturated / montgomery and non-montgomery.

JasonGross:
> maybe report a clear coq performance issue for it being so much slower with records? For example, would synthesizing a record instead of a module (using the strategy here) be a small enough diff to report its performance impact as ridiculous?

I will try to get things into a state where the difference between `transparent assert (H : T); [ tac () | ]` and `transparent assert (H : T); [ transparent_abstract tac () | ]` is absurd.  But first we need to get the entire pipeline into a single top-level goal.

JasonGross:
> figure out the easiest way to get Ocaml 4.02.1 (for coq8.7) on Debian/Ubuntu...

Which version of Debian/Ubuntu?  I have packages for 8.7+beta1 and 8.7+beta2 [on launchpad](https://launchpad.net/~jgross-h/+archive/ubuntu/many-coq-versions/+packages), for trusty and precise, and I can set up packages for other versions if you want support for something newer.  (Alternative route: use opam)

andres-erbsen:
re perf: could we get an example that demonstrates that it is important to use modules internally?

saturated and unsaturated arithmetic can share a goal that describes them as implementations of ring arithmetic. inversion is built purely from ring operations and the fact that the ring is Z_q. unsaturated elliptic curve operations use the *code* from ring operations, but not with the same specification for input and output ranges as it is used with in ring operations. Currently, we inline and reanalyze. This seems like poor form regardless, so I think it is worth thinking about changing it...

JasonGross:
> re perf: could we get an example that demonstrates that it is important to use modules internally?

Modules are not important for performance, per se.  The essential thing for performance is `transparent_abstract`.  The use of modules is purely for ease of notation and minimization of curve-specific python-generated code: when I write
```coq
Definition xzladderstep :
  { xzladderstep : feW -> feW * feW -> feW * feW -> feW * feW * (feW * feW)
  | forall x1 Q Q',
      let xz := xzladderstep x1 Q Q' in
      let eval := B.Positional.Fdecode wt in
      feW_bounded x1
      -> feW_bounded (fst Q) /\ feW_bounded (snd Q)
      -> feW_bounded (fst Q') /\ feW_bounded (snd Q')
      -> ((feW_bounded (fst (fst xz)) /\ feW_bounded (snd (fst xz)))
          /\ (feW_bounded (fst (snd xz)) /\ feW_bounded (snd (snd xz))))
         /\ Tuple.map (n:=2) (Tuple.map (n:=2) phiW) xz = FMxzladderstep (m:=m) (eval (proj1_sig a24_sig)) (phiW x1) (Tuple.map (n:=2) phiW Q) (Tuple.map (n:=2) phiW Q') }.
Proof.
```
I don't want to mention the particular package I'm using for every single instance of `phiW`, `feW_bounded`, `feW`, `FMxzladderstep`, `a24_sig`, `m`, `wt`, etc.  I chose to fix this with modules, because they support mix-in notations, rather than with typeclasses and global notations that I have to be sure to pass along as `Export` statements.  We could instead use typeclasses; would you prefer this?

JasonGross:
Note: I've pushed a commit that brings in most of the reorganization done to support montgomery

JasonGross:
Pushed the commit adding support for montgomery, now that I've finished tracking down where the slowdown came from (it was from `Print Assumptions`)

JasonGross:
> saturated and unsaturated arithmetic can share a goal that describes them as implementations of ring arithmetic

Can they actually?  Unsaturated currently proves a goal of the form:
```coq
ring
     : Hierarchy.ring (Z ^ sz) (B.Positional.eq wt m)
         (` zero_sig) (` one_sig) (` opp_sig) (` add_sig)
         (` sub_sig) (` mul_sig) /\
       Ring.is_homomorphism (F m) eq 1 F.add F.mul (Z ^ sz)
         (B.Positional.eq wt m) (` one_sig) (` add_sig)
         (` mul_sig) (B.Positional.Fencode wt) /\
       Ring.is_homomorphism (Z ^ sz) (B.Positional.eq wt m)
         (` one_sig) (` add_sig) (` mul_sig) (F m) eq 1 F.add F.mul
         (B.Positional.Fdecode wt)
```
But this goal seems false (or at least not easily provable) for saturated arithmetic, where I don't have any theorems that `mul` gives the correct answer (even modulo `m`) unless both of its arguments are less than `m`.  So we'd need to make the carrier type a sigma for saturated; does this seem right to you?  If so, it seems like the best we can do for a goal is something like
```coq
Record SynthesisOutput (curve : RawCurveParameters.CurveParameters) :=
  {
    m := Z.to_pos (RawCurveParameters.s curve - B.Associational.eval (RawCurveParameters.c curve))%Z;
    T : Type;
    eqT : T -> T -> Prop;
    encode : F m -> T;
    decode : T -> F m;
    zero : _;
    one : _;
    opp : _;
    add : _;
    sub : _;
    mul : _;
    ring : @Hierarchy.ring T eqT zero one opp add sub mul;
    encode_homomorphism
    : @Ring.is_homomorphism
        (F m) eq 1 F.add F.mul
        T eqT one add mul
        encode;
    decode_homomorphism
    : @Ring.is_homomorphism
        T eqT one add mul
        (F m) eq 1 F.add F.mul
        decode;
    a24t : if RawCurveParameters.montgomery curve return Type then unit else T;
    xzladderstep : if RawCurveParameters.montgomery curve return Type then unit else (T -> T -> T * T -> T * T -> (T * T) * (T * T));
    xzladderstep_correct
    : (if RawCurveParameters.montgomery curve as mont
          return (if mont return Type then unit else T)
                 -> (if mont return Type then unit else (T -> T -> T * T -> T * T -> (T * T) * (T * T)))
                 -> Prop
       then fun _ _ => True
       else fun a24' xzladderstep
            => forall x1 Q Q',
                Tuple.fieldwise
                  (n:=2) (Tuple.fieldwise (n:=2) eq)
                  (Tuple.map (n:=2) (Tuple.map (n:=2) decode) (xzladderstep a24' x1 Q Q'))
                  (@XZ.M.xzladderstep (F m) F.add F.sub F.mul (F.of_Z m a24) (decode x1) (Tuple.map (n:=2) decode Q) (Tuple.map (n:=2) decode Q')))
        a24t xzladderstep

  }.
```

Does this seem right to you?

andres-erbsen:
Hmm, the Z-level spec of unsaturated arithmetic is indeed surprisingly nice. Once we are down to bounded words, both should have similar specifications (with bounds preconditions). The Z-level goals are an internal-only interface, we should change them in whichever way makes the word-level goals easier.

JasonGross:
Ah, right, these are internal.  But, still, the specifications never line up, currently, until you actually hit reification (post glue), which is far too low level.  The saturated spec would look something like:
```
ring
     : Hierarchy.ring { v : Z ^ sz | 0 <= B.Positional.Fdecode wt v < m } (fun x y => B.Positional.eq wt m (` x) (` y))
         (` zero_sig) (` one_sig) (` opp_sig) (` add_sig)
         (` sub_sig) (` mul_sig) /\
       Ring.is_homomorphism (F m) eq 1 F.add F.mul { v : Z ^ sz | 0 <= B.Positional.Fdecode wt v < m }
         (B.Positional.eq wt m) (` one_sig) (` add_sig)
         (` mul_sig) (fun x => exist _ (B.Positional.Fencode wt x) ????) /\
       Ring.is_homomorphism { v : Z ^ sz | 0 <= B.Positional.Fdecode wt v < m } (fun x y => B.Positional.eq wt m (` x) (` y))
         (` one_sig) (` add_sig) (` mul_sig) (F m) eq 1 F.add F.mul
         (fun x => B.Positional.Fdecode wt (` x))
```
The unsaturated spec does not require this `0 <= v < m` condition, and I'm not entirely sure how to get it from `carry_sig` or whatever.  It's also not the cleanest spec, and has me wondering if we should be talking about an alternate spec of ring homomorphism where you take both a carrier type and a predicate on that type, and keep the predicate preservation properties separate.  We could even imagine a more general spec where you specify the carry operation in addition to the others, and its only carry_add, carry_mul, etc, that need to preserve the predicate.
----
So for the top-level spec, are you thinking something like this?
```coq
Record SynthesisOutput (curve : RawCurveParameters.CurveParameters) :=
  {
    m := Z.to_pos (RawCurveParameters.s curve - B.Associational.eval (RawCurveParameters.c curve))%Z;
    bounds := (expr computing bounds);
    T := { v : feBW bounds | 0 <= eval v < m };
    eqT := ...;
    encode := ...;
    decode := ...;
    zero : _;
    one : _;
    opp : _;
    add : _;
    sub : _;
    mul : _;
    ring : @Hierarchy.ring T eqT zero one opp add sub mul;
    encode_homomorphism
    : @Ring.is_homomorphism
        (F m) eq 1 F.add F.mul
        T eqT one add mul
        encode;
    decode_homomorphism
    : @Ring.is_homomorphism
        T eqT one add mul
        (F m) eq 1 F.add F.mul
        decode;
    a24t : if RawCurveParameters.montgomery curve return Type then unit else T;
    xzladderstep : if RawCurveParameters.montgomery curve return Type then unit else (T -> T -> T * T -> T * T -> (T * T) * (T * T));
    xzladderstep_correct
    : (if RawCurveParameters.montgomery curve as mont
          return (if mont return Type then unit else T)
                 -> (if mont return Type then unit else (T -> T -> T * T -> T * T -> (T * T) * (T * T)))
                 -> Prop
       then fun _ _ => True
       else fun a24' xzladderstep
            => forall x1 Q Q',
                Tuple.fieldwise
                  (n:=2) (Tuple.fieldwise (n:=2) eq)
                  (Tuple.map (n:=2) (Tuple.map (n:=2) decode) (xzladderstep a24' x1 Q Q'))
                  (@XZ.M.xzladderstep (F m) F.add F.sub F.mul (F.of_Z m a24) (decode x1) (Tuple.map (n:=2) decode Q) (Tuple.map (n:=2) decode Q')))
        a24t xzladderstep
  }.
```

JasonGross:
Perhaps we should chat synchronously about what the spec should be?

andres-erbsen:
Okay. I don't have any meetings today.

src/Specific/CurveParameters/x2448_c64_karatsuba.json
```diff
@@ -0,0 +1,12 @@
+{
+    "modulus"          : "2^448-2^224-1",
+    "base"             : "56",
+    "goldilocks"       : "true",
```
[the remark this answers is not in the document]
JasonGross: It means currently means "whatever is coded in `src/Specific/Framework/ArithmeticSynthesis/Karatsuba.v`"; I'm not fully clear on the differences.  Right now, it's whatever is coded for by `goldilocks_mul_cps`.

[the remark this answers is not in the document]
andres-erbsen: Wonderful. Once I get coq to build again, I will investigate why that synthesis is not failing for curve25519.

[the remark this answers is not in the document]
JasonGross: Note that this is currently controlled by the line
```python
                        ('Karatsuba.v', (CP_BASE_DEFAULTS_LIST, 'goldilocks')),
```
in `src/Specific/Framework/ArithmeticSynthesis/remake_packages.py` which is interpreted as "things in `Karatsuba.v` should only be turned on if `goldilocks` is set" (the `CP_BASE_DEFAULTS_LIST` says which files define constants which are expected to be present by the time we do stuff in `Karatsuba.v`).  I suppose we'd probably want to move the goldilocks stuff to `Goldilocks.v` and repurpose `Karatsuba.v`, and then add the relevant lines to `remake_packages.py`.

JasonGross:
Definition of final synthesis target added in `src/Specific/Framework/OutputType.v`:
```coq
Record SynthesisOutput (curve : RawCurveParameters.CurveParameters) :=
  {
    m := Z.to_pos (curve.(s) - Associational.eval curve.(c))%Z;
    rT := ((Tbase (TWord (Z.log2_up curve.(bitwidth))))^curve.(sz))%ctype;
    T' := interp_flat_type rT;
    RT := (Unit -> rT)%ctype;
    wt := wt_gen m curve.(sz);

    encode : F m -> Expr RT
    := fun v var
       => Abs
            (fun _
             => SmartPairf
                  (flat_interp_untuple
                     (T:=Tbase _)
                     (Tuple.map
                        (fun v => Op (OpConst v) TT)
                        (@Positional.Fencode wt curve.(sz) m modulo div v))));
    decode : T' -> F m
    := fun v => @Positional.Fdecode
                  wt (curve.(sz)) m
                  (Tuple.map interpToZ (flat_interp_tuple (T:=Tbase _) v));
    zero : Expr RT;
    one : Expr RT;
    add : Expr (rT * rT -> rT); (* does not include carry *)
    sub : Expr (rT * rT -> rT); (* does not include carry *)
    mul : Expr (rT * rT -> rT); (* includes carry *)
    square : Expr (rT -> rT); (* includes carry *)
    opp : Expr (rT -> rT); (* does not include carry *)
    carry : Expr (rT -> rT);
    carry_add : Expr (rT * rT -> rT)
    := (carry ∘ add)%expr;
    carry_sub : Expr (rT * rT -> rT)
    := (carry ∘ sub)%expr;
    carry_opp : Expr (rT -> rT)
    := (carry ∘ opp)%expr;

    P : T' -> Prop;

    encode_correct : forall v, _;
    encode_sig := fun v => exist P (Interp (encode v) tt) (encode_correct v);

    decode_sig := fun v : sig P => decode (proj1_sig v);

    zero_correct : zero = encode 0%F; (* which equality to use here? *)
    one_correct : one = encode 1%F; (* which equality to use here? *)
    zero_sig := encode_sig 0%F;
    one_sig := encode_sig 1%F;

    opp_correct : forall x, P x -> P (Interp carry_opp x);
    opp_sig := fun x => exist P _ (@opp_correct (proj1_sig x) (proj2_sig x));

    add_correct : forall x y, P x -> P y -> P (Interp carry_add (x, y));
    add_sig := fun x y => exist P _ (@add_correct (proj1_sig x) (proj1_sig y) (proj2_sig x) (proj2_sig y));

    sub_correct : forall x y, P x -> P y -> P (Interp carry_sub (x, y));
    sub_sig := fun x y => exist P _ (@sub_correct (proj1_sig x) (proj1_sig y) (proj2_sig x) (proj2_sig y));

    mul_correct : forall x y, P x -> P y -> P (Interp mul (x, y));
    mul_sig := fun x y => exist P _ (@mul_correct (proj1_sig x) (proj1_sig y) (proj2_sig x) (proj2_sig y));


    T := { v : T' | P v };
    eqT : T -> T -> Prop
    := fun x y => eq (decode (proj1_sig x)) (decode (proj1_sig y));
    ring : @Hierarchy.ring
             T eqT zero_sig one_sig opp_sig add_sig sub_sig mul_sig;
    encode_homomorphism
    : @Ring.is_homomorphism
        (F m) eq 1%F F.add F.mul
        T eqT one_sig add_sig mul_sig
        encode_sig;
    decode_homomorphism
    : @Ring.is_homomorphism
        T eqT one_sig add_sig mul_sig
        (F m) eq 1%F F.add F.mul
        decode_sig
  }.
```

andres-erbsen:
I think the synthesis goal you wrote is good for driving design, assuming satisfying the convertibility it requires is easy. We will probably change it later, but most likely in ways that make it simpler.

src/Specific/Framework/ArithmeticSynthesis/Karatsuba.v
```diff
@@ -0,0 +1,140 @@
...
+Require Import Crypto.Util.Tactics.PoseTermWithName.
+Require Import Crypto.Util.Tactics.CacheTerm.
+
+Local Notation tuple := Tuple.tuple.
+Local Open Scope list_scope.
+Local Open Scope Z_scope.
+Local Coercion Z.of_nat : nat >-> Z.
+Local Infix "^" := Tuple.tuple : type_scope.
+
+(** XXX TODO(jadep) FIXME: Is sqrt(s) the right thing to pass to goldilocks_mul_cps (the original code hard-coded 2^224 *)
```
jadephilipoom: Goldilocks-compatible primes should always have the form `2^(2k) - 2^k - c`, and it should be `2^k` that is used here. So yes, this will give you the right answer on goldilocks-compatible primes, but it also will not fail as it should on incompatible primes. This may be why goldilocks somehow succeeded for curve25519 in the discussion above.

JasonGross:
@jadephilipoom Did you see https://github.com/mit-plv/fiat-crypto/pull/248#discussion_r144113338 ? (fusing the two `basesystem_partial_evaluation_RHS` tactics)

> I'd like to somehow make sure that the output code is still sane/correct; could we rerun some benchmarks and tests

The generated C-code should be character-for-character identical.  You can look at changes in `*Display.log`, most of which should be just moving files between directories.  (Hm, I guess I should fix the lack of unfolding of `Synthesis.P.bound1`...)

JasonGross:
@andres-erbsen The synthesis goal is actually not good enough for ladderstep; we need to provide both the pre-bounds-analyzed code, and the post-bounds-analyzed code...

andres-erbsen:
8.7.0 and trunk on 0356c2012aa83b56375565aae84913df1e3d6fdd give

File "./src/Compilers/CommonSubexpressionEliminationProperties.v", line 172, characters 36-43:
Error: The reference nat_beq was not found in the current environment.

JasonGross:
Note that trunk is very outdated; Coq has switched to master recently.  There is also no V8.7.0 tag yet, so I'm not entirely sure which version of Coq you're using...  Does fiat-crypto master branch build for you on the version of Coq you're using?

JasonGross:
Huh, now I'm getting the same error, both here and on master.  I'm confused what's changed to cause this...

andres-erbsen:
ladderstep32.log https://gist.github.com/andres-erbsen/e3cd3a9e30357b1bf71aded16b877a8f

PR #249, opened 2017-10-07
Hopefully fix travis

JasonGross:
Travis was too slow to run all of the coq target, so we split off the
no-curves-proofs and the curves-proofs targets, and, while we're at it,
also add 8.7+beta1 tests.


JasonGross:
This does not fully fix travis, but it is better than nothing

PR #250, opened 2017-10-10
Curves from JSON with Montgomery

JasonGross:
This is a version of @248 which supports Montgomery / P256.  It's not yet ready for merging, as there's a bit of a slowdown.


JasonGross:
Fused into #248

PR #256, opened 2017-10-13
Factor out truncation_bounds

JasonGross:
This makes it easier to extend the bounds analysis framework.


andres-erbsen:
LGTM. Perhaps also inline `t := bounds`?

JasonGross:
> Perhaps also inline `t := bounds`?

I'm open to doing that, but, why?  This keeps a bit of separation between "`Z`-based bounds (i.e., `zrange`)" and "bounds on reflective things".  In particular, if we do this (and then do you also want to inline `Notation bounds := zrange.`?), we should probably also move most of the code here to `Crypto.Util.ZRange`....

I'll merge this as is for now; would you open an issue for inlining `t := bounds` if you think that's a good idea?

PR #259, opened 2017-10-20
Loops (trying again)

andres-erbsen:
Here is a cleaned up version of the loop framework @jadephilipoom  wrote a while back. I am considering using this for boringssl ed25519, so go, critique. Current point of uncleaniness

1. Handling termination in lemma statements `loop` vs `loop_default`, `while` vs `while_default`
2. Namespacing, naming

Notations would be desirable, but let's worry about that after we figure out the semantics.

### DIFF
--- a/src/Experiments/Loops.v
+++ b/src/Experiments/Loops.v
@@ -73,60 +73,29 @@ Section Loops.
     cbv [terminates loop_default] in *; break_match; congruence.
   Qed.
 
-  Definition invariant_for
-             (inv : continue_state -> Prop)
-             (P : break_state -> Prop) : Prop :=
-    (forall state state',
-        body state = inr state' -> inv state ->
-        inv state') /\ 
-    (forall state state',
-        body state = inl state' -> inv state ->
-        P state').      
-
-  Local Ltac loop_simplify :=
-    repeat match goal with
-           | _ => progress (simpl loop in * )
-           | _ => progress intros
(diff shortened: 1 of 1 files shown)


src/Experiments/Loops.v
```diff
@@ -0,0 +1,264 @@
+Require Import Coq.Lists.List.
+Require Import Lia. 
+Require Import Crypto.Util.ListUtil.
+Require Import Crypto.Util.Decidable.
+Require Import Crypto.Util.Sum.
+Require Import Crypto.Util.Prod.
+Require Import Crypto.Util.Option.
+Require Import Crypto.Util.LetIn.
+
+Notation break := (@inl _ _).
```
JasonGross: I am not a fan.  Either these should be `only parsing`, or they should be `Local Notation`s, or they should be specific to the type, or should live inside a module not exported by default.  Or, perhaps better, use some new `Inductive` rather than `sum`.

src/Experiments/Loops.v
```diff
@@ -24,8 +20,8 @@ Section Loops.
     funapp
     (body_cps start _) (fun next =>
                       match next with
-                      | break state => Some (ret state)
-                      | continue state =>
+                      | inl state => Some (ret state)
```
andres-erbsen: @jadephilipoom do we know how to prove this CPS loop equivalent to the non-cps version here?

PR #264, opened 2017-10-31
Preliminary parameterization of reference algorithms over constant values

JasonGross:


JasonGross:
No longer useful; the relevant bits have been pushed to master

Issue #257, opened 2017-10-13
Boundedness of Fencode


JasonGross:
Suppose I have that `forall i, 0 < wt (S i) / wt i <= wt 1`, and `wt 0 = 1` (and also `forall i, 0 < wt i`).  I want
```coq
forall v,
  Tuple.map (fun x => (Z.max 0 x) mod (wt 1)) (Positional.Fencode wt v)
  = Positional.Fencode wt v
```

@jadephilipoom Can you add a proof of this, or something closely related, or tell me how to prove it?

andres-erbsen:
I confused about how this `wt 1` got in there. What is the higher-level goal for which you need this?

JasonGross:
The higher-level goal is `forall v, decode b (Interp (encode b v) tt) = v` or
```coq
  Positional.Fdecode (wt curve)
    (Tuple.map interpToZ
       (flat_interp_tuple
          (interpf interp_op
             (SmartPairf
                (flat_interp_untuple
                   (Tuple.map (fun v0 : Z => Op (OpConst v0) ()%expr)
                      (Positional.Fencode (wt curve) v))))))) = v
```
I have that `wt i := 2^Qceiling((Z.log2_up m/sz)*i)`, so `wt 1` is how I chose to access `2^Qceiling((Z.log2_up m/sz))` in a generic way.

andres-erbsen:
Do we still need this?

JasonGross:
This would still be nice to have; I expect that either we will need to prove this, or it'll show up as a side-condition of the redesigned mostly-in-gallina synthesis pipeline.

JasonGross:
Yes, I still need this.  I'm currently admitting
```coq
forall (sz : nat) (wt : nat -> Z) (m : positive) (v : F m),
  (forall i : nat, wt i <> 0) ->
  (forall i : nat, 0 <= wt i) ->
  (forall i : nat, wt (S i) / wt i <= wt 1%nat) ->
  @Tuple.map sz Z Z (fun x : Z => Z.max 0 x mod wt 1%nat)
    (@Positional.Fencode wt sz m (@modulo_cps) (@div_cps) v) =
  @Positional.Fencode wt sz m (@modulo_cps) (@div_cps) v
```

I think the primary place it shows up is in proving that we can use encoded word things to form a correct ring.

JasonGross:
Hmmm, I also need a proof that encode is bounded by the tight bounds....

JasonGross:
[...]

PR #267, opened 2017-11-14
Add MontgomeryAPI.encode and two lemmas about it

JasonGross:
One of them is Admitted; @jadephilipoom , can you prove it?  (Note that this is a special case of #257.)


jadephilipoom:
And yes, I am pretty sure I can prove the admitted lemma.

PR #269, opened 2017-11-17
Remove ad-hoc copy of CoqPrime.

maximedenes:
Coq Prime is now available on github, under LGPL, and comes with an OPAM
package.

Fiat-crypto seems to build entirely.


maximedenes:
Note that having copies of dependencies like CoqPrime in fiat-crypto is problematic when we are working on Coq PR because we need to make the fixes twice (once upstream and once in the embedded copy).

That happened for instance in https://github.com/coq/coq/pull/1049

andres-erbsen:
Thanks for bringing this up again; we find the manually tweaked copy of coqprime to be a nuisance as well.

Unfortunately, fiat-crypto still needs lemmas from coqprime. I believe it is Fermat's little theorem, and maybe something else on the same topic. https://github.com/mit-plv/fiat-crypto/blob/master/src/Util/NumTheoryUtil.v fails to build after this change, see https://travis-ci.org/mit-plv/fiat-crypto/jobs/303583624#L8792 . So I don't think we can merge it.

We would be happy to switch to using the actively maintained repository of coqprime as a git submodule, as we are doing for FCF at https://github.com/mit-plv/cross-crypto. This would probably require slight changes to the directory structure of coqprime, similar to https://github.com/adampetcher/fcf/pull/9. Would that be helpful?

maximedenes:
> Unfortunately, fiat-crypto still needs lemmas from coqprime.

Sure, but that's fine. I confirmed that fiat-crypto builds, with CoqPrime installed of course.
What I should probably do is to add a mention to that dependency in the README.

> witch to using the actively maintained repository of coqprime as a git submodule

Why would you need to do that? If you plan to bring changes to CoqPrime, I suggest you open PRs against the upstream repository instead.

andres-erbsen:
From the standpoint of fiat-crypto alone, it is efficient to tie every fiat-crypto commit to a specific coqprime commit, so that updates in the coqprime repository only affect fiat-crypto when we are there to handle it. Using coqprime as a git submodule in COQPATH satisfies this requirement, as does just bundling the entirety of coqprime and putting it in COQPATH. (Especially with the git submodule solution, it would be entirely reasonable to set up a regular job that checks for updates to coqprime, and bumps the submodule commit pointer as long as the fiat-crypto build still succeeds. We haven't done this solely because we have needed to update coqprime once, I think, as a part of a coq version migration.)

Relying on the coqprime in the unix user home directory can cause version inconsistencies that are much more difficult to handle. Sure, we can have the readme say that to build fiat-crypto, you need some particular version of coqprime. However, some other locally-installed library depending on coqprime could be requesting a different and incompatible version, perhaps just because they were slightly faster or slower in reacting to a coqprime change than fiat-crypto. Or perhaps some project knowlingly relies on an old coqprime, like VST bundles a diverged version of FCF. In this scenario, not only are dependency issues more likely than in my proposal, but they also impact developers who have nothing to do with fiat-crypto's use of coqprime.

The usual diamond argument that bundling coqprime would break a project that depends both on fiat-crypto and something else that depends on coqprime is unconvincing as depending on fiat-crypto in its current state would be unwise to begin with. Even when/if fiat-crypto becomes stable enough for that, there should be no reason to require the two coqprime versions in this scenario to match, as fiat-crypto uses coqprime for opaque proofs only (I guess to actually achieve that, we would need to namespace our pinned coqprime under fiat-crypto).

That said, if not pinning dependencies is a requirement for being tested in Coq CI system, we will comply. 

I am failing to build coqprime from github: https://gist.github.com/andres-erbsen/0715ad055cab1246f7d9296c90c9e2d9

andres-erbsen:
I investigated the build failure. It seems that thery/coqprime depends on coq/bignums, and coq/bignums tries to build a coq plugin, so it depends on ocaml and the coq ocaml libraries. However, the proofs required by fiat-crypto seem to be unaffected by this; `make ./PrimalityTest/Zp.vo` succeeds without building the plugin.

There still is the issue of accessing coqprime .vo files from fiat-crypto. Currently, the coqprime repository directory layout does not match its Coq namespace usage, so is not suitable for being on `COQPATH`. I looked through https://github.com/coq/coq/blob/master/lib/envars.ml and https://github.com/coq/coq/blob/master/tools/coq_makefile.ml trying to find a way to set a destination directory for `make install`, but couldn't find any. On the machine that I currently use, `make install` fails because `/usr/lib` is not writable by the fiat-crypto account, and that seems like a good thing.

If thery/coqprime did not depend on coq/bignums, I would just go ahead and make a pull request to make thery/coqprime `COQPATH`-compatible, but currently that would be only useful for projects like fiat-crypto.

Based on this, I can split the options here into three categories

1. accept dependence on system-wide mutable state; require the results of `make install` on coq/bignums and thery/coqprime; figure out how to change fiat-crypto developers' setups to support that
2. make coqprime a submodule, figure out a patching script / Makefile hack to allow its .vo-s to be used in fiat-crypto
3. replace the lemmas imported from coqprime, either with admits or independent proofs

I am inferring that @maximedenes is in favor of option 1, which makes me think that this is the implicit/default option for coq projects (and perhaps we just missed some memo when we started using git-submodule for dependencies). Option 2 seems the most principled, except that I haven't thought about how to practically apply it to coq plugins, and some part of coqprime not used by fiat-crypto depends on one. Option 3 is obviously dissatisfying, but depending on what the "figure out" parts of options 1 and 2 result in, it might be the most reasonable thing to do here.

maximedenes:
@JasonGross can you summarize why this was closed?

JasonGross:
Uh, because @andres-erbsen made a commit which said "closes #269"?  I suspect because we are no longer maintaining an ad-hoc copy of coqprime and we are instead pinning to a particular commit of it with a submodule.

andres-erbsen:
Yes. Poked by another instance where manual updates to fiat-crypto's coqprime were needed (https://github.com/mit-plv/fiat-crypto/pull/316), I went and did https://github.com/thery/coqprime/pull/3 and https://github.com/mit-plv/fiat-crypto/pull/317. At least now it should be possible to mechanically propagate updates from coqprime using `git submodule update --recursive --remote`.

The higher-level judgement call I made here is that the previous situation with an ad-hoc copy of coqprime was obviously undesirable, and is now mitigated. I don't think this answers all the questions (e.g., whether it is acceptable for coq ci-s to pin dependency versions, or to avoid opam), but the changes I made seem like a strict improvement to me. If there is something here that seems unresolved to you, let's track that in a separate issue.

PR #271, opened 2017-11-22
Demo.v without tuples

JasonGross:


src/DemoWithReification.v
```diff
@@ -0,0 +1,210 @@
...
+  Definition eval x := Associational.eval (@to_associational x).
+  Lemma eval_to_associational x :
+    Associational.eval (@to_associational x) = eval x.
+  Proof. trivial.                                             Qed.
+
+  (* SKIP over this: zeros, add_to_nth *)
+  Local Ltac push := autorewrite with push_eval push_map distr_length
+    push_flat_map push_fold_right push_nth_default cancel_pair natsimplify.
+  Definition zeros n : list Z
+    := List.map (fun _ => 0) (seq 0 n).
```
andres-erbsen: List.replicate?

src/DemoWithReification.v
```diff
@@ -0,0 +1,210 @@
...
+  Hint Rewrite eval_reduce : push_eval.
+End Associational.
+
+Module Positional. Section Positional.
+  Context (weight : nat -> Z)
+          (weight_0 : weight 0%nat = 1)
+          (weight_nz : forall i, weight i <> 0).
+
+  Definition to_associational (xs:list Z) : list (Z*Z)
+    := combine (map weight (List.seq 0 (List.length xs))) xs.
```
andres-erbsen: Add a `nat` argument for the length, require it to be equal to `length xs` as a precondition to the correctness proof. (I am not sure about this, but it seems like a direct translation from tuple code).

src/DemoWithReification.v
```diff
@@ -0,0 +1,210 @@
...
+  pose proof place_in_range a (pred n); try omega; try nsatz;
+  apply fold_right_invariant; cbv [zeros add_to_nth];
+  intros; rewrite ?map_length, ?seq_length, ?length_update_nth;
+  try omega.                                                  Qed.
+  Hint Rewrite @eval_from_associational : push_eval.
+
+  Section mulmod.
+    Context (m:Z) (m_nz:m <> 0) (s:Z) (s_nz:s <> 0)
+            (c:list (Z*Z)) (Hm:m = s - Associational.eval c).
+    Definition mulmod (a b:list Z) : list Z
```
andres-erbsen: Again, to blindly match the tuple code, I would add another argument that gets used as `length a`, `length b`, and `max (length a) (length b)`.

_CoqProject
```diff
@@ -3,6 +3,7 @@
 Bedrock/Nomega.v
 Bedrock/Word.v
 src/Demo.v
+src/DemoWithReification.v
```
andres-erbsen: Perhaps src/Experiments/SimplyTypedArithmetic ?

src/DemoWithReification.v
JasonGross: Done

src/DemoWithReification.v
JasonGross: Done

src/DemoWithReification.v
JasonGross: Done.  Turns out that precondition isn't needed to prove `eval_to_associational`.  Should I insert it anyway?  (And `eval_add_to_nth` could take a weaker hypothesis, namely `i < min n (length xs)` rather than `i < length xs` and `n = length xs`; which hypotheses should I use?

_CoqProject
JasonGross: Done

JasonGross:
Quoting https://github.com/mit-plv/fiat-crypto/pull/271#discussion_r152720364 here so it's not collapsed:
@andres-erbsen said:
> Add a `nat` argument for the length, require it to be equal to `length xs` as a precondition to the correctness proof. (I am not sure about this, but it seems like a direct translation from tuple code).

@JasonGross said:
> Done.  Turns out that precondition isn't needed to prove `eval_to_associational`.  Should I insert it anyway?  (And `eval_add_to_nth` could take a weaker hypothesis, namely `i < min n (length xs)` rather than `i < length xs` and `n = length xs`; which hypotheses should I use?

andres-erbsen:
Looks good to me as-is, still.

PR #272, opened 2017-11-23
[Demo] Define [place] in terms of nat_rect

JasonGross:
This will make it easier to reify

This is on top of #271

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -125,14 +125,21 @@ Module Positional. Section Positional.
            end; lia.                                          Qed.
   Hint Rewrite @eval_add_to_nth eval_zeros : push_eval.
 
-  Fixpoint place (t:Z*Z) (i:nat) : nat * Z :=
-    if dec (fst t mod weight i = 0)
-    then (i, let c := fst t / weight i in (c * snd t)%RT)
-    else match i with S i' => place t i' | O => (O, fst t * snd t)%RT end.
+  Definition place (t:Z*Z) (i:nat) : nat * Z :=
+    nat_rect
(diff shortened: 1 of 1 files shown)


src/DemoWithReification.v
```diff
@@ -0,0 +1,221 @@
...
+      (fun _ => (nat * Z)%type)
+      ((O, fst t * snd t)%RT)
+      (fun i' place_i'
+       => let i := S i' in
+          if (fst t mod weight i =? 0)
+          then (i, let c := fst t / weight i in (c * snd t)%RT)
+          else place_i')
+      i.
+
+  Fixpoint place' (t:Z*Z) (i:nat) : nat * Z :=
```
andres-erbsen: Dead code?

src/DemoWithReification.v
```diff
@@ -0,0 +1,221 @@
...
+      (weight 0, 0)) by (push; lia); cbv [ListUtil.splice_nth id].
+    repeat match goal with
+           | _ => progress push
+           | _ => progress break_match
+           | _ => progress (apply Zminus_eq; ring_simplify)
+           | _ => rewrite <-ListUtil.map_nth_default_always
+           end; lia.                                          Qed.
+  Hint Rewrite @eval_add_to_nth eval_zeros : push_eval.
+
+  Definition place (t:Z*Z) (i:nat) : nat * Z :=
```
andres-erbsen: I don't have alternative suggestions but I am not attached to the current form either. I feel that it is not elegant but might not be worth improving on right now.

src/DemoWithReification.v
JasonGross: Yes, now removed, thanks.  (I originally had it to make sure I could prove that the new `place` was pointwise equal to the old `place`.)

PR #273, opened 2017-11-24
[demo] Replace (max (length a) (length b)) with explicit nat arg

JasonGross:
Oops, forgot this in #271.  Planning to merge this soon.


JasonGross:
Build failure seems like a transient instance of coqdep taking too long...

PR #275, opened 2017-11-24
[demo] Add reification in src/Experiments/SimplyTypedArithmetic.v

JasonGross:
It's rather verbose, unfortunately.  The reification also doesn't have
any of the nice debugging features of the version of reification in
Compilers, because that's even more boilerplate.  Not sure if I should
add that back in, at the moment.

Also, for some strange reason, places where `constr`s fail to typecheck
seem to induce backtracking where I don't think they should, and I'm not
sure what's going on...

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -180,57 +180,158 @@ Module Positional. Section Positional.
 End Positional. End Positional.
 
 Module Compilers.
-  Inductive type := Unit | Prod (A B : type) | Arrow (s d : type) | List (A : type) | TNat | TZ | TBool.
-  Delimit Scope ctype_scope with ctype.
-  Bind Scope ctype_scope with type.
-  Notation "()" := Unit : ctype_scope.
-  Notation "A * B" := (Prod A B) : ctype_scope.
-  Notation "A -> B" := (Arrow A B) : ctype_scope.
-  Fixpoint interp_type (t : type)
-    := match t with
-       | Unit => unit
-       | Prod A B => interp_type A * interp_type B
-       | Arrow A B => interp_type A -> interp_type B
-       | List A => list (interp_type A)
-       | TNat => nat
-       | TZ => Z
-       | TBool => bool
-       end%type.
+  Module type.
+    Inductive type := unit | prod (A B : type) | arrow (s d : type) | list (A : type) | nat | Z | bool.
 
-  Inductive op : type -> type -> Set :=
-  | OpConst {t} (v : interp_type t) : op Unit t
-  | NatS : op TNat TNat
-  | Nil {t} : op Unit (List t)
-  | Cons {t} : op (t * List t) (List t)
-  | Fst {A B} : op (A * B) A
-  | Snd {A B} : op (A * B) B
-  | BoolRect {T} : op (T * T * TBool) T
-  | NatRect {P} : op (P * (TNat -> P -> P) * TNat) P
-  | Seq : op (TNat * TNat) (List TNat)
-  | Repeat {A} : op (A * TNat) (List A)
-  | LetIn {tx tC} : op (tx * (tx -> tC)) tC
-  | Combine {A B} : op (List A * List B) (List (A * B))
-  | Map {A B} : op ((A -> B) * List A) (List B)
-  | FlatMap {A B} : op ((A -> List B) * List A) (List B)
-  | Partition {A} : op ((A -> TBool) * List A) (List A * List A)
-  | ListApp {A} : op (List A * List A) (List A)
-  | FoldRight {A B} : op ((B -> A -> A) * A * List B) A
-  | Pred : op TNat TNat
-  | UpdateNth {T} : op (TNat * (T -> T) * List T) (List T)
-  | RuntimeMul : op (TZ * TZ) TZ
-  | RuntimeAdd : op (TZ * TZ) TZ
-  | Zadd : op (TZ * TZ) TZ
-  | Zmul : op (TZ * TZ) TZ
-  | Zpow : op (TZ * TZ) TZ
-  | Zopp : op TZ TZ
-  | Zdiv : op (TZ * TZ) TZ
-  | Zmodulo : op (TZ * TZ) TZ
-  | Zeqb : op (TZ * TZ) TBool
-  | ZofNat : op TNat TZ
-  | App {s d} : op ((s -> d) * s) d.
+    Fixpoint interp (t : type)
+      := match t with
+         | unit => Datatypes.unit
+         | prod A B => interp A * interp B
+         | arrow A B => interp A -> interp B
+         | list A => Datatypes.list (interp A)
+         | nat => Datatypes.nat
+         | Z => BinInt.Z
+         | bool => Datatypes.bool
+         end%type.
+
+    Ltac reify ty :=
+      lazymatch eval cbv beta in ty with
+      | Datatypes.unit => unit
+      | Datatypes.prod ?A ?B
+        => let rA := reify A in
+           let rB := reify B in
+           constr:(prod rA rB)
+      | ?A -> ?B
+        => let rA := reify A in
+           let rB := reify B in
+           constr:(arrow rA rB)
+      | Datatypes.list ?T
+        => let rT := reify T in
+           constr:(list rT)
+      | Datatypes.nat => nat
+      | Datatypes.bool => bool
+      | BinInt.Z => Z
+      end.
+
+    Module Export Notations.
+      Delimit Scope ctype_scope with ctype.
+      Bind Scope ctype_scope with type.
+      Notation "()" := unit : ctype_scope.
+      Notation "A * B" := (prod A B) : ctype_scope.
+      Notation "A -> B" := (arrow A B) : ctype_scope.
+      Notation type := type.
+    End Notations.
+  End type.
+  Export type.Notations.
+
+  Module op.
+    Import type.
+    Inductive op : type -> type -> Set :=
+    | Const {t} (v : interp t) : op unit t
+    | Let_In {tx tC} : op (tx * (tx -> tC)) tC
+    | App {s d} : op ((s -> d) * s) d
+    | S : op nat nat
+    | nil {t} : op unit (list t)
+    | cons {t} : op (t * list t) (list t)
+    | fst {A B} : op (A * B) A
+    | snd {A B} : op (A * B) B
+    | bool_rect {T} : op (T * T * bool) T
+    | nat_rect {P} : op (P * (nat -> P -> P) * nat) P
+    | pred : op nat nat
+    | List_seq : op (nat * nat) (list nat)
+    | List_repeat {A} : op (A * nat) (list A)
+    | List_combine {A B} : op (list A * list B) (list (A * B))
+    | List_map {A B} : op ((A -> B) * list A) (list B)
+    | List_flat_map {A B} : op ((A -> list B) * list A) (list B)
+    | List_partition {A} : op ((A -> bool) * list A) (list A * list A)
+    | List_app {A} : op (list A * list A) (list A)
+    | List_fold_right {A B} : op ((B -> A -> A) * A * list B) A
+    | List_update_nth {T} : op (nat * (T -> T) * list T) (list T)
+    | Z_runtime_mul : op (Z * Z) Z
+    | Z_runtime_add : op (Z * Z) Z
+    | Z_add : op (Z * Z) Z
+    | Z_mul : op (Z * Z) Z
+    | Z_pow : op (Z * Z) Z
+    | Z_opp : op Z Z
+    | Z_div : op (Z * Z) Z
+    | Z_modulo : op (Z * Z) Z
+    | Z_eqb : op (Z * Z) bool
+    | Z_of_nat : op nat Z.
+
+    Notation curry2 f
+      := (fun '(a, b) => f a b).
+    Notation curry3 f
+      := (fun '(a, b, c) => f a b c).
+
+    Definition interp {s d} (opc : op s d) : type.interp s -> type.interp d
+      := match opc in op s d return type.interp s -> type.interp d with
+         | Const t v => fun _ => v
+         | Let_In tx tC => curry2 (@LetIn.Let_In (type.interp tx) (fun _ => type.interp tC))
+         | App s d
+           => fun '((f, x) : (type.interp s -> type.interp d) * type.interp s)
+              => f x
+         | S => Datatypes.S
+         | nil t => fun _ => @Datatypes.nil (type.interp t)
+         | cons t => curry2 (@Datatypes.cons (type.interp t))
+         | fst A B => @Datatypes.fst (type.interp A) (type.interp B)
+         | snd A B => @Datatypes.snd (type.interp A) (type.interp B)
+         | bool_rect T => curry3 (@Datatypes.bool_rect (fun _ => type.interp T))
+         | nat_rect P => curry3 (@Datatypes.nat_rect (fun _ => type.interp P))
+         | pred => Nat.pred
+         | List_seq => curry2 List.seq
+         | List_combine A B => curry2 (@List.combine (type.interp A) (type.interp B))
+         | List_map A B => curry2 (@List.map (type.interp A) (type.interp B))
+         | List_repeat A => curry2 (@List.repeat (type.interp A))
+         | List_flat_map A B => curry2 (@List.flat_map (type.interp A) (type.interp B))
+         | List_partition A => curry2 (@List.partition (type.interp A))
+         | List_app A => curry2 (@List.app (type.interp A))
+         | List_fold_right A B => curry3 (@List.fold_right (type.interp A) (type.interp B))
+         | List_update_nth T => curry3 (@update_nth (type.interp T))
+         | Z_runtime_mul => curry2 runtime_mul
+         | Z_runtime_add => curry2 runtime_add
+         | Z_add => curry2 Z.add
+         | Z_mul => curry2 Z.mul
+         | Z_pow => curry2 Z.pow
+         | Z_modulo => curry2 Z.modulo
+         | Z_opp => Z.opp
+         | Z_div => curry2 Z.div
+         | Z_eqb => curry2 Z.eqb
+         | Z_of_nat => Z.of_nat
+         end.
+
+    Module List.
+      Notation seq := List_seq.
+      Notation repeat := List_repeat.
+      Notation combine := List_combine.
+      Notation map := List_map.
+      Notation flat_map := List_flat_map.
+      Notation partition := List_partition.
+      Notation app := List_app.
+      Notation fold_right := List_fold_right.
+      Notation update_nth := List_update_nth.
+    End List.
+
+    Module Z.
+      Notation runtime_mul := Z_runtime_mul.
+      Notation runtime_add := Z_runtime_add.
+      Notation add := Z_add.
+      Notation mul := Z_mul.
+      Notation pow := Z_pow.
+      Notation opp := Z_opp.
+      Notation div := Z_div.
+      Notation modulo := Z_modulo.
+      Notation eqb := Z_eqb.
+      Notation of_nat := Z_of_nat.
+    End Z.
+
+    Module Export Notations.
+      Notation op := op.
+    End Notations.
+  End op.
+  Export op.Notations.
 
   Inductive expr {var : type -> Type} : type -> Type :=
-  | TT : expr Unit
+  | TT : expr ()
   | Pair {A B} (a : expr A) (b : expr B) : expr (A * B)
   | Var {t} (v : var t) : expr t
   | Op {s d} (opc : op s d) (args : expr s) : expr d
@@ -242,80 +343,21 @@ Module Compilers.
   Notation "( x , y , .. , z )" := (Pair .. (Pair x%expr y%expr) .. z%expr) : expr_scope.
   Notation "( )" := TT : expr_scope.
   Notation "()" := TT : expr_scope.
-  Notation "'llet' x := A 'in' b" := (Op LetIn (Pair A%expr (Abs (fun x => b%expr)))) : expr_scope.
-  Notation "f ‘’ x" := (Op App (f, x)%expr) (at level 200) : expr_scope.
+  Notation "'llet' x := A 'in' b" := (Op op.Let_In (Pair A%expr (Abs (fun x => b%expr)))) : expr_scope.
+  Notation "f ‘’ x" := (Op op.App (f, x)%expr) (at level 200) : expr_scope.
 
   Definition Expr t := forall var, @expr var t.
 
-  Notation curry2 f
-    := (fun '(a, b) => f a b).
-  Notation curry3 f
-    := (fun '(a, b, c) => f a b c).
-
-  Definition interp_op {s d} (opc : op s d) : interp_type s -> interp_type d
-    := match opc in op s d return interp_type s -> interp_type d with
-       | OpConst t v => fun _ => v
-       | NatS => S
-       | Nil t => fun _ => @nil (interp_type t)
-       | Cons t => curry2 (@cons (interp_type t))
-       | BoolRect T => curry3 (@bool_rect (fun _ => interp_type T))
-       | Seq => curry2 seq
-       | LetIn tx tC => curry2 (@Let_In (interp_type tx) (fun _ => interp_type tC))
-       | Combine A B => curry2 (@combine (interp_type A) (interp_type B))
-       | Map A B => curry2 (@List.map (interp_type A) (interp_type B))
-       | Repeat A => curry2 (@List.repeat (interp_type A))
-       | FlatMap A B => curry2 (@List.flat_map (interp_type A) (interp_type B))
-       | Fst A B => @fst (interp_type A) (interp_type B)
-       | Snd A B => @snd (interp_type A) (interp_type B)
-       | Partition A => curry2 (@List.partition (interp_type A))
-       | ListApp A => curry2 (@List.app (interp_type A))
-       | FoldRight A B => curry3 (@List.fold_right (interp_type A) (interp_type B))
-       | NatRect P => curry3 (@nat_rect (fun _ => interp_type P))
-       | Pred => Nat.pred
-       | UpdateNth T => curry3 (@update_nth (interp_type T))
-       | RuntimeMul => curry2 runtime_mul
-       | RuntimeAdd => curry2 runtime_add
-       | Zadd => curry2 Z.add
-       | Zmul => curry2 Z.mul
-       | Zpow => curry2 Z.pow
-       | Zmodulo => curry2 Z.modulo
-       | Zopp => Z.opp
-       | Zdiv => curry2 Z.div
-       | Zeqb => curry2 Z.eqb
-       | ZofNat => Z.of_nat
-       | App s d => curry2 (fun (f : interp_type s -> interp_type d) (x : interp_type s)
-                            => f x)
-       end.
-
-  Fixpoint interp {t} (e : @expr interp_type t) : interp_type t
+  Fixpoint interp {t} (e : @expr type.interp t) : type.interp t
     := match e with
        | TT => tt
        | Pair A B a b => (interp a, interp b)
        | Var t v => v
-       | Op s d opc args => interp_op opc (interp args)
+       | Op s d opc args => op.interp opc (interp args)
        | Abs s d f => fun v => interp (f v)
        end.
 
   Definition Interp {t} (e : Expr t) := interp (e _).
-
-  Ltac reify_type term :=
-    lazymatch eval cbv beta in term with
-    | unit => Unit
-    | prod ?A ?B
-      => let rA := reify_type A in
-         let rB := reify_type B in
-         constr:(Prod rA rB)
-    | ?A -> ?B
-      => let rA := reify_type A in
-         let rB := reify_type B in
-         constr:(Arrow rA rB)
-    | list ?T
-      => let rT := reify_type T in
-         constr:(List rT)
-    | nat => TNat
-    | bool => TBool
-    | Z => TZ
-    end.
 
   Ltac is_known_const_gen term on_success on_failure :=
     let is_known_const term := is_known_const_gen term on_success on_failure in
@@ -335,7 +377,7 @@ Module Compilers.
   Ltac is_known_const_as_bool term :=
     is_known_const_gen term ltac:(fun _ => true) ltac:(fun _ => false).
 
-  Definition Uncurry0 {A var} (opc : op Unit A) : @expr var A
+  Definition Uncurry0 {A var} (opc : op type.unit A) : @expr var A
     := Op opc TT.
   Definition Uncurry1 {A B var} (opc : op A B) : @expr var (A -> B)
     := λ a, Op opc (Var a).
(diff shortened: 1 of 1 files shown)


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+    | xO ?p => is_known_const p
+    | xH => on_success ()
+    | ?term => on_failure term
+    end.
+  Ltac is_known_const term :=
+    is_known_const_gen term ltac:(fun _ => idtac) ltac:(fun term => fail 0 "Not a known const:" term).
+  Ltac is_known_const_as_bool term :=
+    is_known_const_gen term ltac:(fun _ => true) ltac:(fun _ => false).
+
+  Definition Uncurry0 {A var} (opc : op Unit A) : @expr var A
```
andres-erbsen: Confused. I get there is some idea that the AST-s pass arguments as pairs, and the code being reified passes by currying, but I don't understand why `Uncyrry0` and similar are necessary or what they do.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+    | _
+      => let assert_const := match goal with
+                             | _ => is_known_const term
+                             end in
+         let T := type of term in
+         let rT := reify_type T in
+         Uncurry0 (@OpConst rT term)
+    end.
+
+  Inductive context_var_map {var : type -> Type} :=
```
andres-erbsen: Just curious: how does the amount of typing precision in this definition affect reification performance? E.g. if all variales were {T:Type} (var:T), would that change things? What if there was one tactic context entry for each binding, instead of a list of bindings in one Coq context variable?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+  | cnil
+  | ccons {t} (gallina_v : interp_type t) (v : var t) (ctx : context_var_map).
+
+  Ltac refresh n :=
```
andres-erbsen: COQBUG(https://...) ?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+    let n' := fresh n in
+    let n' := fresh n' in
+    let n' := fresh n' in
+    n'.
+
+  Ltac is_type_arg arg_ty :=
```
andres-erbsen: I am not sure what this function is for, and perhaps it could be factored better.
- why `arg` in  tactic name?
- why `match` `eval cbv beta` instead of `unify`?
- why not `_ -> _ => true`?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+    | _ => false
+    end.
+
+  Ltac build_dummy_delayed_arguments_for ty :=
+    lazymatch ty with
+    | forall x : ?T, ?P
+      => let not_x := refresh x in
+         let rest := constr:(
+                       fun x : T
+                       => match P with
```
andres-erbsen: Seems like a redundant match. Why? If COQBUG, please indicate.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+                          | not_x
+                            => ltac:(
+                                 let P := (eval cbv delta [not_x] in not_x) in
+                                 let v := build_dummy_delayed_arguments_for P in
+                                 exact v
+                               )
+                          end) in
+         lazymatch rest with
+         | (fun _ => ?rest)
+           => constr:((tt, rest))
```
andres-erbsen: Does this function just return a `(tt, (tt, (tt, ...)))` according to the arity of the input? If so, why is it useful?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+         end
+    | _ => tt
+    end.
+  Ltac check_delayed_arguments_dummy delayed_arguments if_dummy if_not_dummy :=
+    lazymatch delayed_arguments with
+    | tt => if_dummy ()
+    | (tt, ?delayed_arguments) => check_delayed_arguments_dummy delayed_arguments if_dummy if_not_dummy
+    | (_, _) => if_not_dummy ()
+    end.
+  Ltac plug_delayed_arguments f delayed_arguments :=
```
andres-erbsen: Okay, I now understand here that we are doing substituting in some arguments of a function and not others. However, it is not clear how the bindings being substitutied are encoded. Part of this might be me having trouble reading tactics-in-terms.

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: Again, how about namespacing the constructors? I actually had to look them up when I encountered them below.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+          => let T := type of t in
+             reify_rec (@bool_rect (fun _ => T) t f b)
+        | let x := ?a in @?b x
+          => let A := type of a in
+             let B := lazymatch type of b with forall x, @?B x => B end in
+             reify_rec (@Let_In A B a b)
+        | ?f ?x
+          =>
+          let x_ty := type of x in
+          let x_type_arg := is_type_arg x_ty in
```
andres-erbsen: `is_type_arg` -> `is_type` or `is_sort` or something like that?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+          lazymatch x_type_arg with
+          | true
+            => (* we can't reify things of type [Type], so we save it for later to plug in *)
+            reify_helper var f ctx (x, delayed_arguments)
```
andres-erbsen: Is `delayed_arguments` a list of substitutions to be made during reification?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
   End mulmod.
 End Positional. End Positional.
 
-Import Associational Positional.
+Module Compilers.
+  Inductive type := Unit | Prod (A B : type) | Arrow (s d : type) | List (A : type) | TNat | TZ | TBool.
+  Delimit Scope ctype_scope with ctype.
```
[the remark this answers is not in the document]
JasonGross: `c` stands for `compilers` or `compiler` or `compiled`.  The issue is that `type_scope` is already taken by Coq, and scopes aren't namespaced.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+  Bind Scope ctype_scope with type.
+  Notation "()" := Unit : ctype_scope.
```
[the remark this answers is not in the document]
JasonGross: No, but I can fix that.  Shall I?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
   End mulmod.
 End Positional. End Positional.
 
-Import Associational Positional.
+Module Compilers.
+  Inductive type := Unit | Prod (A B : type) | Arrow (s d : type) | List (A : type) | TNat | TZ | TBool.
```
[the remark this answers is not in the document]
JasonGross: Sure, seems fine.  I'll do that.  Should I also call it `type.interp` rather than `interp_type`?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+  | Abs {s d} (f : var s -> expr d) : expr (s -> d).
+
+  Bind Scope expr_scope with expr.
+  Delimit Scope expr_scope with expr.
+  Notation "'λ'  x .. y , t" := (Abs (fun x => .. (Abs (fun y => t%expr)) ..)) : expr_scope.
+  Notation "( x , y , .. , z )" := (Pair .. (Pair x%expr y%expr) .. z%expr) : expr_scope.
+  Notation "( )" := TT : expr_scope.
+  Notation "()" := TT : expr_scope.
+  Notation "'llet' x := A 'in' b" := (Op LetIn (Pair A%expr (Abs (fun x => b%expr)))) : expr_scope.
+  Notation "f ‘’ x" := (Op App (f, x)%expr) (at level 200) : expr_scope.
```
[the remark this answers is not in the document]
JasonGross: No, but it makes display a lot more readable for nested applications.  Suggestions welcome.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+  | Zmodulo : op (TZ * TZ) TZ
+  | Zeqb : op (TZ * TZ) TBool
+  | ZofNat : op TNat TZ
+  | App {s d} : op ((s -> d) * s) d.
+
+  Inductive expr {var : type -> Type} : type -> Type :=
+  | TT : expr Unit
+  | Pair {A B} (a : expr A) (b : expr B) : expr (A * B)
+  | Var {t} (v : var t) : expr t
+  | Op {s d} (opc : op s d) (args : expr s) : expr d
```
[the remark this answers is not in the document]
JasonGross: There's actually a couple of design questions buried here, which I'd love your thoughts on:
1. Which functions should be required to be fully applied, if any?
2. Do we represent functions as curried or uncurried?

In particular, when writing the partial evaluator, there's a natural abstraction barrier I want to have between types that are generic and types where I need constructors to reduce to the term.  It might make sense to leave everything as curried, draw the line at lambdas, and forbid higher-order type-instantiations (e.g., can't use `nat_rect` to compute a function, can't have a list of functions).  I'll need to patch things up where this is wrong, i.e., I'll have to pay a small cost at every node of the partial evaluator, except function application.  Alternatively, I could add a marker for "should be constructor" at the type level, and pay the cost everywhere except the partial evaluator.  I could also, and maybe this is the best option, define an additional inductive of reduction-annotations, which mark which types can be treated as opaque and which need constructors.

I'm presently leaning towards this last option, and replacing `op` with `ident`, and treating functions as curried.  What do you think?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: This is where I pay the cost for the fact that I'm representing operations as uncurried and representing terms as curried (and Gallina represents terms as curried).  It's necessary because, rather than collecting all the arguments in reification, I need to `App` them one at a time, and then, when I get to the head constant, I need make up for the fact that, actually, I need them all at once.  If I use `ident` rather than `op` as mentioned above, then this goes away.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: > if all variales were {T:Type} (var:T), would that change things?

No idea.  The biggest cause of slowness in reification here was the fact that it backtracked on error, and so I was seeing infinite loops rather than fast failure.  Once I removed that, reification was reasonable, and not slow enough, I'd guess, for me to get sensible timing data.

> What if there was one tactic context entry for each binding

What do you mean by that? (What's an example?)

src/Experiments/SimplyTypedArithmetic.v
JasonGross: https://github.com/coq/coq/issues/5448

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Reification has the wart that some Gallina applications do not become application nodes in the PHOAS syntax tree.  This tactic identifies which applications those are.  In particular, the application of a function to a type (something of type `Type`, `Set`, or `Prop`) or to a function which generates such a thing (e.g., the first argument to `nat_rect`), get applied in Gallina rather than in PHOAS.
- `arg` because we are looking at the argument of a function application
- `unify` doesn't take open terms, so I need `match` and `cbv beta` for the arrow anyway, and I might as well use it for `Type`, `Prop`, and `Set`
- `_ -> _ => true` would catch the function argument to `List.map`, which would be wrong

Given this additional context, do you have more suggestions?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: I use it in more places than we actually need it, but I figure better safe than sorry.  (We don't need it for the branch of the single argument in a match, but I believe we do need it for any additional variables we add to the binders.)

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Newly reported https://github.com/coq/coq/issues/6243

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Yes.  It's useful because some other code relies on the "delayed arguments" list exactly matching the shape of the type.  We could instead add some extra cases to deal separately with `tt` as dummy, and `(tt, _)` as dummy, but I decided to pay the cost in this one place, rather than in all the places where I use it.

The problem I'm solving is that some applications need to stay as Gallina applications, and not turn into PHOAS application nodes.  I could throw an error if not all such applications came at the head of the term (i.e., if there was anything of type `nat -> forall T : Type, ....`), but I currently support such cases.  So I need to save the arguments that need to be plugged in as Gallina applications, and mark the ones that become PHOAS applications.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: The encoding is that there is a tuple of arguments.  `tt` means "this becomes a PHOAS application node", anything else means "this should be substituted in Gallina".  (This is safe, because only things of type `Type`, `Set`, `Prop`, or `_ -> _ -> ... _ -> {Type,Set,Prop}` get substituted.)  Terms are eta-expanded as necessary to fill in arguments, but, importantly, are not eta-expanded if all remaining arguments are dummy (`tt`).  Perhaps there is a cleaner way to do all of this.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Yeah, basically

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Yes, with `tt` marking places where substitutions are reified as applications rather than made

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+  | OpConst {t} (v : interp_type t) : op Unit t
+  | NatS : op TNat TNat
+  | Nil {t} : op Unit (List t)
+  | Cons {t} : op (t * List t) (List t)
+  | Fst {A B} : op (A * B) A
+  | Snd {A B} : op (A * B) B
+  | BoolRect {T} : op (T * T * TBool) T
+  | NatRect {P} : op (P * (TNat -> P -> P) * TNat) P
+  | Seq : op (TNat * TNat) (List TNat)
+  | Repeat {A} : op (A * TNat) (List A)
```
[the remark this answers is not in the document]
JasonGross: Renamed to `List_repeat`, added some namespacing

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+  | Op {s d} (opc : op s d) (args : expr s) : expr d
+  | Abs {s d} (f : var s -> expr d) : expr (s -> d).
+
+  Bind Scope expr_scope with expr.
+  Delimit Scope expr_scope with expr.
+  Notation "'λ'  x .. y , t" := (Abs (fun x => .. (Abs (fun y => t%expr)) ..)) : expr_scope.
+  Notation "( x , y , .. , z )" := (Pair .. (Pair x%expr y%expr) .. z%expr) : expr_scope.
+  Notation "( )" := TT : expr_scope.
+  Notation "()" := TT : expr_scope.
+  Notation "'llet' x := A 'in' b" := (Op LetIn (Pair A%expr (Abs (fun x => b%expr)))) : expr_scope.
```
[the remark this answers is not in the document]
JasonGross: Updated to `expr_let`

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+  Notation "f ‘’ x" := (Op App (f, x)%expr) (at level 200) : expr_scope.
+
+  Definition Expr t := forall var, @expr var t.
+
+  Notation curry2 f
+    := (fun '(a, b) => f a b).
+  Notation curry3 f
+    := (fun '(a, b, c) => f a b c).
+
+  Definition interp_op {s d} (opc : op s d) : interp_type s -> interp_type d
```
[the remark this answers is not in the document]
JasonGross: Yes.  I've moved it up.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+       | RuntimeAdd => curry2 runtime_add
+       | Zadd => curry2 Z.add
+       | Zmul => curry2 Z.mul
+       | Zpow => curry2 Z.pow
+       | Zmodulo => curry2 Z.modulo
+       | Zopp => Z.opp
+       | Zdiv => curry2 Z.div
+       | Zeqb => curry2 Z.eqb
+       | ZofNat => Z.of_nat
+       | App s d => curry2 (fun (f : interp_type s -> interp_type d) (x : interp_type s)
```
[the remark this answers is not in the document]
JasonGross: Done.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+         constr:(Arrow rA rB)
+    | list ?T
+      => let rT := reify_type T in
+         constr:(List rT)
+    | nat => TNat
+    | bool => TBool
+    | Z => TZ
+    end.
+
+  Ltac is_known_const_gen term on_success on_failure :=
```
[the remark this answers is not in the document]

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+    let is_known_const term := is_known_const_gen term on_success on_failure in
```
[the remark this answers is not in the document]

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,435 @@ Module Positional. Section Positional.
...
+    | Zneg ?p => is_known_const p
+    | xI ?p => is_known_const p
+    | xO ?p => is_known_const p
+    | xH => on_success ()
+    | ?term => on_failure term
+    end.
+  Ltac is_known_const term :=
+    is_known_const_gen term ltac:(fun _ => idtac) ltac:(fun term => fail 0 "Not a known const:" term).
+  Ltac is_known_const_as_bool term :=
+    is_known_const_gen term ltac:(fun _ => true) ltac:(fun _ => false).
```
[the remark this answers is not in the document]

src/Experiments/SimplyTypedArithmetic.v

JasonGross:
Eh, responding to the scribbles was not hard.  I appreciate them.  I think you did a double-take at all of the places where my abstraction barriers are leaky, painful, or unclean.  There are still some design questions to be answered (around currying of functions and around what the right interface is for delayed arguments), so I'll wait on your feedback.  (Note also that I made delayed arguments more general than it needs to be for me to use it here.  Perhaps I should just assume/enforce the invariant that all type arguments come first, and that will simplify code a decent amount?)

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: Why do we need to do beta-reduction here at all?

I would rename to `is_template_parameter parameter_type` (even though we call it on argument types).

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: Do we actually support these, though? I don't know what would be passed to `reify_op` for something like this.

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: Why do we pass arguments in tuples anyway? Why does that transformation need to happen during reification?

JasonGross: There are two cases:
1. Somehow, reification needs to know to reify `@List.map ?A ?B`, not `@List.map`.  This could be solved by cps-ing `reify_op`, and then having reification try `reify_op` first, and only try other things if `reify_op` fails.  However, this will stop working if we want to, e.g., define `List.map` in terms of `list_rect` and remove `List.map` from reification in favor of an initial `cbv delta [List.map]`, in which case we'll have:
2. reification of `(fun (A B : Type) => ...) T U`

Because we're not doing `cbv beta`, we need to either commit to never having lambdas that take type arguments, or we need to pass types around in tuples to do controlled beta reduction ourselves, or we need to write special cases for all of the numbers and kinds of type arguments we want to handle.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: These = `nat -> forall T : Type, ...`?  Or just type arguments?  For type arguments, we need something like this for `List.map`, etc.  For things like `nat -> forall T : Type, ...`, we nominally support them.  We can reify something silly like
```coq
(fun (n : nat) (T : Type) (f : T -> nat) (x : T) => n + f x) 0 nat (fun x => x) 1
```
But we don't need to support these things.  Would it be better to drop support for them?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: When we change `let x := v in f` to `@Let_In ?A ?P v (fun x => f)` for reification, it gets typed as `P v`, i.e., `(fun _ => ...) v`.  I definitely hit this case in `type.reify`, and I can't recall if I added it here preemptively, or because I hit it here, too.

JasonGross: This function gets `((fun _ : nat => (nat * Z)%type) (Init.Nat.pred n))`, `(forall i' : nat, (fun _ : nat => (nat * Z)%type) i' -> nat * Z)`, `((fun _ : nat => (nat * Z)%type) i')`.  So removing beta will not break anything here, but it seems like a poor thing to rely on, given that some of our types are not beta-reduced.

Also, I've renamed it.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,480 @@ Module Positional. Section Positional.
...
+      Notation modulo := Z_modulo.
+      Notation eqb := Z_eqb.
+      Notation of_nat := Z_of_nat.
+    End Z.
+
+    Module Export Notations.
+      Notation op := op.
+    End Notations.
+  End op.
+  Export op.Notations.
```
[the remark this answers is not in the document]
JasonGross: Alias could be after the end of the module, but I imagine `op` may get it's own file, or something, and this is preempting that.  Is that a bad idea?

andres-erbsen:
I am still confused why we need to mark non-template arguments with tt -- they should be obvious from the function itself. Is this somehow because we do currying in the same pass?

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: I agree that we do need to implement beta reduction of type arguments.

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: Oh, so this is about being insensitive to ordering of arguments / section variables. That seems like a good property.

andres-erbsen:
I think I am now convinced that the reification should indeed be doing what it is doing right now. However, it seems sketchy that we need to insert `tt` in the list of type arguments to keep things in sync -- I am guessing it is duplicating some other information already available in the context where it is used, and the code would be nicer without it. E.g. we could just keep an ordered list of template arguments, using them on demand.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,480 @@ Module Positional. Section Positional.
...
+        | match ?b with true => ?t | false => ?f end
+          => let T := type of t in
+             reify_rec (@bool_rect (fun _ => T) t f b)
+        | let x := ?a in @?b x
+          => let A := type of a in
+             let B := lazymatch type of b with forall x, @?B x => B end in
+             reify_rec (@Let_In A B a b)
+        | ?f ?x
+          =>
+          let x_ty := type of x in
```
andres-erbsen: actually, we could just inspect the type of `f` here instead of x to make parameter/argument naming less confusing.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +179,480 @@ Module Positional. Section Positional.
...
+            let dummy_args := build_dummy_delayed_arguments_for x_ty in
+            let rx := reify_helper var x ctx dummy_args in
+            (* in rf's delayed_arguments, tt stands for "dummy" *)
+            let rf := reify_helper var f ctx (tt, delayed_arguments) in
+            constr:(Op (var:=var) op.App (Pair (var:=var) rf rx))
+          end
+        | (fun x : ?T => ?f)
+          =>
+          (*let dummy := match goal with _ => idtac "funcase delaying" delayed_arguments end in*)
+          lazymatch delayed_arguments with
```
andres-erbsen: Could we branch on `T` here instead, only consuming the arguments list when `T` is a template parameter?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Ah, that seems wise, I'll do that.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Yes, though then I should change `cbv beta` to something stronger like `hnf`, because we need to avoid the case where the type of the binder unifies with the type of the argument, but is syntactically different in a way that causes us to delay the argument and then miss it here, or vice versa (e.g., if we ever do `Definition Type1 := Type.`).

andres-erbsen: Could we change to consistently inspecting the function binder type, not the argument type?

JasonGross: Sure, but that's not enough for all cases (e.g., `((fun (T : Type) (x : T) => x) : forall T : Type1, T -> T) nat 0)`).  I think I have a decent way of switching to using `unify`, though.

andres-erbsen: Okay, it really does seem that alias unification (in the same sense as it came up for section records) is what we want here. I think I might be okay with just not supporting aliases for `Type` and `Set`, though. Also happy to look at your unify solution.

src/Experiments/SimplyTypedArithmetic.v

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Changed

JasonGross:
> E.g. we could just keep an ordered list of template arguments, using them on demand.

Updated to do that

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Unify solution presently at https://github.com/mit-plv/fiat-crypto/pull/275/files#diff-d6d787cba38e2097ad545be3240b32feR484

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +180,460 @@ Module Positional. Section Positional.
...
+    run_tactic_as_bool ltac:(fun _ => require_template_parameter parameter_type).
+  Ltac get_first_argument_type f :=
+    let f_ty := type of f in
+    lazymatch eval hnf in f_ty with
+    | forall x : ?T, _ => T
+    end.
+
+  Ltac plug_delayed_arguments f delayed_arguments :=
+    lazymatch delayed_arguments with
+    | tt => f
```
andres-erbsen: Is tt acting as a list nil here?

JasonGross: yes

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +180,460 @@ Module Positional. Section Positional.
...
+            lazymatch rf0 with
+            | (fun _ => ?rf)
+              => constr:(@Abs var rT _ rf)
+            | _ => let dummy := match goal with
+                                | _ => fail 1 "Failure to eliminate functional dependencies of" rf0
+                                end in
+                   constr:(I : I)
+            end
+          end
+        | _
```
andres-erbsen: Please add a comment explaining why this needs to be the default case.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +180,460 @@ Module Positional. Section Positional.
...
+                       => ltac:(
+                            let f := (eval cbv delta [not_x2] in not_x2) in
+                            (*idtac "rec call" f "was" term;*)
+                            let rf := reify_helper var f (@var_context.cons var rT x not_x ctx) delayed_arguments in
+                            exact rf)
+                     end) in
+            lazymatch rf0 with
+            | (fun _ => ?rf)
+              => constr:(@Abs var rT _ rf)
+            | _ => let dummy := match goal with
```
andres-erbsen: Please add a comment explaining when this case might happen.

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: Please document the `delayed_arguments` substitution context representation in a comment.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +180,460 @@ Module Positional. Section Positional.
...
+      | true
+        => plug_delayed_arguments (f arg) delayed_arguments'
+      | false
+        => constr:(fun x : T
+                   => ltac:(let v := plug_delayed_arguments (f x) delayed_arguments in
+                            exact v))
+      end
+    end.
+
+  Ltac reify_helper var term ctx delayed_arguments :=
```
andres-erbsen: Proposed rename: `value_ctx` and `template_ctx`

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +180,460 @@ Module Positional. Section Positional.
...
+        | match ?b with true => ?t | false => ?f end
+          => let T := type of t in
+             reify_rec (@bool_rect (fun _ => T) t f b)
+        | let x := ?a in @?b x
+          => let A := type of a in
+             let B := lazymatch type of b with forall x, @?B x => B end in
+             reify_rec (@Let_In A B a b)
+        | ?f ?x
+          =>
+          let ty := get_first_argument_type f in
```
andres-erbsen: `type_of_first_argument_of`

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +180,460 @@ Module Positional. Section Positional.
...
+          | unify parameter_type Type
+          | lazymatch eval hnf in parameter_type with
+            | forall x : ?T, @?P x
+              => let check := constr:(fun x : T
+                                      => ltac:(require_template_parameter (P x);
+                                               exact I)) in
+                 idtac
+            end ].
+  Ltac is_template_parameter parameter_type :=
+    run_tactic_as_bool ltac:(fun _ => require_template_parameter parameter_type).
```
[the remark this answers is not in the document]
JasonGross: Mmm, I'm not sure about this.  The semantics are that it runs a tactic during tactic expression evaluation, and returns a constr indicating success or failure.  I use it to build a more general `constr_tryif_then_else`:
https://github.com/mit-plv/fiat-crypto/blob/6c7913c216f117d2069400d6dc81cbed407007e2/src/Util/Tactics/RunTacticAsConstr.v#L16-L23

I could use this (though it may be a bit awkward).  Or maybe `is_success_run_tactic`?

andres-erbsen:
This might be the last set of comments I have. I think at this point handling of template-arguments is essentially as systematic as the handling of runtime arguments in `ctx`, and what remains is to make that convention obvious from code and comments.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +180,460 @@ Module Positional. Section Positional.
...
+    let n' := fresh n in
+    let n' := fresh n' in
+    let n' := fresh n' in
+    n'.
+
+  Ltac require_template_parameter parameter_type :=
+    first [ unify parameter_type Prop
+          | unify parameter_type Set
+          | unify parameter_type Type
+          | lazymatch eval hnf in parameter_type with
```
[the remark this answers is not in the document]
JasonGross: `unify` doesn't work on uconstrs, so `unify parameter_type (_ -> _)` will always fail.  I could make an `open_constr` first, but that's clunky and not exactly fine-grained; I guess one way is:
```coq
let T := open_constr:(_) in
let arrow := open_constr:(_ -> T) in
unify parameter_type arrow;
require_template_parameter T
```
Is that better?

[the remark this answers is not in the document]
andres-erbsen: What is the difference beteewen uconstr and openconstr? What does the change from one to another require? Is using an arrow instead of a forall intentional?

src/Experiments/SimplyTypedArithmetic.v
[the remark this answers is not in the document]
andres-erbsen: `is_success_run_tactic` sounds okay, I also don't really care

src/Experiments/SimplyTypedArithmetic.v
JasonGross: I'm not sure what to say in the comment.  "This must come last because we're using `lazymatch`"?  Or "This must be here because we do in fact need to reify [op]s"?  Or something else?  I'm not sure what I'm explaining: that this case is present, that it's last, that it's run whenever the only cases don't match, that it's run only when the other cases don't match, or that there's no other cases to handle when none of the others match (or something else)

src/Experiments/SimplyTypedArithmetic.v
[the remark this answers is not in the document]
JasonGross: If I write `unify p (_ -> _)`, `_ -> _` gets interpreted as a ... constr, I think?  The implicit cast from uconstr to constr requires that all holes be fillable by type inference.  The "cast" from uconstr to open_constr fills uninferrable holes with evars.  (`constr` and `uconstr` roughly correspond to actual datatypes, while `open_constr` is just a description of how to fill holes, roughly)  What I am saying here is that I need the extra two `let` lines because the default `unify` syntax will try to fill the holes too early.

The change from a forall to an arrow is suboptimal, but we already can't handle any of the cases where I can imagine it coming up (things that look like recursion principles of length-indexed lists).  If we use `forall`, then we need to reintroduce tactics in terms.  I'm confused by your objection to `hnf` + `match`, though; can you add a bit more?

src/Experiments/SimplyTypedArithmetic.v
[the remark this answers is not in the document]

src/Experiments/SimplyTypedArithmetic.v

src/Experiments/SimplyTypedArithmetic.v

src/Experiments/SimplyTypedArithmetic.v

src/Experiments/SimplyTypedArithmetic.v

src/Experiments/SimplyTypedArithmetic.v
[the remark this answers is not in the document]
andres-erbsen: `hnf` is one of the things I assume to be arbitrarily slow (in case the next match fails, in particular). Unification too, but I am less angry at unification in types. Feel free to change back to hnf if you feel better about it.

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: Why can't we write an actual pattern in this case, instead of `_`? Why do we need to use a catch-all instead of something more specific.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +180,467 @@ Module Positional. Section Positional.
...
+    let n' := fresh n' in
+    n'.
+
+  Ltac type_of_first_argument_of f :=
+    let f_ty := type of f in
+    lazymatch eval hnf in f_ty with
+    | forall x : ?T, _ => T
+    end.
+
+  (** Template parameters (things whose type or codomain is [Type],
```
andres-erbsen: Suggestion: Forms of abstraction in Gallina that our reflective language cannot handle get handled by specializing the code "template" to each particular application of that abstraction. In particular, type arguments (`Type`, `Set`, `Prop`) get substituted into lambdas and treated as a integral part of primitive operations (such as `@List.app T`). During reification...

src/Experiments/SimplyTypedArithmetic.v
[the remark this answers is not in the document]
JasonGross: `hnf` is differently arbitrarily slow than unification.  (Note that I am doing `hnf` in a type, here.  It'd probably even be safe to do `cbv` / `compute`.)  In particular, if `unify` is faster than `hnf` in this case, it will be because `unify` performs more beta reduction and leaves over fewer things of the form `(fun _ => <small term>) <big term>`.  In this case, I'm making use of the fact that I absolutely need to know whether or not the term is headed by `forall`, and that `hnf` stops evaluating when it hits `forall`.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Updated to
```diff
-  (** Template parameters (things whose type or codomain is [Type],
-      [Set] or [Prop]) get handled specially, because our PHOAS
-      representation does not support dependent types.  During
+  (** Forms of abstraction in Gallina that our reflective language
+      cannot handle get handled by specializing the code "template" to
+      each particular application of that abstraction. In particular,
+      type arguments (nat, Z, (λ _, nat), etc) get substituted into
+      lambdas and treated as a integral part of primitive operations
+      (such as [@List.app T], [@list_rect (λ _, nat)]).  During
```

src/Experiments/SimplyTypedArithmetic.v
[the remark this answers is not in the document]
andres-erbsen: Yeah, I think what I was saying is that we don't need to know what the head symbol is if it is not a function type constructor. Use your judgement for which command to use.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -178,22 +180,470 @@ Module Positional. Section Positional.
...
+                            (*idtac "rec call" f "was" term;*)
+                            let rf := reify_helper var f (@var_context.cons var rT x not_x value_ctx) template_ctx in
+                            exact rf)
+                     end) in
+            lazymatch rf0 with
+            | (fun _ => ?rf)
+              => constr:(@Abs var rT _ rf)
+            | _
+              => (* This will happen if the reified term still
+              mentions the non-var variable.  By chance, [cbv delta]
```
andres-erbsen: What do casts have to do with this?

andres-erbsen:
Feel free to change comments as you wish / in response to things I may have said and forgotten about, but I think it is also fair to call this done. I am actually really happy with how this turned out, including the initial version which was much more digestible than what I was afraid of at last meeting.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: If you do something really silly like try to reify `(fun x : nat => S (O : (fun _ => nat) x))`, by chance, `cbv delta` strips away the cast and has us reifying `S O` under the binder.  If instead we passed `S (O : (fun _ => nat) x)` to `reify_op`, the term we get back is `@op.Const type.nat (S (O : (fun _ => nat) x))`.  This term would trigger this error message, because it would still mention the binder `x`.

I can't think of any other way to get `x` into the reified term, because we pull everything except for constants apart to reify them, but that doesn't mean I thought of everything.

src/Experiments/SimplyTypedArithmetic.v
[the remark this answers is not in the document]
JasonGross: Sure.  I'm leaving it as `hnf`.  I think I see the algorithm you're proposing (first check if it's a `forall`, which should be fast, since you only need a very limited form of beta, then, if it is, pull out the type).  However, there isn't actually a primitive for this, and `unify` doesn't follow this strategy, I believe.

PR #277, opened 2017-11-29
Partial work on implementing partial reduction

JasonGross:
Here is the code that implements what we discussed today.  It doesn't do the right thing.  In particular, the recursive call to `partial_reduce_cps` in the `Abs` node is wrong; it should not always be passing `Var`, but instead be computing something of shape `expr -> expr` that gets handled correctly elsewhere.  I'm too deep into the code to make sense of the exact issue right now; the encoding in my head is "the var types for different pieces of machinery don't line up the right way; has something to do with cps, probably; has something to do with arrows, probably", which is unhelpful.

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -687,6 +687,51 @@ Module Compilers.
     | Z : arguments type.Z
     | bool : arguments type.bool.
 
+    Definition preinvertT (t : type) :=
+      match t with
+      | type.unit => Datatypes.unit
+      | type.prod A B => arguments A * arguments B
+      | type.arrow s d => arguments d
+      | type.list A => arguments A
+      | type.nat => Datatypes.unit
+      | type.Z => Datatypes.unit
+      | type.bool => Datatypes.unit
+      end%type.
+    Definition invertT (t : type) :=
+      option (* [None] means "generic" *) (preinvertT t).
+
+    Definition invert {t : type} (P : arguments t -> Type)
+               (generic_case : P generic)
+               (non_generic_cases
+                : forall v : preinvertT t,
+                   match t return forall v : preinvertT t, (arguments t -> Type) -> Type with
+                   | type.unit
+                     => fun v P => P unit
+                   | type.prod A B
+                     => fun '((a, b) : arguments A * arguments B) P
+                        => P (prod a b)
+                   | type.arrow s d => fun v P => P (arrow v)
+                   | type.list A => fun v P => P (list v)
+                   | type.nat => fun v P => P nat
+                   | type.Z => fun v P => P Z
+                   | type.bool => fun v P => P bool
+                   end v P)
+               (a : arguments t)
+      : P a.
+    Proof.
+      destruct a;
+        try specialize (fun a b => non_generic_cases (a, b));
+        cbn in *;
+        [ exact generic_case | apply non_generic_cases; apply tt .. ].
+    Defined.
+
+    Definition invert_arrow {s d} (a : arguments (type.arrow s d)) : arguments d
+      := @invert (type.arrow s d) (fun _ => arguments d) generic (fun ad => ad) a.
+
+    Definition invert_prod {A B} (a : arguments (type.prod A B)) : arguments A * arguments B
+      := @invert (type.prod A B) (fun _ => arguments A * arguments B)%type (generic, generic) (fun '(a, b) => (a, b)) a.
+
+
     Fixpoint ground {t : type} : arguments t
       := match t with
          | type.unit => unit
@@ -997,39 +1042,54 @@ Module Compilers.
   Section partial_reduce.
     Context {var : type -> Type}.
 
-    Fixpoint partial_reduce_cps {T} {t} (e : @expr (@expr var) t)
-      : (@expr var t -> @expr var T) -> @expr var T
-      := match e in expr t return (expr t -> expr T) -> expr T with
-         | TT => fun k => k TT
+    Local Notation partial_reduceT t a
+      := ((@expr var t * arguments.type.option.interp (@expr var) (@expr var) a)%type)
+           (only parsing).
+
+    Fixpoint partial_reduce' {t} (e : @expr (@expr var) t)
+      : forall a : arguments t, partial_reduceT t a
+      := match e in expr t return (forall a : arguments t, partial_reduceT t a) with
+         | TT
+           => arguments.invert
+                (fun a : arguments type.unit => partial_reduceT type.unit a)
+                (TT, TT)
+                (fun u => (TT, u))
          | Pair A B a b
-           => fun k
-              => @partial_reduce_cps
-                   T A a
-                   (fun a'
-                    => @partial_reduce_cps
-                         T B b
-                         (fun b' => k (Pair a' b')))
-         | Var t v => fun k => k v
-         | Abs s d f
-           => fun k
-              => k (Abs (fun x => @partial_reduce_cps _ d (f (Var x)) id))
+           => arguments.invert
+                (fun a => partial_reduceT (type.prod A B) a)
+                (let ab := (fst (@partial_reduce' A a arguments.generic),
+                            fst (@partial_reduce' B b arguments.generic))%expr in
+                 (ab, ab))
+                (fun '(aA, aB)
+                 => let '(a0, a1) := @partial_reduce' A a aA in
+                    let '(b0, b1) := @partial_reduce' B b aB in
+                    ((a0, b0)%expr, Some (a1, b1)))
+         | Var t v
+           => fun a => (v, arguments.expr.interp _ _ v)
          | Op s d opc args
-           => fun k
-              => @partial_reduce_cps
-                   T s args
-                   (fun args'
-                    => k
-                         match arguments.op.rewrite
-                                 opc
-                                 (arguments.expr.interp _ _ args')
-                         with
-                         | Some e => arguments.expr.reify _ _ e
-                         | None => Op opc args'
-                         end)
+           => let '(args0, args1) := @partial_reduce' s args (arguments.op.lookup_src opc) in
+              let e :=
+                  match arguments.op.rewrite opc args1 with
+                  | Some e => arguments.expr.reify _ _ e
+                  | None => Op opc args0
+                  end in
+              fun a => (e, arguments.expr.interp _ _ e)
+         | Abs s d f
+           => fun a
+              => let e' := Abs (fun x => fst (@partial_reduce' d (f (Var x)) (arguments.invert_arrow a))) in
+                 arguments.invert
+                   (fun a => partial_reduceT (type.arrow s d) a)
+                   (e', e')
+                   (fun ad
+                    => (e',
+                        (fun x =>
+                           snd (@partial_reduce' d (f x) ad))))
+                   a
          end.
 
+
     Definition partial_reduce {t} (e : @expr (@expr var) t) : @expr var t
-      := @partial_reduce_cps t t e id.
+      := snd (@partial_reduce' t e arguments.generic).
   End partial_reduce.
 
   Definition PartialReduce {t} (e : Expr t) : Expr t
@@ -1332,6 +1392,13 @@ Example base_25_5_mul (*(f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 g0 g1 g2 g3 g4 g5 g6 g7 g
   apply (f_equal (fun F => F f g)).
   cbv [n].
   cbv delta [mulmod w to_associational mul to_associational reduce from_associational add_to_nth zeros place split].
+  assert True.
+  { let v := Reify ((fun x => 2^x) 255)%Z in
+    pose v as E.
+    vm_compute in E.
+    pose (PartialReduce E) as E'.
+    vm_compute in E'.
+    constructor. }
   Reify_rhs ().
   let e := match goal with |- _ = Interp ?e => e end in
   pose e as E.
@@ -1339,7 +1406,6 @@ Example base_25_5_mul (*(f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 g0 g1 g2 g3 g4 g5 g6 g7 g
   Timeout 2 vm_compute in E.
   pose (PartialReduce E) as E'.
   Timeout 2 vm_compute in E'.
-
   (*cbv -[runtime_mul runtime_add]; cbv [runtime_mul runtime_add].
   ring_simplify_subterms.*)
 (* ?fg =
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -2311,7 +2311,7 @@ Example base_51_carry_mul (*(f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 g0 g1 g2 g3 g4 g5 g6 
 (*        (g9,g8,g7,g6,g5,g4,g3,g2,g1,g0)) mod (2^255 - 19) *)
   etransitivity; (* work around [rewrite] being stupid about evars *)
     [
-    | rewrite <- eval_chained_carries with (s:=2^255) (c:=[(1,19)]) (idxs:=(seq 0 (pred n) ++ [0; 1])%list%nat) (modulo:=fun x y => Z.modulo x y) (div:=fun x y => Z.div x y)
+    | rewrite <- eval_chained_carries with (s:=2^255) (c:=[(1,19)]) (idxs:=(seq 0 n ++ [0; 1])%list%nat) (modulo:=fun x y => Z.modulo x y) (div:=fun x y => Z.div x y)
       by (try assumption; auto using Z.div_mod; try (intros; eapply pow_ceil_mul_nat_divide_nonzero); try eapply pow_ceil_mul_nat_nonzero; try vm_decide);
       reflexivity ].
   eapply f_equal2; [|trivial]. eapply f_equal.
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -138,8 +138,8 @@ Module Positional. Section Positional.
     rewrite <- (seq_length n 0) at 2.
     generalize dependent (List.seq 0 n); intro xs.
     induction xs; simpl; nsatz.                               Qed.
-  Definition add_to_nth i x : list Z -> list Z
-    := ListUtil.update_nth i (runtime_add x).
+  Definition add_to_nth i x (ls : list Z) : list Z
+    := ListUtil.update_nth i (fun y => runtime_add x y) ls.
   Lemma eval_add_to_nth (n:nat) (i:nat) (x:Z) (xs:list Z) (H:(i<length xs)%nat)
         (Hn : length xs = n) (* N.B. We really only need [i < Nat.min n (length xs)] *) :
     eval n (add_to_nth i x xs) = weight i * x + eval n xs.
@@ -179,9 +179,9 @@ Module Positional. Section Positional.
   Hint Rewrite weight_place : push_eval.
 
   Definition from_associational n (p:list (Z*Z)) :=
-    List.fold_right (fun t =>
+    List.fold_right (fun t ls =>
       let p := place t (pred n) in
-      add_to_nth (fst p) (snd p)    ) (zeros n) p.
+      add_to_nth (fst p) (snd p) ls ) (zeros n) p.
   Lemma eval_from_associational {n} p (n_nz:n<>O \/ p = nil) :
     eval n (from_associational n p) = Associational.eval p.
   Proof. destruct n_nz; [ induction p | subst p ];
@@ -254,7 +254,7 @@ Module Positional. Section Positional.
 
       will produce [fun a b c d => (a + (b + (c + d)))].*)
     Definition chained_carries {n} s c p (idxs : list nat) :=
-      fold_right (@carry_reduce n s c) p (rev idxs).
+      fold_right (fun a b => @carry_reduce n s c a b) p (rev idxs).
 
     Lemma eval_chained_carries {n} s c p idxs :
       (s <> 0) -> (s - Associational.eval c <> 0) -> (n <> 0%nat) ->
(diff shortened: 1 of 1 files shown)


JasonGross:
I think I may be able to fix this by fusing `partial_reduce_cps` with `arguments.expr.interp`.  I'll try this tomorrow.

JasonGross:
Update from discussion with @andres-erbsen / todo list for me:
1. Rewrite reification and the syntax tree to be curried
2. Define a fixedpoint `interp_arguments_or_expr_for : type -> Type` which is something like
````
interp_arguments_or_expr_for type.nat := expr type.nat + nat
interp_arguments_or_expr_for (type.list T) := expr (type.list T) + list (interp_arguments_or_expr_for T)
interp_arguments_or_expr_for (A -> B) := expr (A -> B) + (interp_arguments_or_expr_for A -> interp_arguments_or_expr_for B))
```
I'm not completely sure about the `+` in the arrow case; I think it may screw with the `nat -> P -> P` argument of `nat_rect`, for example, but I hope it will be all-right.
3. rewrite `partial_reduce'` to return `interp_arguments_or_expr_for t`, and write a function `interp_arguments_or_expr_for t -> expr t`

JasonGross:
@andres-erbsen I now have a working version based on what we did today.  It works!  I might try to make the 200 lines of `Let` definitions be a single (and shorter) fixedpoint, but probably not in the next day or two.  Ready for your suggestions and review

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -325,44 +326,834 @@ Module Compilers.
...
+        Let defaulted_App_ee_p {A B C} (idc : ident (A -> B -> C))
+            (F : interp var A
+                 -> interp var B
+                 -> interp_prestep (interp var) C)
+          : interp var A
+            -> interp var B
+            -> @expr var C + interp_prestep (interp var) C
+          := fun x y => inr (F x y).
+
+        Definition interp {t} (idc : ident t) : interp var t
```
andres-erbsen: Just another idea that might work for cleaning up case analysis: `Either` as an error handing monad?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -325,44 +326,834 @@ Module Compilers.
...
+                        | None => inl v
+                        end
+                end.
+      End reify.
+
+      Section SmartLetIn.
+        Context {var : type -> Type} {tC : type}.
+        (** TODO: Find a better name for this *)
+        (** N.B. This always inlines functions; not sure if this is the right thing to do *)
+        (** TODO: should we handle let-bound pairs, let-bound lists?  What should we do with them?  Here, I make the decision to always inline them; not sure if this is right *)
```
andres-erbsen: I don't know. Whether or not to inline lets seems rather different from partial evaluation... I think the general rule is something like "pattern matching on partially evaluated terms should transparently see through vars and operate on expressions bound to them as if they were inlined, let-binding subparts of let-bound expressions right before the original binder. then later unused binders would be cleaned up."

The example I am thinking of is
~~~
let x := 1::2::3::4::nil in
let y := match x with nil => _ | cons a b => a end in
let z := match x with nil => _ | cons a b => b end in
(x,y,z)
~~~
--(partial)-->
~~~
let one := 1 in
let x' := 2::3::4::nil in
let x := one::x' in
let y := one in
let z := x' in
(x,y,z)
~~~
--(no-brainer 0 1 size1)-->
~~~
let x' := 2::3::4::nil in
(1::x',1,x')
~~~

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -325,44 +326,834 @@ Module Compilers.
...
+                          n)
+             | ident.List_seq as idc
+               => defaulted_App_pp_p2 idc List.seq
+             | ident.List_repeat A as idc
+               => defaulted_App_ep_p idc (@List.repeat (interp var A))
+             | ident.List_combine A B as idc
+               => defaulted_App_pp_p2 idc (@List.combine (interp var A) (interp var B))
+             | ident.List_map A B as idc
+               => defaulted_App_ep_p idc (@List.map (interp var A) (interp var B))
+             | ident.List_flat_map A B as idc
```
andres-erbsen: Why is this case particularly big?

JasonGross: Because it's the one case that handles three levels of evaluation: (1) if I get `expr (list A)` as the list I'm flat-mapping over, return `expr` using `ident.List.flat_map`; (2) if I get `list (expr A)` to flat-map over, and some of the applications of the function return `expr (list B)`, apply the function to all elements, return applications of `ident.List.app` to the resulting expression lists, which is also `expr (list B)`; (3) if all functions return `list (expr B)` and the list I'm flat-mapping over is `list (expr A)`, then I can return `list (expr B)`.  Almost all other cases handle only the equivalents of (1) and (3); some of the `Z` operations also handle a case where only one argument is an expression, but the logic is simpler there.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -474,11 +1264,10 @@ Module Compilers.
     | cons {t} (gallina_v : type.interp t) (v : var t) (ctx : list).
   End var_context.
 
-  (* cf COQBUG(https://github.com/coq/coq/issues/5448) *)
+  (* cf COQBUG(https://github.com/coq/coq/issues/5448) , COQBUG(https://github.com/coq/coq/issues/6315) *)
```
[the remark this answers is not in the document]
JasonGross: XD

I am losing track of which things forced me to use which variations of `fresh`.  I seem to recall something in the original, larger, code-base failing with the `refresh` I have now, but this code works, so I'll keep it until it stops working.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -626,13 +1415,63 @@ Module Compilers.
...
+Proof.
+  cbv [expand_list_helper]; revert idx H.
+  induction n as [|n IHn]; cbn; intros.
+  { reflexivity. }
+  { rewrite IHn by omega.
+    erewrite (@skipn_nth_default _ idx ls) by omega.
+    reflexivity. }
+Qed.
+
+Lemma expand_list_correct (n : nat) {A} (default : A) (ls : list A) (H : List.length ls = n)
```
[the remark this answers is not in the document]
JasonGross: Ah, good.  It took me like five tries to settle on the variant I ended up using, trying a bunch of different ways of list-indexing that all ended up being hard to prove.  Perhaps I should remove `ident.Nat.sub` and `ident.List.rev`, which were left over from previous tries; what do you think?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -325,44 +326,834 @@ Module Compilers.
...
+      := match v with
+         | inl v => inl (f v)
+         | inr v => inr (g v)
+         end.
+    Infix "+" := sum_arrow : function_scope.
+
+    Module ident.
+      Section interp.
+        Context {var : type -> Type}.
+        Let defaulted_App_p_e {A B} (idc : ident (A -> B))
```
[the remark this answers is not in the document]
JasonGross: Let's chat about this at the meeting today.  I can cover most of these by writing an inductive type that describes the structure here, and changing the `p_e` and `pp_p` and etc suffixes to a proper argument which is recursed over.  This would bring back the old inductive arguments type, basically, though it might be slightly simpler.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: This is quite tricky, and I don't see immediately how to do it.  It would mean the partial type wouldn't be a simple sum, but would be presented with the interface of a `force` method which gives one more layer of structure, and a `reify` method which hands back an `expr`.  The `force` method would have to be cps'd, at least for non-arrow non-unit types, because it would operate underneath let-binders....

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -325,44 +326,834 @@ Module Compilers.
...
+           | type.list _ as t
+           | type.Z as t
+           | type.nat as t
+           | type.bool as t
+             => @expr var t + interp_prestep interp t
+           end%type.
+      Fixpoint interp (t : type)
+        := interp_step interp t.
+
+      Definition unprestep {t : type} : interp_prestep interp t -> interp t
```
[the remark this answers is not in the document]
JasonGross: Roughly, they are `Some : t -> option t` and `Some ∘ Some : t -> option (option t)`, except that we're using `expr +` rather than `option`, and we're not using it uniformly, so it's a bit more complicated.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -325,44 +326,834 @@ Module Compilers.
...
+    Fixpoint invert_nat_full (e : @expr var type.nat) : option nat
+      := match e return option nat with
+         | App type.nat _ (Ident _ ident.S) args
+           => option_map S (invert_nat_full args)
+         | Ident _ (ident.Const type.nat v)
+           => Some v
+         | _ => None
+         end.
+    (* oh, the horrors of not being able to use non-linear deep pattern matches *)
+    Fixpoint invert_list_full {t} (e : @expr var (type.list t))
```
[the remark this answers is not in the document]
JasonGross: Yes.  I'll also mention https://github.com/coq/coq/issues/6320, which I just opened.  I could possibly compress this some with `Notation`s as I do in https://github.com/coq/coq/issues/6320; do you think that's worth it?

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: I don't see how to do it either, I am just using it as a reference for how I wish we could do it. Currently I can just suggest doing whatever approximates this suggestion for the code we are dealing with.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -325,44 +326,834 @@ Module Compilers.
...
+    Definition reify_list_by_app {t} (ls : list (@expr var (type.list t))) : @expr var (type.list t)
+      := list_rect
+           (fun _ => _)
+           (Ident ident.nil)
+           (fun ls1 _ ls2 => App (App (Ident ident.List.app) ls1) ls2)
+           ls.
+  End gallina_reify.
+
+  (** TODO: should this even be named [arguments]? *)
+  Module arguments.
```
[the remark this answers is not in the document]

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -325,44 +326,834 @@ Module Compilers.
...
+             | type.nat as t
+             | type.Z as t
+             | type.bool as t
+               => fun x : expr t + type.interp t
+                  => match x with
+                     | inl v => v
+                     | inr v => expr_const (t:=t) v
+                     end
+             end
+        with eta_expand {t : type}
```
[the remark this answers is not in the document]

src/Experiments/SimplyTypedArithmetic.v
JasonGross: And then write things in monadic syntax?  I think this will break in a few cases, where we want access to both branches, notably `runtime_add`, `runtime_mul`, and `flat_map`.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: One possibility is that we let-bind arguments to constructors and inline the constructors, consts, and vars.  Then `let x := expr1::expr2::expr3::expr4::nil in f x` becomes `let x1 := expr1 in let x2 := expr2 in let x3 := expr3 in let x4 := expr4 in f (x1::x2::x3::x4::nil)`.  This was a lot of code to write (because https://github.com/coq/coq/issues/6320) given that, if I recall correctly, we don't let-bind lists anywhere.  I think it might help to have `mul` or whatever in hand, to know what sorts of patterns we have to handle and what sorts we don't.  (I'd rather not write 100 lines of pattern matching return statements for a case we're never going to see in practice anyway.)

src/Experiments/SimplyTypedArithmetic.v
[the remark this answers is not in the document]
andres-erbsen: No.

JasonGross:
@andres-erbsen Ready for review.  I've made the changes we've talked about.  Alas, the pretty notations for list-indexing no longer work in the output; perhaps we should preserve uses of `nth_default` in addition to uses of `list_rect` and `nat_rect`?

andres-erbsen:
I pushed a new commit and I want github to show it on this conversation so I am commenting on it.

JasonGross:
@andres-erbsen Ready for more review.  I expect you will have things to say about `collect_translation_function_types` which exists only to to make the convenience wrapper `translate_under_binders` possible to write without having a formal description of what a "type with no arrows in it" is.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -273,9 +273,11 @@ End Positional. End Positional.
 
 Module Compilers.
   Module type.
-    Variant opaque := Z.
-    Inductive type := type_opaque (_:opaque) | unit | prod (A B : type) | arrow (s d : type) | list (A : type) | nat | bool.
-    Global Coercion type_opaque : opaque >-> type.
+    Variant opaque := unit | Z | nat | bool.
```
andres-erbsen: LGTM

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1326,43 +1326,6 @@ Module Compilers.
     End expr.
   End CPS.
```
andres-erbsen: LGTM

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -649,8 +649,9 @@ Module Compilers.
           | @List.repeat ?A
             => let rA := type.reify A in
                constr:(@ident.List_repeat rA)
-          | @LetIn.Let_In ?A (fun _ => ?B)
-            => let rA := type.reify A in
+          | @LetIn.Let_In ?A ?B
```
andres-erbsen: Would it be reasonable to change the input code instead? I think having no reduction in reification / only very simple and absolutely necessary reduction is a desirable goal.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1872,100 +1945,200 @@ Example base_25_5_mul (*(f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 g0 g1 g2 g3 g4 g5 g6 g7 g
...
+                             (2251799813685248 * (x [[2]] * x0 [[4]]) +
+                              (2 * (x [[3]] * x0 [[1]]) +
+                               (67108864 * (x [[3]] * x0 [[2]]) +
+                                (4503599627370496 * (x [[3]] * x0 [[3]]) +
+                                 (151115727451828646838272 * (x [[3]] * x0 [[4]]) +
+                                  (x [[4]] * x0 [[0]] +
+                                   (67108864 * (x [[4]] * x0 [[1]]) +
+                                    (2251799813685248 * (x [[4]] * x0 [[2]]) +
+                                     (151115727451828646838272 * (x [[4]] * x0 [[3]]) +
+                                      5070602400912917605986812821504 * (x [[4]] * x0 [[4]])))))))))))))) in
```
andres-erbsen: Why are there large numbers in the output code? I don't think there should be anything bigger than the 51-bit mask.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1297,41 +1309,75 @@ Module Compilers.
...
-        Context {var : type -> Type}
-                {R : type}.
+        Context {var : type -> Type}.
         Notation var' R := (fun t => var (type.translate R t)).
-
-        Fixpoint translate {t}
+        Section with_R.
+          Context {R : type}.
+
+          Fixpoint translate {t}
```
andres-erbsen: I don't understand what changed and why.

JasonGross: I indented this block so that this code could be in a section specifying `R` and the code immediately afterwards can specialize `R`

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Yes, the easiest thing to do here is to add a `dlet_nd` notation, which might need its own definition.  Shall I do that?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Yeah, I find this suspicious too.  Not sure how to debug, though.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Fixed

src/Experiments/SimplyTypedArithmetic.v
JasonGross: The large numbers are:
- 67108864 = 2^26
- 2251799813685248 = 2^51
- 151115727451828646838272 = 2^77

The issue was that I was using 25.5 as a base and asking for only 10 limbs; I didn't realize that `w` didn't autocalculate the base from the number of limbs.  Will be fixed momentarily.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: This was so that I could add `translate_under_binders`

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Now fixed

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -273,10 +273,10 @@ End Positional. End Positional.
 
 Module Compilers.
   Module type.
-    Variant opaque := unit | Z | nat | bool.
-    Inductive type := type_opaque (_:opaque) | prod (A B : type) | arrow (s d : type) | list (A : type).
+    Variant primitive := unit | Z | nat | bool.
```
andres-erbsen: LGTM.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -661,11 +661,13 @@ Module Compilers.
           | @List.repeat ?A
             => let rA := type.reify A in
                constr:(@ident.List_repeat rA)
-          | @LetIn.Let_In ?A ?B
-            => let B := lazymatch (eval cbv beta in B) with fun _ => ?B => B end in
-               let rA := type.reify A in
+          | @LetIn.Let_In ?A (fun _ => ?B)
```
andres-erbsen: LGTM.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1954,194 +1955,148 @@ Eval cbv [proj1_sig base_51_carry_mul] in (fun f g Hf Hg => proj1_sig (base_51_c
...
-                                   (67108864 * (x [[4]] * x0 [[1]]) +
-                                    (2251799813685248 * (x [[4]] * x0 [[2]]) +
-                                     (151115727451828646838272 * (x [[4]] * x0 [[3]]) +
-                                      5070602400912917605986812821504 * (x [[4]] * x0 [[4]])))))))))))))) in
-           expr_let _ := Ident ident.Z.div @ y @ 67108864 in
-           expr_let _ := Ident ident.Z.modulo @ y @ 67108864 in
+                         (x [[1]] * x0 [[3]] +
+                          (x [[2]] * x0 [[2]] +
+                           (x [[3]] * x0 [[1]] + x [[4]] * x0 [[0]]))) in
+           expr_let _ := Ident ident.Z.div @ y @ 2251799813685248 in
```
andres-erbsen: are these unused binders created by the CPS conversion?

JasonGross: I believe CPS conversion does not introduce any new let-binders, only new abstractions and applications.  So they must already be there in the source.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -316,269 +340,122 @@ Module Compilers.
...
+        | xH => on_success ()
+        | ?term => on_failure term
+        end.
+      Ltac require_primitive_const term :=
+        is_primitive_const_cps2 term ltac:(fun _ => idtac) ltac:(fun term => fail 0 "Not a known const:" term).
+      Ltac is_primitive_const term :=
+        is_primitive_const_cps2 term ltac:(fun _ => true) ltac:(fun _ => false).
+
+      Module var_context.
+        Inductive list {var : type -> Type} :=
```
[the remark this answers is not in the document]
JasonGross: What's the suggestion here?  Replace `var_context.list` with `reification_context`, and unqualify `nil` and `cons`?  Or replace `var_context.list` with `var_context.reification_context` (seems redundant)?  Or replace `var_context.*` with `reification_context.*`?  Or something else?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -587,168 +464,1949 @@ Module Compilers.
...
+      Section translate.
+        Fixpoint translate (t : Compilers.type.type) : type
+          := match t with
+             | A * B => (translate A * translate B)%cpstype
+             | s -> d => (translate s * (translate d --->) --->)%cpstype
+             | Compilers.type.list A => type.list (translate A)
+             | Compilers.type.type_primitive t
+               => t
+             end%ctype.
+        Fixpoint untranslate (R : Compilers.type.type) (t : type)
```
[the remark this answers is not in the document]
JasonGross: No, it is not.  I could call it `continuation_to_arrow` or `cps_arrow_to_arrow`?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -587,168 +464,1949 @@ Module Compilers.
...
+                          (* ignore this line; it's to work around lack of fixpoint refolding in type inference *) var (type.untranslate _ (type.translate tx) * (type.untranslate _ (type.translate tx) * (type.untranslate _ (type.translate tC) -> R) -> R) * (type.untranslate _ (type.translate tC) -> R))%ctype) ,
+                     ident.Let_In
+                       @@ (ident.fst @@ (ident.fst @@ (Var xyk)),
+                           (λ (x :
+                                 (* ignore this line; it's to work around lack of fixpoint refolding in type inference *) var (type.untranslate _ (type.translate tx))) ,
+                            (ident.snd @@ (ident.fst @@ (Var xyk)))
+                              @ (Var x, ident.snd @@ Var xyk)))
+                | ident.nat_rect P
+                  => λ (PO_PS_n_k :
+                          (* ignore this line; it's to work around lack of fixpoint refol
```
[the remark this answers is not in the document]
JasonGross: What do you mean "unused"?  How can `(* ignore this line; it's to work around lack of fixpoint refol` be unused?  (And why is it truncated?)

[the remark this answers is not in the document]
JasonGross: What do you mean "unused"?  How can `(* ignore this line; it's to work around lack of fixpoint refol` be unused?  (And why is it truncated?)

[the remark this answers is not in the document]
JasonGross: What do you mean "unused"?  How can `(* ignore this line; it's to work around lack of fixpoint refol` be unused?  (And why is it truncated?)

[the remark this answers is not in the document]
JasonGross: What do you mean "unused"?  How can `(* ignore this line; it's to work around lack of fixpoint refol` be unused?  (And why is it truncated?)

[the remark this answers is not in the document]
JasonGross: Is this also a mis-placed comment?  The line does say "ignore this line"...

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -762,8 +2420,171 @@ Example base_25_5_mul (*(f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 g0 g1 g2 g3 g4 g5 g6 g7 g
...
+         (_ : length (Datatypes.snd fg) = 5%nat) =>
+       expr.Interp (@interp)
+         (fun var : type -> Type =>
+          (λ x : var (type.list type.Z * type.list type.Z)%ctype,
+           expr_let y := fst @@ x [[0]] * snd @@ x [[4]] +
+                         (fst @@ x [[1]] * snd @@ x [[3]] +
+                          (fst @@ x [[2]] * snd @@ x [[2]] +
+                           (fst @@ x [[3]] * snd @@ x [[1]] +
+                            fst @@ x [[4]] * snd @@ x [[0]]))) in
+           expr_let _ := Z.div @@ (y, 2251799813685248) in
```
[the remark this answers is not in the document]
JasonGross: I'm not sure, so I would like you to take a look at it.  I will note that every `expr_let` comes from an explicit `dlet ... in ...`, so I think, if there is an issue, either the issue is that we're mis-marking the expressions we want not inlined, or we're doing something wrong in the cps transform that results in lets not being lifted in the right places.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -587,168 +464,1949 @@ Module Compilers.
...
-            let rf := reify_helper var f value_ctx template_ctx in
-            constr:(Op (var:=var) op.App (Pair (var:=var) rf rx))
+            => constr:(fun x : T
+                       => ltac:(let v := plug_template_ctx (f x) template_ctx in
+                                exact v))
           end
-        | (fun x : ?T => ?f)
+        end.
+
+      Ltac reify_helper ident reify_ident var term value_ctx template_ctx :=
```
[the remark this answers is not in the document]
JasonGross: Changed to `reify_in_context`

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -316,269 +340,122 @@ Module Compilers.
...
+             | AppIdent s d idc args => interp_ident s d idc (@interp s args)
+             | App s d f x => @interp _ f (@interp _ x)
+             | Pair A B a b => (@interp A a, @interp B b)
+             | Abs s d f => fun v => interp (f v)
+             end.
+
+        Definition Interp {t} (e : Expr t) := interp (e _).
+      End with_ident.
+
+      Ltac is_primitive_const_cps2 term on_success on_failure :=
```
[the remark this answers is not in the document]
JasonGross: done

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -273,7 +273,23 @@ End Positional. End Positional.
...
+      Global Coercion type_primitive : primitive >-> type.
+    End Coercions.
+
+    Fixpoint final_codomain (t : type) : type
+      := match t with
+         | arrow _ d => final_codomain d
+         | t => t
+         end.
+
+    Fixpoint under_arrows (t : type) (f : type -> type) : type
```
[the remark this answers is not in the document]
JasonGross: removed

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -273,7 +273,23 @@ End Positional. End Positional.
...
 Module Compilers.
   Module type.
-    Inductive type := unit | prod (A B : type) | arrow (s d : type) | list (A : type) | nat | Z | bool.
+    Variant primitive := unit | Z | nat | bool.
+    Inductive type := type_primitive (_:primitive) | prod (A B : type) | arrow (s d : type) | list (A : type).
+    Module Export Coercions.
+      Global Coercion type_primitive : primitive >-> type.
+    End Coercions.
+
+    Fixpoint final_codomain (t : type) : type
```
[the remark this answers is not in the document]
JasonGross: removed

JasonGross:
>  Do you have a plan for debugging this?

I do not, currently.  I'm not actually sure exactly what the output is supposed to look like.  Perhaps we should meet sometime and pair-debug?

andres-erbsen:
Okay. I don't know what happened to the actual review comments (all hail github), and I don't really care. I think pair debugging would make sense, but I am currently on the other side of the pond so it would need to be a video call. Monday? Or I could just look into it myself... I'm worried that "substitution" of unused binders may be necessary, and not just for getting the right output also for keeping the computational complexity managable.

JasonGross:
How about you look into it yourself, and then if it's not solved by Monday, we pair-debug online.  I'm free most of the day, until around 6pm EST.

Is the issue just that we have unused binders?  Or also that we have duplicated code or things that are inlined too much?  If it's just the unused binders, we can run dead code elimination after partial evaluation...  (Alternatively, if we can somehow figure out ahead of time which let binders will be unused, we can have partial evaluation unfold those particular let-binders.)

andres-erbsen:
After dead code elimination (by hand):

~~~~~~~~~~~~
(λ x : var (type.list type.Z * type.list type.Z)%ctype,
expr_let y11 := fst@@x[[0]] * snd@@x[[0]] + (19 * (fst@@x[[1]] * snd@@x[[4]]) + (19 * (fst@@x[[2]] * snd@@x[[3]]) + (19 * (fst@@x[[3]] * snd@@x[[2]]) + 19 * (fst@@x[[4]] * snd@@x[[1]])))) in
expr_let y12 := Z.div@@(y11, 2251799813685248) in
expr_let y13 := Z.modulo@@(y11, 2251799813685248) in
expr_let y23 := y12 + (fst@@x[[0]] * snd@@x[[1]] + (fst@@x[[1]] * snd@@x[[0]] + (19 * (fst@@x[[2]] * snd@@x[[4]]) + (19 * (fst@@x[[3]] * snd@@x[[3]]) + 19 * (fst@@x[[4]] * snd@@x[[2]]))))) in
expr_let y24 := Z.div@@(y23, 2251799813685248) in
expr_let y25 := Z.modulo@@(y23, 2251799813685248) in
expr_let y34 := y24 + (fst@@x[[0]] * snd@@x[[2]] + (fst@@x[[1]] * snd@@x[[1]] + (fst@@x[[2]] * snd@@x[[0]] + (19 * (fst@@x[[3]] * snd@@x[[4]]) + 19 * (fst@@x[[4]] * snd@@x[[3]]))))) in
expr_let y35 := Z.div@@(y34, 2251799813685248) in
expr_let y36 := Z.modulo@@(y34, 2251799813685248) in
expr_let y44 := y35 + (fst@@x[[0]] * snd@@x[[3]] + (fst@@x[[1]] * snd@@x[[2]] + (fst@@x[[2]] * snd@@x[[1]] + (fst@@x[[3]] * snd@@x[[0]] + 19 * (fst@@x[[4]] * snd@@x[[4]]))))) in
expr_let y45 := Z.div@@(y44, 2251799813685248) in
expr_let y46 := Z.modulo@@(y44, 2251799813685248) in
expr_let y62 := Z.div@@(y13, 2251799813685248) in
expr_let y63 := Z.modulo@@(y13, 2251799813685248) in
expr_let y71 := y62 + y25 in
expr_let y72 := Z.div@@(y71, 2251799813685248) in
expr_let y73 := Z.modulo@@(y71, 2251799813685248) in
           y63::y73::(y72+y36)::y46::(y45+(fst@@x[[0]] * snd@@x[[4]] + (fst@@x[[1]] * snd@@x[[3]] + (fst@@x[[2]] * snd@@x[[2]] + (fst@@x[[3]] * snd@@x[[1]] + fst@@x[[4]] * snd@@x[[0]]))))) :: [])%expr
~~~~~~~~~~~~~~

andres-erbsen:
I think this code is incorrect. In particular, the nil end of the list has an unreduced limb, and `y13` is reduced twice. Now investigating whether we could blame the templates instead of the compiler.

andres-erbsen:
New ouput looks more reasonable (at https://github.com/JasonGross/fiat-crypto/blob/0ac8aa53a3cde2409b6275a910af1d8d346c32c8/src/Experiments/SimplyTypedArithmetic.v#L2581)

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2272,6 +2273,103 @@ Module Compilers.
...
+         | TT => (cur_idx, PositiveSet.empty)
+         | AppIdent s d idc args
+           => let default := @compute_live' _ args cur_idx in
+              match args in expr.expr t return ident.ident t d -> _ with
+              | Pair A B x (Abs s d f)
+                => fun idc
+                   => match idc with
+                      | ident.Let_In _ _
+                        => let '(idx, live) := @compute_live' A x cur_idx in
+                           let '(_, live) := @compute_live' _ (f (PositiveSet.add idx live)) (Pos.succ idx) in
```
andres-erbsen: Is it actually necessary to special-case let in (and not other higher-order functions) when assigning numbers to binders? I think could just work as if the variables bound using a let are accessible outside of its body (but happen to be unused).

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2272,6 +2273,103 @@ Module Compilers.
...
+                                           -> match s return Type with
+                                              | _ * (s -> d) => (expr s -> expr d)%type
+                                              | _ => unit
+                                              end%ctype
+                                           -> positive * expr d
+                                           -> positive * expr d)
+                             with
+                             | ident.Let_In _ _
+                               => fun x' f' _
+                                  => if PositiveSet.mem idx live
```
andres-erbsen: I understand that we need to special-case LetIn here, since we want to eliminate it if useful.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: The important thing that happens here is that we pass in `PositiveSet.add idx live` as the variable node to `f`.  I don't see a way to do that without special-casing at least let in (though we could special-case all identifiers, and basically write this as an `interp`).

andres-erbsen: Perhaps it would make sense to `add` live variables at all `Abs` nodes, not just let binders? Do I understand correctly that then the let-in special case would not be necessary?

JasonGross: What do we `add` the live variables to, and how do we make sure we have that set at `Abs` nodes?

andres-erbsen:
Okay, I'll stop trying to figure out something more principled for dead code elimination because the current one seems sufficient for the purpose we are going to be using it for. let's revisit it if it ever starts to influence larger design decisions, though.

JasonGross:
That sounds good; the more principled way is to wait until after we've stripped out the higher-order combinators to do dead code elimination (we in fact already have a proven-correct version that goes through the Named representation).  This is more a placeholder to have something in place.

Shall we merge this, and then I can write a pass to strip out all the things we don't want to handle in bounds analysis (like `list_rect`), pull in the bounds analysis pass, and then we can figure out what pipeline interface should actually look like?

andres-erbsen:
sgtm.

PR #278, opened 2017-11-29
Curves.Montgomery.XZ: add+check boringssl ladderstep

andres-erbsen:
This PR is to check that the elliptic curve arithmetic at https://boringssl.googlesource.com/boringssl/+/7ce2378750f9f56459ddba67947c3c4ba001a526/third_party/fiat/curve25519.c#4945 does what we want it to do. In particular, we prove it equivalent to the version in https://cr.yp.to/ecdh/curve25519-20060209.pdf appendix B, which has already been proven to correspond to addition in affine coordinates at https://github.com/mit-plv/fiat-crypto/compare/boringladderstep?diff=split&expand=1&name=boringladderstep#diff-d9c50f42ae6bdbec2520e416b2848608R150

Apparently the previous decision to name the small hardcoded constant `a24` was a poor one: the two ladderstep functions use two different hardcoded constants (`(a+2)/4` and `(a-2)/4`) compensating for it with slight changes in the formula.

@davidben please take a look :)

### DIFF
--- a/src/Curves/Montgomery/XZ.v
+++ b/src/Curves/Montgomery/XZ.v
@@ -85,6 +85,31 @@ Module M.
         ((x2, z2), (x3, z3))%core
       end.
 
+    Context {ap2d4:F} {ap2d4_correct:(1+1+1+1)*a24 = a+1+1}.
+    Definition boringladderstep (x1:F) (Q Q':F*F) : (F*F)*(F*F) :=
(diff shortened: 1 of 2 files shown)


davidben:
With the caveat that I don't really understand everything here, I agree that this PR looks like a plausible transcription of the BoringSSL code you linked to. :-)

PR #280, opened 2017-12-06
Curves.Weierstrass.Jacobian: boringssl p256 formulas

andres-erbsen:
This contains transcriptions of the remaining two cases of group operations from https://boringssl-review.googlesource.com/c/boringssl/+/23244/13/third_party/fiat/p256.c#980. In particular, we state the invariant for Jacobian coordinates and prove the correctness of addition and doubling in these coordinates.

The Travis failures are timeouts (this file takes 10min, crossing over the 50min limit).

@davidben let's read this together.


src/Curves/Weierstrass/Jacobian/Basic.v
```diff
@@ -0,0 +1,187 @@
...
+      Hint Unfold double negb andb : points_as_coordinates.
+      Program Definition add (P Q : point) : point :=
+        match proj1_sig P, proj1_sig Q return F*F*F with
+        | (x1, y1, z1), (x2, y2, z2) =>
+          let z1nz := if dec (z1 = 0) then false else true in
+          let z2nz := if dec (z2 = 0) then false else true in
+          let z1z1 := z1^2 in
+          let z2z2 := z2^2 in
+          let u1 := x1 * z2z2 in
+          let ftmp5 := z1 + z2 in
```
davidben: Nit: s/ftmp5/two_z1z2/ to match the naming now.

src/Curves/Weierstrass/Jacobian/Basic.v
```diff
@@ -0,0 +1,187 @@
...
+          let gamma := gamma^2 in
+          let y_out := alpha * y_out in
+          let gamma := gamma + gamma in
+          let y_out := y_out - gamma in
+          (x_out, y_out, z_out)
+        end.
+      Next Obligation. Proof. t. Qed.
+
+      Hint Unfold double negb andb : points_as_coordinates.
+      Program Definition add (P Q : point) : point :=
```
davidben: This only does the mixed = 0 case, though the mixed = 1 case is pretty obviously equivalent. Maybe I'll toy with it as a really easy exercise...

andres-erbsen: Sure. https://github.com/mit-plv/fiat-crypto/blob/cf51e032986312e5638d911cb9f3749c87c47219/src/Curves/Weierstrass/Jacobian/Precomputed.v#L72 is probably the best place to do it.

davidben: Oh heh, I didn't see this message and ended up doing something much more ridiculous. (Emailed the diff to you.)

andres-erbsen:
@JasonGross please take a look at this after Travis has resolved. I don't know how to fix Travis better than what i attempted now.

JasonGross:
Do we want to try to optimize this file?  I started looking into it, and I can get a pretty decent speed-up on the contradictory goals with ad-hoc recognition rules, and the other ones can often get a 2x or more speedup by selectively clearing unused hypotheses first.  In any case, I'll push a commit to this branch that separates out this file into yet another travis job.

JasonGross:
Travis now fails (about 3 seconds before it would time out) with
```
File "./src/Curves/Weierstrass/Jacobian.v", line 214, characters 48-69:
Error:
Anomaly "Uncaught exception Unix.Unix_error(Unix.ENOMEM, "fork", "")." 
Please report at http://coq.inria.fr/bugs/.
Command exited with non-zero status 129
src/Curves/Weierstrass/Jacobian (real: 594.55, user: 410.89, sys: 1.85, mem: 3126312 ko)
```

andres-erbsen:
If you have heuristics that push this file below the build timeout, please add them in a way that it clear that these are just performance optimizations and makes them trivial to remove. Another rather generic optimization I think might give a benefit would be to canonicalize (ACI) and deduplicate goals.

andres-erbsen:
I don't see the OOM on travis, just a timeout. Am I looking at something else?

JasonGross:
https://travis-ci.org/mit-plv/fiat-crypto/jobs/315759463#L1771 and https://travis-ci.org/mit-plv/fiat-crypto/jobs/315759463#L1654  Apparently coq just sits there hanging if the worker thread rather than the main thread gets killed with ENOMEM?

JasonGross:
If we can fix that, it looks like we'll have 15-20 minutes to spare on this Travis job.

andres-erbsen:
@JasonGross what should we do about this one? I don't know how much work it would be to make it faster or less memoruy-hungry, and I don't want to sink a day just to find out.

JasonGross:
What happens if we just remove the `par:`?

src/Curves/Weierstrass/Jacobian.v
```diff
@@ -0,0 +1,219 @@
...
+      Proof. t. Qed.
+
+      Lemma add_mixed_eq_add (P : point) (Q : point) (H : z_is_zero_or_one Q) :
+        eq (add P Q) (add_mixed P Q H).
+      Proof. t. Qed.
+
+      Lemma Proper_add : Proper (eq ==> eq ==> eq) add. Proof. t. Qed.
+      Import BinPos.
+      Lemma to_affine_add
+            {char_ge_12:@Ring.char_ge F Feq Fzero Fone Fopp Fadd Fsub Fmul 12} (* TODO: why do we need 12 instead of 3? *)
```
JasonGross: Have you checked recently if we still need 12?  It seems to go through fine for me when I replace 12 with 3 (at least in my optimized version; I'm still waiting to see if it all goes through without my ad-hoc helper scripts which are basically all "clear some hypotheses and use `fsatz` to prove that the given versions of others imply simplified ones", which, if `fsatz` works correctly, shouldn't change the dependence on the characteristic).

JasonGross: Oh, hrm, it seems like it's probably a bug in `nsatz`; the failure is at `Qed` time, and seems to have something to do with nested inverses or something

JasonGross:
@andres-erbsen I've pushed some commits that speed up `Jacobian.v` from 13m36s to 1m06s.  I would have factored out the tactic script into a separate file, except this makes the tactic even more unreadable (and is very painful to do) because the way we are doing context variables makes it impossible to reuse notations in tactics across sections.

JasonGross:
@andres-erbsen I will leave it to you whether or not to merge with my changes that speed up Jacobian, or to do something else with them.

andres-erbsen:
Just to make sure I understand why the shortcut tactics are in Jacobian instead of an external file: would the issues be fixed if it was possible to introduce a notation for a ltac-bound gallina term such as `Fmul`?

JasonGross:
Yes

andres-erbsen:
Well, it's a mess. I don't know how to clean it up, so we may as well keep it on master instead of this branch.

JasonGross:
Created https://github.com/coq/coq/issues/6569

Also, it occurs to me that we could possibly move the script to another file by doing `Set Printing All. Print Ltac ...`, and then copy paste.  It'd still be unreadable.  Do you think this is worth doing?

PR #282, opened 2017-12-13
Curves.Montgomery.XZProofs overhaul

andres-erbsen:
@davidben was not able to jump into this file earlier, so a cleanup is warranted. Unfortunately this change causes a slowdown in the biggest proof, 19s -> 160s. It is possible that the speed could be re-gained with less mess than this file contained before...

### DIFF
--- a/src/Curves/Montgomery/XZProofs.v
+++ b/src/Curves/Montgomery/XZProofs.v
@@ -40,13 +40,14 @@ Module M.
       match goal with
       | _ => progress intros
       | _ => progress autounfold with points_as_coordinates in *
-      | _ => progress destruct_head' @unit
-      | _ => progress destruct_head' @bool
-      | _ => progress destruct_head' @prod
-      | _ => progress destruct_head' @sig
-      | _ => progress destruct_head' @sum
-      | _ => progress destruct_head' @and
-      | _ => progress destruct_head' @or
+      | _ => progress destruct_head'_True
+      | _ => progress destruct_head'_unit
+      | _ => progress destruct_head'_prod
+      | _ => progress destruct_head'_sig
+      | _ => progress destruct_head'_bool
+      | _ => progress destruct_head'_sum
+      | _ => progress destruct_head'_and
+      | _ => progress destruct_head'_or
       | H: context[dec ?P] |- _ => destruct (dec P)
       | |- context[dec ?P]      => destruct (dec P)
       | |- ?P => lazymatch type of P with Prop => split end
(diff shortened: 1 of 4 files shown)


JasonGross:
Can we leave the mess in a `MessyTactics.v` file and chain it with `t` afterwards, and say "look, you can remove `fast_ad_hoc_t; t` and replace it with `t`, and the proofs still all go through, they're just slower?

src/Curves/Montgomery/XZProofs.v
```diff
@@ -192,70 +133,51 @@ Module M.
...
-      :
-           eq (to_xz (Madd Q Q )) (fst (xzladderstep x1 xz xz))
-        /\ eq (to_xz (Madd Q Q')) (snd (xzladderstep x1 xz x'z'))
-        /\ projective (fst (xzladderstep x1 xz x'z'))
+    Lemma ladder_invariant_preserved x1 Q Q' (H:ladder_invariant x1 Q Q')
+      : ladder_invariant x1 (Madd Q Q) (Madd Q Q').
+    Proof.
+      cbv [ladder_invariant] in *.
+      pose proof difference_preserved Q Q' as Hrw.
+      (* TODO: rewrite with in match argument with [sumwise (fieldwise (n:=2) Feq) (fun _ _ => True)] *)
```
andres-erbsen: @JasonGross 
I tried to prove this one today, and did not find any nice way. I am guessing you have seen a bunch of situations like this and can quickly guess a good strategy for tackling it. Or if it would be very painful now and less painful after some change we have been considering making, that would also be useful to hear. The general situation here is that I am just not worried that what I mean by this lemma might be false, but I'm still stuck trying to come up with a proof. This is probably the only place where we have had to connect between points-as-coordinates and points-as-group-elements without just blasting both sides into coordinates..

~~~~~~~~~~~
  H : match M.coordinates (Madd Q (Mopp Q')) with
      | (x, _) => x = x1 /\ x1 <> 0
      | ∞ => False
      end
  Hrw : M.eq (Madd (Madd Q Q) (Mopp (Madd Q Q'))) (Madd Q (Mopp Q'))
  ============================
  match M.coordinates (Madd (Madd Q Q) (Mopp (Madd Q Q'))) with
  | (x, _) => x = x1 /\ x1 <> 0
  | ∞ => False
  end
~~~~~~~~~~~

andres-erbsen: I got it (in a hacky way).

JasonGross: Use `sum_rect` rather than `match` (you can write a tactic that folds it) and then make `forall A B T L R IMPL, Proper ((L ==> IMPL) ==> (R ==> IMPL) ==> sumwise L R ==> IMPL) (@sum_rect A B (fun _ => T))`.  That's how I'd do it.

PR #287, opened 2018-01-03
Prove montladder correct in the zero case.

davidben:
This is largely copy-pasted from the other one. I also have no clue what I'm doing, so feel free to ask me to change things arbitrarily.


andres-erbsen:
Todo tracked in #292

PR #289, opened 2018-01-05
Reenable TWord 0 and print it as uint8_t

### LINKED ISSUE #288: Perform bool to uint8_t conversion at the notation level
For @JasonGross, by way of a discussion with me and @andres-erbsen. Any mistakes in the transcription are mine.

The synthesized code works with one-bit integers a lot for carries and the like. This is nominally a `bool`, but C didn't get support for this until C99, so instead things got rounded up to `uint8_t` at some point down the line. This was done by way of making `TWord 3` a `uint8_t` while leaving `TWord 0` a `bool`, then somewhere telling the bounds-checker to round to `TWord 3` instead of `TWord 0`.

This has two problems. First, the unused `TWord 0` notations are not actually correct. Adding two `bool`s together is *not* addition mod 2. It is actually saturating. It's because C first casts the two `bool`s to `int`, adds them, and casts back down to `bool` by checking `2 != 0`.

Second, converting to `TWord 3` forces `AddWithCarry` and `SubWithBorrow` implementations to be pessimistic. They must take in a full 8 bits of precision in the carry bit and output 8 bits. This means one can't use the [real compiler intrinsics](https://software.intel.com/en-us/cpp-compiler-18.0-developer-guide-and-reference-addcarry-u32-addcarry-u64), which treat any non-zero input as a 1. Also, for instance, [this](https://boringssl.googlesource.com/boringssl/+/11850d5f618f2e39ad80fedb42feafa9047330b7/third_party/fiat/p256.c#65) `subborrow_u64` implementation isn't correct.

The effect is particularly pronounced for, say, `addcarryx_u26` and `subborrow_u26`. In theory, these could be implemented as:

```
static uint8_t addcarryx_u26(uint8_t c, uint32_t a, uint32_t b, uint32_t *low) {
  uint32_t x = a + b + c;
  *low = x & ((1 << 26) - 1);
  return (x >> 26) & 1;
}

static uint8_t subborrow_u26(uint8_t c, uint32_t a, uint32_t b, uint32_t *low) {
  uint32_t x = a - b - c;
  *low = x & ((1 << 26) - 1);
  return x >> 31;
}
```

These functions need only extract 26+1 = 27 bits of output, so a `uint32_t` is a sufficient temporary. However, if the `uint8_t` output is treated as a true `uint8_t`, rather than a stand-in for a `uint1_t`, we need 26+8 = 34 bits of output, and we must go up a word size, even though that's not actually necessary.

Proposal:

1. Fix the `TWord 0` notations to, rather than using `bool`, which doesn't work, use `uint8_t` with manual truncations. So, adding two bools would be `uint8_t c = (a + b)&1`. (Though I'd be surprised if you ever emitted that one.)

2. Remove the `TWord 0` to `TWord 3` conversion and leave all `TWord 0`s as-is.

3. Remove the `TWord 3` notation cases for `addcarryx` and `subborrow`. Only select those intrinsics when the carry bits are `TWord 0`.

JasonGross:
This closes #288.  (Handling bullet point 1 was kind-of painful.)

I can't seem to request @davidben as a reviewer?

Potentially interesting things:
- Request 3 (don't use addcarryx notations for TWord 3) was fundamentally a 1-line change
- Request 2 was the only non-notation change; it happens in the place where we fill in default permissible bitwidths.  In the new pipeline, it might be worth considering where things like this should live.
- I'm not completely confident I got request 1 correct.  (For example, if `&` does not bind more tightly than multiplication, addition, and subtraction, then some of the cast notations might be wrong.)


src/Compilers/Z/CNotations.v
```diff
@@ -718,34 +729,34 @@ Notation "( x + y )" := (Op (Add _ _ _) (Pair (Var x) y)) (format "( x  +  y )")
...
-Notation "( x + '(bool)' y )" := (Op (Add (TWord _) (TWord _) (TWord 0)) (Pair (Var x) y)) (at level 50, y at level 9, format "( x  +  '(bool)' y )") : expr_scope.
-Notation "( '(bool)' x + '(bool)' y )" := (Op (Add (TWord _) (TWord _) (TWord 0)) (Pair (Var x) y)) (at level 50, x at level 9, y at level 9, format "( '(bool)' x  +  '(bool)' y )") : expr_scope.
-Notation "( '(bool)' x + y )" := (Op (Add (TWord _) (TWord _) (TWord 0)) (Pair (Var x) (Var y))) (at level 50, x at level 9, format "( '(bool)' x  +  y )") : expr_scope.
-Notation "( x + '(bool)' y )" := (Op (Add (TWord _) (TWord _) (TWord 0)) (Pair (Var x) (Var y))) (at level 50, y at level 9, format "( x  +  '(bool)' y )") : expr_scope.
-Notation "( '(bool)' x + '(bool)' y )" := (Op (Add (TWord _) (TWord _) (TWord 0)) (Pair (Var x) (Var y))) (at level 50, x at level 9, y at level 9, format "( '(bool)' x  +  '(bool)' y )") : expr_scope.
+Notation "( '1&(uint8_t/*bool*/)' x + y )" := (Op (Add (TWord _) (TWord _) (TWord 0)) (Pair x y)) (at level 50, x at level 9, format "( '1&(uint8_t/*bool*/)' x  +  y )") : expr_scope.
```
davidben: Hrm. Is that quite right? I think we want the `&1` to be applied after the operation, rather than before. The reason it's there is because if you have

```
   uint8_t/*bool*/ a = 1;
   uint8_t/*bool*/ b = 1;
   uint8_t/*bool*/ c = a + b;
```

you'll get 2 rather than 0. In other cases, you rely on the language taking the modulus for you. But here, the language only knows to reduce by 2^8, so you need to do the rest of overflow handling yourself.

Of course, this only matters if we ever overflow and hit the reduction case, which won't happen since we're starting out with Z anyway. But the proofs do nominally apply to the reducing semantics.

davidben: If it's simpler to convince notations to do it, one could also imagine transforming it into `c = a ^ b`. I.e. special-case all the boolean ones with the actual bit operations they are. However, then you need to worry about, say, `Add (TWord 5) (TWord 5) (TWord 0)`. That should transform to:

```
   uint32_t a;
   uint32_t b;
   uint8_t/*bool*/ c = (1&(uint8_t/*bool*/)a) ^ (1&(uint8_t/*bool*/)b);
```

(Not that that case would happen either because there are no actual overflows here...)

Alternatively, do you ever actually emit this thing? The only cases where there aren't overflows are when you know a priori that `a` or `b` is zero. You could just punt these operations. :-)

JasonGross: We don't emit them, but I don't think we can punt them either, because there is a default notation that will render them as `+`. :-P  I'll fix it so the cast happens after.

davidben: `Add (TWord _) (TWord _) (TWord 0) => SYNTAX_ERROR_IM_PRETTY_SURE_WE_NEVER_EMIT_THIS`? :-)

JasonGross: Actually, that's not needed.  If you look somewhat further down, you will see that this notation is overridden by the following additions:
```coq
Notation "( '1&(' x * y ')' )" := (Op (Mul (TWord 0) (TWord _) (TWord _)) (Pair x y)) (at level 40, format "( '1&(' x  *  y ')' )").
Notation "( '1&(' x * y ')' )" := (Op (Mul (TWord 0) (TWord _) (TWord _)) (Pair x (Var y))) (at level 40, format "( '1&(' x  *  y ')' )").
Notation "( '1&(' x * y ')' )" := (Op (Mul (TWord 0) (TWord _) (TWord _)) (Pair (Var x) y)) (at level 40, format "( '1&(' x  *  y ')' )").
Notation "( '1&(' x * y ')' )" := (Op (Mul (TWord 0) (TWord _) (TWord _)) (Pair (Var x) (Var y))) (at level 40, format "( '1&(' x  *  y ')' )").
Notation "( '1&(' x + y ')' )" := (Op (Add (TWord 0) (TWord _) (TWord _)) (Pair x y)) (at level 50, format "( '1&(' x  +  y ')' )").
Notation "( '1&(' x + y ')' )" := (Op (Add (TWord 0) (TWord _) (TWord _)) (Pair x (Var y))) (at level 50, format "( '1&(' x  +  y ')' )").
Notation "( '1&(' x + y ')' )" := (Op (Add (TWord 0) (TWord _) (TWord _)) (Pair (Var x) y)) (at level 50, format "( '1&(' x  +  y ')' )").
Notation "( '1&(' x + y ')' )" := (Op (Add (TWord 0) (TWord _) (TWord _)) (Pair (Var x) (Var y))) (at level 50, format "( '1&(' x  +  y ')' )").
Notation "( '1&(' x - y ')' )" := (Op (Sub (TWord 0) (TWord _) (TWord _)) (Pair x y)) (at level 50, format "( '1&(' x  -  y ')' )").
Notation "( '1&(' x - y ')' )" := (Op (Sub (TWord 0) (TWord _) (TWord _)) (Pair x (Var y))) (at level 50, format "( '1&(' x  -  y ')' )").
Notation "( '1&(' x - y ')' )" := (Op (Sub (TWord 0) (TWord _) (TWord _)) (Pair (Var x) y)) (at level 50, format "( '1&(' x  -  y ')' )").
Notation "( '1&(' x - y ')' )" := (Op (Sub (TWord 0) (TWord _) (TWord _)) (Pair (Var x) (Var y))) (at level 50, format "( '1&(' x  -  y ')' )").
```

JasonGross: Oh, urgh, did I get my underscores mixed up?  I think the `0` needs to be moved to the third slot for these...

davidben: Ah, okay. I don't know how to read the notation syntax, so I'll take your word for it that it's correct. If it ever ends up in synthesized code, I suppose we can just visually inspect it to make sure it's not totally crazy.

JasonGross: Updated

davidben:
This LGTM, but I don't understand the notations code and can't really review it beyond seeing the checked-in output looks right and random snippets seem plausible. :-)

JasonGross:
I'll take that as "good enough to merge", and as evidence that we should move away from this clunky notation system for outputting valid C.

Issue #294, opened 2018-01-10
Align/combine solinas{32,64}_2e255m19_{10,5}limbs and X25519/C{32,64}


davidben:
@JasonGross @andres-erbsen 

I started trying to adapt the 51-bit limb schedule for BoringSSL, as that's a considerable perf win. Things I found:

* The solinas* versions are missing an fecarry implementation.
* The X25519/C{32,64} versions are missing C files and also include said C files in .gitignore, which makes them a little harder to use.
* The X25519/C{32,64} versions don't seem to emit quite the same code as the solinas* versions. The input and output are funny and there's some surrounding bits.
* I gather the C{32,64} versions are nominally the same, but have some manual instruction scheduling borrowed from an curve25519-donna, unlike the solinas* versions which are purely automated? I ran some benchmarks. It appears that C32 vs solinas32* doesn't make any difference, but C64 is about 16% faster than solinas64*.

This suggests the C32 version isn't very useful, but the C64 one is. I'm not sure how you all prefer to organize it, either fold the manual scheduling of C64 into solinas64*, or keep having two of them around? Either way, it would be nice for fecarry to be easily available and for the C code to be like the solinas* one in whichever instances are ultimately supported.

JasonGross:
@davidben I've added fecarry for all the solinas implementations in a4789ec72a32b6d3b1d40154626e8c6674d4aa7d.  Feel free to make the `*Display.log` and `*.c` files that you need and add them. 

x25519_c64 and x25519_c32 both have manually specified replacement C code (which is parsed into Coq) in the .json files:
https://github.com/mit-plv/fiat-crypto/blob/a4789ec72a32b6d3b1d40154626e8c6674d4aa7d/src/Specific/CurveParameters/x25519_c64.json#L12-L73

I don't see a way to add that to the fully-autogenerated versions, which generate everything based on just the prime number.

Currently, the method we use to go from the display logs to the C code is different for X25519/* vs solinas*.  The former uses a different method of benchmarking, etc, and also has as test binary.   The easiest, most kludgy thing to do is to add variants of `solinas64_2e255m19_*limbs.json` which mention donna and have the hand-crafted C code.  I've done this in #295.  Perhaps it makes more sense to unify the two ways we have of benchmarking, though. @andres-erbsen what do you think?

davidben:
Thanks!

[...]

PR #295, opened 2018-01-10
Add x25519 donna versions with the new way of generating C code

JasonGross:
See #294.

This is kludgy, and perhaps we should update `X25519/C*` versions instead.


davidben:
By the way, I haven't fully unpacked what's going on here, but I think the difference between the synthesized code and donna is that donna does the 0x13 multiplication earlier as a `uint64_t` (which fits because 51 bits leaves enough room), while the synthesized code does that multiplication after we've done enough other work that it's a `uint128_t`.

That is, the synthesized code computes:

```
uint64_t a, b, c, d;
uint128_t ab = (uint128_t)a * b;
uint128_t cd = (uint128_t)c * d;
uint128_t abcd = ab + cd;
uint128_t abcd19 = 0x13 * abcd;
```

Whereas donna computes:

```
uint64_t a, b, c, d;
uint64_t a19 = 0x13 * a;  // This is okay because a is 51-bit
uint64_t c19 = 0x13 * c;  // This is okay because c is 51-bit
uint128_t ab19 = (uint128_t)a19 * b;
uint128_t cd19 = (uint128_t)c19 * d;
uint128_t abcd19 = ab19 + cd19;
```

Notably the synthesized code multiples one `uint128_t` by 19 rather than two `uint64_t`s. This seems to frustrate the compiler's ability to optimize multiplying by 19, since `uint128_t` is not the natural word size.

davidben:
I believe the compiler also can't do this transformation itself because it doesn't know `19 * a` as a `uint64_t` does not overflow. (Though if I add a check, it still doesn't perform this transformation. I suspect `uint128_t`s are only minimally optimized.)

davidben:
Oh, and a second trick donna does is once you perform that transformation, CSE gets to take another pass because `a19` and `c19` are used a couple of times.

JasonGross:
We very nearly have a CSE pass; I believe the main issue is that I haven't read Adam's proof of correctness from lambda-tamer closely enough to see why it is easy, so there are a few lingering cases that I haven't yet pushed through in the proof.

The other transformation is a bit tricker to specify, and it's not entirely clear to me what it should be.  Something like "if we find a node `c * x` for some constant `c`, and we find that all references to `x` across the entire program are `c * x`, and `x` is defined as `y + z` where `y` and `z` are both of the form `a * b` (not necessarily for the same `a` and `b`) and the bit-width on `c * a * b` is the same as the bitwidth of `a * b` and larger than either the bitwidth of `c * a` or `c * b`, and this holds for both the constituents of `y` and the constituents of `z`, then we can bind a fresh variable to `c * a` (or `c * b`) for each of `y` and `z`, bind fresh variables for `cy` and `cz`, and replace the binding of `c * x` with a binding of `cy + cz`" would work, but is very convoluted.  Moreover, it requires at least two passes: one pass to annotate each variable with the number of occurrences, and a second pass to do the replacement.  Perhaps a better strategy would involve a different representation of arithmetic expressions and some sort of bounds-aware pass out of that?

davidben:
Do you actually need to count occurrences? Even if it didn't get to do CSE, I expect two 19 * u64 -> u64 operations is already going to faster than one 19 * u128 -> u128, when the native word size is u64. In the latter, I still need to multiply 19 by low and high u64s, but I have to deal with carries from the former to the latter rather than discarding the top word. (The tricks the compiler does to multiply by 19 don't give back the high word.)

davidben:
(To be clear, I'm perfectly happy with just using the manual scheduling. If closing that gap isn't very interesting, don't take this as a request or anything.)

JasonGross:
No, we don't need to count occurences.  We'd still want to do dead code elimination, but, hm, I guess if each variable node stored it's value one and two levels deeper, it wouldn't be too hard to identify `19 * (a * b + c * d)` and change it to `19 * a * b + 19 * c * d`.  I'd expect it to take me about two days to write the PHOAS pass and another week or two to prove it correct.  It's not currently on my todo list, but feel free to open an issue to track this, or to undertake it yourself :-)

PR #303, opened 2018-01-18
Pass -mbmi2 to gcc

JasonGross:
This will hopefully fix the issue where travis gcc complains about not
being able to inline mulx.  I hope.


JasonGross:
@andres-erbsen Will anything go wrong if we pass this flag?

andres-erbsen:
I read the gcc man page and I think that we should have been passing it all the time when using mulx.

PR #305, opened 2018-02-07
[Work in Progress] Experiment with Bounds Analysis based on lists

JasonGross:
We now do bounds analysis in experiments.  This is the minimum product necessary to prototype the idea; there are a number of admits, mostly arounds bounds proofs, but also around casting.  My hope is that, with feedback, we can turn this into something more reasonable.  Please question everything that seems the least bit unfortunate.

I created this as a pull request to take advantage of GitHub's review infrastructure.

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -3536,9 +3536,6 @@ Module X25519_64.
 
   Derive base_51_carry_mul
          SuchThat (forall
-                      (*(f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 g0 g1 g2 g3 g4 g5 g6 g7 g8 g9 : Z)
-        (f:=(f0 :: f1 :: f2 :: f3 :: f4 :: f5 :: f6 :: f7 :: f8 :: f9 :: nil)%list)
-        (g:=(f0 :: f1 :: f2 :: f3 :: f4 :: f5 :: f6 :: f7 :: f8 :: f9 :: nil)%list)*)
                       (fg : list (BoundsAnalysis.type.BoundedZ _ _)
                             * list (BoundsAnalysis.type.BoundedZ _ _))
                       (f := fst fg) (g := snd fg)
@@ -3728,9 +3725,6 @@ Module X25519_32.
 
   Derive base_25p5_carry_mul
          SuchThat (forall
-                      (*(f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 g0 g1 g2 g3 g4 g5 g6 g7 g8 g9 : Z)
-        (f:=(f0 :: f1 :: f2 :: f3 :: f4 :: f5 :: f6 :: f7 :: f8 :: f9 :: nil)%list)
-        (g:=(f0 :: f1 :: f2 :: f3 :: f4 :: f5 :: f6 :: f7 :: f8 :: f9 :: nil)%list)*)
                       (fg : list (BoundsAnalysis.type.BoundedZ _ _)
                             * list (BoundsAnalysis.type.BoundedZ _ _))
                       (f := fst fg) (g := snd fg)
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -12,6 +12,7 @@ Require Import Crypto.Util.ZRange.
 Require Import Crypto.Util.ZRange.Operations.
 Require Import Crypto.Util.Tactics.RunTacticAsConstr.
 Require Import Crypto.Util.Tactics.Head.
+Require Import Crypto.Util.Option.
 Require Import Crypto.Util.Notations.
 Import ListNotations. Local Open Scope Z_scope.
 
@@ -2936,6 +2937,11 @@ Module Compilers.
                | Pair A B a b => fun _ => b
                | _ => fun x => x
                end (AppIdent snd e).
+          Definition smart_cast {A B : type.primitive} (e : @expr ident A) : @expr ident B
+            := match transport (@expr ident) A B e with
+               | Some e' => e'
+               | None => AppIdent (cast A B) e
+               end.
 
           Fixpoint list_map {A B : type.primitive}
                    (f : @expr ident A -> @expr ident B)
@@ -2984,6 +2990,12 @@ Module Compilers.
                                       (rec_xs f)
                             | _ => fun _ _ _ => None
                             end a' (fun f => @list_map _ _ f b') f
+                    | TT
+                      => match idc with
+                         | ident.nil _
+                           => fun _ _ => Some (AppIdent ident.nil TT)
+                         | _ => fun _ _ => None
+                         end
                     | _ => fun _ _ => None
                     end idc
                | _ => fun _ => None
@@ -3022,10 +3034,10 @@ Module Compilers.
                                => existT _ (Datatypes.fst r :: Datatypes.snd r)
                                          (AppIdent
                                             cons
-                                            (cast _ _ @@ (smart_fst args),
+                                            (smart_cast (smart_fst args),
                                              snd_args)))
                               (list_map
-                                 (AppIdent (cast _ _))
+                                 smart_cast
                                  (smart_snd args))
                        | None => None
                        end
@@ -3061,11 +3073,14 @@ Module Compilers.
                       (fun '(existT r args)
 		       => existT _ (land_bounds r r[mask~>mask]%zrange)
                                  (AppIdent (land _ _ mask) args))
-               | cast _ _
+               | cast _ (type.ZBounded l u)
                  => option_map
-                      (fun '(existT r args)
-		       => existT _ admit (AppIdent (cast _ _) args))
-               end.
+                      (fun '(existT r[l'~>u'] args)
+		       => existT _ r[Z.max l l' ~> Z.min u u']
+                                 (AppIdent (cast _ _) args))
+               | cast _ _
+                 => fun _ => None
+               end%zrange.
         End with_relax.
       End ident.
 
@@ -3077,7 +3092,7 @@ Module Compilers.
           Local Notation RangeCtx := (PositiveMap.t { t : type & ident.range t }).
           Local Notation expr := (@expr ident).
           Local Notation type_for_range := (type_for_range relax_zrange).
-          Local Notation "( a , b )" := (existT _ a b).
+          Local Notation "( a ; b )" := (existT _ a b) : core_scope.
 
           Definition lookupb (ctx : RangeCtx) {t} (p : positive) : option (range t)
             := match PositiveMap.find p ctx with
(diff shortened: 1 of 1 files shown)


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2433,660 @@ Module Compilers.
...
+      Definition resize2uc {T1 T2 T3 T1' T2' T3' : type.primitive}
+        : (type.interp (T1 * T2) -> type.interp T3) -> (type.interp (T1' * T2') -> type.interp T3')
+        := fun f '((x, y) : type.interp T1' * type.interp T2')
+           => resize (f (resize x, resize y)).
+
+      Fixpoint default {t : type} : type.interp t
+        := match t with
+           | unit => tt
+           | Z => -1
+           | ZBounded lower upper => {| value := lower ; value_bounded := admit |}
```
andres-erbsen: re `admit`: What if lower=1, upper=0?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2433,660 @@ Module Compilers.
...
+           : nexpr_scope.
+      Notation "'x_' n" := (Var _ n) (at level 10, format "'x_' n") : nexpr_scope.
+    End Notations.
+
+    Module AdjustBounds.
+      Import Indexed.
+      Import type expr ident.
+      Module ident.
+        Section with_relax.
+          Context (relax_zrange : zrange -> option zrange).
```
andres-erbsen: Would `relax_zrange` later round up to fastest remaining word size available in the hardware?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Right, so we have to make a design choice here.  We could represent ranges as lower : Z and size : option positive, or we could interpret ranges by reorienting them if they're wrong in interp (with min and max), or we could have the proof component of boundedness be order agnostic and fix bounds analysis to impose ordering.  Each of these has complications.  Do you have any thoughts on design here?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Yes

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2433,660 @@ Module Compilers.
...
+               | nil _ as idc
+                 => fun _ => Some (existT _ Datatypes.nil (AppIdent nil TT))
+               | cons t
+                 => option_map
+                      (fun '(existT bs args)
+                       => existT _ (Datatypes.fst bs :: Datatypes.snd bs)
+                                 (AppIdent
+                                    cons
+                                    (cast _ _ @@ (fst @@ args),
+                                     cast_list _ _ @@ (snd @@ args))))
```
andres-erbsen: What is the logic for the target type of the cast? I don't think we can make this work by requiring all elements of a list to have the same bounds.

JasonGross: Right, this is why I thought that we'd need to switch to tuples, and then realized that we can actually just carry bounds separately.  We do not require that the bounds be the same, but we do require that the word size be the same.  (Said another way, we do require that the bounds carried in the type be the same, but we track the tighter bounds of each element separately.)

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2433,660 @@ Module Compilers.
...
+      | cons {t : type.primitive} : ident (t * list t) (list t)
+      | fst {A B} : ident (A * B) A
+      | snd {A B} : ident (A * B) B
+      | List_nth {T : type.primitive} (n : nat) : ident (list T) T
+      | mul (T1 T2 Tout : type.primitive) : ident (T1 * T2) Tout
+      | add (T1 T2 Tout : type.primitive) : ident (T1 * T2) Tout
+      | shiftr (T1 Tout : type.primitive) (offset : BinInt.Z) : ident T1 Tout
+      | land (T1 Tout : type.primitive) (mask : BinInt.Z) : ident T1 Tout
+      | cast (T1 Tout : type.primitive) : ident T1 Tout
+      | cast_list (T1 Tout : type.primitive) : ident (list T1) (list Tout).
```
JasonGross: Note to self: Make this a `Fixpoint`, return `None` in the case where we do not have a bunch of cons cells

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2566,30 +3317,40 @@ Proof.
...
         (g:=(f0 :: f1 :: f2 :: f3 :: f4 :: f5 :: f6 :: f7 :: f8 :: f9 :: nil)%list)*)
                     (fg : list Z * list Z)
                     (f := fst fg) (g := snd fg)
-                    (n:=5%nat)
                     (Hf : length f = n) (Hg : length g = n)
+                    (fg_bounds := (f_bounds, f_bounds))
                     (fg' := base_51_carry_mul fg),
-                    (eval w n fg') mod (2^255-19)
-                    = (eval w n f * eval w n g) mod (2^255-19))
+                    (eval w n fg') mod (s - Associational.eval c)
```
JasonGross: Note to self: `Derive` an `expr`

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2433,660 @@ Module Compilers.
...
+           match e with
+           | Some e
+             => let e := AdjustBounds.expr.adjust_bounds
+                           relax_zrange e
+                           (PositiveMap.add 1%positive (existT _ _ s_bounds) (PositiveMap.empty _)) in
+                e
+           | None => None
+           end.
+
+      Definition cast
```
JasonGross: Note to self: This should go away

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2435,784 @@ Module Compilers.
...
+                 {bs : AdjustBounds.ident.range (Indexed.OfPHOAS.type.compile d)}
+                 (args : type.interp (AdjustBounds.ident.type_for_range relax_zrange s_bounds))
+                 (e : @expr ident (AdjustBounds.ident.type_for_range relax_zrange bs))
+        : option (Compilers.type.interp d)
+        := let ctx := PositiveMap.add
+                        1%positive
+                        (existT _ _ args)
+                        (PositiveMap.empty _) in
+           let v := Indexed.expr.interp (@ident.interp) e ctx in
+           option_map (cast_back relax_zrange) v.
```
andres-erbsen: Could `option_map (cast_back relax_zrange)` be moved into the caller instead of here? Would we then be able to generalize the type of `Interp`?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2435,784 @@ Module Compilers.
...
+          Context {ident : type -> type -> Type}
+                  (interp_ident : forall s d, ident s d -> type.interp s -> type.interp d).
+
+          Fixpoint interp {t} (e : @expr ident t)
+                   (ctx : PositiveMap.t { t : type & type.interp t })
+            : option (type.interp t)
+            := match e with
+               | Var t n
+                 => match PositiveMap.find n ctx with
+                    | Some (existT t' v) => type.transport type.interp t' t v
```
andres-erbsen: please factor out `find` and `transport` into a single function that returns option

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2435,784 @@ Module Compilers.
...
+
+      Module expr.
+        Import ident.
+        Section with_relax.
+          Context (relax_zrange : zrange -> option zrange).
+
+          Local Notation RangeCtx := (PositiveMap.t { t : type & ident.range t }).
+          Local Notation expr := (@expr ident).
+          Local Notation type_for_range := (type_for_range relax_zrange).
+          Local Notation "( a ; b )" := (existT _ a b) : core_scope.
```
andres-erbsen: perhaps `&` to match the type-level notation?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2435,784 @@ Module Compilers.
...
+
+          Definition lookupb (ctx : RangeCtx) {t} (p : positive) : option (range t)
+            := match PositiveMap.find p ctx with
+               | Some (existT t' r)
+                 => type.transport range t' t r
+               | None => None
+               end.
+          Definition extendb (ctx : RangeCtx) {t} (p : positive) (r : range t)
```
andres-erbsen: inline this

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2435,784 @@ Module Compilers.
...
+        Import ident.
+        Section with_relax.
+          Context (relax_zrange : zrange -> option zrange).
+
+          Local Notation RangeCtx := (PositiveMap.t { t : type & ident.range t }).
+          Local Notation expr := (@expr ident).
+          Local Notation type_for_range := (type_for_range relax_zrange).
+          Local Notation "( a ; b )" := (existT _ a b) : core_scope.
+
+          Definition lookupb (ctx : RangeCtx) {t} (p : positive) : option (range t)
```
andres-erbsen: `find_of_type`? `find_at_type`?

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: Or perhaps `Bind sigT_scope with sigt` and then overload comma?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2435,784 @@ Module Compilers.
...
+                      let ctx' := extendb ctx x range_x in
+                      bC <- @adjust_bounds ctx' tC eC;
+                        let '(range_C; eC') := bC in
+                        Some (range_C; Let_In x ex' eC')
+               end.
+        End with_relax.
+      End expr.
+    End AdjustBounds.
+
+    Module OfPHOAS.
```
andres-erbsen: why is this module `OfPHOAS`?  Perhaps a range-analysis-related name?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2435,784 @@ Module Compilers.
...
+                         | ident.nil _
+                           => fun _ _ => Some (AppIdent ident.nil TT)
+                         | _ => fun _ _ => None
+                         end
+                    | _ => fun _ _ => None
+                    end idc
+               | _ => fun _ => None
+               end f.
+
+          Definition adjust_bounds {s d} (idc : ident s d)
```
andres-erbsen: perhaps definitions like this one would benefit from the notations used below?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2435,784 @@ Module Compilers.
...
+               | Pair A B a b => fun _ => b
+               | _ => fun x => x
+               end (AppIdent snd e).
+          Definition smart_cast {A B : type.primitive} (e : @expr ident A) : @expr ident B
+            := match transport (@expr ident) A B e with
+               | Some e' => e'
+               | None => AppIdent (cast A B) e
+               end.
+
+          Fixpoint list_map {A B : type.primitive}
```
andres-erbsen: Based on the type, I would guess a trivial implementation. Is the code below actually doing that (and having non-linear pattern matching troubles), or does it do something else? A comment might be useful if something short can communicate the answer to this question.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2370,6 +2435,784 @@ Module Compilers.
...
+                                        end
+                            -> @expr ident match t with
+                                           | type.prod A B => A
+                                           | _ => A
+                                           end
+               with
+               | Pair A B a b => fun _ => a
+               | _ => fun x => x
+               end (AppIdent fst e).
+          Definition smart_snd {A B} (e : @expr ident (A * B)) : @expr ident B
```
andres-erbsen: As far as input language is concerned, I would be fine with just not supporting `let p := (h, t) in cons p`. If supporting it makes implementation+proofs overall easier, so be it.

andres-erbsen:
Why do we have a `unit` type in the embedded language?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: The general version is `Indexed.expr.interp`.  This is a thin wrapper specifically designed so that we don't have to write the context-adding and the cast_back in the caller

src/Experiments/SimplyTypedArithmetic.v
JasonGross: The `Bind Scope` would not work well, because the pairing notation is in `core_scope`, and most functions don't explicitly take sigma types, and, e.g., the sigma scope wouldn't be opened for the argument to `Some` even if we were looking for an `option (sigT _)`.  I prefer `;` to `&`, in part because it matches with HoTT so I'm used to it, in part because this one doesn't collide with parenthesizing `Z.land`.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Why?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: It's already called `BoundsAnalysis.OfPHOAS`.  It's because these are the wrapper functions that do bounds analysis on PHOAS syntax tree inputs

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Maybe a bit.  I'll update the one case that it seems relevant in

src/Experiments/SimplyTypedArithmetic.v
JasonGross: If you strip out all of the convoy patterning and return types, this is just
```coq
          Fixpoint list_map {A B : type.primitive}
                   (f : @expr ident A -> @expr ident B)
                   (e : @expr ident (type.list A))
            : option (@expr ident (type.list B))
            := match e with
               | AppIdent _ _ (ident.nil _) _
                 => Some (AppIdent ident.nil TT)
               | AppIdent _ _ (ident.cons _) (Pair _ _ x xs)
                 => option_map
                      (fun f_xs => AppIdent ident.cons (f x, f_xs))
                      (@list_map _ _ f xs)
               | _ => None
               end.
```
Of course, that doesn't typecheck.  I can insert a comment to this effect.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: I don't see an easy way to forbid `ident`s from returning `A * list B`, which is what we'd need, in addition to restricting `var` to primitive types, to eliminate this.  (Note also that the function argument is of type `list Z * list Z`, so we'd need to do something if that's forbidden as a `Var` node.)

JasonGross:
You suggested inlining `Definition extendb`; I believe merging is primarily waiting on a reason to do that.

andres-erbsen:
I'm fine with not inlining it, feel free to merge.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3506,143 +3506,150 @@ Proof.
...
+  Definition c := [(1, 19)].
+  Definition n := 5%nat.
+  Definition idxs := (seq 0 n ++ [0; 1])%list%nat.
+  Definition f_bounds := List.repeat r[0~>(2^Qceiling limbwidth + 2^(Qceiling limbwidth - 3))%Z]%zrange n.
+  Definition relax_zrange : zrange -> option zrange
+    := relax_zrange_gen [64; 128].
+
+  Derive base_51_carry_mul
+         SuchThat (forall
+                      (*(f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 g0 g1 g2 g3 g4 g5 g6 g7 g8 g9 : Z)
```
andres-erbsen: this comment is out of date, please delete it from both locations

JasonGross: Removed

PR #306, opened 2018-02-12
Write a C string generator for synthesized code (in Experiments)

JasonGross:
Note that all return values get an extra level of indirection, so that
we can uniformly support returning types like `Z`, `list Z`, `list Z *
list Z`, `Z * Z`, etc.

Sorry to steal your thunder @davidben 

The function I'm most concerned about is `infix_to_string`, which does some logic about where to insert casts and where to not insert casts.

Note that I added string helper functions to `Util` in 711c020174dc922d0f45ae7afe28e1c7ea8c9168.  This code will hopefully make it into the Coq stdlib via https://github.com/coq/coq/pull/6599 and https://github.com/coq/coq/pull/6597

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -32,6 +32,8 @@ Require Import Crypto.Util.ZUtil.Tactics.DivModToQuotRem.
 Require Import Crypto.Util.Tactics.SpecializeBy.
 Require Import Crypto.Util.Tactics.SplitInContext.
 Require Import Crypto.Util.Tactics.SubstEvars.
+Require Import Crypto.Util.Strings.Decimal.
+Require Import Crypto.Util.Strings.HexString.
 Require Import Crypto.Util.Notations.
 Require Import Crypto.Util.ZUtil.Definitions.
 Require Import Crypto.Util.ZUtil.CC Crypto.Util.ZUtil.Rshi.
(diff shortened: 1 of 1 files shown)


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2879,6 +2891,308 @@ Module Compilers.
...
+        End Z.
+
+        Module primitive.
+          Definition to_UL_postfix (t : type.primitive) : string
+            := match t with
+               | type.unit => ""
+               | type.Z => ""
+               | type.ZBounded lower upper
+                 => let u := (if lower >=? 0 then "U" else "") in
+                    let sz := Z.log2_up (Z.max (Z.abs upper + 1) (Z.abs lower)) in
```
andres-erbsen: I think the `sz` logic may not be correct. How about you just test whether the range of `t` is included in the range specified at <http://en.cppreference.com/w/cpp/language/types#Range_of_values> for each macro in <http://en.cppreference.com/w/cpp/types/integer#Function_macros_for_minimum-width_integer_constants>?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2879,6 +2891,308 @@ Module Compilers.
             := compile (e _).
         End expr.
       End OfPHOAS.
+
+      Module ToString.
+        Import Coq.Strings.String.
+        Import Coq.Strings.Ascii.
+        Local Open Scope string_scope.
+
+        Module Z.
```
andres-erbsen: Maybe this should go in coq#6597?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2879,6 +2891,308 @@ Module Compilers.
...
+                    else if sz <=? 64
+                         then u ++ "L"
+                         else if sz <=? 128
+                              then u ++ "LL"
+                              else " /* " ++ Z.to_hex_string lower ++ " <= val <= " ++ Z.to_hex_string upper ++ " */"
+               end.
+
+          Definition to_string {t} (v : type.primitive.interp t) : string
+            := match t return type.primitive.interp t -> string with
+               | type.unit => fun 'tt => "tt"
```
andres-erbsen: `tt` might be valid C in some context... how about `#error "tt"`?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2879,6 +2891,308 @@ Module Compilers.
...
+               | type.Z => Z.to_hex_string
```
andres-erbsen: Again, outputting a `Z` is should be error imho

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2879,6 +2891,308 @@ Module Compilers.
...
+            Definition to_string (t : type.primitive) : string
+              := match t with
+                 | type.unit => "unit"
+                 | type.Z => "ℤ"
+                 | type.ZBounded lower upper
+                   => let lg2u := Z.log2_up (upper + 1) in
+                      let default := "ℤ[" ++ Z.to_hex_string lower ++ "," ++ Z.to_hex_string upper ++ "]" in
+                      if (2^lg2u - 1 =? upper) && (lower =? 0)
+                      then "uint" ++ Z.to_decimal_string lg2u ++ "_t"
+                      else if (-2^lg2u =? lower) && (2^lg2u - 1 =? upper)
```
andres-erbsen: This logic seems correct to me. We may want to make a shared definition `Z.log2_exact : Z -> option Z`.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2879,6 +2891,308 @@ Module Compilers.
...
+
+          Fixpoint string_interp (t : type) : Set
+            := match t with
+               | type.type_primitive x => string
+               | type.prod A B => string_interp A * string_interp B
+               | type.list A => list string + string
+               end.
+        End type.
+
+        Module ident.
```
andres-erbsen: Here is an idea of how to do this in a way that seems less sketchy to me.

We would implement "common type"
 computation as in <http://en.cppreference.com/w/c/language/conversion#Usual_arithmetic_conversions> case `4)`, including conversion ranks. This would be a fair bit of transcription, but the only thinking required would be to make sure our rank order is transitive. Then we could handle each operation node by trying various combinations of input casts and picking the nicest-looking one for which the "common type" can fit the values of the input 1, input 2, and output. If the common type is not the output type, insert an additional output cast. As the base case, casting each inputs to a type that fits values of all of them should ensure that the common type is that type.

andres-erbsen: In particular, I am worried about unintuitive signed-unsigned conversion.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2879,6 +2891,308 @@ Module Compilers.
...
+                    => [(type.primitive.to_string A ++ " *" ++ (if by_ref then "*" else ""))
+                          ++ match ls with
+                             | inl ls
+                               => List.hd "" ls
+                             | inr sls
+                               => sls
+                             end]
+               end.
+
+          Fixpoint join (sep : string) (ls : list string) : string
```
andres-erbsen: Is this actually not in the standard library? :/

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3697,6 +4013,40 @@ Module X25519_64.
  x_19 :: x_22 :: x_21 +₃₂₋₆₄₋₆₄ x_10 :: x_13 :: x_16 :: [])%nexpr
      : expr
          (BoundsAnalysis.AdjustBounds.ident.type_for_range relax_zrange f_bounds)
+*)
+  Compute BoundsAnalysis.Indexed.ToString.expr.to_function_string
+          "femul"
+          arg_type
+          base_51_carry_mul.
+  (*
+void femul(uint64_t *x₁, uint64_t *x₂, uint64_t **ret) {
```
andres-erbsen: Would it be easy for bounds analysis to determine the concrete length of a list and put it in a type?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2879,6 +2891,308 @@ Module Compilers.
...
+                            => make_assignment true ret_prefix _ retval)).
+
+          Definition to_string (indent : string) {t} (e : @expr ident t)
+            : string
+            := List.fold_right
+                 (fun a b => a ++ NewLine ++ b)
+                 ""
+                 (to_lines indent e).
+
+          Fixpoint make_arg_list (by_ref : bool)
```
andres-erbsen: I think we should allow allow only primitives (passed by value) and constant-length lists (passed by pointer to first element) as arguments. Pairs should be flattened as they are, lists of lists and variable-length lists should be given error messages.
I think we should only allow scalars (passed by pointer) and constant-length lists (passed by pointer to first element) as outputs. Everything else should be flattened. If you feel like it, you can implement either "if the output is only one scalar, return it directly" or "if there is only one scalar in the output, return it directly", but these are not important. If at all possible, maintain the invariant that none of the inputs are read after any of the outputs have been written (perhaps generate output-writing code completely separately, even at the cost of rejecting `(a, let x := a in b)`).

Maaaybe we could try to support lists of lists by passing a list of pointers to first elements, but I do not see an use case. List of pairs of lists would be even more painful. Later-later we may want to support variable-length buffers, but I don't think we have enough context to deal with them for now.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2879,6 +2891,308 @@ Module Compilers.
...
+                                     ls n
+                       | inr sls => sls ++ "[" ++ Z.to_decimal_string (Z.of_nat n) ++ "]"
+                       end
+               | ident.mul T1 T2 Tout
+                 => fun '(x, y)
+                    => mul_to_string T1 T2 Tout x y
+               | ident.add T1 T2 Tout
+                 => fun '(x, y)
+                    => add_to_string T1 T2 Tout x y
+               | ident.shiftr T1 Tout offset
```
andres-erbsen: I don't understand the following 2 cases, so I lean even more towards implementing common type computation.

andres-erbsen: Oh, apparently it is even more complicated than that: https://stackoverflow.com/questions/42787336/c-bitwise-shift-right-operand-considered-for-implicit-type-conversion

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2879,6 +2891,308 @@ Module Compilers.
...
+                 => fun k
+                    => k (var_to_string (x ++ decimal_string_of_pos p) t)
+               | expr.TT => fun k => k "()"
+               | expr.AppIdent s d idc args
+                 => fun k
+                    => @to_string'
+                         x s args
+                         (fun v
+                          => k (ident.to_string idc v))
+               | expr.Pair A B a b
```
andres-erbsen: I think we should make it obvious that we do not support pairs for anything other than return values, for which we flatten them.

andres-erbsen:
Well, C syntax is a mess. I don't know what to say, other than whatever this PR is doing is most likely way better than Coq notation hackery :).

src/Experiments/SimplyTypedArithmetic.v
JasonGross: You mean for the return type?  Or in general, have lists with length annotations without any checking on interpretation?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: No, but I can go make a PR to add it

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Good idea

src/Experiments/SimplyTypedArithmetic.v
JasonGross: How about I give it a suffix of `ℤ`, so that it still makes sense to human readers, is easy to grep for, and gives invalid C

src/Experiments/SimplyTypedArithmetic.v

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Note that all lists are of primitives, by the time we get here.  This made a lot of things easier.  So we don't support lists of lists or of pairs.  I've updated the output logic to not have double pointers.

JasonGross:
@davidben I'd like to assign the remainder of this to you.  In particular, I think that I've done most of the bits that are hard for Coq reasons, and basically all that is left is getting the C right.  Are you willing to hack on this from here?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: https://github.com/coq/coq/pull/6744

JasonGross:
Oh, ew, and by the time we're making this decision, we don't even have the relevant information around anymore, because we've already relaxed the bounds to the relevant integer types.  So we have no way of knowing if our `uint32_t` is actually small enough to fit into an `int32_t`.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -6483,6 +6485,971 @@ Module Compilers.
...
+
+      Definition arith_bin_arith_expr_of_PHOAS_ident
+                 (s:=(Compilers.type.Z * Compilers.type.Z)%ctype)
+                 (d:=Compilers.type.Z)
+                 (idc : ident (type.Z * type.Z) type.Z)
+        : ZRange.type.option.interp d -> arith_expr_for s -> option (arith_expr_for d)
+        := fun rout '((e1, r1), (e2, r2))
+           => let r1 := option_map integer_promote_range r1 in
+              let r2 := option_map integer_promote_range r2 in
+              let ct := (r1 <- r1; r2 <- r2; Some (common_type r1 r2))%option in
```
JasonGross: Note to self: this needs to be swapped with the following line if we want things to be correct

Issue #300, opened 2018-01-15
Handle 19 * (x * y + z * w) -> (19 * x) * y + (19 * z) * w optimization reflectively


JasonGross:
See https://github.com/mit-plv/fiat-crypto/pull/295#issuecomment-356892413 for more details

davidben:
Specifically this is valuable when 19 * x and 19 * z still fit in the native word size.

PR #307, opened 2018-02-12
[experiments] Use smaller multiplication for 19*x

JasonGross:
See also #300.

We take the simple path of pulling all of the constant multiplication of
sufficiently small size to the end of any expression like a * b * c * d
* ..., and then we associate all addition and multiplication all the way
to the right.  This might be very much the wrong choice in some cases;
I invite suggestions that don't require knowing any bounds, which
perform better.  (Doing things that require bounds would complicate this
significantly.)

I have no idea why this reassociation results in this part of the diff:
```diff
  expr_let 15 := (uint64)(x_14 >> 51) in
  expr_let 16 := ((uint64)x_14 & 2251799813685247) in
- expr_let 17 := x_4 +₆₄₋₆₄₋₆₄ (19) *₃₂₋₆₄₋₆₄ x_15 in
- expr_let 18 := (uint32)(x_17 >> 51) in
+ expr_let 17 := x_4 +₆₄ x_15 *₆₄₋₆₄₋₆₄ (19) in
+ expr_let 18 := (uint64)(x_17 >> 51) in
  expr_let 19 := ((uint64)x_17 & 2251799813685247) in
- expr_let 20 := x_18 +₃₂₋₆₄₋₆₄ x_7 in
- expr_let 21 := (uint32)(x_20 >> 51) in
+ expr_let 20 := x_18 +₆₄ x_7 in
+ expr_let 21 := (uint64)(x_20 >> 51) in
```

cc @davidben

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -2436,6 +2436,98 @@ Module Compilers.
       := fun var => eliminate_dead (ComputeLive e) (e _).
   End DeadCodeElimination.
 
+  Module ReassociateSmallConstants.
+    Import Compilers.Uncurried.expr.default.
+
+    Section with_var.
+      Context (max_const_val : Z)
+              {var : type -> Type}.
+
+      Fixpoint to_mul_list (e : @expr var type.Z) : list (@expr var type.Z)
+        := match e in expr.expr t return list (@expr var t) with
+           | AppIdent s type.Z ident.Z_runtime_mul (Pair type.Z type.Z x y)
+           | AppIdent s type.Z ident.Z_mul (Pair type.Z type.Z x y)
+             => to_mul_list x ++ to_mul_list y
+           | Var _ _ as e
+           | TT as e
+           | App _ _ _ _ as e
+           | Abs _ _ _ as e
+           | Pair _ _ _ _ as e
+           | AppIdent _ _ _ _ as e
+             => [e]
+           end.
+
+      Fixpoint to_add_mul_list (e : @expr var type.Z) : list (list (@expr var type.Z))
+        := match e in expr.expr t return list (list (@expr var t)) with
+           | AppIdent s type.Z ident.Z_runtime_add (Pair type.Z type.Z x y)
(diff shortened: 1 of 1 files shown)


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3799,231 +3922,183 @@ Module X25519_32.
...
-                     ((2) *₃₂₋₆₄₋₆₄
-                      ((19) *₃₂₋₆₄₋₆₄
-                       (fst @@ x_1 [[7]] *₃₂₋₃₂₋₆₄ snd @@ x_1 [[3]])) +₆₄₋₆₄₋₆₄
-                      ((19) *₃₂₋₆₄₋₆₄
-                       (fst @@ x_1 [[8]] *₃₂₋₃₂₋₆₄ snd @@ x_1 [[2]]) +₆₄₋₆₄₋₆₄
-                       (2) *₃₂₋₆₄₋₆₄
-                       ((19) *₃₂₋₆₄₋₆₄
-                        (fst @@ x_1 [[9]] *₃₂₋₃₂₋₆₄ snd @@ x_1 [[1]])))))))))) in
+(expr_let 2 := fst @@ x_1 [[0]] *₃₂₋₃₂₋₆₄ snd @@ x_1 [[0]] +₆₄
+               (fst @@ x_1 [[1]] *₃₂₋₃₂₋₆₄
```
andres-erbsen: Ideally we would associate all `2` factors with one input field element and all `19` factors with the other, and then use all non-multiplied inputs before all multiplied inputs. (recalling what Tanja said at HACS#1).

andres-erbsen:
The unexplained diff looks vaguely like what we would get if we were multiplying addition outputs instead of multiplication outputs, but I have no idea how it could occur here other than due to a bug.

In general, I don't think we can make a simple optimization that would not significantly hurt performance of any code but still make both x25519-c64 and x25519-c32 use the correct instructions. I think it is okay to merge a slight hack as long as it is documented as such, but in general I think we should be looking int o making the higher-level template include code for appropriately grouping operands for performance rather than trying to rediscover the structure in the compiler.

JasonGross:
Merging tentatively, after chatting in person.  We may backtrack on this or improve it at some point.

PR #308, opened 2018-02-13
Prove another Barrett reduction variant.

davidben:
This variant comes from http://www.ridiculousfish.com/blog/posts/labor-of-division-episode-i.html. It was useful for https://boringssl-review.googlesource.com/#/c/boringssl/+/25887.

Andres suggested it could live in fiat-crypto and that I should make a PR so you all can tell me all the ways I did it inefficiently/non-idiomatically. (This is probably the first `repeat match` I've ever written by myself?) It was also originally a standalone file and I don't know what utilities you have, so I'm sure a lot of the support bits can be replaced.

### DIFF
--- a/src/Arithmetic/BarrettReduction/RidiculousFish.v
+++ b/src/Arithmetic/BarrettReduction/RidiculousFish.v
@@ -1,8 +1,10 @@
 Require Import Crypto.Util.Notations.
-Require Import Coq.NArith.BinNat.
+Require Import Crypto.Util.ZUtil.
+Require Import Crypto.Util.ZUtil.Tactics.DivModToQuotRem.
+Require Import Coq.ZArith.ZArith.
 Require Import Coq.micromega.Lia.
 
-Open Scope N_scope.
+Open Scope Z_scope.
 
 
 (** This file proves an implementation of the variant of Barrett reduction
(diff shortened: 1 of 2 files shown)


src/Arithmetic/BarrettReduction/RidiculousFish.v
```diff
@@ -0,0 +1,449 @@
...
+           (op : N -> N -> N)
+           (a : word aw) (b : word bw) : word (N.max aw bw) :=
+  to_word (op (to_N a) (to_N b)).
+
+Definition word_add {aw bw : N} := word_binop aw bw N.add.
+Definition word_sub {aw bw : N} := word_binop aw bw (sub_wrapped (N.max aw bw)).
+Definition word_mul {aw bw : N} := word_binop aw bw N.mul.
+Definition word_div {aw bw : N} := word_binop aw bw N.div.
+Definition word_shiftl {aw bw : N} := word_binop aw bw N.shiftl.
+Definition word_shiftr {aw bw : N} := word_binop aw bw N.shiftr.
```
davidben: Oh, apparently shifts don't actually behave like `word_binop` and need their own thing. The proof should otherwise still work.

src/Arithmetic/BarrettReduction/RidiculousFish.v
```diff
@@ -0,0 +1,449 @@
...
+    rewrite N.add_0_r.
+    lia. }
+  lia.
+Qed.
+
+(* Peel off one innermost [wrap], and replace it with the expected
+   bounds. Before doing so, introduce a lemma into the context that
+   the value is bounded. This is allows later goals to make use of
+   earlier bounds. *)
+Ltac replace_innermost_to_word expr :=
```
davidben: `replace_innermost_wrap`.

(Remnant of an earlier version where `wrap` was called `to_word` and I didn't have the `word 32` stuff.)

src/Arithmetic/BarrettReduction/RidiculousFish.v
```diff
@@ -0,0 +1,449 @@
+Require Import Coq.NArith.BinNat.
+Require Import Coq.micromega.Lia.
+
+Open Scope N_scope.
+
+
+(* This file proves an implementation of the variant of Barrett
```
JasonGross: Nit: If you start lines with `(**` rather than `(*`, then coqdoc will pick up your comments as text and show them in html/pdf if we ever make the `html` or `pdf` targets.

src/Arithmetic/BarrettReduction/RidiculousFish.v
```diff
@@ -0,0 +1,449 @@
...
+   reduction described in
+   http://www.ridiculousfish.com/blog/posts/labor-of-division-episode-i.html *)
+
+
+(* To simulate C behavior with fixed-width words, we introduce a word
+   data type and associated operations that automatically take
+   modulus. We'll then prove the moduli are all no-ops (save one which
```
JasonGross: plurality mismatch: "take moduli" or "take a modulus"

src/Arithmetic/BarrettReduction/RidiculousFish.v
```diff
@@ -0,0 +1,449 @@
...
+   is a subtraction).
+
+   TODO(davidben): This does not perfectly match C's behavior. In C,
+   operations on types smaller than [int], notably [uint16_t], promote
+   both types to [int] first. Assuming a standard 32-bit [int],
+   multiplying two [uint32_t]s is always defined, but multiplying two
+   [uint16_t]s may not be! However, we never multiply [uint16_t]s, and
+   we would otherwise need an unreasonable large expression to
```
JasonGross: unreasonable -> unreasonably?

src/Arithmetic/BarrettReduction/RidiculousFish.v
```diff
@@ -0,0 +1,449 @@
...
+Definition N_to_u32 := @to_word 32.
+Definition N_to_u64 := @to_word 64.
+
+Delimit Scope word_scope with word.
+Bind Scope word_scope with word.
+Infix "+" := word_add : word_scope.
+Infix "-" := word_sub : word_scope.
+Infix "*" := word_mul : word_scope.
+Infix "/" := word_div : word_scope.
+Infix "<<" := word_shiftl (at level 30, no associativity) : word_scope.
```
JasonGross: Please instead `Require Import Crypto.Util.Notations` at the top, and drop `(at level 30, no associativity)` here; this will ensure that we update precedences globally, because precedences that are different in different files make it impossible to import (perhaps even `Require`) both files at once.

src/Arithmetic/BarrettReduction/RidiculousFish.v
```diff
@@ -0,0 +1,449 @@
...
+  assert (b mod 2^a' < 2^a') by (apply N.mod_upper_bound; apply N.pow_nonzero; lia).
+  assert (2^a' <= 2^a) by (apply N.pow_le_mono_r; lia).
+  lia.
+Qed.
+
+Lemma remove_inner_wrap (a a' b : N) (H : a' >= a) :
+  wrap a (wrap a' b) = wrap a b.
+Proof.
+  unfold wrap.
+  rewrite <- (N.sub_add a a') by lia.
```
JasonGross: If you want, you can write this as
```coq
rewrite <- (N.sub_add a a'),
 N.pow_add_r,
 N.mul_comm,
 N.mod_mul_r,
 N.mul_comm,
 N.mod_add
 by (try apply N.pow_nonzero; lia).
```

Alternatively, you can probably do the entire proof with
```coq
unfold wrap.
pose proof (N.pow_nonzero <relevant arguments>).
pose proof (N.mod_mod <relevant arguments).
lia.
```

src/Arithmetic/BarrettReduction/RidiculousFish.v
```diff
@@ -0,0 +1,449 @@
...
+    lia. }
+  { apply N.ltb_ge in H.
+    rewrite N.min_r by lia.
+    apply remove_outer_wrap.
+    lia. }
+Qed.
+
+(* [lia] and [nia] need some help with exponents, and [cbn] alone
+   often unfolds too much. *)
+Ltac unfold_exponents :=
```
JasonGross: Might want to make this `Local Ltac`, unless you want other files to be able to use it.

andres-erbsen:
`ZUtil/Tactics/DivModToQuotRem.v` that Jason mentions is actually very useful in combination with `nia`, and might be a good reason to port this PR to `Z`. However, that is not a blocker in any sense, and I don't feel like doing the work now.

davidben:
> `ZUtil/Tactics/DivModToQuotRem.v` that Jason mentions is actually very useful in combination with `nia`, and might be a good reason to port this PR to `Z`. However, that is not a blocker in any sense, and I don't feel like doing the work now.

Eh, I can see about doing that with Jason's other comments. Hopefully it's a good exercise both for understanding `Z` vs `N` and that's a good exercise for understanding how to use that file.

andres-erbsen:
@davidben Should we just merge this so it doesn't break, or are you planning to work on it in the immediate future?

davidben:
Made some progress, but haven't quite finished it yet. I'll continue chugging through it tomorrow. Having to add in all my missing `0 <= x` hypotheses is a little annoying, but `Z.div_mod_to_quot_rem` sure is a nice sledgehammer!

davidben:
Alright, ported it over to `Z`. Do you all generally prefer adding a new commit for each iteration or just clobbering the existing one?

src/Arithmetic/BarrettReduction/RidiculousFish.v
davidben: Done.

src/Arithmetic/BarrettReduction/RidiculousFish.v
davidben: Done.

src/Arithmetic/BarrettReduction/RidiculousFish.v
davidben: Done.

src/Arithmetic/BarrettReduction/RidiculousFish.v
davidben: Done.

src/Arithmetic/BarrettReduction/RidiculousFish.v
davidben: Done.

src/Arithmetic/BarrettReduction/RidiculousFish.v
davidben: I don't think `N.mod_mod` works here because of the different modulus. Rewrote this with `Znumtheory.Zmod_div_mod` which is I think a bit easier to follow.

src/Arithmetic/BarrettReduction/RidiculousFish.v
davidben: (Since removed with the port to `Z`. Seems `lia` can handle such constants in `Z` but not `N`.)

src/Arithmetic/BarrettReduction/RidiculousFish.v
davidben: Done.

davidben:
Any other issues remaining with this PR? (No rush if you're busy.)

andres-erbsen:
I think this can be merged, @JasonGross? One potential todo item would be to move generic Z lemmas like division_shrinks into somewhere in ZUtil, but I think either me or Jason doing it would be less work than @davidben doing it.

PR #309, opened 2018-02-13
[experiments] Implement much of the bounds pipeline (no proofs)

JasonGross:
Most of what we talked about today is in here.  I haven't translated @jadephilipoom 's python script to Gallina.  I also haven't assembled things together into a ring.  I'm (almost) quite happy with the tactic that executes the pipeline for a specific choice of parameters / prime:
```coq
Ltac solve_rcarry_mul _ :=
  eapply rcarry_mul_correct;
  lazy; reflexivity.
```
My biggest issue is that the `lazy` runs twice: once during the proof, and again during `Qed`.

Once Jade's script is translated to Gallina, the curve-specific code for a single operation will look like:
```coq
Derive fecarry_mul SuchThat (rcarry_mul_correctT "2^255 - 19" 64) As fecarry_mul_correct.
Proof. Time solve_rcarry_mul (). Time Qed.
```

A large chunk of the code here is to about allowing a more general sort of transformation out of CPS-land.  Previously we were restricted to things of the form `s -> d` where `d` could not have any arrows, and we had to manually correct for arrows in `s`.  Now we can handle anything of the form `x -> y -> ... -> z` where `z` has no arrows and `x`, `y`, etc. are of the form `a -> b -> ... -> d` where `a`, `b`, ..., `d` must not contain arrows.  (Is this any first-order function? second-order?)  We make use of this expressivity; we need to handle the weight function (of type `nat -> Z`) as an argument.  It is not very pretty, and I possibly filled in too much of it by making tyes line up, but everything that it does with continuations actually makes sense to me.  (I don't know how to handle continuations in higher-order things, which is why this isn't more general.)
I did this so we could remove the manual tuple-munging that we were previously doing.

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -4162,12 +4162,12 @@ Qed.
 
 (** XXX TODO: Translate Jade's python script *)
 Section rcarry_mul.
-  Context (limbwidth : Q)
+  Context (n : nat)
           (s : Z)
           (c : list (Z * Z))
           (machine_wordsize : Z).
 
-  Let n := Z.to_nat (Qceiling (Z.log2_up (s - Associational.eval c) / limbwidth)).
+  Let limbwidth := (Z.log2_up (s - Associational.eval c) / Z.of_nat n)%Q.
   Let idxs := (seq 0 n ++ [0; 1])%list%nat.
   Let f_bounds := List.repeat r[0~>(2^Qceiling limbwidth + 2^(Qceiling limbwidth - 3))%Z]%zrange n.
 
@@ -4255,7 +4255,7 @@ Section rcarry_mul.
       [ | clear -Hrv; cbv [check_args] in Hrv; break_innermost_match_hyps; discriminate ].
     erewrite <- carry_mul_gen_correct.
     eapply Pipeline.BoundsPipeline_correct in Hrv'.
-    apply check_args_success_id in Hrv; inversion Hrv; subst.
+    apply check_args_success_id in Hrv; inversion Hrv; subst rv.
     rewrite Hrv'.
     cbv [expr.Interp].
     cbn [expr.interp].
@@ -4350,13 +4350,13 @@ End PrintingNotations.
 
 
 Module X25519_64.
-  Definition limbwidth := 51%Q.
+  Definition n := 5%nat.
   Definition s := 2^255.
   Definition c := [(1, 19)].
   Definition machine_wordsize := 64.
 
   Derive base_51_carry_mul
-         SuchThat (rcarry_mul_correctT limbwidth s c machine_wordsize base_51_carry_mul)
+         SuchThat (rcarry_mul_correctT n s c machine_wordsize base_51_carry_mul)
          As base_51_carry_mul_correct.
   Proof. Time solve_rcarry_mul (). Time Qed.
 
@@ -4428,13 +4428,13 @@ End X25519_64.
 
 (*
 Module X25519_32.
-  Definition limbwidth := (25 + 1/2)%Q.
+  Definition n := 10%nat.
   Definition s := 2^255.
   Definition c := [(1, 19)].
   Definition machine_wordsize := 32.
 
   Derive base_25p5_carry_mul
-         SuchThat (rcarry_mul_correctT limbwidth s c machine_wordsize base_25p5_carry_mul)
+         SuchThat (rcarry_mul_correctT n s c machine_wordsize base_25p5_carry_mul)
          As base_25p5_carry_mul_correct.
   Proof. Time solve_rcarry_mul (). Time Qed.
 
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -2206,25 +2206,35 @@ Module Compilers.
         : @Compilers.Uncurried.expr.default.Expr R
         := expr.CallWithContinuation (@ident.untranslate _) (@ident.fst) (@ident.snd) (@ident.bool_rect) e k.
 
-      Module type_descr.
+      (** It's not clear how to "plug in the identity continuation"
+          for the CPS'd form of an expression of type [((A -> B) -> C)
+          -> D].  So we must describe types of at most second order
+          functions, so that we can write a uniform "plug in the
+          identity continuation" transformation. *)
+      Module second_order.
         Import Compilers.type.
-        Inductive flat_type :=
-        | type_primitive (_ : primitive)
-        | prod (_ : flat_type) (_ : flat_type)
-        | list (_ : flat_type).
-        Inductive argtype :=
-        | flat_arg (_ : flat_type)
-        | arrow_arg (s : flat_type) (d : argtype)
-        | prod_arg (_ _ : argtype).
-        Inductive type :=
-        | flat (_ : flat_type)
-        | arrow (s : argtype) (d : type).
+        Module Import type.
+          Inductive flat_type :=
+          | type_primitive (_ : primitive)
+          | prod (_ : flat_type) (_ : flat_type)
+          | list (_ : flat_type).
+          Inductive argtype :=
+          | flat_arg (_ : flat_type)
+          | arrow_arg (s : flat_type) (d : argtype)
+          | prod_arg (_ _ : argtype).
+          Inductive type :=
+          | flat (_ : flat_type)
+          | arrow (s : argtype) (d : type).
+        End type.
 
         Module Export Coercions.
           Coercion type_primitive : primitive >-> flat_type.
           Coercion flat_arg : flat_type >-> argtype.
           Coercion flat : flat_type >-> type.
         End Coercions.
+        Notation flat_type := flat_type.
+        Notation argtype := argtype.
+        Notation type := type.
 
         Fixpoint flat_to_type (t : flat_type) : Compilers.type.type
           := match t with
@@ -2254,7 +2264,7 @@ Module Compilers.
                   | Some A, Some B => @Some flat_type (prod A B)
                   | _, _ => None
                   end
-             | type.arrow s d => None
+             | Compilers.type.arrow s d => None
              | Compilers.type.list A
                => option_map list (flat_of_type A)
              end.
@@ -2269,7 +2279,7 @@ Module Compilers.
                     => @Some argtype (prod_arg A B)
                   | _, _, _ => None
                   end
-             | type.arrow s d
+             | Compilers.type.arrow s d
                => match flat_of_type s, arg_of_type d with
                   | Some s, Some d => Some (arrow_arg s d)
                   | _, _ => None
@@ -2281,7 +2291,7 @@ Module Compilers.
 
         Fixpoint of_type (t : Compilers.type.type) : option type
           := match t with
-             | type.arrow s d
+             | Compilers.type.arrow s d
                => match arg_of_type s, of_type d with
                   | Some s, Some d => Some (arrow s d)
                   | _, _ => None
@@ -2308,7 +2318,7 @@ Module Compilers.
                           end
                      | None => None
                      end
-             | type.arrow s d => fun _ => None
+             | Compilers.type.arrow s d => fun _ => None
              | Compilers.type.list A
                => fun v
                   => option_map
@@ -2334,7 +2344,7 @@ Module Compilers.
                           | None => None
                           end
                      end
-             | type.arrow s d
+             | Compilers.type.arrow s d
                => fun v
                   => match try_transport_flat_of_type (fun s => P (s -> _)%ctype) s v with
                      | Some (existT s v)
@@ -2356,7 +2366,7 @@ Module Compilers.
         Fixpoint try_transport_of_type P (t : Compilers.type.type)
           : P t -> option { t' : _ & P (to_type t') }
           := match t with
-             | type.arrow s d
+             | Compilers.type.arrow s d
                => fun v
                   => match try_transport_arg_of_type (fun s => P (s -> _)%ctype) s v with
                      | Some (existT s v)
@@ -2375,22 +2385,22 @@ Module Compilers.
                        (fun '(existT t v) => existT (fun t => P (to_type t)) (flat t) v)
                        (try_transport_flat_of_type P t v)
              end.
-      End type_descr.
-      Import type_descr.Coercions.
+      End second_order.
+      Import second_order.Coercions.
 
       Fixpoint untranslate_translate_flat
                (P : Compilers.type.type -> Type)
                {R}
-               {t : type_descr.flat_type}
-               (e : P (type_descr.to_type t))
+               {t : second_order.flat_type}
+               (e : P (second_order.to_type t))
                {struct t}
-        : P (type.untranslate R (type.translate (type_descr.to_type t)))
-        := match t return P (type_descr.to_type t)
-                          -> P (type.untranslate R (type.translate (type_descr.to_type t)))
+        : P (type.untranslate R (type.translate (second_order.to_type t)))
+        := match t return P (second_order.to_type t)
+                          -> P (type.untranslate R (type.translate (second_order.to_type t)))
            with
-           | type_descr.type_primitive x => id
-           | type_descr.prod A B
-             => fun ab : P (type_descr.flat_to_type A * type_descr.flat_to_type B)%ctype
+           | second_order.type.type_primitive x => id
+           | second_order.type.prod A B
+             => fun ab : P (second_order.flat_to_type A * second_order.flat_to_type B)%ctype
                 => @untranslate_translate_flat
                      (fun a => P (a * _)%ctype)
                      R A
@@ -2398,7 +2408,7 @@ Module Compilers.
                         (fun b => P (_ * b)%ctype)
                         R B
                         ab)
-           | type_descr.list A
+           | second_order.type.list A
              => @untranslate_translate_flat
                   (fun t => P (Compilers.type.list t))
                   R A
@@ -2407,17 +2417,17 @@ Module Compilers.
       Fixpoint untranslate_translate_flat'
                (P : Compilers.type.type -> Type)
                {R}
-               {t : type_descr.flat_type}
-               (e : P (type.untranslate R (type.translate (type_descr.to_type t))))
+               {t : second_order.flat_type}
+               (e : P (type.untranslate R (type.translate (second_order.to_type t))))
                {struct t}
-        : P (type_descr.to_type t)
-        := match t return P (type.untranslate R (type.translate (type_descr.to_type t)))
-                          -> P (type_descr.to_type t)
+        : P (second_order.to_type t)
+        := match t return P (type.untranslate R (type.translate (second_order.to_type t)))
+                          -> P (second_order.to_type t)
            with
-           | type_descr.type_primitive x => id
-           | type_descr.prod A B
+           | second_order.type.type_primitive x => id
+           | second_order.type.prod A B
              => fun ab :
-                      (* ignore this line *) P (type.untranslate R (type.translate (type_descr.flat_to_type A)) * type.untranslate R (type.translate (type_descr.flat_to_type B)))%ctype
+                      (* ignore this line *) P (type.untranslate R (type.translate (second_order.flat_to_type A)) * type.untranslate R (type.translate (second_order.flat_to_type B)))%ctype
                 => @untranslate_translate_flat'
                      (fun a => P (a * _)%ctype)
                      R A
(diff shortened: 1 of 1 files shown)


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -276,6 +278,44 @@ Module Positional. Section Positional.
...
+                            (Hw_div_nz : forall i : nat, weight (S i) / weight i <> 0),
+                        (eval n carry_mulmod) mod (s - Associational.eval c)
+                        = (eval n f * eval n g) mod (s - Associational.eval c))
+           As carry_mulmod_correct.
+    Proof.
+      intros.
+      erewrite <-eval_mulmod with (s:=s) (c:=c)
+        by (subst; try assumption; try reflexivity).
+      etransitivity;
+        [ | rewrite <- @eval_chained_carries with (s:=s) (c:=c) (idxs:=idxs) (modulo:=fun x y => Z.modulo x y) (div:=fun x y => Z.div x y)
```
andres-erbsen: `expand_list` is an artifact of how our partial evaluation pipeline works... if there was a way to have the pipeline automatically work as if all known-length lists were expanded, that would be even nicer.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1467,6 +1528,59 @@ Module Compilers.
                => (untranslate R t -> R)%ctype
              | type.list A => Compilers.type.list (untranslate R A)
              end%cpstype.
+        Fixpoint try_untranslate (t : type)
```
andres-erbsen: What is this fucntion for?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1878,23 +1992,332 @@ Module Compilers.
...
-              e
-              (fun _ => id))
-             (only parsing).
+                 {t : type_descr.type}
+                 (e : Expr (type.translate (type_descr.to_type t)))
+        : @Compilers.Uncurried.expr.default.Expr (type_descr.to_type t)
+        := fun var => @call_fun_with_id_continuation' _ t (e _).
+
+      Definition CallFunWithIdContinuation
+                 {t}
```
andres-erbsen: Could we require the input expression to have type of `translate t` here, and would doing so avoid the transport?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3576,67 +4024,140 @@ Derive carry_mul_gen
...
+  Inductive ErrorMessage :=
+  | Computed_bounds_are_not_tight_enough
+      {t} (computed_bounds expected_bounds : BoundsAnalysis.Indexed.Range.range t)
+  | Bounds_analysis_failed
+  | Return_type_mismatch {T'} (found expected : T')
+  | Value_not_le (descr : string) {T'} (lhs rhs : T')
+  | Values_not_provably_distinct (descr : string) {T'} (lhs rhs : T')
+  | Values_not_provably_equal (descr : string) {T'} (lhs rhs : T').
+
+  Inductive ErrorT {T} :=
```
andres-erbsen: I like this type; mostly based on good experiences with <https://doc.rust-lang.org/std/result/>

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3576,67 +4024,140 @@ Derive carry_mul_gen
...
+
+  Definition BoundsPipeline
+             relax_zrange
+             {s d}
+             arg_bounds
+             out_bounds
+             (E : Expr (s -> d))
+  : ErrorT (BoundsAnalysis.Indexed.expr.Notations.expr (BoundsAnalysis.AdjustBounds.ident.type_for_range relax_zrange (t:=BoundsAnalysis.Indexed.OfPHOAS.type.compile d) out_bounds))
+    := let E := PartialReduce E in
+       let E := DeadCodeElimination.EliminateDead E in
```
andres-erbsen: perhaps dead code elimination should belong with partial reduction? it is currently just a hack to clean up after it

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3576,67 +4024,140 @@ Derive carry_mul_gen
...
+        (BoundsAnalysis.OfPHOAS.Interp
+           (s:=s)
+           (d:=d)
+           relax_zrange
+           arg_bounds
+           (bs:=out_bounds)
+           arg
+           rv
+         = Some res)
+        /\ res = Interp E arg'.
```
andres-erbsen: substitute this equality, eliminate the existential?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3576,67 +4024,140 @@ Derive carry_mul_gen
...
+  Proof.
+  Admitted.
+End Pipeline.
```
andres-erbsen: This pipeline interface looks good to me, we will figure out something for notations later if these become a pain.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3576,67 +4024,140 @@ Derive carry_mul_gen
...
+
+Record mul_rargs :=
```
andres-erbsen: What would go wrong if we removed this record and passed individual arguments every time we currently used this record?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3654,6 +4175,184 @@ Definition relax_zrange_gen (possible_values : list Z) : zrange -> option zrange
                          (round_up_bitwidth_gen possible_values (Z.log2_up (u+1)))
          else None)%zrange.
 
+(** TODO: Move me? *)
+Definition SmartApp {s d} (f : Expr (s -> d)) (x : Expr s) : Expr d
```
andres-erbsen: `SmartApp` -> `app_and_maybe_cancel`
`SmartFst` -> `fst_and_maybe_cancel`
`SmartSnd` -> `snd_and_maybe_cancel`

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3576,67 +4024,140 @@ Derive carry_mul_gen
...
+  pose (ordering := LHS).
   etransitivity.
   Focus 2.
-  { subst fg'.
-    repeat match goal with H : _ |- _ => clear H end; revert args'.
-    lazymatch goal with
-    | [ |- forall args, ?ev = @?RHS args ]
-      => refine (fun args => f_equal (fun F => F args) (_ : _ = RHS))
-    end.
+  { repeat match goal with
```
andres-erbsen: Why do we need more than a couple of lines of code to go from mul to reified mul, or from weight to reified weight (not counting Q unfolding)?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3654,6 +4175,184 @@ Definition relax_zrange_gen (possible_values : list Z) : zrange -> option zrange
...
+  Context (limbwidth : Q)
+          (s : Z)
+          (c : list (Z * Z))
+          (machine_wordsize : Z).
+
+  Let n := Z.to_nat (Qceiling (Z.log2_up (s - Associational.eval c) / limbwidth)).
+  Let idxs := (seq 0 n ++ [0; 1])%list%nat.
+  Let f_bounds := List.repeat r[0~>(2^Qceiling limbwidth + 2^(Qceiling limbwidth - 3))%Z]%zrange n.
+
+  Definition make_carry_mul_rargs
```
andres-erbsen: What if we defined an up-arrow notation for `GallinaReify.reify` that did the necessary type inference in Ltac?

andres-erbsen: Not blocking in any sense

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3654,6 +4175,184 @@ Definition relax_zrange_gen (possible_values : list Z) : zrange -> option zrange
...
+          rn var := GallinaReify.reify (t:=type.nat) n;
+          ridxs var := GallinaReify.reify (t:=type.list type.nat) idxs;
+          rlen_c var := GallinaReify.reify (t:=type.nat) (List.length c);
+          rlen_idxs var := GallinaReify.reify (t:=type.nat) (List.length idxs);
+          relax_zrange := relax_zrange_gen [machine_wordsize; 2 * machine_wordsize]%Z;
+          arg_bounds := (f_bounds, f_bounds);
+          out_bounds := f_bounds;
+       |}.
+
+  Definition rcarry_mul
```
andres-erbsen: Looks good for now, later we will probably want do something about precondition checks.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3654,6 +4175,184 @@ Definition relax_zrange_gen (possible_values : list Z) : zrange -> option zrange
...
+              (Hg : List.length (snd arg) = n),
+      exists res,
+        (BoundsAnalysis.OfPHOAS.Interp
+           (opts.(relax_zrange))
+           (opts.(arg_bounds))
+           (bs:=opts.(out_bounds))
+           arg
+           rv
+         = Some res)
+        /\ res = carry_mulmod (Interp opts.(rw)) s c n (Interp opts.(rlen_c)) idxs (Interp opts.(rlen_idxs)) arg'.
```
andres-erbsen: substitute this equality and remove existential? (as above)

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3654,6 +4175,184 @@ Definition relax_zrange_gen (possible_values : list Z) : zrange -> option zrange
...
+
+  Lemma rcarry_mul_correct
```
andres-erbsen: Didn't we say this proof would be ~two lines? what happened?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Is it currently unused? It's the more simply typed version of try_transport_untranslate or whatever which is used to invert translate so that when we get a cpsexpr, we don't have to manually fill in the type that it came from for the transformations out of cps which require their input to have come from the translation of a PHOAS type, if I recall correctly.  ( Alternatively, I wrote this to make the types line up with what I needed, or with problems that type inference is capable of solving.)

src/Experiments/SimplyTypedArithmetic.v
JasonGross: I am concerned that it would break type inference and mean we need to pass `t` by hand, but, yes, it would eliminate the need for the transport.

andres-erbsen:
then do it, type inference is not a good constraint to drive design from

andres-erbsen:
I think it's currently unused, and there should be no need for it as i explained below.

andres-erbsen:
Perhaps list lengths could be handled the same way as integer ranges?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Let me answer a slightly different question, and see if that addresses your concern.

I think this is the wrong interface, but I think it is close.  I think we want the Gallina equivalent of the python script to generate a record.  It should do all of the default computing.  I think it is probably fine to chain reification methods with projection from that record.  However, the abstraction barriers are not entirely clear to me.  Here are the pieces that I see: parse the prime string to a sorted list of taps; from the taps and the word size, compute the possible limb widths and numbers of limbs and carry chains; from the limb width and the number of limbs, compute the input and output relaxed and tight bounds; from the limb width, compute the reified weight function; reify all the things that need to be reified.  It's unclear to be where to draw the lines, but I'd like to share as much as possible/reasonable between femul, feadd, etc.  What do you think?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Because `Reify_rhs` only handles reification of closed terms / 0-argument functions.  The type signature of the derivation imposes an ordering on the arguments of the reified term, and the extra code is to communicate that information to the reification routine in a way it can handle.

For rweight, there is the additional complication that we need a function out of Q, but can only handle Z, so we need to pull tricks to handle the Q argument.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: I inlined the proof that the checks we have are sufficient to prove that the default values we provide for the curve specific arguments are sufficiently nice.  This part should probably be factored out?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: > Perhaps list lengths could be handled the same way as integer ranges?

Urg.  Integer ranges require already a decent amount of boilerplate, and that's after we've specialized to only handling things of shape `s -> d` with no arrows in `d`.  We'd need to do list expansion much, much earlier for this to work.  I could alternatively write a tactic which looks for context hypotheses of the form `length ls = n` and rewrites with list-expansion, and run this tactic right before reification in the caching derive?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Removed by rebasing this on top of #311

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Should we factor it out into a separate `Util` file?

src/Experiments/SimplyTypedArithmetic.v

src/Experiments/SimplyTypedArithmetic.v
JasonGross: I don't see `SmartFst` or `SmartSnd` anywhere, but I've done the `SmartApp` renaming

src/Experiments/SimplyTypedArithmetic.v

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Defined as `GallinaReify.AutoReify`

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: Okay, the tactic solution sounds reasonable.

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: I do think we should share things between femul and feadd, but I don't think a record is necessary for doing so. feadd and femul will still take different parameters, just some of them will be computed in the same way for both funcitons.

andres-erbsen:
It seems that we actually do not know how to get rid of the transports, so let's live with them for now. I think my comment about removing the parameters record is the only remaining todo on this one.

JasonGross:
Record removed.  However, this is not yet ready for merging.  For reasons I'm not entirely clear on, this PR makes the pipeline slower.  The final derive went from about 5 s to about 15 s (in 64-bit 25519), and from about 45 s to about 185 seconds.  My current suspicion is that there's something unfortunate going on with how we're plugging in arguments...

JasonGross:
Performance restored.  I am mystified, but this is ready for merge if you are happy with it @andres-erbsen .

Details:
Apparently using `refine eq_refl` rather than `subst <name for evar>; reflexivity` was resulting in βδ reduction of `chained_carries` in `carry_mulmod`.  The β reduction resulted in us getting a different
cps'd term.

I do not know why this particular β reduction resulted in a 3x slowdown in partial reduction; it seems like anything that cared about sharing should either get sharing from the top-level in `carry_mulmod`, or should have no difference in sharing between the terms
```coq
(fun n s c p idxs
  => fold_right (fun a b => @carry_reduce n s c a b) p (rev idxs))
 n s c p idxs
```
and
```coq
fold_right (fun a b => @carry_reduce n s c a b) p (rev idxs))
```

This feels fragile, and I am mystified.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Record removed

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1496,6 +1549,56 @@ Module Compilers.
...
+                => (ident.primitive x @@ TT)%expr
+           end.
+    End reify.
+
+    Definition Reify {t : type} (v : forall var, value var t) : Expr t
+      := fun var => reify (v _).
+
+    (** [AutoReify] does Ltac type inference to get the type *)
+    Notation AutoReify v
+      := (Reify (t:=type.reify_type_of v) (fun _ => v)) (only parsing).
```
andres-erbsen: consider `Reify` -> `Reify_as` and `AutoReify` -> `Reify`, or `AutoReify` -> `infer_type_and_Reify`

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2103,23 +2206,328 @@ Module Compilers.
         : @Compilers.Uncurried.expr.default.Expr R
         := expr.CallWithContinuation (@ident.untranslate _) (@ident.fst) (@ident.snd) (@ident.bool_rect) e k.
 
+      Module type_descr.
```
andres-erbsen: perhaps `secondorder.type`, with a comment about why we are doing this?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3579,88 +3967,117 @@ End test6.
...
-  subst args' args; cbn [fst snd].
-  subst fg'.
+  let LHS := match goal with |- ?LHS = _ => LHS end in
+  lazymatch LHS with
+  | context LHS[expr.Interp _ _]
+    => let LHS := context LHS[Interp E''] in
+       transitivity LHS
+  end;
+    [ clear E | exact admit ].
+  subst carry_mul_gen.
```
andres-erbsen: There is nothing operation-specific here, correct? In the future we will probably want to factor out this tactic.

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: Let's do that if we end up having nontrivial amounts of code for this general pattern?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -278,6 +278,41 @@ Module Positional. Section Positional.
...
+  Section carry_mulmod.
+    Context (m:Z) (s:Z)
+            (c:list (Z*Z))
+            (n : nat)
+            (len_c : nat)
+            (idxs : list nat)
+            (len_idxs : nat)
+            (fg : list Z * list Z).
+
+    Derive carry_mulmod
```
andres-erbsen: After addressing my comments below, it might be a good idea to add a test that runs the entire pipeline in one go, without caching the generic code using a separate derive. This is how we will want to use the pipeline when prototyping arithmetic code or figuring out what arguments we want to pass to it. More generally, while caching is totally justified here, I would prefer if we did most of the design with the direct interface in mind and treated caching as a nice extra.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Correct.  Once this PR is merged, I'll open a new PR for my work on getting a `ring` theorem, which adds the other operations and imposes constraints on what the correctness conditions can look like.  I do factor this bit into a separate tactic there.  I expect we'll have to go back and forth a few times to get that into a satisfactory state, and that's the place to discuss how generic this is.

src/Experiments/SimplyTypedArithmetic.v

src/Experiments/SimplyTypedArithmetic.v

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3678,6 +4095,204 @@ Definition relax_zrange_gen (possible_values : list Z) : zrange -> option zrange
...
+              (Hg : List.length (snd arg) = n),
+      BoundsAnalysis.OfPHOAS.Interp
+        (relax_zrange)
+        (arg_bounds)
+        (bs:=out_bounds)
+        arg
+        rv
+      = Some (carry_mulmod (Interp rw) s c n (Interp rlen_c) idxs (Interp rlen_idxs) arg').
+
+  Lemma rcarry_mul_correct
```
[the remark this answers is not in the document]
JasonGross: Generalize, yes.  Nice interface, maybe.  Let's punt this to the PR I'll make about the ring goal.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3678,6 +4095,204 @@ Definition relax_zrange_gen (possible_values : list Z) : zrange -> option zrange
...
+  Let ridxs := GallinaReify.AutoReify idxs.
+  Let rlen_c := GallinaReify.AutoReify (List.length c).
+  Let rlen_idxs := GallinaReify.AutoReify (List.length idxs).
+  Let relax_zrange := relax_zrange_of_machine_wordsize.
+  Let arg_bounds : BoundsAnalysis.Indexed.Range.range (BoundsAnalysis.Indexed.OfPHOAS.type.compile (type.list type.Z * type.list type.Z))
+    := (f_bounds, f_bounds).
+  Let out_bounds : BoundsAnalysis.Indexed.Range.range (BoundsAnalysis.Indexed.OfPHOAS.type.compile (type.list type.Z))
+    := f_bounds.
+
+  Definition check_args {T} (res : Pipeline.ErrorT T)
```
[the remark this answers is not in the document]
JasonGross: Yes, in the PR for the ring goal, there will be a separate computation to sanity check the curve, which will be required to prove the ring goal, but is unused in the derivations themselves.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3678,6 +4095,204 @@ Definition relax_zrange_gen (possible_values : list Z) : zrange -> option zrange
...
+  Let n := Z.to_nat (Qceiling (Z.log2_up (s - Associational.eval c) / limbwidth)).
+  Let idxs := (seq 0 n ++ [0; 1])%list%nat.
+  Let f_bounds := List.repeat r[0~>(2^Qceiling limbwidth + 2^(Qceiling limbwidth - 3))%Z]%zrange n.
+
+  Definition relax_zrange_of_machine_wordsize
+    := relax_zrange_gen [machine_wordsize; 2 * machine_wordsize]%Z.
+  Local Arguments relax_zrange_of_machine_wordsize / .
+
+  Let rw := rweight limbwidth.
+  Let rs := GallinaReify.AutoReify s.
```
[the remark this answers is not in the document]
JasonGross: Meh.  I could make an `@` for composing `Expr`.  Let's punt this for the ring PR, where we actually use this multiple times in the section.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3678,6 +4095,204 @@ Definition relax_zrange_gen (possible_values : list Z) : zrange -> option zrange
...
+Qed.
+
+(** XXX TODO: Translate Jade's python script *)
+Section rcarry_mul.
+  Context (limbwidth : Q)
+          (s : Z)
+          (c : list (Z * Z))
+          (machine_wordsize : Z).
+
+  Let n := Z.to_nat (Qceiling (Z.log2_up (s - Associational.eval c) / limbwidth)).
```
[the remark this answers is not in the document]
JasonGross: Is `Qceiling` unneeded here?  I can do
```coq
Let limbwidth := (Z.log2_up (s - Associational.eval c) / Z.of_nat n)%Q.
```
if that's what you're asking for?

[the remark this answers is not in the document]

src/Experiments/SimplyTypedArithmetic.v
JasonGross: I'd like to punt this to the ring-goal PR.  In particular, I still have uncertainties about the just-below-top-level interface to be designing for, and I think questions about caching (or not caching) make more sense once that's resolved, and I think it's easier (at least for me) to resolve that when we have the way in which the interface needs to fit together into a ring goal in mind.

(For example, it seems plausible that we'll have a "do the caching derive" tactic and that the solution here is actually to write a thing that chains it with the higher-level goal; I can mostly-generally chain them with a `cut`.  I suspect this runs afoul of something, but I'm not sure what.)

(I also don't want to keep rebasing the ring-goal work on top of changes to this PR in places that conflict with changes that I've already made in the ring-goal branch.)

PR #310, opened 2018-02-16
[WIP] Montgomery Reduction in non-CPS pipeline

jadephilipoom:
What's done:
- Support for reifying additional operations (`Z.leb`, `Z.mul_split`, `Z.add_get_carry`, `Z.add_with_get_carry`, `Z.sub_get_borrow`, `Z.zselect`, and `Z.add_modulo`)
- Non-CPS versions of `Arithmetic/Saturated/Core.v` and `Arithmetic/Saturated/MulSplit.v`
- A preliminary testing version of Montgomery reduction for p256, which assumes we have a 256x256=>256x256 multiplication instruction. This version has been pushed through the parts of the pipeline that were in master at the beginning of the week (namely, no bounds analysis yet).

What still needs doing:
- [x] Currently, when I use `dlet` to assign the result of a two-output operation like `Z.add_get_carry`, it's actually inlined during reification--I think because `(Z * Z)` is not a primitive type. Figure out what to do for this and fix it. (@JasonGross Is just adding `(Z * Z)` to the primitive set  the correct approach?)
- [x] Get (the testing version of) Montgomery reduction through the bounds-analysis part
- [x] Make (the testing version of) Montgomery reduction use the saturated arithmetic operations, resulting in marginally more reasonable widths of multiplications
- [ ] See if I can get the code I want out of the word-by-word Montgomery instead of the hand-coded testing version I am messing with


src/Util/ZUtil.v
```diff
@@ -964,6 +964,18 @@ Module Z.
   Qed.
   Hint Resolve div_sub_mod_cond : zarith.
 
+  Lemma div_add_mod_cond_l : forall x y d, d <> 0 -> (x + y) / d = (x mod d + y) / d + x / d.
```
JasonGross: Could you instead pick (or make) a file in `ZUtil/` which it makes sense to put these in?  (`Div.v` or `DivAdd.v`, for example)  I'd rather make `ZUtil.v` smaller over time, since it's currently a bit slow and a bit of a block to compilation.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2120,6 +2522,7 @@ Module Compilers.
     Module ident.
       Section interp.
         Context {var : type -> Type}.
+        Eval compute in (fun a b => value var (type.Z * type.Z)).
```
JasonGross: This line should disappear

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1659,11 +2032,18 @@ Module Compilers.
...
+                  | ident.Z_add_get_carry as idc
+                  | ident.Z_add_with_get_carry as idc
+                  | ident.Z_sub_get_borrow as idc
+                  | ident.Z_zselect as idc
+                  | ident.Z_add_modulo as idc
                     => cps_of (Uncurried.expr.default.ident.interp idc)
                   | ident.Let_In tx tC
                     => fun '((x, f) : (interp R (type.translate tx)
-                                       * (interp R (type.translate tx) * (interp R (type.translate tC) -> R) -> R)))
+                                       * (interp R (type.translate tx) * (interp R (type.translate tC) -> R) -> R )))
```
JasonGross: Useless extra space?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -278,6 +298,283 @@ Module Positional. Section Positional.
...
+
+Module Columns.
+  Section Columns.
+    Context (weight : nat->Z)
+            {weight_0 : weight 0%nat = 1}
+            {weight_nonzero : forall i, weight i <> 0}
+            {weight_positive : forall i, weight i > 0}
+            {weight_multiples : forall i, weight (S i) mod weight i = 0}
+            {weight_divides : forall i : nat, weight (S i) / weight i > 0}.
+    (* This div and modulo will be instantiated with versions that use bitshifts if the denominator/modulus is a power of 2 *)
```
andres-erbsen: I think we can do this reflectively now. @JasonGross ?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -278,6 +298,283 @@ Module Positional. Section Positional.
...
+    Proof.
+      cbv [eval]; intros; subst. rewrite map_app. simpl map.
+      apply Positional.eval_snoc; distr_length.
+    Qed.
+
+    Section compact_digit.
+      Context (n : nat).
+
+      (* Outputs (sum, carry) *)
+      Fixpoint compact_digit (digit : list Z) : (Z * Z) :=
```
andres-erbsen: How about calling this function `flatten_column` or something with that makes it clear that the list passed in is a single column?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -278,6 +298,283 @@ Module Positional. Section Positional.
...
+        match digit with
```
andres-erbsen: How about passing in `bound := (weight (S n) / weight n)` instead of `n`?

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: Hmm, `flatten` seems more intuitive to me now but then we should probably rename all `compact` functions to `flatten`, so idk if worth it.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Yes, I do this already in the `ident.Z_div` case of partial reduction (in `partial.ident.interp`).

src/Experiments/SimplyTypedArithmetic.v
jadephilipoom: Now is a good time for renaming things, if we're going to replace the old code. So I think I will change the name here.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2897,7 +3313,14 @@ Module Compilers.
                | ident.Z_div
                | ident.Z_modulo
                | ident.Z_eqb
+               | ident.Z_leb
                | ident.Z_of_nat
+               | ident.Z.mul_split (* TODO: is this the right thing? *)
```
JasonGross: No, this is not the right thing.  `None` means we expect the operation to be fully evaluated by partial reduction before we hit bound analysis.  I expect that instead of this, you want to follow the path of compiling `Z.modulo` to `runtime_land` (or `nth_default` to `nth_default_concrete`, or `Z.div` to `runtime_shiftr`), where, in partial reduction, you do something special if the bitwidth / splitting location is a concrete `Z`, and then you pipe these operations through the bounds pipeline.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -4120,3 +4543,200 @@ Module X25519_32.
...
+
+Require Import Crypto.Arithmetic.MontgomeryReduction.Definition.
+Require Import Crypto.Arithmetic.MontgomeryReduction.Proofs.
+Require Import Crypto.Util.ZUtil.EquivModulo.
+Require Import Crypto.Util.ZUtil.Modulo Crypto.Util.ZUtil.Tactics.
+Require Import Crypto.Util.ZUtil.Zselect Crypto.Util.ZUtil.AddModulo.
+
+Module Montgomery256.
+
+  (* written in a way that is easier to reify *)
```
JasonGross: What makes this version easier to reify? / Why not extend reification to handle whatever version is above?

jadephilipoom: This version doesn't use the Saturated part yet, just extremely large multiplications. It seemed prudent to work with a minimal and controllable example like this at first, until I get that working and understand the pipeline a bit better. But don't worry, I intend to eventually use a suitable generic version.

src/Experiments/SimplyTypedArithmetic.v
jadephilipoom: Hmm. I'm not sure I have a clear idea how to do this, but I'll give it a try following the examples you mentioned and bother you if I can't figure it out.

JasonGross:
I'd like to get this merged sooner rather than later.  I'm going to be rewriting bounds analysis to happen as part of partial reduction in #315, and I think things will be nicer if I do it on top of this PR rather than underneath it, so that you don't have to adapt to the changing pipeline too much.  I am happy with reflective parts of this PR.

@andres-erbsen Do you have thoughts on this PR?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -4653,3 +5254,305 @@ Module X25519_32.
...
+End MontgomeryReduction.
+
+Ltac solve_rmontred _ :=
+  eapply MontgomeryReduction.rmontred_correct;
+  lazy; reflexivity.
+
+Module Montgomery256.
+
+  Definition N := (2^256-2^224+2^192+2^96-1).
+  Definition N':= (115792089210356248768974548684794254293921932838497980611635986753331132366849).
```
JasonGross: This number is magic, and I thought we had a way of computing it?

jadephilipoom: We do, and I will once I start using a general template for Montgomery rather than my ad-hoc testing version.

jadephilipoom:
I'd like to push some more changes before it's merged--I've now mostly completed the third item on my checklist (using `Columns` and smaller multiplications). I should have it cleaned up and pushed before the end of today; if not, we can merge without it.

jadephilipoom:
Okay, finished with the changes I want to insert before Jason's refactor. It's ready for another final sanity check, and then I'm happy to merge it. Regarding dead code elimination, I wanted it now but am guessing it's already in the plans for the pipeline anyway. I'm happy for the dead code elimination in this commit to either be replaced or integrated with the other passes.

The main things remaining to do (mostly a note to myself, none of these are blocking for anyone else afaik) are:
 - Get what I've done here to come out of a generic template (like the WBW Montgomery in `Arithmetic/`) rather than my ad-hoc example
- Make the `flatten` code generate `add_with_get_carry` instructions when the previous column had only two elements
- Generalize the convert-multiply-convert multiplication strategy I needed for my example and prove it correct (see TODO)

andres-erbsen:
We actually removed dead code elimination from the main pipeline once we figured out what caused dead code to be generated in cps conversion of if statements that partial evaluation gets rid of. Do you think you will need dead code elimination even after these fixes?

Do you want me/Jason to review the Montgomery reduction code now, or would it be better if we let it be until you are done generalizing it?

jadephilipoom:
I added the dead code elimination because `flatten` produces some dead code (for instance, it might accumulate carries for the last column, only to have them discarded by the caller who knows from input preconditions that the output fits in the columns allocated). It also just seems like a generally good thing to have--is there a reason we shouldn't have it?

The Montgomery reduction code is not really useful to review right now, although I don't mind your passing through if you're curious.

JasonGross:
The dead code elimination you have seems quadratic; I'll revive the old DCE code, and we can make sure that partial reduction handle inlining the "silly" lets.  I'm going to merge this so I can get started.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -4653,3 +5363,520 @@ Module X25519_32.
...
+Local Open Scope nexpr_scope.
+
+Print Montgomery256.montred256.
+(*
+c.Mul128x128($r2, Lower128 @@ $r1_lo, RegPinv >> 128) << 128;
+c.Mul128x128($r3, ($r1_lo >> 128), Lower128{RegPinv}) << 128;
+c.Mul128x128($r8, Lower128 @@ $r1_lo, Lower128{RegPinv});
+c.Add256($r9, $r2, $r3);
+c.Add256($r10, $r8, $r9_lo);
+(c.Mul128x128($r20, Lower128 @@ $r10_lo, RegMod << 128) << 128;
```
JasonGross: @jadephilipoom Do you actually want these parentheses?  If not, you can put all of your notations that end with `; f` `at level 200, f at level 200` to get rid of the parentheses around blocks of code.

PR #311, opened 2018-02-16
Fix CPSify of bool_rect to eliminate dead code

JasonGross:
The proper cps transform of
```
  if b then (let x := y in z) else w
```
is
```
  cps b (fun b' =>
  if b'
  then cps y (fun y'
       => let x := y' in cps z k)
  else cps w k
```
not
```
  cps b (fun b' =>
  cps y (fun y' =>
  let x := y' in
  cps z (fun z' =>
  cps w (fun w' =>
  if b' then k z' else k w')
```

So we fix the cps transform, and remove the now useless dead code
elimination.  The partial reduction pipeline is now 2x faster.

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -1345,6 +1345,133 @@ Module Compilers.
     Export expr.default.
   End Uncurried.
 
+  Import Uncurried.
+  Section invert.
+    Context {var : type -> Type}.
+
+    Definition invert_Var {t} (e : @expr var t) : option (var t)
+      := match e with
+         | Var t v => Some v
+         | _ => None
+         end.
+
+    Local Notation if_arrow f
+      := (fun t => match t return Type with
+                   | type.arrow s d => f s d
+                   | _ => True
+                   end) (only parsing).
+    Local Notation if_arrow_s f := (if_arrow (fun s d => f s)) (only parsing).
+    Local Notation if_arrow_d f := (if_arrow (fun s d => f d)) (only parsing).
+    Local Notation if_prod f
+      := (fun t => match t return Type with
+                   | type.prod A B => f A B
+                   | _ => True
+                   end).
+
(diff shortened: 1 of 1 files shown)


andres-erbsen:
This looks like a good way to handle `if _ then _ else _`. I'm wondering what we could do about `list_rect`, or more generally, non-cbv constructions. The current `invert`-based special-casing structure is not really instructive for that. Perhaps we should never treat eliminators as app nodes at all, and special-case `list_rect` and `bool_rect` all the way? List recurstion cleanup would still convert app nodes to applications of `list_tect` I guess?

JasonGross:
> The current invert-based special-casing structure is not really instructive for that

If non-linear deep pattern matches were not an issue (and I weren't being generic of the type of identifiers, I could write
```coq
| AppIdent (A * A * type.bool) A ident.bool_rect (Ptrue, Pfalse, b)
  => b' <-- @translate _ b;
     App_bool_rect (@translate _ Ptrue) (@translate _ Pfalse) b'
| AppIdent s d idc args
  => args' <-- @translate _ args;
     k <- Output.expr.Abs (fun r => Halt r);
     p <- (args', k);
     f <- Output.expr.Ident (translate_ident s d idc);
     f @ p
```
) I could add this in a comment?...

I'd rather not special-case list_rect and bool_rect and nat_rect more generally ... most things don't care that they're special.

andres-erbsen:
Yes, please add that comment. With that, LGTM.
I still think it would be a good idea to add similar cases for `list` and `nat` eliminators, just because we think of them as eliminators and who knows when we would trip over the fact that cps translation does not treat them as such. I would be not that surprised if that change improved performance further.

JasonGross:
Comment added.

> I still think it would be a good idea to add similar cases for list and nat eliminators,

I disagree.  The only case where it should make a difference is if we use `list_rect` or `nat_rect` in a way that results in not calling the base case, and we let-bind something in the base case; because functions don't really get evaluated until they are passed arguments (and cps encodes something like this), nothing bad happens with the recursive calls to `list_rect` and `nat_rect`.

PR #313, opened 2018-02-17
Add some truly terribly ad-hoc notations for montgomery wbw

JasonGross:


andres-erbsen:
I decided to go with a more measured approach for now, essentially just added a \lambda notation for `fun` and managed to fit the code in anyway. Thanks, though.

PR #315, opened 2018-02-19
[experiments] Add a ring goal to the pipeline

JasonGross:
Note: We do not yet synthesize sub / opp; these are waiting on @jadephilipoom to fill in `sub` at the arithmetic level (and decide what to do about balance, and where to insert / compute it).

Additionally, unfortunately, the ring proofs are a bit messy.

This is on top of #312.

There is, I expect, a great deal that could be better about this code, but I think it's enough for insightful comments from @andres-erbsen about how the interfaces should look.


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -9,20 +9,24 @@ Require Import Crypto.Util.Tuple Crypto.Util.Prod Crypto.Util.LetIn.
...
 Require Import Crypto.Util.Notations.
 Import ListNotations. Local Open Scope Z_scope.
 
 Definition runtime_mul := Z.mul.
 Definition runtime_add := Z.add.
+Definition runtime_opp := Z.opp.
 Delimit Scope runtime_scope with RT.
 Infix "*" := runtime_mul : runtime_scope.
 Infix "+" := runtime_add : runtime_scope.
+Notation "- a" := (runtime_opp a%RT) : runtime_scope.
```
andres-erbsen: `runtime_scope` is meaningless now, right? Or does partial evaluation still special-case it? I don't think it needs to.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -219,14 +229,47 @@ Module Positional. Section Positional.
...
+  Section add.
+    Context (m:Z) (m_nz:m <> 0) (s:Z) (s_nz:s <> 0)
+            (c:list (Z*Z)) (Hm:m = s - Associational.eval c).
+    Definition add (n:nat) (a b:list Z) : list Z
+      := let a_a := to_associational n a in
+         let b_a := to_associational n b in
+         from_associational n (a_a ++ b_a).
+    Lemma eval_add n (f g:list Z)
+          (Hf : length f = n) (Hg : length g = n) :
+      eval n (add n f g) mod m = (eval n f + eval n g) mod m.
```
andres-erbsen: Why `mod`? should be able to prove equality of inputs to `mod`.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -219,14 +229,47 @@ Module Positional. Section Positional.
...
+  End add.
+
+  Section sub.
+    (** TODO(jadep): Fill me in *)
+    Axiom sub : forall (n:nat) (a b:list Z), list Z. (* should be balanced *)
+    Axiom eval_sub
+      : forall (s:Z)
+               (c:list (Z*Z))
+               n (f g:list Z)
+               (Hf : length f = n) (Hg : length g = n),
```
andres-erbsen: ditto.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -219,14 +229,47 @@ Module Positional. Section Positional.
...
+        = (eval n f - eval n g) mod (s - Associational.eval c).
+    Hint Rewrite eval_sub : push_eval.
+    Definition opp (n:nat) (a:list Z) : list Z
+      := sub n (zeros n) a.
+    Lemma eval_opp
+          (s:Z)
+          (c:list (Z*Z))
+          n (f:list Z)
+          (Hf : length f = n)
+      : eval n (opp n f) mod (s - Associational.eval c)
```
andres-erbsen: ditto.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Partial evaluation special-cases `runtime_mul` and `runtime_add`; these two are the only ones that get additional simplification like `1 * x -> x`.  I can remove the special-casing, though, and fold them into the `Z` operators.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Because copy-paste.  I can drop it.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: No, this one needs the `mod`, because we don't have any guarantee that `f` is bounded, and so we don't have any guarantee that the balance will be enough, right?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: No, I think this one needs to have `mod` to deal with the balance, or something like that?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Removed

JasonGross:
Updated.  We now fully reify `sub` and `opp`.

andres-erbsen:
Is there anything in particular you would like to have reviewed here? I think the github list of changes may be being unhelpful.

JasonGross:
There were some number of things that I punted from the last PR to this one.  While I'm happy to have reviews on the earlier bits of code (the factored ring goal, the changes to compilers), what I'd really like reviews on is the pipeline interface, from the reify-cache on forward.  I'd like to use this PR to get the pipeline into a state where we're both happy with what it looks like to use.  I think you wanted it to be easy to play around with, and it's currently not?

JasonGross:
(I think the only thing on my stack for improving this independent of reviews is that I want to fix the 2x performance regression, and I think that I can fuse bounds relaxation with partial evaluation if I also fold in bounds checking, and I hope that this will get rid of the 2x slowdown.)

JasonGross:
Pipeline is now nearly as fast as it was, again

andres-erbsen:
The code view on github is too slow to be useful for reviewing this PR, so here are free-form comments from reviewing in vimdiff:

Section rcarry_mul except for make_ring:
I think this is roughly the right code, but I do not find the current factoring quite satisfying. I would spend more effort make the interface to the general pipeline nicer to use to reduce the amount of boilerplate in this section, currently it just looks like you factored out recurring term pieces (which indeed is more readable than just a soup of parenthesis, but not particularly nice still).

prime_bound: does this bound also claim to apply to more significant limbs that don't have c subtracted?
beginnnig of section rcarry_mul: inline reification Let-s 
calls to  Pipeline.BoundsPipeline: make a notation that infers s and d, fix dead code elimination to true, in the beginning of all sections that have relax_zrange overload BoundsPipeline to a copy with the relaxation function fixed. Then inline BoundsPipeline12 and etc.
rexpr_1_correctT_Interp, rexpr_21_correctT and similar: inline, then maybe add some notations to smaller like being in bounds and reifying rype arguments

ring stuff:
check_args could perhaps be defined using Decidable.dec to make use_curve_good simpler?
Module Ring does not make sene to me. I think it would be much better to keep subsetoid-to-sig wrapping separate from the ring operations.
After that, I think you could prove that the pre-reification mul, add, sub, whatever form a ring that is isomoprhic got F_q.
Finally, you would show that compiled versions are related to pre-reification versions, thus extending the ring isomorphism.

cache_reify:
I think it should be possible to
1) reify the type of mul (as for GallinaReify)
2) compute from that the type of the intermediate representation of mul
3) compute the relation back to the original code
4) compute the intermediate representation of mul
5) prove the equivalence
Overall interface:
Lemma ir_add : IntermediateRepresentationFor add. t. Qed. Hint Immediate ir_add : ir_cache.
Then, instead of referencing add_gen by name in radd, we would again pass add to a tactic that would be willing to look in the cache but also able to perform compute intermediate representation in case it is not in the cache.
The hintdb parts are not actually important *right now*, but the optional (and thus boring!) nature of caching is what drives my advice here.

Also, do not refer to expand_list as id.

phoas stuff:
please rename is gen_interp (I don't understand it enough to say to what)

Why is everything downstream from partial_reduce_with_bounds1' specialized to a number of arguments?
I think bounds propogation rules should not be called interp, it given that there already are enough interps.
I do not understand why is_var_like  is necessary
I do not understand the type.Z case of reflect, in particular the match statement with invert_Z_cast and reflect_primitive
I do not understand Module option (any of them) and their callers

JasonGross:
Notes/todo for me from video chat with @andres-erbsen :
1. write end-to-end correctness lemma including cpsify
2. add a comment around every module with `Interp` or `interp` explaining what the `interp` function there does
3. rebase on top of Jade's stuff

JasonGross:
I've rebased this on top of master, and I'd like to get this merged soon, so that I can start work on cleaning up the compilers part.  I'd like to not have to rebase the montgomery part of this PR again (in particular, I don't want to have to fix the notation merge conflicts with #324).

@andres-erbsen Do you want to take a look at the way I'm handling casts on `Z * Z`, or should we just merge this and save further things for the next PR?

@jadephilipoom This breaks the specific code format of montred, because you're relying on the code not inlining `expr_let x := fst (Var y) in ...`, and I don't think that's a reasonable assumption.  Is this something you can not rely on?  (Otherwise, I could go back and fix this....)  Also, please take a look at notational changes around montred.

andres-erbsen:
I think it is better to merge this regardless of review since Jade is already working on top of this.

JasonGross:
This now builds on my travis: https://travis-ci.org/JasonGross/fiat-crypto/builds/355201817  .  The travis build here is confused, and is somehow using 62b0fc216af0dd8453c794edccb112c199a05bfe which is a merge of 6e0d95f4405843ad2d5cb0ffad7f76f09fe004b1 into 768dc8d4524b0e48b54fd56876312032e626484a.  I'm going to merge now.

PR #317, opened 2018-02-24
coqprime as a submodule (closes #269)

andres-erbsen:
This is on top of <https://github.com/thery/coqprime/pull/3>.


JasonGross:
Please ensure that this doesn't break the opam package; I seem to recall opam having trouble with uninitialized submodules

andres-erbsen:
I don't know opam and don't really want to learn...

JasonGross:
See https://github.com/coq/opam-coq-archive/pull/262#pullrequestreview-95680712

Can we update the Makefile to update the submodule automatically?  Otherwise we should figure out how to tell opam to update the submodule

andres-erbsen:
Hmm, I'm not a fan of doing this in the Makefile. Our Makefile already does way too much, and fetching things over the network is clearly outside its department.

Does opam not support something like `unpack` or `prepare` scripts that are run once per checkout and not once per builld?

andres-erbsen:
Wait, if submodules break opam, why hasn't our use of bbv already broken it?

andres-erbsen:
Perhaps https://github.com/ocaml/opam/issues/1422#issuecomment-222951257 helps us?

JasonGross:
Checking the commit log on opam, that was https://github.com/ocaml/opam/commit/c030dcbff6aada1c8fa32e868268d4eda676d0c7 from Jan 2016, which shows up no earlier than opam 2.0-alpha.  Travis seems to be using opam 1.2.2, and the most recent commit on both master and opam 1.3 is https://github.com/ocaml/opam/commit/dc3f801a7faf61aeba7981dff5e1f144a0a499d9, from Aug 2015.  Let me see if I can modify the opam script to init submodules.

JasonGross:
Ah, I guess we are already broken on opam.  I will fix opam separately, and we can merge this.

Issue #320, opened 2018-02-26
Reification fails if let-bound variables have sequential names


jadephilipoom:
Here's my minimal example of the issue (tried at the bottom of `src/Experiments/SimplyTypedArithmetic.v`, at commit e11e696).
```
Module Wat.
  Axiom X : Z.

  (* fails : Unable to unify "test" with "5 + 5" *)
  Definition test0: Z :=
    let a0 := 3 in
    let a1 := 5 in
    a0 + a1.
  
Axiom X : Z.
Goal (X = test4).
    etransitivity.
    Focus 2. {
    Reify_rhs().
  (* succeeds *)
  Definition test1 : Z :=
    let a0_x := 3 in
    let a1_y := 5 in
    a0_x + a1_y.

  (* fails : Unable to unify "test" with "5 + 5" *)
  Definition test2: Z :=
    let n0 := 3 in
    let n1 := 5 in
    n0 + n1.
  (* succeeds *)
  Definition test3: Z :=
    let a0 := 3 in
    a0 + 5.
  (* succeeds *)
  Definition test4: Z :=
    let n0 := 3 in
    let m0 := 5 in
    n0 + m0.

```

JasonGross:
Yet another bug in `fresh`, most likely.  I'll look into figuring it out and opening another bug report in Coq later today, but could you check to see if replacing all calls to `refresh <arg1> <arg2>` with `fresh` (without arguments) fixes this?

jadephilipoom:
Yes, that fixes it.

JasonGross:
Given that this works fine in master but fails in 8.7, I suspect this is https://github.com/coq/coq/issues/6534

PR #322, opened 2018-03-01
More work on Montgomery

jadephilipoom:
This change:
- Factors out `mul_converted` and generalizes it, which required:
  * adding base conversion
  * adding a new kind of correctness proof for `flatten` saying that we actually do div/mod on every limb
- Makes my Montgomery implementation carry correctly after multiplication (in the sense of associational carries, not carry flags--that's still in progress and pending on row-wise flatten), which required
  * **adding Nat.mul and Nat.add to the pipeline** -- this was a change that touched the Compilers module, hence the PR to not conflict with Jason
  * changing `refresh` to `fresh` in a few places to avoid a weird bug (see #320) -- this should be reverted once the issue is fixed, but otherwise things don't reify now
- Fixes some notations


jadephilipoom:
I now have a substantial amount of work on top of this PR (in branch `rowwise`). The first large chunk of that branch only touches arithmetic-level stuff, so I'll probably do the strategy of waiting for this PR to be merged, then just committing the arithmetic-only stuff myself, then making another PR for the few things that touch compiler-level code.

PR #324, opened 2018-03-08
row-wise flatten

jadephilipoom:
Most of the stuff in this PR is purely arithmetic stuff, but it touches compiler bits in the following ways:
- adds `Nat.max`, `List.hd`, and `List.tl` to the pipeline
- simplifies away `add_with_get_carry` to `add_get_carry` when the carry is a constant 0

@JasonGross, please review the compiler stuff when you get a chance. Just scroll down to where `Compilers` starts.

@andres-erbsen, now is a good time to take a look at the row-wise `flatten` stuff I've added if you feel like it.

Some context about row-wise vs column-wise flatten, mostly for documentation purposes and can be skipped: 
The procedure we have been calling `flatten` or `compact` is essentially a saturated-arithmetic-friendly `Positional.from_associational`. One strategy (the one we have been using so far in `Columns.flatten`) involves adding together all the terms with the same weight and accumulating their carries together, then adding those carries to the group of terms with the next weight and repeating. This is "column-wise" in the sense that we proceed one column at a time. However, we can make better use of add-with-carry by making groups of terms with *sequential* weights--"rows". We can then take two rows, add together the term from each row with the lowest weight, and then consume the carry when we add the next two terms.

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -18,6 +18,10 @@ Require Import Crypto.Util.ZUtil.Tactics.LtbToLt.
 Require Import Crypto.Util.ZUtil.Tactics.PullPush.Modulo.
 Require Import Crypto.Util.Notations.
 Require Import Crypto.Util.ZUtil.Definitions.
+Require Import Crypto.Util.ZUtil.AddGetCarry Crypto.Util.ZUtil.MulSplit.
+Require Import Crypto.Util.ZUtil Crypto.Util.ZUtil.Hints.Core.
+Require Import Crypto.Util.ZUtil.Modulo Crypto.Util.ZUtil.Div.
+Require Import Crypto.Util.ZUtil.Hints.PullPush.
 Import ListNotations. Local Open Scope Z_scope.
 
 Module Associational.
@@ -445,8 +449,6 @@ Module Positional. Section Positional.
   End carry_mulmod.
 End Positional. End Positional.
 
-Require Import Crypto.Util.ZUtil.AddGetCarry Crypto.Util.ZUtil.MulSplit.
-
 Module BaseConversion.
   Import Positional.
   Section BaseConversion.
@@ -470,10 +472,6 @@ Module BaseConversion.
 
   End BaseConversion.
 End BaseConversion.
-
-Require Import Crypto.Util.ZUtil.
-Require Import Crypto.Util.ZUtil.Modulo Crypto.Util.ZUtil.Div Crypto.Util.ZUtil.Hints.Core.
-Require Import Crypto.Util.ZUtil.Hints.PullPush.
 
 Module Saturated.
   Section Weight.
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -1811,6 +1811,7 @@ Module Compilers.
           | Nat_max : ident (nat * nat) nat
           | Nat_mul : ident (nat * nat) nat
           | Nat_add : ident (nat * nat) nat
+          | Nat_sub : ident (nat * nat) nat
           | nil {t} : ident () (list t)
           | cons {t} : ident (t * list t) (list t)
           | fst {A B} : ident (A * B) A
@@ -1875,6 +1876,7 @@ Module Compilers.
                | Let_In tx tC => curry2 (@LetIn.Let_In (type.interp tx) (fun _ => type.interp tC))
                | Nat_succ => Nat.succ
                | Nat_add => curry2 Nat.add
+               | Nat_sub => curry2 Nat.sub
                | Nat_mul => curry2 Nat.mul
                | Nat_max => curry2 Nat.max
                | nil t => curry0 (@Datatypes.nil (type.interp t))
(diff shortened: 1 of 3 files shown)


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -471,14 +471,51 @@ Module BaseConversion.
   End BaseConversion.
 End BaseConversion.
 
-(* Non-CPS version of Arithmetic/Saturated/MulSplit.v *)
-Module MulSplit.
+Require Import Crypto.Util.ZUtil.
```
JasonGross: Please move these up to the top of the file; `coqtop` seems to slow down exponentially when backtracking across `Require` multiples times

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5659,6 +6228,37 @@ Module RemoveDeadLets.
         end
       end.
 
+    (* TODO(jadep): this is pretty ad-hoc and should eventually be removed *)
```
JasonGross: This should be fused with partial reduction; update the check for `Z` (in `interp_let_in`, I think) to inline lets let bind the structure you care about.

jadephilipoom: I'm not sure we want to do this in the general case. For my particular use case I want to inline those lets, but this is not even something I'm sure I want to keep doing--the main utility is making the output easier to read for me.

JasonGross: Add a parameter to partial reduction of type `ident -> bool` which classifies identifiers which are permissible to inline.  Fill it with `fun _ => false` in the default case, and with `fun idc => match idc with | ident.shiftl | ident.shiftr => true | _ => false end` in the case you want.

andres-erbsen:
I looked at `Section Rows` and `flatten`. It looks like code that does the thing, and I did not find any issues. I did not do a oh-I'm-now-auditing-crypto-level review, but if the proofs check out and the final code is fast enough, I guess I don't have to :+1: .

src/Experiments/SimplyTypedArithmetic.v
andres-erbsen: I don't think it is obvious that we want to do this in partial reduction. In particular, I think the real rule we want here includes only inlining a let if it is used at most once. The separate ad-hoc pass seems no more problematic to me than an ad-hoc parameter inside the partial reduction pass. Either way, I don't think we should block this PR on resolving the inlining issue (but I guess if @JasonGross cares, we could merge this without the inliner).

JasonGross: It is trivial to extend the proof of partial reduction correctness to handle this case.  I would rather not have to write the additional boilerplate proofs to do this extra pass.  If we expect this to disappear by the time I get around to proofs, then I'm fine with having it as a separate ad-hoc pass.  Otherwise, I'd like it merged into partial reduction

jadephilipoom: The particular reason I want to inline these lets is that the code I'm referencing has a single instruction for shift + multiply, and it's easier for me to match up lines this way. So it's not quite true that inlining lets used only once would capture my use case, and it's also not really something I think we need to generalize.

JasonGross: Generalization makes the proof easier / have fewer cases :-P

andres-erbsen:
Perhaps we should just accept that some compilation targets will have ad-hoc touchup passes. I don't think what Jade wants here is generally useful, and I don't think the generalizations we have been thinking of would be used in any other case.

Jason, would you be more okay with the current variant if it was clear that you don't need to do proofs for it?

JasonGross:
> Jason, would you be more okay with the current variant if it was clear that you don't need to do proofs for it?

Yes

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3580,7 +4144,20 @@ Module Compilers.
...
                        inr (inr (fst result), inr (snd result))
                      | inr (inr (inr (inr x, y), z), a)
-                        => expr.reflect (AppIdent (ident.Z.add_with_get_carry_concrete x) (expr.reify (t:=type.Z*type.Z*type.Z) (inr (inr (y, z), a))))
+                       => let default_with_carry := expr.reflect (AppIdent (ident.Z.add_with_get_carry_concrete x) (expr.reify (t:=type.Z*type.Z*type.Z) (inr (inr (y, z), a)))) in
+                          let default_no_carry := expr.reflect (AppIdent (ident.Z.add_get_carry_concrete x) (expr.reify (t:=type.Z*type.Z) (inr (z,a)))) in
+                          let default := match y with
+                                         | inr xx=> if Z.eqb xx 0 then default_no_carry else default_with_carry
+                                         | _ => default_with_carry
+                                         end in
+                          match (z,a) with
```
JasonGross: Note that you don't need parentheses here (and if you remove them, they also need to be removed from the cases

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -3580,7 +4144,20 @@ Module Compilers.
                        let result := ident.interp idc (x, y, z, a) in
                        inr (inr (fst result), inr (snd result))
                      | inr (inr (inr (inr x, y), z), a)
-                        => expr.reflect (AppIdent (ident.Z.add_with_get_carry_concrete x) (expr.reify (t:=type.Z*type.Z*type.Z) (inr (inr (y, z), a))))
+                       => let default_with_carry := expr.reflect (AppIdent (ident.Z.add_with_get_carry_concrete x) (expr.reify (t:=type.Z*type.Z*type.Z) (inr (inr (y, z), a)))) in
```
JasonGross: Note that, when you rebase, the default values have changed a bit, to support bounds analysis.

andres-erbsen:
Is there any reason not to merge this right now?

JasonGross:
> Is there any reason not to merge this right now?

Because "This branch has conflicts that must be resolved"

andres-erbsen:
Continued in #334.

PR #325, opened 2018-03-11
Add Algebra.SubsetoidRing

JasonGross:


andres-erbsen:
This looks like what I had in mind (although I did not read super closely). Thanks.

PR #329, opened 2018-03-22
[experiments] Uncurry in phoas

JasonGross:
Now things like `carry_mulmod` no longer have manual uncurrying.  This also unifies the "constant" and "1-argument" versions of the pipeline.  Uncurrying happens in the second stage of the pipeline, after caching; this makes it a bit easier to phrase the correctness and caching lemmas, because it means that we don't need to translate between curried and uncurried things when phrasing the correctness of the cache.

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -2859,16 +2859,6 @@ Module Compilers.
 
       Definition uncurry (t : type) : type
         := type.arrow (uncurried_domain t) (type.final_codomain t).
-
-      Section with_var.
-        Context (var : type -> Type).
-        Fixpoint value (t : type)
-          := match t with
-             | type.arrow s d
-               => value s -> value d
-             | t => @expr var t
-             end.
-      End with_var.
     End type.
 
     Fixpoint app_curried {t : type}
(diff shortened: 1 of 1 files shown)


andres-erbsen:
I cannot make up my mind on whether having a wrapper pass like this is a good idea overall. I agree that it removes some boilerplate from the definitions, and that's good. I think we might have a better chance of understanding whether this is a good idea if we did away with rweight first. And the recent performance improvements may have changed the tradeoff between different amounts of caching as well.

More specifically, I think it seems silly that we lose the tupled form somewhere between the CPS input and the final output, and need this pass to emulate tupled form again.

andres-erbsen:
I don't understand the example with `A -> B -> C -> D`. If that is our source program, say `a*b+c`, it should have types
- `A -> B -> C -> D` after translating recursion to eliminators
-  `A * B * C -> D` after converting partial applications to tuple formation
- `A * B * C -> forall {R}, (D -> R) -> R` after CPS conversion, for which I think the appropriate notation would be `(A * B * C * (D -->)) -->`.

If our pipeline currently does not achieve that, I think we should be trying to fix it, rather than make a wrapper that makes the structure look okay from the outside.

JasonGross:
@achlipala could you review the first commit in this PR (currently a4ca575)?  It's an implementation of uncurrying based on what we talked about in group meeting.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2839,6 +2845,189 @@ Module Compilers.
...
+             | type.arrow s d
+               => value s -> value d
+             | t => @expr var t
+             end.
+      End with_var.
+    End type.
+
+    Fixpoint app_curried {t : type}
+      : type.interp t -> type.interp (type.uncurried_domain t) -> type.interp (type.final_codomain t)
+      := match t return type.interp t -> type.interp (type.uncurried_domain t) -> type.interp (type.final_codomain t) with
```
achlipala: Is the `return` annotation here necessary in the final, working version?  Do you just leave it there to avoid bad error messages while editing the code?

JasonGross: Almost certainly unnecessary.  I put them in so that I get localized error message, and so that I can write them with `refine` one case at a time.

JasonGross:
@achlipala 

> Is the idea that primitive functions should stay curried for simplicity, while we want to make all derived functions tuple-based? Why exactly?

Writing a curried version of, e.g., `List.map` is complicated.  In particular, if start by passing in a function of type `A -> B -> C`, and a list of type `A`, and we need to transform this into a `map` that takes in a function of type `A * B -> C` and a list of type `A`, this requires a new primitive identifier.  Similarly with `list_rect` or `nat_rect` returning functions.  I don't want the uncurrying pass to have to change the list of identifiers being used, because that's complicated.  (I'm also still not clear on the exact reason we want to uncurry more deeply; it does reduce the number of continuations that show up in the cps form, but it's not clear to me that this is a bottleneck, and the transformation that I really care about here is just uncurrying the top-level lambda, without touching the body.)

JasonGross:
Updated with a simpler form of uncurrying that does less work, after in-person discussion with @andres-erbsen .  Now merging.

PR #330, opened 2018-03-22
Thunk let-bound default values, unfold bool_rect

JasonGross:
This makes the pipeline faster to run under vm_compute than under lazy.
I suspect mainly because I've replaced [bool_rect] with [if].

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -6009,39 +6009,20 @@ Section rcarry_mul.
 
   Let limbwidth := (Z.log2_up (s - Associational.eval c) / Z.of_nat n)%Q.
   Let idxs := (seq 0 n ++ [0; 1])%list%nat.
+  Let coef := 2.
   Let upperbound_tight := (2^Qceiling limbwidth + 2^(Qceiling limbwidth - 3))%Z.
-  Let upperbound_loose := (3 * upperbound_tight)%Z.
-  Let f_bounds_tight := List.repeat r[0~>upperbound_tight]%zrange n.
-  Let f_bounds_loose := List.repeat r[0~>upperbound_loose]%zrange n.
   Let prime_bound : ZRange.type.interp (type.Z)
     := r[0~>(s - Associational.eval c - 1)]%zrange.
 
   Definition relax_zrange_of_machine_wordsize
     := relax_zrange_gen [machine_wordsize; 2 * machine_wordsize]%Z.
-  Local Arguments relax_zrange_of_machine_wordsize / .
 
   Let rw := rweight limbwidth.
-  Let rs := GallinaReify.Reify s.
-  Let rc := GallinaReify.Reify c.
-  Let rn := GallinaReify.Reify n.
-  Let ridxs := GallinaReify.Reify idxs.
-  Let rlen_c := GallinaReify.Reify (List.length c).
-  Let rlen_idxs := GallinaReify.Reify (List.length idxs).
-  Let rcoef := GallinaReify.Reify 2.
   Let relax_zrange := relax_zrange_of_machine_wordsize.
   Let tight_bounds : ZRange.type.interp (type.list type.Z)
-    := f_bounds_tight.
-  Let tight2_bounds : ZRange.type.interp (type.list type.Z * type.list type.Z)
-    := (tight_bounds, tight_bounds).
+    := List.repeat r[0~>upperbound_tight]%zrange n.
   Let loose_bounds : ZRange.type.interp (type.list type.Z)
-    := f_bounds_loose.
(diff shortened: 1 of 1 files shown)


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -6376,11 +6376,25 @@ Ltac do_inline_cache_reify do_if_not_cached :=
...
+  lazymatch goal with
+  | [ |- ?LHS = ?RHS ]
+    => is_evar RHS;
+       let x := (eval vm_compute in LHS) in
+       unify RHS x;
+       vm_cast_no_check (eq_refl x)
+  end.
+(* TODO: MOVE ME *)
+Ltac transitivity_vm_compute_lhs :=
+  etransitivity; [ | lazy ];
```
andres-erbsen: why `lazy` on the right?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5114,11 +5113,12 @@ Module Compilers.
            | Var t v => (cur_idx, v)
            | TT => (cur_idx, TT)
            | AppIdent s d idc args
-             => let default := @eliminate_dead' _ args cur_idx in
-                let default := (fst default, AppIdent idc (snd default)) in
-                match args in expr.expr t return ident.ident t d -> positive * expr d -> positive * expr d with
+             => let default _
+                    := let default := @eliminate_dead' _ args cur_idx in
+                       (fst default, AppIdent idc (snd default)) in
```
andres-erbsen: perhaps also remove the shadowing here, when the code is not strictly linear it is much harder to read

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -4566,115 +4566,114 @@ Module Compilers.
...
-                  => let default0 := AppIdent idc (expr.reify (t:=_*_) x_y) in
-                     let default := expr.reflect default0 in
+                  => let default0 _ := AppIdent idc (expr.reify (t:=_*_) x_y) in
+                     let default _ := expr.reflect (default0 tt) in
                      match x_y return _ * expr _ + type.interp _ with
                      | inr (inr x, inr y) => inr (ident.interp idc (x, y))
                      | inr (inr x, inl (data, e) as y)
                      | inr (inl (data, e) as y, inr x)
-                       => let data' := ZRange.ident.option.interp idc (data, Some r[x~>x]%zrange) in
+                       => let data' _ := ZRange.ident.option.interp idc (data, Some r[x~>x]%zrange) in
```
andres-erbsen: why not just inline tihs let?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -4566,115 +4566,114 @@ Module Compilers.
...
+                                                   e)
+                                    | None => default0 tt
+                                    end)
                      | inr (inl (dataa, a), inl (datab, b))
-                       => let data' := ZRange.ident.option.interp idc (dataa, datab) in
-                          match invert_Z_opp a, invert_Z_opp b with
-                          | Some a, Some b
-                            => inl (data',
-                                    AppIdent
+                       => inl (ZRange.ident.option.interp idc (dataa, datab),
```
andres-erbsen: do I understand correctly that this is a performance-irrelevant change?

src/Experiments/SimplyTypedArithmetic.v
JasonGross: This tactic should probably just be inlined, because it doesn't form a coherent block, but we do need `lazy` or similar; the rhs is `Pipeline.Success rv` where `rv` is a context variable whose body is an evar, and I believe `reflexivity` won't instantiate the evar if it's hidden behind a context variable.

JasonGross: Updated

src/Experiments/SimplyTypedArithmetic.v
JasonGross: The code is linear, because the outer let is a `let` not a `let rec`, and so the inner `default` is unbound before I bind it, and disappears when the outer one comes into effect.  But I've changed the inner one to `default'`.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Current thought: because it is clutter, and I'd rather not clutter the body of the `if` statement with things independent of the conditional (note that `data' tt` is equal to `data` in the first branch of the `if`)

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Yes, I believe so.  It might have minor effects when using the `lazy` strategy, because it might allow skipping the computation of the data in cases where the partial values are special (e.g., `Z.add @@ (0, 0)`), but it should be negligible.

PR #332, opened 2018-03-22
NoBrainer cps 1

JasonGross:
On top of #330 

This adds a pass which inline let-bound variables which are used at most once.

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -5291,6 +5291,18 @@ Module Compilers.
   End NoBrainer1.
 
   Module GeneralizeVar.
+    (** In both lazy and cbv evaluation strategies, reduction under
+        lambdas is only done at the very end.  This means that if we
+        have a computation which returns a PHOAS syntax tree, and we
+        plug in two different values for [var], the computation is run
+        twice.  This module provides a way of computing a
+        representation of terms which does not suffer from this issue.
+        By computing a flat representation, and then going back to
+        PHOAS, the cbv strategy will fully compute the preceeding
+        PHOAS passes only once, and the lazy strategy will share
+        computation among the various uses of [var] (because there are
+        no lambdas to get blocked on) and thus will also compute the
+        preceeding PHOAS passes only once. *)
     Module Flat.
       Inductive expr : type -> Set :=
       | Var (t : type) (n : positive) : expr t
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -5376,7 +5376,7 @@ Module Compilers.
   End GeneralizeVar.
 
   Module ReassociateSmallConstants.
-      Import Compilers.Uncurried.expr.default.
+    Import Compilers.Uncurried.expr.default.
 
     Section with_var.
       Context (max_const_val : Z)
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -5371,7 +5371,7 @@ Module Compilers.
     Definition FromFlat {t} (e : Flat.expr t) : Expr t
       := let e' := @from_flat t e in
          fun var => e' var (PositiveMap.empty _).
-    Definition GeneralizeVar {t} (e : @expr _ t) : Expr t
+    Definition GeneralizeVar {t} (e : @expr (fun _ => PositiveMap.key) t) : Expr t
       := FromFlat (to_flat e).
   End GeneralizeVar.
 
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -4069,6 +4069,11 @@ Module Compilers.
   End ZRange.
 
   Module DefaultValue.
+    (** This module provides "default" inhabitants for the
+        interpretation of PHOAS types and for the PHOAS [expr] type.
+        These values are used for things like [nth_default] and in
+        other places where we need to provide a dummy value in cases
+        that will never actually be reached in correctly used code. *)
     Module type.
       Module primitive.
         Definition default {t : type.primitive} : type.interp t
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -4087,6 +4087,23 @@ Module Compilers.
            | type.list A => @nil (type.interp A)
            end.
     End type.
+
+    Module expr.
+      Section with_var.
+        Context {var : type -> Type}.
+        Fixpoint default {t : type} : @expr var t
+          := match t with
+             | type.type_primitive x
+               => AppIdent (ident.primitive type.primitive.default) TT
+             | type.prod A B
+               => (@default A, @default B)
+             | type.arrow s d => Abs (fun _ => @default d)
+             | type.list A => AppIdent ident.nil TT
+             end.
+      End with_var.
+
+      Definition Default {t} : Expr t := fun _ => default.
+    End expr.
   End DefaultValue.
 
   Module partial.
@@ -5066,7 +5083,6 @@ Module Compilers.
       exact admit. (* boundedness *) }
   Qed.
 
-
   Module DeadCodeElimination.
     Fixpoint compute_live' {t} (e : @expr (fun _ => PositiveSet.t) t) (cur_idx : positive)
     : positive * PositiveSet.t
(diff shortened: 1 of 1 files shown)


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -7376,29 +7644,40 @@ Local Open Scope expr_scope.
...
-c.Add64($v14, $v13_hi, $v12_hi);
-c.Add256($v15, $v8, RegMod << 128 * $v6);
-c.Add256($v16, $v10, $v15_lo);
-c.Add256($v17, $v14, $v16_lo);
-c.Add256($v18, $v_lo, $v13_lo);
-c.Addc($v19, $v_hi, $v17_lo);
-c.Selc($v20,RegZero, RegMod);
-c.Sub($v21, $v19_lo, $v20);
-c.AddM($ret, $v21, RegZero, RegMod);
+c.ShiftR($x0, $x_lo, 128);
```
andres-erbsen: I don't understand what changed here. Is the new code supposed to be the same as the old code?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -6646,62 +6921,44 @@ Module X25519_64.
...
- expr_let v44 := 19 *₆₄ v42 in
- expr_let v45 := v27 +₆₄ v44 in
- expr_let v46 := (uint64)(v45 >> 51) in
- expr_let v47 := ((uint64)(v45) & 2251799813685247) in
- expr_let v48 := v46 +₆₄ v34 in
- expr_let v49 := (uint64)(v48 >> 51) in
- expr_let v50 := ((uint64)(v48) & 2251799813685247) in
- expr_let v51 := v49 +₆₄ v37 in
- v47 :: v50 :: v51 :: v40 :: v43 :: [])%expr
+ expr_let x0 := x₁ [[0]] *₁₂₈ x₂ [[0]] +₁₂₈
```
andres-erbsen: this output makes sense to me

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5166,8 +5182,185 @@ Module Compilers.
...
+      Fixpoint nobrainer1' {t} (e : @expr (@expr var) t) (cur_idx : positive)
+        : positive * @expr var t
+        := match e with
+           | Var t v => (cur_idx, v)
+           | TT => (cur_idx, TT)
+           | AppIdent s d idc args
+             => let default _
+                    := let default := @nobrainer1' _ args cur_idx in
+                       (fst default, AppIdent idc (snd default)) in
+                match args in expr.expr t return ident.ident t d -> (unit -> positive * expr d) -> positive * expr d with
```
andres-erbsen: is this just deep pattern matching for LetIn?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5166,8 +5182,185 @@ Module Compilers.
...
+                                              | _ * (s -> d) => (expr s -> expr d)%type
+                                              | _ => unit
+                                              end%ctype
+                                           -> (unit -> positive * expr d)
+                                           -> positive * expr d)
+                             with
+                             | ident.Let_In _ _
+                               => fun x' f' _
+                                  => if match PositiveMap.find idx live with
+                                        | Some n => (1 <? n)%nat
```
andres-erbsen: please flip the meanings of the boolean here: `true` = should substitute makes much more sense to me, for example because the default would be in the else case then

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5166,8 +5182,185 @@ Module Compilers.
       := fun var => eliminate_dead (ComputeLive e) (e _).
   End DeadCodeElimination.
 
+  Module NoBrainer1.
```
andres-erbsen: Hmm, I'm not sure how I feel about the naming here. I would suggest `subst01` with a comment "substitute variables used 0 or 1 times". And if we want to reference nobrainer paper, let's just add a link instead of naming after it. And even I am guessing the reason this does not supersede dead code elimination is that dead code elimination is the fixed point of subst0, not just one subst0?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5166,8 +5182,185 @@ Module Compilers.
...
+      | Var (t : type) (n : positive) : expr t
+      | TT : expr type.unit
+      | AppIdent {s d} (idc : ident s d) (arg : expr s) : expr d
+      | App {s d} (f : expr (s -> d)) (x : expr s) : expr d
+      | Pair {A B} (a : expr A) (b : expr B) : expr (A * B)
+      | Abs (s : type) (n : positive) {d} (f : expr d) : expr (s -> d).
+    End Flat.
+
+    Definition ERROR {T} (v : T) : T := v.
+    Global Opaque ERROR.
```
andres-erbsen: Are you sure you want reduction-hint opaque instead of `exact v. Qed.`?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5166,8 +5182,185 @@ Module Compilers.
...
+         end.
+
+    Definition to_flat {t} (e : expr t) : Flat.expr t
+      := to_flat' e 1%positive.
+    Definition ToFlat {t} (E : Expr t) : Flat.expr t
+      := to_flat (E _).
+    Definition FromFlat {t} (e : Flat.expr t) : Expr t
+      := let e' := @from_flat t e in
+         fun var => e' var (PositiveMap.empty _).
+    Definition GeneralizeVar {t} (e : @expr _ t) : Expr t
```
andres-erbsen: please add input var type annotation here for clarity

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5166,8 +5182,185 @@ Module Compilers.
...
+
+      Definition nobrainer1 {t} e : expr t
+        := snd (@nobrainer1' t e 1).
+    End with_var.
+
+    Definition NoBrainer1 {t} (e : Expr t) : Expr t
+      := fun var => nobrainer1 (ComputeLiveCounts e) (e _).
+  End NoBrainer1.
+
+  Module GeneralizeVar.
```
andres-erbsen: please add a module comment explaining what this does and what performance issue we work around using this

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5633,16 +5870,23 @@ Module Pipeline.
...
+             (with_nobrainer1 : bool)
              {s d}
              (E : Expr (s -> d))
              arg_bounds
   : Expr (s -> d)
     := let E := PartialEvaluate true E in
        (* Note that DCE evaluates the expr with two different [var]
-          arguments, and so will likely result in a pipeline that is
-          2x slower *)
+          arguments, and so results in a pipeline that is 2x slower
```
andres-erbsen: Perhaps we should merge the var-mangling pass into a wrapper of the function that needs it, i.e. dce and subst1?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5662,18 +5906,20 @@ Module Pipeline.
        E.
 
   Definition BoundsPipeline
-             (with_dead_code_elimination : bool)
+             (with_dead_code_elimination : bool := true)
```
andres-erbsen: why did this become hardcoded now? because performance issues were fixed?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -4087,6 +4087,23 @@ Module Compilers.
            | type.list A => @nil (type.interp A)
            end.
     End type.
+
+    Module expr.
+      Section with_var.
+        Context {var : type -> Type}.
+        Fixpoint default {t : type} : @expr var t
```
andres-erbsen: these are reather meaningless values... perhaps add a comment (or even rename) to explain that the only thing we achieve here is that the type is inhabited.

JasonGross: Comment added.

src/Experiments/SimplyTypedArithmetic.v

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Added

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Yes, because performance issues were fixed.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Changed to `subst01`.  re DCE is the fixed point: yes.

PR #333, opened 2018-03-23
Add a constant-time binary extended GCD algorithm.

davidben:
Here's my latest foray into this amusement. I suspect some of this could be factored out better, but I wasn't sure what the right structure was.

---

Thankfully, most modular inverses involve a prime or at least odd modulus, but RSA key generation requires inverting with an even modulus (the totient). Implementing that in constant-time is tricky.

This file adapts the binary extended GCD algorithm in the Handbook of Applied Cryptography. This is to verify the C implementation in https://boringssl-review.googlesource.com/c/boringssl/+/26370.

### DIFF
--- a/src/Arithmetic/BinaryExtendedGCD.v
+++ b/src/Arithmetic/BinaryExtendedGCD.v
@@ -1,0 +1,873 @@
+Require Import Coq.Bool.Bool.
+Require Import Coq.micromega.Lia.
+Require Import Coq.ZArith.ZArith.
+Require Import Coq.ZArith.Znumtheory.
+Require Import Crypto.Experiments.Loops.
+Require Import Crypto.Util.Tactics.DestructHead.
+Require Import Crypto.Util.ZUtil.
+Require Import Crypto.Util.ZUtil.Tactics.DivModToQuotRem.
+
+Local Open Scope Z_scope.
+
+Module BinaryExtendedGCD.
+  Section BinaryExtendedGCD.
+    (** This section implements a constant-time version of the binary extended
+        GCD algorithm, for computing modular inverses. It is the fully general
+        version, supporting even moduli. In most cases, more specific algorithms
+        should be used. For instance, inverting by a prime is simpler to do with
+        Fermat's Little Theorem. However, RSA key generation requires the most
+        general version.
+
+        This algorithm is adapted from the Handbook of Applied Cryptography,
+        section 14.4.3, algorithm 14.61. It was modified slightly to better
+        bound coefficients, then recast into constant-time selects:
+
+        1. Negate [B] and [C] so they are positive. The invariant now involves a
+           subtraction.
+
+        2. If step 2 (both [x] and [y] are even) runs, abort immediately. This
+           algorithm only cares about [x] and [y] relatively prime.
+
+        3. Subtract copies of [x] and [y] as needed in step 6 (both [u] and [v]
+           are odd) so coefficients stay in bounds.
+
+        4. Replace the [u >= v] check with [u > v]. This changes the end
+           condition to [v = 0] rather than [u = 0]. This saves an extra
+           subtraction due to which coefficients were negated.
+
+        5. Rename x and y to a and n, to capture that one is a modulus.
+
+        6. Rearrange steps 4 through 6 slightly. Merge the loops in steps 4 and
+           5 into the main loop (step 7's goto), and move step 6 to the start of
+           the loop iteration, ensuring each loop iteration halves at least one
+           value.
+
+        Note this algorithm does not handle either input being zero. *)
+
+    (** This is a constant-time algorithm, so first define fixed-width words to
+        compute over. In particular, we want to capture the non-trivial bounds on
+        the intermediate values. A [word] is an integer with specified bit
+        width. The actual implementation will represent this with an array of
+        smaller machine words. *)
+    Inductive word (bits : Z) :=
+    | Word : Z -> word bits.
+
+    Definition wrap (bits val : Z) : Z := val mod 2^bits.
+
+    Definition to_Z {bits : Z} (a : word bits) : Z :=
+      match a with Word v => v end.
+    Definition to_word {bits : Z} (val : Z) :=
+      Word bits (wrap bits val).
+
+    (** Our algorithm is implemented over the following constant-time
+        primitives. [add] and [sub] return a carry and borrow bit,
+        respectively. *)
+    Definition add {bits : Z} (a b : word bits) : bool * word bits :=
+      let a' := to_Z a in
+      let b' := to_Z b in
+      (2^bits <=? a' + b', to_word (a' + b')).
+    Definition sub {bits : Z} (a b : word bits) : bool * word bits :=
+      let a' := to_Z a in
+      let b' := to_Z b in
+      (a' <? b', to_word (a' - b')).
+    Definition rshift1 {bits : Z} (a : word bits) : word bits :=
+      to_word ((to_Z a) / 2).
+    Definition select {bits : Z} (t : bool) (a b : word bits) : word bits :=
+      if t then a else b.
+    Definition odd {bits : Z} (a : word bits) : bool := Z.odd (to_Z a).
+    Definition is_one {bits: Z} (a : word bits) : bool := to_Z a =? 1.
+
+
+    (** These helpers are implemented with temporaries. They return the final
+        values of the temporaries, to model clobbering them in imperative
+        code. *)
+    Definition maybe_rshift1_carry {bits : Z}
+               (t carry : bool) (a : word bits) : word bits * word bits :=
+      let tmp := rshift1 a in
+      let a := select t tmp a in
+      let a := if t && carry then to_word ((to_Z a)  + 2^(bits - 1))
+               else a in
+      (a, tmp).
+
+    Definition maybe_rshift1 {bits : Z} (t : bool) :=
+      @maybe_rshift1_carry bits t false.
+
+    Definition maybe_add {bits : Z}
+               (t : bool) (a b : word bits) : bool * word bits * word bits :=
+      let (carry, tmp) := add a b in
+      let a := select t tmp a in
+      (carry && t, a, tmp).
+
+    (** We ultimately wish to compute [a] inverse mod [n]. Parameterize over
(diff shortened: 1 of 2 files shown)


src/Arithmetic/BinaryExtendedGCD.v
```diff
@@ -0,0 +1,873 @@
...
+      let (v, tmp) := maybe_rshift1 v_even v in
+      let C_or_D_odd := odd C || odd D in
+      let '(C_carry, C, tmp) := maybe_add (C_or_D_odd && v_even) C n in
+      let '(D_carry, D, tmp) := maybe_add (C_or_D_odd && v_even) D a in
+      let (C, tmp) := maybe_rshift1_carry v_even C_carry C in
+      let (D, tmp) := maybe_rshift1_carry v_even D_carry D in
+      (a, n, A, B, C, D, u, v).
+
+    Ltac destruct_state s :=
+      destruct s as [[[[[[[[a] [n]] [A]] [B]] [C]] [D]] [u]] [v]].
```
JasonGross: You might be interested in things like `destruct_head'_prod` and `destruct_head' word` from `Crypto.Util.Tactics.DestructHead`.  Also, I hope you don't depend on these names anywhere.

src/Arithmetic/BinaryExtendedGCD.v
```diff
@@ -0,0 +1,873 @@
...
+    Proof. destruct t; reflexivity. Qed.
+
+    Lemma if_mod_distr (t : bool) (a b n : Z) :
+      (if t then a mod n else b mod n) = (if t then a else b) mod n.
+    Proof. destruct t; reflexivity. Qed.
+
+    Ltac simplify_words :=
+      cbv [select odd add sub rshift1 maybe_add maybe_rshift1
+                  maybe_rshift1_carry to_Z to_word wrap is_one];
+      try rewrite !if_func_distr.
```
JasonGross: `try rewrite !foo` is the same as `rewrite ?foo`

src/Arithmetic/BinaryExtendedGCD.v
```diff
@@ -0,0 +1,873 @@
...
+        repeat first [bool_to_prop_once | split];
+        repeat match goal with
+               | H : ?a < ?m |- context[?a mod ?m] =>
+                 rewrite (Z.mod_small a m) by lia
+               | _ => lia
+               end.
+    Qed.
+
+    (** Prove the left version of [Z.gcd_sub_diag_r]. *)
+    Lemma Zgcd_sub_diag_l (n m : Z) : Z.gcd (n - m) m = Z.gcd n m.
```
JasonGross: This belongs in some ZUtil file (maybe make a new `ZUtil.GCD` file and stick a `Module Z` in it and put this lemma there)

src/Arithmetic/BinaryExtendedGCD.v
```diff
@@ -0,0 +1,873 @@
...
+    Proof.
+      rewrite (Z.gcd_comm (n - m) m).
+      rewrite (Z.gcd_comm n m).
+      apply Z.gcd_sub_diag_r.
+    Qed.
+
+    (** The following bounds on [B + D] come out of [nia], but [nia] takes a
+        long time to discover this in the context of the full proof. *)
+    Lemma AC_lt_BD_le (a n A B C D : Z) :
```
JasonGross: This should be `Local Lemma`, because if I import this module, I don't want to see unqualified names like `AC_lt_BD_le`.

src/Arithmetic/BinaryExtendedGCD.v
```diff
@@ -0,0 +1,873 @@
...
+      - rewrite Zmod_odd in H;
+          destruct (Z.odd z); try trivial; lia.
+    Qed.
+
+    Ltac fail_if_mod t m := match t with
+                            | _ mod m => fail 1
+                            | _ => idtac
+                            end.
+
+    Ltac push_mod_down_in t m :=
```
JasonGross: Have you seen the `push_Zmod` tactic in one of the ZUtil files?  It's probably better to make a copy of that tactic, in that file, and adapt it to work in hypotheses.  Unless you want to not consider things modulo things other than `m`?

src/Arithmetic/BinaryExtendedGCD.v
```diff
@@ -0,0 +1,873 @@
...
+    Proof.
+      destruct_state s.
+      unfold even_step, invariant in *.
+      simplify_words.
+
+      destruct_odd_u_v u v;
+        [ (* The odd/odd case is a no-op. *)
+          rewrite !andb_false_r; cbn; rewrite Hu; rewrite Hv;
+          apply H
+        | destruct (Z.odd C) eqn:HC;
```
JasonGross: You're depending on the name you give in `destruct_state` here.  Can you instead do some sort of context matching?

src/Arithmetic/BinaryExtendedGCD.v
```diff
@@ -0,0 +1,873 @@
...
+        repeat match goal with
+               | _ => rewrite Zplus_mod_idemp_l
+               | _ => rewrite if_mod_distr
+               | _ => rewrite add_carry_div2
+               end;
+
+        repeat split;
+        (* Keep only the relevant half of the problem in the context. *)
+        match goal with
+        | [ |- context[u]] => idtac
```
JasonGross: Could you instead `clear dependent` all variables of type `Z` (or `word`) which do not show up in the goal?  (So that you're not depending on `destruct_state` names)  Or even just try to `clear dependent` everything of type `Z` / `word`?

src/Arithmetic/BinaryExtendedGCD.v
```diff
@@ -0,0 +1,873 @@
...
+    Proof.
+      destruct_state s.
+      unfold even_step, u_v_prod, invariant, u_or_v_even in *.
+      simplify_words.
+
+      destruct_odd_u_v u v;
+      [ (* [u] and [v] cannot both be even by [Heven]. *)
+        destruct_head'_and;
+        destruct_head'_or;
+        match goal with
```
JasonGross: Is there a reason to not just use `discriminate` (or `congruence`) here?  I sometimes do this match by hand for speed in tactics with a `repeat match`, but that doesn't seem to be the case here.

src/Arithmetic/BinaryExtendedGCD.v
```diff
@@ -0,0 +1,873 @@
...
+      state_get_a (step s) = state_get_a s /\
+      state_get_n (step s) = state_get_n s.
+    Proof.
+      destruct_state s.
+      unfold step, state_get_a, state_get_n.
+      simplify_words.
+      auto.
+    Qed.
+
+    Lemma gcd_even_even (a b : Z)
```
JasonGross: This belongs in ZUtil.GCD or something

src/Arithmetic/BinaryExtendedGCD.v
```diff
@@ -0,0 +1,873 @@
...
+      | Some (Word ainv) =>
+        0 <= ainv < n /\
+        (ainv * a) mod n = 1
+      | None => Z.gcd a n <> 1
+      end.
+    Proof.
+      (* Tidy up the loop. *)
+      unfold mod_inverse_consttime.
+      simplify_words.
+      rewrite !Z.mod_small by
```
JasonGross: I think I have a tactic like `Z.rewrite_mod_small` that does this and handles side conditions nicely in some ZUtil file.  Idk if it'll serve your purposes here.

src/Arithmetic/BinaryExtendedGCD.v
davidben: I don't think so, or is there a really clever version of `clear dependent`? The goal may only involve, say, `A`, but it relies on knowing something about `u` and a hypothesis that links them. (E.g. a lot of the mod 2 bashing.) `A` and `B` are the coefficients that make `u`, and `C` and `D` make `v`.

I can see about matching on that pattern instead. Likewise with the above.

Is the risk with the names that Coq might introduce an `A` before `destruct_state` and get in the way? The names were handy since, well, they're the names of the bits of state. :-)

src/Arithmetic/BinaryExtendedGCD.v
davidben: I didn't know about them. :-) Will play with those. I did think all my `false = true`, etc., matchers were kinda silly...

src/Arithmetic/BinaryExtendedGCD.v
JasonGross: I would say: if the thing you're proving relies on knowing which bit of the state you're talking about, see if you can factor out more lemmas.  Ideally, there are some lemmas which are math, and then there are some lemmas which are about sequences of operations (e.g., this function preserves this invariant), and then this lemma would put them all together with something like `auto`.  This may not be feasible here, depending on how complicated tings are.

As for a clever version of clear dependent...
```coq
Ltac revert_if_in_goal v := lazymatch goal with |- context[v] => revert dependent v end.
repeat match goal with
       | [ v : Z |- _ ] => revert_if_in_goal v
       | [ v : word _ |- _ ] => revert_if_in_goal v
       | [ v : bool |- _ ] => revert_if_in_goal v
       end;
clear; intros.
```
?

> Is the risk with the names that Coq might introduce an

I don't actually know the rules for global names specified in `destruct` in `Ltac`.  It's probably safe...

Note that if you make `state` a primitive record with named fields, you won't need to destruct it, and you can actually talk about named fields of your state.

jadephilipoom:
I've been referencing this proof for my own work-related reasons recently, and I think it's a bit of a shame we haven't merged it. Maintaining this proof so that it's easy to download, check, and cite would cost us virtually nothing and perform a valuable service to at least one member of the cryptographic community. Does anyone object to me taking ownership of this enough to fix the merge conflict and any other (e.g. Coq version) compatibility issues, and then merging it? I might not have time to implement all the suggested tactic improvements, but as long as the proofs don't take too long I don't think we should continue blocking on that.

JasonGross:
> Does anyone object to me taking ownership of this enough to fix the merge conflict and any other (e.g. Coq version) compatibility issues, and then merging it? I might not have time to implement all the suggested tactic improvements, but as long as the proofs don't take too long I don't think we should continue blocking on that.

Sounds good, let's merge and then improve later

PR #334, opened 2018-03-26
rowwise flatten (with merge conflicts fixed)

jadephilipoom:
From #324; I just rebased on a new branch. Let me know if there is some other preferred strategy to resolve all those conflicts. Otherwise I'll just merge this in a day or so.


src/Util/ListUtil.v
```diff
@@ -1845,3 +1869,10 @@ Ltac expand_lists _ :=
     | _ => idtac
     end;
     subst v; reflexivity ].
+
+Lemma list_rect_to_match A (P:list A -> Type) (Pnil: P nil) (PS: forall a tl, P (a :: tl)) ls :
+  @list_rect A P Pnil (fun a tl _ => PS a tl) ls = match ls with
```
andres-erbsen: Please either add a comment describing that this lemma applies to non-recursive uses of `list_rect`, or change the name to something that implies this, or both.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -586,20 +623,90 @@ Module MulSplit.
...
-      repeat match goal with
-              | _ => progress (autorewrite with push_eval in * )
-              | _ => progress simpl flat_map
-              | _ => rewrite IHp
-              | _ => ring_simplify; omega
-              end.
+        cbv [sat_mul]; induction p; [reflexivity|].
+        repeat match goal with
+               | _ => progress (autorewrite with push_eval in * )
+               | _ => progress simpl flat_map
```
andres-erbsen: `rewrite flat_map_cons`? or `autorerwite with push_flat_map`?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -586,20 +623,90 @@ Module MulSplit.
...
       Hint Rewrite eval_sat_mul : push_eval.
     End Associational.
   End Associational.
-End MulSplit.
+
+  Section DivMod.
+    Lemma mod_step a b c d: 0 < a -> 0 < b ->
+                            c mod a + a * ((c / a + d) mod b) = (a * d + c) mod (a * b).
+    Proof.
+      intros; rewrite Z.rem_mul_r by omega. push_Zmod.
```
andres-erbsen: `Z.div_mod_to_quot_rem; nia`? Also for the next couple of lemmas.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -586,20 +623,90 @@ Module MulSplit.
...
+               | _ => rewrite IHp
+               | _ => ring_simplify; omega
+               end.
       Qed.
       Hint Rewrite eval_sat_mul : push_eval.
     End Associational.
   End Associational.
-End MulSplit.
+
+  Section DivMod.
```
andres-erbsen: I do not understand the motivation behind the lemmas in this section. Probably you factored them out from some place below? Why?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -617,290 +724,703 @@ Module Columns.
...
-      length inp = n ->
-      (Positional.eval weight n (fst (flatten inp))
-       = (eval n inp) mod (weight n))
-        /\ (snd (flatten inp) = eval n inp / weight n).
-    Proof.
-      (* to make the invariant take the right form, we make everything depend on output length, not input length *)
-      intro. subst n. rewrite <-(flatten_length inp). cbv [flatten].
-      unfold flatten_step; fold flatten_step.
-      induction inp using rev_ind;
+      Lemma flatten_partitions inp:
```
andres-erbsen: Perhaps it would be easier to move this up and use it to prove `flatten_div_mod`? Either way is fine by me.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -617,290 +724,703 @@ Module Columns.
...
+      
+      (* from_associational *)
+      Definition from_associational n (p:list (Z*Z)) : list (list Z) :=
+        List.fold_right (fun t ls =>
+                           let p := Positional.place weight t (pred n) in
+                           cons_to_nth (fst p) (snd p) ls ) (nils n) p.
+      Lemma length_from_associational n p : length (from_associational n p) = n.
+      Proof. cbv [from_associational]. apply fold_right_invariant; intros; distr_length. Qed.
+      Hint Rewrite length_from_associational: distr_length.
+      Lemma eval_from_associational n p (n_nonzero:n<>0%nat\/p=nil):
```
andres-erbsen: I feel like I have seen this proof before. Perhaps it is the case that the positional representation is naturally parametric over the representation used for each position -- be it a single Z or a column list?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -617,290 +724,703 @@ Module Columns.
...
+  End Columns.
+End Columns.
 
-    Lemma flatten_snoc x inp : flatten (inp ++ [x]) = flatten_step x (flatten inp).
-    Proof. cbv [flatten]. rewrite rev_unit. reflexivity. Qed.
+Module Rows.
+  Import Saturated.
+  Section Rows.
+    Context (weight : nat->Z)
+            {weight_0 : weight 0%nat = 1}
```
andres-erbsen: If you would like, the properties required from a uniform `weight` could be aggregated into a record, the record taken as a context variable, and `Let pf := r.(pf)` used to extract the proofs. This would make instantiating these sections easier.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -617,290 +724,703 @@ Module Columns.
...
+      | H: _ \/ _ |- _ => destruct H
+      | _ => progress break_match; try discriminate
+      end.
+
+    Section Flatten.
+      Section flatten_column.
+        Context (fw : Z). (* maximum size of the result *)
+
+        (* Outputs (sum, carry) *)
+        Definition flatten_column (digit: list Z) : (Z * Z) :=
```
andres-erbsen: I can't parse this definition. Perhaps better notation would help. @JasonGross would it be practical to transform well-behaved fixpoints and matches into corresponding uses of eliminators during reification?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -617,290 +724,703 @@ Module Columns.
...
+             | H: forall x, In x (?y :: ?ls) -> ?P |- _ =>
+               unique pose proof (H y ltac:(apply in_eq));
+               unique assert (forall x, In x ls -> P) by auto
              end.
-    Qed.
 
-    (* TODO: move to ZUtil *)
-    Lemma Z_divide_div_mul_exact' a b c : b <> 0 -> (b | a) -> a * c / b = c * (a / b).
-    Proof. intros. rewrite Z.mul_comm. auto using Z.divide_div_mul_exact. Qed.
+    Section FromAssociational.
```
andres-erbsen: `Section FromColumns`?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -617,290 +724,703 @@ Module Columns.
...
+        rewrite Columns.length_from_associational in *; auto.
+      Qed.
+    End FromAssociational.
+
+    Section Flatten.
+      Local Notation fw := (fun i => weight (S i) / weight i) (only parsing).
+
+      Section SumRows.
+        Definition sum_rows' start_state (row1 row2 : list Z) : list Z * Z :=
+          fold_right (fun next (state : list Z * Z) =>
```
andres-erbsen: `fold_right` on `rev` is the same as `fold_left`, right?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -617,290 +724,703 @@ Module Columns.
...
+        match goal with H: _ |- _ => apply length_from_columns in H end.
+        rewrite Columns.length_from_associational in *; auto.
+      Qed.
+    End FromAssociational.
+
+    Section Flatten.
+      Local Notation fw := (fun i => weight (S i) / weight i) (only parsing).
+
+      Section SumRows.
+        Definition sum_rows' start_state (row1 row2 : list Z) : list Z * Z :=
```
andres-erbsen: `fold_adc`?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -617,290 +724,703 @@ Module Columns.
...
+            length row2' = m ->
+            (nm = n + m)%nat ->
+            let eval := Positional.eval weight in
+            is_div_mod (eval m) start_state (eval m row1' + eval m row2') (weight m) ->
+            length (fst (sum_rows' start_state row1 row2)) = nm
+            /\ is_div_mod (eval nm) (sum_rows' start_state row1 row2)
+                          (eval nm (row1' ++ row1) + eval nm (row2' ++ row2))
+                          (weight nm).
+        Proof.
+          induction row1 as [|x1 row1]; destruct row2 as [|x2 row2]; intros; subst nm; push.
```
andres-erbsen: Please make the distinction between the base case and the inductive case syntactically clear. If at some point in a `;` line there should be only one goal, you can assert that using `; [];`.

andres-erbsen:
Note: I am guessing some of the intermediate commits do not build; in that case I think this is best merged using a true merge commit, not rebase+fast-forward. Or if you think the immediate commits are totally arbitrary, feel free to squash them together instead.

andres-erbsen:
Also, I am trusting that you and Jason have reached consensus on the changes requested in #324.

jadephilipoom:
I believe we have consensus on #324, yes. I even removed the inlining part we disagreed on, so my scan shows no loose threads from that PR. (Jason, let me know if you still have any concerns.)

About the intermediate commits--I did build as I was rebasing, any time there were conflicts, so I think the intermediate ones are actually okay. Jason and I were mostly working on separate pieces of the code.

src/Experiments/SimplyTypedArithmetic.v
jadephilipoom: I factored them out from some of the more complex `_partitions` proofs in `Columns` and `Rows`. A lot of those proofs have elements in common--the general pattern being that at each step they start with a chunk of the input they currently represent, and a modulus, and the values of that chunk mod the modulus and div the modulus. Each step, they increase the modulus and take another chunk of input. There probably is a more elegant factorization that will actually make these proofs significantly nicer (or maybe just a broader application of these) but I haven't yet found it. It's on my radar, though, which I can formalize by opening an issue if you'd like.

src/Experiments/SimplyTypedArithmetic.v
jadephilipoom: It solves `div_step` but not the others.

src/Experiments/SimplyTypedArithmetic.v
jadephilipoom: The issue here really is using `list_rect` instead of a match. I could put the match-based definition in comments?

src/Experiments/SimplyTypedArithmetic.v
jadephilipoom: I did briefly try that, but for now found it easier to prove them separately. However, maybe this could be revisited when I try to find a more elegant factorization for these proofs.

src/Experiments/SimplyTypedArithmetic.v
jadephilipoom: Somewhat. Actually, what would make the most sense is to have `Positional.from_associational` go through `Columns`--just naively summing each column. It makes sense that the columns representation, which just collects and doesn't combine the associational terms, would be the basic step in converting from associational. 

The downside, though, is that it makes our positional-associational story, which we have been using as our primary demo, more complicated.

src/Experiments/SimplyTypedArithmetic.v
jadephilipoom: This is a good idea. I'll open an issue for it but I don't want to block the merge on it--sound fair?

src/Experiments/SimplyTypedArithmetic.v
jadephilipoom: Well, we only define `from_columns` in order to get `from_associational`. And it has more symmetry with the sections in `Columns` this way.

src/Experiments/SimplyTypedArithmetic.v
jadephilipoom: But `fold_left` is not in our reification pipeline, and it didn't seem simple to add it either. Would be fine to change this if `fold_left` is added.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Handling `match` is quite simple.  We already do this for `bool_rect`.  Simply add a case that looks like
```coq
| match ?ls with nil => ?N | cons x xs => ?C end
=> let T := type of term in
   reify_rect (@list_rect _ (fun _ => T) N (fun x xs _ => C))
```
I'm not sure how good Ltac is at matching `fix`, and I don't know if it can match on various choices of structural argument, but if you know how many arguments your `fix` has, and which one is structural, you're free to add such a case.

JasonGross: Oh, urg, actually, it's a bit more complicated.  But I can handle it.  I'll add a non-recursive `list_case` and rewrite it to `list_rect` during list canonicalization.

JasonGross: I've added https://github.com/mit-plv/fiat-crypto/commit/111acd333fef4681779e57eaa2579e18568beaa5, which adds support for `match` on `list`.  Note that with this change, you no longer need to add `tl` and `hd` separately.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1683,6 +2195,8 @@ Module Compilers.
           | List_partition {A} : ident ((A -> bool) * list A) (list A * list A)
           | List_app {A} : ident (list A * list A) (list A)
           | List_rev {A} : ident (list A) (list A)
+          | List_tl {A} : ident (list A) (list A)
```
JasonGross: No longer needed; please remove

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1683,6 +2195,8 @@ Module Compilers.
...
+          | List_hd {A} : ident (A * list A) A
```
JasonGross: No longer needed; please remove

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1748,6 +2264,8 @@ Module Compilers.
                | List_partition A => curry2 (@List.partition (type.interp A))
                | List_app A => curry2 (@List.app (type.interp A))
                | List_rev A => @List.rev (type.interp A)
+               | List_tl A => @List.tl (type.interp A)
```
JasonGross: No longer needed; please remove

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1748,6 +2264,8 @@ Module Compilers.
...
+               | List_hd A => curry2 (@List.hd (type.interp A))
```
JasonGross: No longer needed; please remove

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1856,6 +2376,12 @@ Module Compilers.
             | @List.rev ?A ?ls
               => let rA := type.reify A in
                  mkAppIdent (@ident.List_rev rA) ls
+            | @List.tl ?A ?ls
```
JasonGross: No longer needed; please remove

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1856,6 +2376,12 @@ Module Compilers.
...
+              => let rA := type.reify A in
+                 mkAppIdent (@ident.List_tl rA) ls
+            | @List.hd ?A ?x ?ls
```
JasonGross: No longer needed; please remove

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1910,6 +2436,8 @@ Module Compilers.
             Notation partition := List_partition.
             Notation app := List_app.
             Notation rev := List_rev.
+            Notation tl := List_tl.
```
JasonGross: No longer needed; please remove

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1910,6 +2436,8 @@ Module Compilers.
...
+            Notation hd := List_hd.
```
JasonGross: No longer needed; please remove

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2459,6 +3001,28 @@ Module Compilers.
                                      (fun x l' rev_l' => List_app rev_l' [x])
                                      ls) in
                     let v := app_and_maybe_cancel v in exact v)
+             | for_reification.ident.List_tl A
```
JasonGross: No longer needed; please remove

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -2459,6 +3001,28 @@ Module Compilers.
...
+                    let v := reify
+                               (@expr var)
+                               (fun ls
+                                => list_rect
+                                     (fun _ => list (type.interp A))
+                                     nil
+                                     (fun _ l' _ => l')
+                                     ls) in
+                    let v := app_and_maybe_cancel v in exact v)
+             | for_reification.ident.List_hd A
```
JasonGross: No longer needed; please remove

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -4498,7 +5068,20 @@ Module Compilers.
                        let result := ident.interp idc (x, y, z, a) in
                        inr (inr (fst result), inr (snd result))
                      | inr (inr (inr (inr x, y), z), a)
-                        => default_interp (ident.Z.add_with_get_carry_concrete x) (inr (inr (y, z), a))
+                       => let default_with_carry := default_interp (ident.Z.add_with_get_carry_concrete x) (inr (inr (y, z), a)) in
```
JasonGross: Please thunk the defaults, so that they behave well in the cbv strategy.  See `let default _ := ...` above.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -4498,7 +5068,20 @@ Module Compilers.
...
+                          let default_no_carry := default_interp (ident.Z.add_get_carry_concrete x) (inr (z,a)) in
```
JasonGross: Please thunk the defaults, so that they behave well in the cbv strategy.  See `let default _ := ...` above.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -4498,7 +5068,20 @@ Module Compilers.
...
+                          let default := match y with
```
JasonGross: Please thunk the defaults, so that they behave well in the cbv strategy.  See `let default _ := ...` above.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -4578,10 +5161,12 @@ Module Compilers.
                                AppIdent idc (expr.reify (t:=type.Z) x))
                      end
              | ident.Nat_add as idc
+             | ident.Nat_sub as idc
              | ident.Nat_mul as idc
+             | ident.Nat_max as idc
+             | ident.Z_pow as idc
```
JasonGross: Nit: `ident.Z_pow` doesn't need to be reordered in this PR

JasonGross:
Actually, there is no hope of reifying arbitrary fixpoints, as
```coq
Goal True.
  lazymatch (eval cbv delta [Nat.add] in Nat.add) with
  | (fix F (n : nat) (v : ?A) := _)
    => idtac
  end.
```
gives `Error: Non supported pattern.`

andres-erbsen:
re intermediate commits: okay, great. nevermind my concerns then. 

FYI: I would like to start reworking the interface to the compilers late this week, so you will probably avoid a bunch of rebasing if you merge it before. If you would like me to hold off merging my stuff for a bit, that is an option, but at a least at a glance this PR does not look far from being ready. (and if there will be conflicts, I am sure we will manage to figure them out -- this note is just a micro-optimization).

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Addendum: and Ltac doesn't even support matching on `fix`, see https://github.com/coq/coq/issues/7092

jadephilipoom:
Jason's comments should be resolved now.

PR #335, opened 2018-03-26
comprehensive loops framework with complete proof theory

andres-erbsen:
All the removals are dangling experiments about modeling loops. This one incorporates all modeling goodness from them, but does not include all nice notations we have previously created. Unlike previous attempts, this version includes a proof that the proof theory of loop invariants and decreasing measures is complete:

~~~~~~~~
    Lemma invariant_iff P f s0 :
      (exists b, loop f s0 = inr b /\ P b)
      <->
      (exists inv measure,
          (inv s0 /\ measure s0 <= f)
          /\ forall s, inv s -> match body s with
                                | inl s' => inv s' /\ measure s' < measure s
                                | inr s' => P s'
                                end).
~~~~~~~~


JasonGross:
Is there something in particular you want me to review here?

andres-erbsen:
I am primarily asking for a sanity-check of the interface and semantics this file provides. I think you should only read lemma statements + definitions, excluding `Local`, and perhaps start skimming once it gets monotonous (it should).

You might also want to check whether we should keep anything (notations) from the removed code.

andres-erbsen:
If nobody objects by Monday, I will merge this.

PR #337, opened 2018-03-27
Remove rweight 

JasonGross:


andres-erbsen:
~~~~~~~~~~~~
File "./src/Experiments/SimplyTypedArithmetic.v", line 7450, characters 14-21:
Error: The reference rweight was not found in the current environment.
~~~~~~~~~~~~~~

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -420,118 +420,174 @@ Module Positional. Section Positional.
...
+          (n : nat)
+          (len_c : nat)
+          (idxs : list nat)
+          (len_idxs : nat)
+          (m_nz:s - Associational.eval c <> 0) (s_nz:s <> 0)
+          (Hn_nz : n <> 0%nat)
+          (Hc : length c = len_c)
+          (Hidxs : length idxs = len_idxs).
+  Definition weight (i : nat)
+    := 2^(-(-(limbwidth_num * i) / limbwidth_den)).
```
andres-erbsen: is the double negation used here to make it round up? I don't know whether the trick here works, but I would believe you if you said it does. In C I often use (n + d - 1)/d for rounding up. Edit: hah, you proved it correct below. I approve of both the trick and the fact that the code was already set up in a way to make it the right thing to do to prove it.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -420,118 +420,174 @@ Module Positional. Section Positional.
...
-    Proof.
-      intros.
-      etransitivity;
-        [ | rewrite <- @eval_chained_carries with (s:=s) (c:=c) (idxs:=idxs)
-            by auto; reflexivity ].
-      eapply f_equal2; [|trivial]. eapply f_equal.
-      expand_lists ().
-      subst carrymod; reflexivity.
-    Qed.
+  Local Lemma weight_0 : weight 0 = 1.
```
andres-erbsen: Could you get this lemma and the next from lemmas about `2^Qceiling(limbwidth*i)` ? Or are your proofs intentionally duplicative because they look simpler?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -6419,21 +6418,21 @@ Section rcarry_mul.
   Definition rcarry_mul
     := BoundsPipeline
          (carry_mul_gen
-            @ rw @ GallinaReify.Reify s @ GallinaReify.Reify c @ GallinaReify.Reify n @ GallinaReify.Reify (length c) @ GallinaReify.Reify idxs @ GallinaReify.Reify (length idxs))
+            @ GallinaReify.Reify (Qnum limbwidth) @ GallinaReify.Reify (Z.pos (Qden limbwidth)) @ GallinaReify.Reify s @ GallinaReify.Reify c @ GallinaReify.Reify n @ GallinaReify.Reify (length c) @ GallinaReify.Reify idxs @ GallinaReify.Reify (length idxs))
```
andres-erbsen: It might be easier to take in `(Z*Z)` instead of `Q` for the limbwidth. either is fine by me.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -6591,6 +6584,7 @@ Section rcarry_mul.
                      | intros; apply eval_oppmod
                      | intros; apply eval_encodemod
                      | eassumption
+                     | apply conj
```
andres-erbsen: I remember having issues with non-conditional (e?)apply conj looping. if you think it's fine here then it's most likely fine

src/Experiments/SimplyTypedArithmetic.v
JasonGross: I got the "trick" by unfolding `Qceiling (Z.pos (Z.to_pos num) / Z.to_pos den)` and then removing the matches

src/Experiments/SimplyTypedArithmetic.v
JasonGross: Huh?  `limbwidth` is computed in this section.  I could also let-bind the numerator and denominator....

src/Experiments/SimplyTypedArithmetic.v
JasonGross: If it starts looping, I'll get rid of it.  But this is really a very simple proof that goes by `eauto` modulo beta-iota-delta.

src/Experiments/SimplyTypedArithmetic.v
JasonGross: I was originally trying to be as permissive as possible.  I'll strengthen the hypotheses to require `0 < limbwidth_den` which is required for basing it on `2^Qceiling(limbwidth*i)` to actually be a simplification.

JasonGross:
Updated to build to the end by plugging in a concrete weight function and proving a few extra lemmas.

JasonGross:
@jadephilipoom Are you okay with the changes to montred that inline the weight function?  I tried to do things in a way that minimally impacted your proof (so definitely feel free to later rewrite it more or whatever).

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -7357,19 +7355,43 @@ Module MontgomeryReduction.
...
+    Let w_nonzero : forall i, w i <> 0
+      := weight_nz _ _ log2R_good.
+    Let w_positive : forall i, w i > 0
+      := weight_positive _ _ log2R_good.
+    Let w_multiples : forall i, w (S i) mod w i = 0
+      := weight_multiples _ _ log2R_good.
+    Let w_divides : forall i : nat, w (S i) / w i > 0
+      := weight_divides _ _ log2R_good.
+    Let w_1_gt1 : w 1 > 1 := weight_1_gt_1 _ _ log2R_good.
+    Let w_half_1_gt1 : w_half 1 > 1 := weight_1_gt_1 _ _ half_log2R_good.
```
jadephilipoom: Ugh, this is quite a lot of boilerplate. I was thinking of maybe making a record or something for saturated weight functions, and maybe this is another point for that strategy.

PR #339, opened 2018-03-29
Do some pipeline consolidation and reorganization

JasonGross:
- Since we pass through a flat representation, we no longer need to
  invoke `vm_compute` twice, so we don't need to separate out bounds
  checking from bounds analysis

- Only call partial evaluation once, not twice

- Remove the call to constant reassociation, as it no longer has a place
  (it only works post-partial-evaluation pre-bounds-analysis)

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -4231,6 +4231,91 @@ Module Compilers.
     End expr.
   End DefaultValue.
 
+  Module GeneralizeVar.
+    (** In both lazy and cbv evaluation strategies, reduction under
+        lambdas is only done at the very end.  This means that if we
+        have a computation which returns a PHOAS syntax tree, and we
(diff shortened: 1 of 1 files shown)


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -6025,62 +6027,67 @@ Module Pipeline.
       expr.Interp (@for_reification.ident.interp) E.
   Admitted.
 
-  Definition BoundsPipelineNoCheck
+  Definition BoundsPipeline_0_or_1
```
andres-erbsen: What does this name mean?

JasonGross: "0 or 1 arguments to the expression".  It'll go away when the two variants are consolidated when I finish the uncurrying pass.

JasonGross:
> subst01 is purely aesthetic, right?

I think so

PR #340, opened 2018-04-01
Fix the pipeline to not insert extra casts

JasonGross:
So long as we have subst01, it must come before bounds analysis (or
followed by an extra pass of partial evaluation), because bounds
analysis inserts casts at every let-in node, and so inlining some of
them results in duplicated cast nodes, which we don't want to have.


JasonGross:
From signal:
@andres-erbsen said:
> i currently don't know how they are generated, maybe let's look into it sometime not so immediate future

I said:
> they are generated when bounds analysis runs on something of the form
>
>     let x := <expr> in
>     let y := <something mentioning x exactly once> in
>     ...
>
> which becomes
>
>     let x := Z.cast <expr> in
>     let y := <something mentioning Z.cast x exactly once> in
>     ...
>
> and when subst01 inlines x, y now mentions `Z.cast (Z.cast <expr>)` exactly once
> we insert the cast in both places because variables do not carry bounds information and so, e.g., if the pretty-printer wants to know variable widths from local information, it needs to have Z.cast in front of both arguments to + (or whatever)

@andres-erbsen said:
> okay, I don't see an obvious solution right now

and okay'ed the merge.

PR #343, opened 2018-04-06
barrett quotient error

jadephilipoom:
The estimated quotient in Barrett reduction (`q`, here) is known to have an error of at most 2--that is, using the variable names in the `Generalized.v` file, `q = a  / n - e` and `0 <= e <= 2`.

However, it turns out that for primes that fit a certain condition (see `n_good` in this PR), we can prove `0 <= e <= 1`, which lets us skip a modular reduction.

@JasonGross These proofs are significantly messier than the ones you wrote in `Generalized`--if you have tips for better utilizing the automation I'd appreciate it.

### DIFF
--- a/src/Arithmetic/BarrettReduction/Generalized.v
+++ b/src/Arithmetic/BarrettReduction/Generalized.v
@@ -142,55 +142,40 @@ Section barrett.
 
       Lemma helper_1 : b ^ (2 * k) * ((a / n) - 1) <= m * (n * (a / n) - b ^ (k - offset)).
       Proof.
-        autorewrite with push_Zmul. ring_simplify.
-        replace (a / n * m * n) with (m * n * (a / n)) by ring.
-        assert (a/n < b ^ k) by auto using Z.div_lt_upper_bound with zarith.
-        assert (b ^ (2 * k) - m * n = b ^ (2 * k) mod n).
-        { subst m.
-          rewrite Z.mul_div_eq' by (auto using Z.lt_gt with zarith).
-          auto with zarith. }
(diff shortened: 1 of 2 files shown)


JasonGross:
Seems to not build with Coq master...

JasonGross:
> if you have tips for better utilizing the automation I'd appreciate it.

#345 .  The "insight" was "`Z.div_mod_to_quot_rem` is powerful".

jadephilipoom:
That's weird--it looks like the build failed, even though the same build succeeded on #345. Error is:

`Compiled library Crypto.Spec.WeierstrassCurve (in file /home/travis/build/mit-plv/fiat-crypto/src/Spec/WeierstrassCurve.vo) makes inconsistent assumptions over library Coq.Classes.CRelationClasses`

...which suggests to me that perhaps this is some issue with the build rather than the code? There's no difference between the code at those two commits.

jadephilipoom:
Adding a comment apparently fixed the build. We may want to investigate that, but I'm not really sure where to start and am pretty confident it has little to do with the code in this PR, so I'll merge it.

JasonGross:
The issue is definitely something to do with how we're caching vo files (which we do so that we can build more of the project than 1 hour's worth).  The error message is because the version of Coq master that got installed here was different from the version of master used to build the cached .vo.  I suspect that different PRs building at the same time interfere with each other's build caches, or (or maybe and) the daily-master build on launchpad got updated in the middle of the build.  (Except I already restarted the entire build once, and that didn't fix it, but there was another build still going at the same time.)

PR #347, opened 2018-04-11
Fix fesub

JasonGross:
We now compute tight bounds as 1.1*(s-1), and make reassociation optional, so that synthesizing sub works.

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -431,7 +431,10 @@ Module Positional. Section Positional.
   End sub.
   Hint Rewrite @eval_opp @eval_sub : push_eval.
   Hint Rewrite @length_sub @length_opp : distr_length.
-End Positional. End Positional.
+End Positional.
+(* Hint Rewrite disappears after the end of a section *)
+Hint Rewrite length_zeros length_add_to_nth length_from_associational @length_add @length_carry_reduce @length_chained_carries @length_encode @length_sub @length_opp : distr_length.
+End Positional.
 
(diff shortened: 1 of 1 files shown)


andres-erbsen:
why would reassociation mess things up here? I would say it should be fixed not to affect sub, but perhaps there is something I am missing.

andres-erbsen:
IIRC we talked about this in person and concluded reassociation should be limited further instead of working it around this case by making it conditional.

JasonGross:
Yes, that's correct.  Now updated.  The change to reassocation is now +4,-31

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -6776,18 +6749,29 @@ Section rcarry_mul.
...
   Definition relax_zrange_of_machine_wordsize
     := relax_zrange_gen [machine_wordsize; 2 * machine_wordsize]%Z.
 
   Let relax_zrange := relax_zrange_of_machine_wordsize.
+  Definition adjust_list_to_length {A} (default : A) (ls : list A) (len : nat)
+    : list A
+    := List.map
+         (fun i => List.nth_default default ls i)
+         (seq 0 len).
+  Let tight_upperbounds : list Z
```
andres-erbsen: Why do we need this? Isn't the original list already length `n`?

JasonGross: Indeed, I forgot that `length_encode` was already proven.  I'll fix it.

JasonGross: Updated to remove the adjusting definition

Issue #344, opened 2018-04-07
fiat-crypto has been failing on Coq Jenkins for 2 months


andres-erbsen:
(from https://github.com/coq/coq/issues/7188)

There seems to be a good mix of different reasons, many of which look like resource limit issues.

Example failure:
https://ci.inria.fr/coq/job/opam-install-coq-fiat-crypto/653/consoleFull
~~~~~~~~~~~~
- "coqc"  -q   -w "-notation-overridden"  src/Specific/X25519/C64/fecarryDisplay.v | sed s'/\r\n/\n/g' > src/Specific/X25519/C64/fecarryDisplay.log.tmp && mv -f src/Specific/X25519/C64/fecarryDisplay.log.tmp src/Specific/X25519/C64/fecarryDisplay.log
- "coqc"  -q   -w "-notation-overridden"  src/Specific/X25519/C64/feaddDisplay.v | sed s'/\r\n/\n/g' > src/Specific/X25519/C64/feaddDisplay.log.tmp && mv -f src/Specific/X25519/C64/feaddDisplay.log.tmp src/Specific/X25519/C64/feaddDisplay.log
- "coqc"  -q   -w "-notation-overridden"  src/Specific/X25519/C64/fesubDisplay.v | sed s'/\r\n/\n/g' > src/Specific/X25519/C64/fesubDisplay.log.tmp && mv -f src/Specific/X25519/C64/fesubDisplay.log.tmp src/Specific/X25519/C64/fesubDisplay.log
- "coqc"  -q   -w "-notation-overridden"  src/Specific/NISTP256/AMD64/fesubDisplay.v | sed s'/\r\n/\n/g' > src/Specific/NISTP256/AMD64/fesubDisplay.log.tmp && mv -f src/Specific/NISTP256/AMD64/fesubDisplay.log.tmp src/Specific/NISTP256/AMD64/fesubDisplay.log
- File "./src/Specific/NISTP256/AMD64/fesubDisplay.v", line 1, characters 15-51:
- Error:
- Cannot find a physical path bound to logical path matching suffix
- Crypto.Specific.NISTP256.AMD64.
- 
- File "./src/Specific/X25519/C64/fesubDisplay.v", line 1, characters 15-47:
- Error:
- Cannot find a physical path bound to logical path matching suffix
- Crypto.Specific.X25519.C64.
- 
- File "./src/Specific/X25519/C64/fecarryDisplay.v", line 1, characters 15-49:
- Error:
[...]

PR #348, opened 2018-04-15
Error if a display target fails

JasonGross:
Because pipes eat error codes, we were previously succeeding when
display targets fail.  This meant that we didn't catch
https://github.com/mit-plv/fiat-crypto/issues/344#issuecomment-381422896
on Travis.  Now we will.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -226,7 +226,8 @@ $(DISPLAY_NON_JAVA_VO:.vo=.log) : %Display.log : %.vo src/Compilers/Z/CNotations
 
 $(DISPLAY_NON_JAVA_VO:.vo=.log) : %.log : %.v
 	$(SHOW)'COQC $< > $@'
-	$(HIDE)$(TIMER) $(COQC) $(COQDEBUG) $(COQFLAGS) $< | sed s'/\r\n/\n/g' > $@.tmp && mv -f $@.tmp $@
+	$(HIDE)$(TIMER) $(COQC) $(COQDEBUG) $(COQFLAGS) $< > $@.tmp
+	$(HIDE)sed s'/\r\n/\n/g' $@.tmp > $@ && rm -f $@.tmp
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -226,7 +226,9 @@ $(DISPLAY_NON_JAVA_VO:.vo=.log) : %Display.log : %.vo src/Compilers/Z/CNotations
 
 $(DISPLAY_NON_JAVA_VO:.vo=.log) : %.log : %.v
 	$(SHOW)'COQC $< > $@'
-	$(HIDE)$(TIMER) $(COQC) $(COQDEBUG) $(COQFLAGS) $< | sed s'/\r\n/\n/g' > $@.tmp && mv -f $@.tmp $@
+	$(HIDE)rm -f $@.ok
+	$(HIDE)($(TIMER) $(COQC) $(COQDEBUG) $(COQFLAGS) $(COQLIBS) $< && touch $@.ok) | sed s'/\r\n/\n/g' > $@.tmp && mv -f $@.tmp $@
+	$(HIDE)rm $@.ok
```
andres-erbsen: How about instead first redirect to a tmp without sed and then sed from tmp to output? No pipes, so all errors should be propagated.

JasonGross: Good idea, updated

andres-erbsen:
~~~~~
COQC src/Curves/Weierstrass/Projective.v
File "./src/Curves/Weierstrass/Projective.v", line 1, characters 0-44:
Error:
Compiled library Crypto.Spec.WeierstrassCurve (in file /home/travis/build/mit-plv/fiat-crypto/src/Spec/WeierstrassCurve.vo) makes inconsistent assumptions over library Coq.PArith.BinPos
~~~~~
restarted

JasonGross:
I'm going to wait to make sure that the "error if a display target fails" commit errors on https://travis-ci.org/JasonGross/fiat-crypto/builds/367331442 on the display target.

PR #350, opened 2018-04-17
Add some instructions in preparation for reifying barrett

jadephilipoom:
I reprinted the code for Montgomery and X25519_64, and nothing is affected. This should be pretty mechanical, but I'll make a PR to pass it under @JasonGross's sanity check.

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -2422,8 +2422,11 @@ Module Compilers.
           | Z_add_get_carry : ident (Z * Z * Z) (Z * Z)
           | Z_add_with_get_carry : ident (Z * Z * Z * Z) (Z * Z)
           | Z_sub_get_borrow : ident (Z * Z * Z) (Z * Z)
+          | Z_sub_with_get_borrow : ident (Z * Z * Z * Z) (Z * Z)
           | Z_zselect : ident (Z * Z * Z) Z
           | Z_add_modulo : ident (Z * Z * Z) Z
+          | Z_rshi : ident (Z * Z * Z * Z) Z
+          | Z_cc_m : ident (Z * Z) Z
           .
 
           Notation curry0 f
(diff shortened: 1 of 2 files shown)


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -4172,11 +4222,18 @@ Module Compilers.
                 | ident.Z_add_get_carry as idc
                 | ident.Z_sub_get_borrow as idc
                 | ident.Z_add_with_get_carry_concrete _ as idc
+                | ident.Z_sub_with_get_borrow_concrete _ as idc
                   => λ (xyk :
                           (* ignore this line; it's to work around lack of fixpoint refolding in type inference *) var (type.Z * type.Z * type.Z * ((type.Z * type.Z) -> R))%ctype) ,
                      (ident.snd @@ (Var xyk))
                        @ ((idc : default.ident _ (type.Z * type.Z))
                             @@ (ident.fst @@ (Var xyk)))
+                | ident.Z_rshi_concrete _ _ as idc
```
JasonGross: This has type `Z * Z -> Z`, and belongs in the same block as `ident.Z_div`, `ident.Z_cc_m`, etc.  Please move it there rather than introducing a new case.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5160,6 +5240,25 @@ Module Compilers.
                        => default_interp (ident.Z.mul_split_concrete x) (inr (y, z))
                      | _ => default_interp idc x_y_z
                      end
+             | ident.Z_rshi as idc
+               => fun (x_y_z_a : (_ * expr (_ * _ * _ * _) +
```
JasonGross: please put the type signature all on one line, or, better, on a separate line like
```coq
(x_y_z_a :
(* ignore this line, it's for issues with type inference / fixpoint refolding *) ...)
```

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5219,6 +5318,19 @@ Module Compilers.
                        => default_interp (ident.Z.sub_get_borrow_concrete x) (inr (y, z))
                      | _ => default_interp idc x_y_z
                      end
+             | ident.Z_sub_with_get_borrow as idc
+               => fun (x_y_z_a : (_ * expr (_ * _ * _ * _) +
```
JasonGross: same comment as above about type signature

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -5219,6 +5318,19 @@ Module Compilers.
...
+                                  (_ * expr (_ * _ * _) +
+                                   (_ * expr (_ * _) + (_ * expr _ + type.interp _) * (_ * expr _ + type.interp _)) *
+                                   (_ * expr _ + type.interp _)) * (_ * expr _ + type.interp _))%type)
+                  => match x_y_z_a return (_ * expr _ + (_ * expr _ + type.interp _) * (_ * expr _ + type.interp _)) with
+                     | inr (inr (inr (inr x, inr y), inr z), inr a) =>
+                       let result := ident.interp idc (x, y, z, a) in
```
JasonGross: Any reason not to use `let '(_, _) := ... in ...` rather than `fst` and `snd`?

PR #355, opened 2018-04-22
More experiments with partial reduction

JasonGross:


JasonGross:
Merging this, after much pair programming, so we can port the experiment back over to the full pipeline.  More experiments on rewriting to come.

PR #357, opened 2018-04-26
Saturated solinas reduction

jadephilipoom:
Seems to work for P-192, but P-256 is a huge, huge term that takes longer to process than I have so far been willing to wait. 

It makes sense that P-256 is huge--by my back-of-the-napkin calculation, it should have these numbers of terms:
pre-multiplication: 8 terms per operand
post-multiplication : 128 partial products (8 * 8 * 2 -- the x2 is because we use 2-output multiplication)
reduction 1 : 64 + 4*64 terms (since 64 are in the "low" part, and c has 4 terms)
...and so on, until after the 8 required reductions we have 1649 terms. I think this can be mediated with some cancelling-out of additions and subtractions of the same partial product, but haven't quite figured that part out yet.

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -673,6 +673,49 @@ Module Saturated.
                end.
       Qed.
       Hint Rewrite eval_sat_mul : push_eval.
+
+      Definition sat_multerm_const s (t t' : (Z * Z)) : list (Z * Z) :=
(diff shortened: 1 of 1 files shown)


jadephilipoom:
I want to merge it so it doesn't break with new changes. However, I am aware that we don't really want `SimplyTypedArithmetic.v` to get even longer/slower, so I would accept "move to a separate file" as a condition for the merge.

PR #359, opened 2018-04-27
Reifying Barrett code

jadephilipoom:
This doesn't need an extensive review--most of it doesn't touch parts of code we're all using--but I'd appreciate a quick glance over by @andres-erbsen or @JasonGross before I merge.


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -8145,22 +8194,169 @@ Module BarrettReduction.
...
+  Derive barrett_red_gen
+         SuchThat (forall (k M muLow : Z)
+                          (n nout: nat)
+                          (xLow xHigh : Z),
+                      Interp (t:=type.reify_type_of barrett_reduce)
+                             barrett_red_gen k M muLow n nout xLow xHigh
+                      = barrett_reduce k M muLow n nout xLow xHigh)
+         As barrett_red_gen_correct.
+  Proof. Time cache_reify (). exact admit. (* correctness of initial parts of the pipeline *) Time Qed.
+  (* TODO : reification here is still quite slow (~40s on a beefy machine). Possibly just due to size of term, but warrants further investigation. *)
```
JasonGross: Which part?  Is it reification itself, or the subsequent reduction / reflexivity?

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -1245,13 +1245,17 @@ Module Rows.
...
-                          (fst state ++ [fst sum_carry], snd sum_carry)) start_state (rev (combine row1 row2)).
-        Definition sum_rows := sum_rows' (nil,0).
+        Definition sum_rows' start_state (row1 row2 : list Z) : list Z * Z * nat :=
+          fold_right (fun next (state : list Z * Z * nat) =>
+                        let i := snd state in
+                        let low_high' :=
+                            dlet_nd low_high := fst state in
+                            let low := fst low_high in
+                            let high := snd low_high in
+                          dlet_nd sum_carry := Z.add_with_get_carry_full (fw i) high (fst next) (snd next) in
```
JasonGross: The indentation here is misleading

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -8145,22 +8194,169 @@ Module BarrettReduction.
...
+  Proof. Time solve_rbarrett_red machine_wordsize. Time Qed.
+
+  Import PrintingNotations.
+  Open Scope expr_scope.
+  Set Printing Width 100000.
+
+  Print barrett_red256.
+  (* TODO: the ADD/ADDC instructions containing Z.opp should be translated to SUB/SUBB in partial evaluation *)
+  (* Note: the aforementioned ADD/ADDC instructions currently *do* fail to bounds-check, although the pipeline doesn't give an error.
+     This is why their results are not cast (because the carry has range [-1~>0]). *)
```
JasonGross: Do you want the pipeline to give an error on this?

jadephilipoom: It would definitely be clearer to me if it had--if we can't coerce to any of the allowed bitwidths, I don't think that should be silent. However, I only want this if it's possible to make an error message that shows clearly which instruction fails to cast. Otherwise, I'd prefer to just look at the output.

src/Experiments/SimplyTypedArithmetic.v
jadephilipoom: Reduction/reflexivity, last I checked.

Issue #352, opened 2018-04-18
Translating add-with-carry to sub-with-borrow


jadephilipoom:
I want to put cases in partial evaluation so that
```
expr_let a := (ident.Z.add_get_carry_concrete s) @@ (x, ident.Z.opp @@ y) in
expr_let b := (ident.Z.add_with_get_carry_concrete s) @@ (ident.snd  @@ a, x', ident.Z.opp @@ y') in
expr_let c := ((Z.add_with_get_carry_concrete s) @@ (ident.snd @@ b, x'', ident.Z.opp @@ y'') in
(ident.fst @@ a, ident.fst @@ b, ident.fst @@ c)
```
becomes
```
expr_let a := (ident.Z.sub_get_borrow s) @@ (x, y) in
expr_let b := (ident.Z.sub_with_get_borrow_concrete s) @@ (ident.snd @@ a, x', y') in
expr_let c := (Z.sub_with_get_borrow_concrete s) @@ (ident.snd @@ b, x'', y'') in
(ident.fst @@ a, ident.fst @@ b, ident.fst @@ c)
```

Now, once upon a time we had this cute idea that we would make `Z.sub_get_borrow s x y` return `- ((x - y) / s` as the borrow, instead of just `(x - y) / s`. So, if `Z.add_get_carry s x (-y) = (a, c)`, then `Z.sub_get_borrow s x y = (a, -c)`. The `Z.opp` on `c` "disappears" because `Z.sub_with_get_borrow` *subtracts* incoming borrows, while `Z.add_with_get_carry` adds them.

Because of this choice, when I try to translate add-opp => sub in the pipeline, I end up with the issue that, to produce an equivalent expression, I have to negate the borrow that comes out of `Z.sub_{with_}get_borrow`. That means I have to destruct the pair, which is really not ideal--especially because the ouput is supposed to be flattened by that point. Ultimately, the problem is that I want to actually not produce an exactly equivalent expression, and I don't have the context to argue that all my transformations together actually come out equivalent.

Thoughts? My current ideas are a) to make a definition of `Z.sub_{with_}get_carry` that actually returns a non-negated borrow and b) to address this at the arithmetic level--basically creating an alternate version of `Rows.sum_rows` that subtracts rows. The latter option would suck, but would work. The former option might fail bounds-checking--at that point we would be passing around a negative number as our "borrow".

jadephilipoom:
[...]

PR #362, opened 2018-05-02
Straightline

jadephilipoom:
Pass that converts code to syntactically-straightline form and a second pass that forces code to use only a particular set of operations. Seems to work for Montgomery reduction and most of Barrett reduction (it fails to translate the last few lines because of #352). No proofs yet.

I would like a review from @andres-erbsen or @JasonGross of my pass to straightline (`Module Straightline`), before I devote too much time to proofs of this version. `Module PreFancy`, the second pass, is simpler and less critical to get a review on (in particular, I imagine anything that could use improvement there would pop up in `Straightline` also).

### DIFF
--- a/src/Experiments/SimplyTypedArithmetic.v
+++ b/src/Experiments/SimplyTypedArithmetic.v
@@ -5529,11 +5529,12 @@ Module Compilers.
              | ident.Z_mul_split_concrete _ as idc
              | ident.Z.sub_get_borrow_concrete _ as idc
                => fun (x_y : _ * expr (_ * _) + (_ * expr _ + type.interp _) * (_ * expr _ + type.interp _))
-                  => match x_y return (_ * expr _ + (_ * expr _ + type.interp _) * (_ * expr _ + type.interp _)) with
+                  => let default _ := default_interp idc x_y in
+                     match x_y return (_ * expr _ + (_ * expr _ + type.interp _) * (_ * expr _ + type.interp _)) with
                      | inr (inr x, inr y) =>
                        let result := ident.interp idc (x, y) in
                        inr (inr (fst result), inr (snd result))
-                     | _ => default_interp idc x_y
+                     | _ => default tt
                      end
              | ident.Z.add_get_carry_concrete _ as idc
                => fun (x_y : _ * expr (_ * _) + (_ * expr _ + type.interp _) * (_ * expr _ + type.interp _))
@@ -7746,7 +7747,7 @@ Module X25519_32.
   Import PrintingNotations.
   Print base_25p5_carry_mul.
 (*
-base_25p5_carry_mul = 
+base_25p5_carry_mul =
 fun var : type -> Type =>
 (λ x : var (type.list (type.type_primitive type.Z) * type.list (type.type_primitive type.Z))%ctype,
  expr_let x0 := x₁ [[0]] *₆₄ x₂ [[0]] +₆₄
@@ -7885,7 +7886,7 @@ fun var : type -> Type =>
              type.list (type.type_primitive type.Z)))
 *)
 End X25519_32.
-*)
+ *)
 
 Module BarrettReduction.
   (* TODO : generalize to multi-word and operate on (list Z) instead of T; maybe stop taking ops as context variables *)
@@ -8618,7 +8619,7 @@ Module P192_64.
   Open Scope expr_scope.
   Set Printing Width 100000.
   Set Printing Depth 100000.
-  
+
   Local Notation "'mul64' '(' x ',' y ')'" :=
     (Z.cast2 (uint64, _)%core @@ (Z.mul_split_concrete 18446744073709551616 @@ (x , y)))%expr (at level 50) : expr_scope.
   Local Notation "'add64' '(' x ',' y ')'" :=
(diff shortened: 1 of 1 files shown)


src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -7885,7 +7915,500 @@ fun var : type -> Type =>
...
 *)
 End X25519_32.
-*)
+ *)
+
+Module Straightline.
+  Module expr.
+    Section with_var.
+      Context {var : type.type -> Type}.
+      Context {dummy_var : forall t, var t}.
```
JasonGross: This line should not be here, it's not satisfiable in general.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -7885,7 +7915,500 @@ fun var : type -> Type =>
...
+             | Snd {a b} : scalar (a * b) -> scalar b
+             | Shiftr : Z -> scalar type.Z -> scalar type.Z
+             | Shiftl : Z -> scalar type.Z -> scalar type.Z
+             | Land : Z -> scalar type.Z -> scalar type.Z
+             | CC_m : Z -> scalar type.Z -> scalar type.Z
+             | Primitive {t} : type.interp (type.type_primitive t) -> scalar t
+        .
+      End with_ident.
+
+      Fixpoint dummy t : @expr ident.ident t := Scalar (Var t (dummy_var t)).
```
JasonGross: This should be a genuine fixpoint over the type, and it should be defined as something like `Scalar (Primitive DefaultValue.default)` (I may be misremembering what I called the thing that says `forall t, type.interp (type.type_primitive t)`) in the base case, and should `Pair` in the other cases.  You may run into trouble with arrows.  Maybe ask for `forall s d, var (s -> d)` as your context variable, and use that in the arrow case; the next version of the compilers stuff will remove the ability talk about pairs of arrows at the type level, and then this function can just not deal with arrows.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -7885,7 +7915,500 @@ fun var : type -> Type =>
...
+    match e with
+    | Uncurried.expr.Var _ _ => O
+    | Uncurried.expr.TT => O
+    | Uncurried.expr.AppIdent _ _ idc args => S (depth dummy_var args)
+    | Uncurried.expr.App _ _ f x => S (Nat.max (depth dummy_var f) (depth dummy_var x))
+    | Uncurried.expr.Pair _ _ x y => S (Nat.max (depth dummy_var x) (depth dummy_var y))
+    | Uncurried.expr.Abs _ _ f => S (depth dummy_var (f (dummy_var _)))
+    end.
+
+  (* TODO : Can I avoid having dummy_var appear here? *)
```
JasonGross: Replace `depth dummy_var (f x)` with `depth (var:=fun _ => unit) (fun _ => tt) (e (fun _ => unit))`.  The `dummy_var` argument of `expr.of_uncurried` and `expr.dummy` will go away, as per above, when we drop support for pairs of arrows.

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -7885,7 +7915,500 @@ fun var : type -> Type =>
...
+Module StraightlineTest.
+  Definition test : Expr (type.Z -> type.Z) :=
+    fun var =>
+      Abs
+        (fun (x : var type.Z) =>
+           AppIdent (var:=var) ident.Let_In
+               (Pair (AppIdent (var:=var) (ident.Z.cast r[0~>4294967295]%zrange) (AppIdent (var:=var) (ident.Z.shiftr 8) (Var x)))
+                     (Abs (fun x : var type.Z => expr.Var x)))).
+
+  Eval vm_compute in (Straightline.of_Expr test).
```
JasonGross: You might want to `Check eq_refl : (expected) = (computed)`

src/Experiments/SimplyTypedArithmetic.v
```diff
@@ -7885,7 +7915,500 @@ fun var : type -> Type =>
...
+            end) args
+         | _ as e =>
+           (fun default =>
+              match scalar_of_uncurried e with
+              | Some s => Scalar s
+              | None => default
+              end)
+         end.
+
+      (* TODO : uses fuel; ideally want a cleaner termination proof *)
```
JasonGross: You can avoid fuel by (effectively) cps'ing the `invert_LetInAppIdent args` case and moving it outside the `match` over `idc`.  Basically, you need to first match (directly) on `args` to expose the structure where you'd need to make the recursive call, then make the recursive call on that, and then convoy the (thunked) recursive call through the `match` on `idc` to get the types to line up.  Something like (untested)
```coq
match e with
| (idc @@ args)
  => let rec_arg := match args in expr s return option match s return Type with a * (b -> c) => expr a * (var b -> expr c) | _ => unit end with
                    | Pair x (Abs f) => Some (x, fun v => rec_call (f v))
                    | _ => None
                    end in
      match idc in ident s d return option match s return Type with a * (b -> c) => expr a * (var b -> expr c) | _ => unit end -> _ with
      | ident.LetIn s d => fun v => match v with Some (x, rec_call_on_f) => ... | None => default end
      | _ => _
      end rec_arg
```
An alternative, when you do have a default to return, is to use deep pattern matching, and then use something like `type.try_transport` to cast everything to the right type, and return the default in cases where the types don't line up.

src/Experiments/SimplyTypedArithmetic.v
jadephilipoom: Can I take in a context variable that gives me dummy primitives? I'm not sure I want to commit to using the interpretation of types provided by `type.interp` (although this aversion might also be misguided).

PR #366, opened 2018-05-13
Compatibility with respect to Coq PR#262

herbelin:
Dear fiat-crypto developers, this is a patch for compatibility with PR coq/coq#262.

PR coq/coq#262 makes the inference of return clauses more uniform and general but unification is sometimes not strong enough to deal with this generality. See coq/coq#5107 for details.

One reduces the search space for the return clause by forbidding it to be obtained by small inversion (using a `return _`).

Note that the patch is backwards compatible and can be applied as soon as now, even with PR coq/coq#262 not yet merged.

### DIFF
--- a/src/Curves/Montgomery/XZ.v
+++ b/src/Curves/Montgomery/XZ.v
@@ -124,7 +124,7 @@ Module M.
             let (x2, x3) := cswap swap x2 x3 in
             let (z2, z3) := cswap swap z2 z3 in
             dlet swap := b in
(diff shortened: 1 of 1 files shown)


JasonGross:
@herbelin I just ran into this again (https://github.com/coq/coq/issues/9955) can you explain (here or maybe better, there), why we need to restrict the search space?  It seems like Coq should always be able to find the solution `fun _ => bool` for the unification problem of `bool` and `?1 x`.  Even if it tries small inversion first (which seems perhaps wrong given that both branches are of type `bool`), it should just backtrack on that heuristic and try the dumb firstorder thing eventually, no?

PR #368, opened 2018-05-30
Add an experiment for rewrite rules

JasonGross:
This is a demo of (nearly*) all the features that I think need to make it into rewrite rules.  It has the following unfortunate properties, and I'd appreciate comments / code review / questions about the confusing or poorly designed parts to make it better.  (The next step, I think, is to integrate this back into the full pipeline.)

Unfortunate things:
- The rewrite rules are more verbose than I'd like them to be.  There is cruft from dealing with the fact that we are working underneath the cast, option, and LetIn monad (transformers), but are applying them somewhat manually and somewhat inconsistently.  (Also because we need to specify types at the pattern variables, and can't infer them from unification or whatever.)  Rewrite rules are copied at the bottom of this PR for convenience.**
- There is a lot of boilerplate.  Luckily this only has to be written once, but it's still fairly long
- The output of the compiler-compiler-compiler has many useless casts and suboptimal binds.  If the rewrite rules were written more carefully, many of these could be eliminated, but I don't think we should require optimally written rewrite rules to give nice output.  Current output is in a comment at the bottom of the file.
- All type variables in identifiers must not permit arrows in them.  We don't support (at all) having a `nat_rect` identifier which can support `nat_rect (fun _ => nat -> nat)`.

Things that this rewriter handles:
- nat, bool, list, prod, arrow
- `repeat` (rules can be tagged as "run again")
- LetIn, let-lifting
- side conditions in rewrite rules (return `None` rather than `Some`)
- let-lifting in the return of rewrite rules (return `UnderLet (thing to lift) (rest)`)
- ad-hoc transformation of `expr list` to `option (list expr)` by a hand-written `Fixpoint` which can be invoked in the rewrite rules
- discriminating between literals and non-literals
- python-script generation of almost everything that is special-cased to the type of identifiers

*The only one that is missing is abstract interpretation, but since we already need two passes, my plan for that is to strip abstract state when passing things into the rewriter, and then add back the abstract state that we would have had if we weren't rewriting anything to whatever the rewriter gives us.  Then, in the first pass, we will eta-expand the input lists (which have known length) and do all of the rewriting, and in the second pass, we will do no rewriting, but do bounds propagation.

**Rewrite rules:
```coq
  Definition rewrite_rules : rewrite_rulesT' value
    := [make_rewrite (0 + ??ℕ) (fun x => x);
          make_rewrite (??ℕ + 0) (fun x => x);
          make_rewrite (#? + #?) (fun x y => ##(x + y));
          make_rewrite (#? + ??ℕ.+1) (fun x y => ##(Datatypes.S x) + y);
          make_rewrite (??ℕ.+1 + #?) (fun x y => x + ##(Datatypes.S y));
          make_rewrite (??ℕ.+1 + ??ℕ.+1) (fun x y => (x+y).+1.+1);
          make_rewrite (??ℕ.+1 + ??ℕ) (fun x y => (x+y).+1);
          make_rewrite (??ℕ + ??ℕ.+1) (fun x y => (x+y).+1);
          make_rewrite (#pNatEqb @ #? @ #?) (fun x y => if Nat.eqb x y return topexpr _ then #iTrue else #iFalse);
          make_rewrite (#pFst @ (??, ??)) (fun tx x ty y => x);
          make_rewrite (#pSnd @ (??, ??)) (fun tx x ty y => y);
          make_rewrite (#pBoolRect @ ??{Unit -> pBase ??} @ ??{Unit -> pBase ??} @ #piTrue) (fun _ t _ f => LiftWrapExpr (t #TT));
          make_rewrite (#pBoolRect @ ??{Unit -> pBase ??} @ ??{Unit -> pBase ??} @ #piFalse) (fun _ t _ f => LiftWrapExpr (f #TT));
          make_rewrite_cps
            (#pMatchPair @ ??{?? -> ?? -> pBase ??} @ (??, ??))
            (fun _ _ _ f _ x _ y
             => x <-- cast x;
                  y <-- cast y;
                  oret (LiftWrapExpr (push_value_lets f x y)));
          make_rewrite_cps
            (??{pList ??} ++ ??{pList ??})
            (fun _ xs _ ys
             => xs <-- @cast expr _ (List _) xs;
                xs <-- reflect_list_cps xs;
                ys <-- reflect_list_cps ys;
                oret (UnderLets.Base (wrap (reify_list (List.app xs ys)))));
          make_rewrite_step_cps
            (#pListFlatMap @ ??{?? -> pList ??} @ ??{pList ??})
            (fun _ _ f _ xs
             => xs <-- @cast expr _ (List _) xs;
                  xs <-- reflect_list_cps xs;
                  oret (fxs <--- List.map f xs;
                          UnderLets.Base
                            (wrap (#ListFoldRight @ (λ ls1 ls2, $ls1 ++ $ls2) @ (λ _, []) @ $(reify_list fxs))%expr)));
          make_rewrite_step_cps
            (#pListPartition @ ??{?? -> pBool} @ ??{pList ??})
            (fun _ f _ xs
             => xs <-- @cast expr _ (List _) xs;
                  xs <-- reflect_list_cps xs;
                  oret (v <-- (list_rect
                                 _
                                 (UnderLets.Base ([], []))
                                 (fun x tl partition_tl
                                  => partition_tl <-- partition_tl;
                                       fx <-- f x;
                                       UnderLets.Base
                                         (#MatchPair
                                           @ (λ g d, #BoolRect
                                                      @ (λ _, ($x :: $g, $d))
                                                      @ (λ _, ($g, $x :: $d))
                                                      @ $fx)
                                           @ partition_tl))
                                 xs)%expr;
                          UnderLets.Base (wrap v))%under_lets);
          make_rewrite_cps
            (#pListFoldRight @ ??{?? -> ?? -> ??} @ ??{pBase ??} @ ??{pList ??})
            (fun _ _ _ f A init B xs
             => f <-- @cast value _ (B -> A -> A)%ctype f;
                  xs <-- reflect_list_cps xs;
                  oret (v <-- (fold_right
                                 (fun x y => y <-- y; push_value_lets f x y)
                                 (UnderLets.Base init)
                                 xs);
                          UnderLets.Base (wrap v))%under_lets);
          make_rewrite_cps
            (#pListRect @ ??{Unit -> pBase ??} @ ??{?? -> ?? -> ?? -> ??} @ ??{pList ??})
            (fun P Pnil _ _ _ _ Pcons A xs
             => Pcons <-- @cast value _ (A -> List A -> P -> P)%ctype Pcons;
                  xs <-- reflect_list_cps xs;
                  oret (v <-- (list_rect
                                 (fun _ => UnderLets (expr P))
                                 (Pnil #TT)
                                 (fun x' xs' rec => rec <-- rec; push_value_lets Pcons x' (reify_list xs') rec)
                                 xs);
                          UnderLets.Base (wrap v))%under_lets);
          make_rewrite_cps
            (#pListMap @ ??{?? -> pBase ??} @ ??{pList ??})
            (fun _ _ f _ xs
             => xs <-- @cast expr _ (List _) xs;
                  xs <-- reflect_list_cps xs;
                  oret (fxs <--- List.map f xs;
                          UnderLets.Base (wrap (reify_list fxs)))%under_lets);
          make_rewrite_cps
            (#pListMap @ ??{?? -> pBase ??} @ (?? :: ??))
            (fun _ _ f _ x _ xs
             => xs <-- @cast expr _ (List _) xs;
                  x <-- cast x;
                  oret (fx <-- f x;
                          UnderLets.Base (wrap (fx :: #ListMap @ (λ v , UnderLets.to_expr (f ($v))) @ xs))%expr)%under_lets)
       ]%list%pattern%continuation%option%under_lets.
```

### DIFF
--- a/src/Experiments/RewriteRulesDecisionTreeNatProdListPHOASCurriedWithLets.v
+++ b/src/Experiments/RewriteRulesDecisionTreeNatProdListPHOASCurriedWithLets.v
@@ -1,0 +1,2788 @@
+Require Import Coq.ZArith.ZArith.
+Require Import Coq.Lists.List.
+Require Import Crypto.Util.Option.
+Require Import Crypto.Util.Notations.
+Require Import Crypto.Util.ListUtil.
+Require Import Crypto.Util.LetIn.
+Import ListNotations.
+
+Set Boolean Equality Schemes.
+Inductive base_type := Unit | Bool | Nat | Prod (A B : base_type) | List (A : base_type).
+Inductive type := Base (t : base_type) | Arrow (s : type) (d : type).
+Coercion Base : base_type >-> type.
+Bind Scope ctype_scope with type.
+Bind Scope ctype_scope with base_type.
+Delimit Scope ctype_scope with ctype.
+Infix "->" := Arrow : ctype_scope.
+Infix "*" := Prod : ctype_scope.
+Notation "( )" := Unit : ctype_scope.
+
+About partition.
+Inductive ident : type -> Type :=
+| O : ident Nat
+| S : ident (Nat -> Nat)
+| NatRect {P : base_type} : ident ((Unit -> P) -> (Nat -> P -> P) -> Nat -> P)
+| NatEqb : ident (Nat -> Nat -> Bool)
+| Add : ident (Nat -> Nat -> Nat)
+| Pair {A B : base_type} : ident (A -> B -> A * B)
+| Fst {A B} : ident (A * B -> A)
+| Snd {A B} : ident (A * B -> B)
+| MatchPair {A B P : base_type} : ident ((A -> B -> P) -> A * B -> P)
+| Nil {A} : ident (List A)
+| Cons {A : base_type} : ident (A -> List A -> List A)
+| ListMap {A B : base_type} : ident ((A -> B) -> List A -> List B)
+| ListApp {A} : ident (List A -> List A -> List A)
+| ListFlatMap {A B : base_type} : ident ((A -> List B) -> List A -> List B)
+| ListRect {A : base_type} {P : base_type} : ident ((Unit -> P) -> (A -> List A -> P -> P) -> List A -> P)
+| ListFoldRight {A : base_type} {B : base_type} : ident ((B -> A -> A) -> (Unit -> A) -> List B -> A)
+| ListPartition {A : base_type} : ident ((A -> Bool) -> List A -> List A * List A)
+| TT : ident Unit
+| iTrue : ident Bool
+| iFalse : ident Bool
+| BoolRect {P : base_type} : ident ((Unit -> P) -> (Unit -> P) -> Bool -> P)
+| Literal (n : nat) : ident Nat.
+
+Show Match ident.
+(*
+<<<
+#!/usr/bin/env python2
+
+show_match_ident = r"""match # with
+ | O =>
+ | S =>
+ | NatRect P =>
+ | NatEqb =>
+ | Add =>
+ | Pair A B =>
+ | Fst A B =>
+ | Snd A B =>
+ | MatchPair A B P =>
+ | Nil A =>
+ | Cons A =>
+ | ListMap A B =>
+ | ListApp A =>
+ | ListFlatMap A B =>
+ | ListRect A P =>
+ | ListFoldRight A B =>
+ | ListPartition A =>
+ | TT =>
+ | iTrue =>
+ | iFalse =>
+ | BoolRect P =>
+ | Literal n =>
+ end
+
+""".replace('=>', '=>')
+ctors = [i.strip('|=> ').split(' ') for i in show_match_ident.split('\n') if i.strip().startswith('|')]
+pctors = ['p' + i[0] for i in ctors]
+print(r"""Inductive pident : Type :=
+| %s.
+""" % '\n| '.join(pctors))
+print(r"""Definition pident_ident_beq {t} (X : pident) (Y : ident t) : bool
+  := match X, Y with
+     | %s
+       => true
+     | %s
+       => false
+     end.
+""" % ('\n     | '.join(pctor + ', ' + ' '.join([ctor[0]] + ['_'] * (len(ctor)-1))
+                        for pctor, ctor in zip(pctors, ctors)),
+       '\n     | '.join(pctor + ', _' for pctor in pctors)))
+print(r"""Definition eta_ident_cps {T t} (idc : ident t)
+           (f : forall t, ident t -> T t)
+  : T t
+  := match idc with
+     | %s
+     end.
+""" % ('\n     | '.join(' '.join(ctor) + ' => f _ '
+                        + (('%s' if len(ctor) == 1 else '(@%s)')
+                           % ' '.join(ctor))
+                        for ctor in ctors)))
+print(r"""Definition eta_option_pident_cps {T} (f : pident -> option T)
+  : option (pident -> T)
+  := (%s;
+      Some (fun c
+            => match c with
+               | %s
+               end))%%option.
+""" % (';\n      '.join('f' + ctor[0] + ' <- f p' + ctor[0] for ctor in ctors),
+       '\n               | '.join('p' + ctor[0] + ' => f' + ctor[0] for ctor in ctors)))
+print(r"""Definition pident_of_ident {t} (idc : ident t) : pident
+  := match idc with
+     | %s
+     end.
+""" % ('\n     | '.join(' '.join(ctor) + ' => ' + pctor for ctor, pctor in zip(ctors, pctors))))
+print(r"""Definition orb_pident (f : pident -> bool) : bool
+  := (%s)%%bool.
+""" % ' || '.join('f ' + pctor for pctor in pctors))
+>>>
+*)
+Inductive pident : Type :=
+| pO
+| pS
+| pNatRect
+| pNatEqb
+| pAdd
+| pPair
+| pFst
+| pSnd
+| pMatchPair
+| pNil
+| pCons
+| pListMap
+| pListApp
+| pListFlatMap
+| pListRect
+| pListFoldRight
+| pListPartition
+| pTT
+| piTrue
+| piFalse
+| pBoolRect
+| pLiteral.
+
+Definition pident_ident_beq {t} (X : pident) (Y : ident t) : bool
+  := match X, Y with
+     | pO, O
+     | pS, S
+     | pNatRect, NatRect _
+     | pNatEqb, NatEqb
+     | pAdd, Add
+     | pPair, Pair _ _
+     | pFst, Fst _ _
+     | pSnd, Snd _ _
+     | pMatchPair, MatchPair _ _ _
+     | pNil, Nil _
+     | pCons, Cons _
+     | pListMap, ListMap _ _
+     | pListApp, ListApp _
+     | pListFlatMap, ListFlatMap _ _
+     | pListRect, ListRect _ _
+     | pListFoldRight, ListFoldRight _ _
+     | pListPartition, ListPartition _
+     | pTT, TT
+     | piTrue, iTrue
+     | piFalse, iFalse
+     | pBoolRect, BoolRect _
+     | pLiteral, Literal _
+       => true
+     | pO, _
+     | pS, _
+     | pNatRect, _
+     | pNatEqb, _
+     | pAdd, _
+     | pPair, _
+     | pFst, _
+     | pSnd, _
+     | pMatchPair, _
+     | pNil, _
+     | pCons, _
+     | pListMap, _
+     | pListApp, _
+     | pListFlatMap, _
+     | pListRect, _
+     | pListFoldRight, _
+     | pListPartition, _
+     | pTT, _
+     | piTrue, _
+     | piFalse, _
+     | pBoolRect, _
+     | pLiteral, _
+       => false
+     end.
+
+Definition eta_ident_cps {T t} (idc : ident t)
+           (f : forall t, ident t -> T t)
+  : T t
+  := match idc with
+     | O => f _ O
+     | S => f _ S
+     | NatRect P => f _ (@NatRect P)
+     | NatEqb => f _ NatEqb
+     | Add => f _ Add
+     | Pair A B => f _ (@Pair A B)
+     | Fst A B => f _ (@Fst A B)
+     | Snd A B => f _ (@Snd A B)
+     | MatchPair A B P => f _ (@MatchPair A B P)
+     | Nil A => f _ (@Nil A)
+     | Cons A => f _ (@Cons A)
+     | ListMap A B => f _ (@ListMap A B)
+     | ListApp A => f _ (@ListApp A)
+     | ListFlatMap A B => f _ (@ListFlatMap A B)
+     | ListRect A P => f _ (@ListRect A P)
+     | ListFoldRight A B => f _ (@ListFoldRight A B)
+     | ListPartition A => f _ (@ListPartition A)
+     | TT => f _ TT
+     | iTrue => f _ iTrue
+     | iFalse => f _ iFalse
+     | BoolRect P => f _ (@BoolRect P)
+     | Literal n => f _ (@Literal n)
+     end.
+
+Definition eta_option_pident_cps {T} (f : pident -> option T)
+  : option (pident -> T)
+  := (fO <- f pO;
+      fS <- f pS;
+      fNatRect <- f pNatRect;
+      fNatEqb <- f pNatEqb;
+      fAdd <- f pAdd;
+      fPair <- f pPair;
+      fFst <- f pFst;
+      fSnd <- f pSnd;
+      fMatchPair <- f pMatchPair;
+      fNil <- f pNil;
+      fCons <- f pCons;
+      fListMap <- f pListMap;
+      fListApp <- f pListApp;
+      fListFlatMap <- f pListFlatMap;
+      fListRect <- f pListRect;
+      fListFoldRight <- f pListFoldRight;
+      fListPartition <- f pListPartition;
+      fTT <- f pTT;
+      fiTrue <- f piTrue;
+      fiFalse <- f piFalse;
+      fBoolRect <- f pBoolRect;
+      fLiteral <- f pLiteral;
+      Some (fun c
+            => match c with
+               | pO => fO
+               | pS => fS
+               | pNatRect => fNatRect
+               | pNatEqb => fNatEqb
+               | pAdd => fAdd
+               | pPair => fPair
+               | pFst => fFst
+               | pSnd => fSnd
+               | pMatchPair => fMatchPair
+               | pNil => fNil
+               | pCons => fCons
+               | pListMap => fListMap
+               | pListApp => fListApp
+               | pListFlatMap => fListFlatMap
+               | pListRect => fListRect
+               | pListFoldRight => fListFoldRight
+               | pListPartition => fListPartition
+               | pTT => fTT
+               | piTrue => fiTrue
+               | piFalse => fiFalse
+               | pBoolRect => fBoolRect
+               | pLiteral => fLiteral
+               end))%option.
+
+Definition pident_of_ident {t} (idc : ident t) : pident
+  := match idc with
+     | O => pO
+     | S => pS
+     | NatRect P => pNatRect
+     | NatEqb => pNatEqb
+     | Add => pAdd
+     | Pair A B => pPair
+     | Fst A B => pFst
+     | Snd A B => pSnd
+     | MatchPair A B P => pMatchPair
+     | Nil A => pNil
+     | Cons A => pCons
+     | ListMap A B => pListMap
+     | ListApp A => pListApp
+     | ListFlatMap A B => pListFlatMap
+     | ListRect A P => pListRect
+     | ListFoldRight A B => pListFoldRight
+     | ListPartition A => pListPartition
+     | TT => pTT
+     | iTrue => piTrue
+     | iFalse => piFalse
+     | BoolRect P => pBoolRect
+     | Literal n => pLiteral
+     end.
+
+Definition orb_pident (f : pident -> bool) : bool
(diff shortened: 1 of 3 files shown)


src/Experiments/RewriteRulesDecisionTreeNatProdListPHOASCurriedWithLets.v
```diff
@@ -0,0 +1,1581 @@
...
+     | pListRect, ListRect
+     | pListFoldRight, ListFoldRight
+     | pListPartition, ListPartition
+     | pTT, TT
+     | piTrue, iTrue
+     | piFalse, iFalse
+     | pBoolRect, BoolRect
+     | pLiteral, Literal _
+       => true
+     | pO, _
```
achlipala: Why is it desirable to list all the `false` cases explicitly here, rather than relying on one wildcard pattern?  Are you trying to guard against bugs in the generator, where you didn't include all the `true` cases where you meant them to appear?

achlipala:
Responding to your list of issues:

# Verbosity of rewrite rules

> The rewrite rules are more verbose than I'd like them to be.

Let's be sure to separate our gripes based on whether it is clear that an OCaml plugin could implement appropriate patterns automatically.  Ideally, we want an interface in the style of `Hint Rewrite`, using Vernacular commands to inform the system of hints stated and proved as normal theorems.

The first few example rules, stated without using any `cps` identifier, are really beautiful!  So we have surely made some progress, if those all work.

Let me now talk through the patterns that I find undermotivated in the example rules, and we can see if commonalities emerge.

*Asymmetry between rules for `map` vs. `partition`*: why does the former call `List.map` while the latter doesn't just call `List.partition` in a similar way?  Is it just because of the use of pairs?

*Asymmetry between rules for `map` vs. `fold_right`*: the former calls `List.map` while the latter doesn't just call `List.fold_right`... and other details seem simpler than for the last one.

# Boilerplate

> There is a lot of boilerplate. Luckily this only has to be written once, but it's still fairly long

I'd like to clarify: are you talking about code that expands when we add new rewrite rules?  If not, then it seems like not such a high-priority issue.

# Useless casts & suboptimal binds

> The output of the compiler-compiler-compiler has many useless casts and suboptimal binds.

Can you point me to one example of a useless cast and one example of a suboptimal bind?

# Restriction to first-order 

> All type variables in identifiers must not permit arrows in them. We don't support (at all) having a `nat_rect` identifier which can support `nat_rect (fun _ => nat -> nat)`.

Do we have a concrete use case in mind that needs a higher-order interface?  I don't see this issue as a clear deal-breaker for now.

# Overall

I'm pretty optimistic about the current state of things.  The generated code doesn't seem obviously problematic.  Once we see some performance-benchmark results, then we can reconsider if any of the current design choices are fatal.

src/Experiments/RewriteRulesDecisionTreeNatProdListPHOASCurriedWithLets.v
JasonGross: Almost.  This guards against me forgetting to re-run the generator code.

JasonGross:
> The first few example rules, stated without using any cps identifier, are really beautiful! So we have surely made some progress, if those all work.

I've made these even better.  In https://github.com/mit-plv/fiat-crypto/pull/371/files, the first rewrite rules become
```coq
        := Eval cbn [Make.interp_rewrite_rules List.app] in
            Make.interp_rewrite_rules
              ++ [
              make_rewrite (#pident.fst @ (??, ??)) (fun _ x _ y => x);
              make_rewrite (#pident.snd @ (??, ??)) (fun _ x _ y => y);
              make_rewrite (#pident.List_repeat @ ?? @ #?ℕ) (fun _ x n => reify_list (List.repeat x n));
...
```

> Asymmetry between rules for map vs. partition: why does the former call List.map while the latter doesn't just call List.partition in a similar way? Is it just because of the use of pairs?

`List.partition` requires that the function return an actual boolean on every element of the list.  There's no way to guarantee this when checking the rewrite rule.

> Asymmetry between rules for map vs. fold_right: the former calls List.map while the latter doesn't just call List.fold_right... and other details seem simpler than for the last one.

Huh?  map is
```coq
          make_rewrite_cps
            (#pListMap @ ??{?? -> pBase ??} @ ??{pList ??})
            (fun _ _ f _ xs
             => xs <-- @cast expr _ (List _) xs;
                  xs <-- reflect_list_cps xs;
                  oret (fxs <--- List.map f xs;
                          UnderLets.Base (wrap (reify_list fxs)))%under_lets);
```
while fold_right is
```coq
          make_rewrite_cps
            (#pListFoldRight @ ??{?? -> ?? -> ??} @ ??{pBase ??} @ ??{pList ??})
            (fun _ _ _ f A init B xs
             => f <-- @cast value _ (B -> A -> A)%ctype f;
                  xs <-- reflect_list_cps xs;
                  oret (v <-- (fold_right
                                 (fun x y => y <-- y; push_value_lets f x y)
                                 (UnderLets.Base init)
                                 xs);
                          UnderLets.Base (wrap v))%under_lets);
```
Maybe you are looking at flat_map (which has the same issue as partition)...?

Btw, here is the new version of the rest of the rewrite rules (modulo add_with_carry &co), which I am a bit happier with:
```coq
              make_rewrite
                (#pident.bool_rect @ ??{() -> ??} @ ??{() -> ??} @ #?𝔹)
                (fun _ t _ f b
                 => ret
                     (if b return UnderLetsExpr (type.base (if b then _ else _))
                      then t ##tt
                      else f ##tt));
              make_rewrite
                (#pident.pair_rect @ ??{?? -> ?? -> ??} @ (??, ??))
                (fun _ _ _ f _ x _ y
                 => x <- castbe x; y <- castbe y; lret (f x y));
              make_rewrite
                (??{pattern.base.type.list ??} ++ ??{pattern.base.type.list ??})
                (fun _ xs _ ys => rlist_rect_cast ys (fun x _ xs_ys => x :: xs_ys) xs);
              make_rewrite
                (#pident.List_rev @ ??{pattern.base.type.list ??})
                (fun _ xs
                 => xs <- reflect_list_cps xs;
                      ret (reify_list (List.rev xs)));
              make_rewrite_step
                (#pident.List_flat_map @ ??{?? -> pattern.base.type.list ??} @ ??{pattern.base.type.list ??})
                (fun _ B f _ xs
                 => rlist_rect_cast
                     []
                     (fun x _ flat_map_tl => fx <-- f x; UnderLets.Base ($fx ++ flat_map_tl))
                     xs);
              make_rewrite_step
                (#pident.List_partition @ ??{?? -> base.type.bool} @ ??{pattern.base.type.list ??})
                (fun _ f _ xs
                 => rlist_rect_cast
                     ([], [])
                     (fun x tl partition_tl
                      => fx <-- f x;
                          (#ident.pair_rect
                            @ (λ g d, #ident.bool_rect
                                       @ (λ _, ($x :: $g, $d))
                                       @ (λ _, ($g, $x :: $d))
                                       @ $fx)
                            @ partition_tl))
                     xs);
              make_rewrite
                (#pident.List_fold_right @ ??{?? -> ?? -> ??} @ ?? @ ??{pattern.base.type.list ??})
                (fun _ _ _ f B init A xs
                 => f <- @castv _ (A -> B -> B)%etype f;
                     rlist_rect
                       init
                       (fun x _ y => f x y)
                       xs);
              make_rewrite
                (#pident.list_rect @ ??{() -> ??} @ ??{?? -> ?? -> ?? -> ??} @ ??{pattern.base.type.list ??})
                (fun P Pnil _ _ _ _ Pcons A xs
                 => Pcons <- @castv _ (A -> base.type.list A -> P -> P) Pcons;
                     rlist_rect
                       (Pnil ##tt)
                       (fun x' xs' rec => Pcons x' (reify_list xs') rec)
                       xs);
              make_rewrite
                (#pident.list_case @ ??{() -> ??} @ ??{?? -> ?? -> ??} @ []) (fun _ Pnil _ _ _ Pcons => lret (Pnil ##tt));
              make_rewrite
                (#pident.list_case @ ??{() -> ??} @ ??{?? -> ?? -> ??} @ (?? :: ??))
                (fun _ Pnil _ _ _ Pcons _ x _ xs
                 => x <- castbe x; xs <- castbe xs; lret (Pcons x xs));
              make_rewrite
                (#pident.List_map @ ??{?? -> ??} @ ??{pattern.base.type.list ??})
                (fun _ _ f _ xs
                 => rlist_rect_cast
                     []
                     (fun x _ fxs => fx <-- f x; fx :: fxs)
                     xs);
              make_rewrite
                (#pident.List_nth_default @ ?? @ ??{pattern.base.type.list ??} @ #?ℕ)
                (fun _ default _ ls n
                 => default <- castbe default;
                     ls <- reflect_list_cps ls;
                     nth_default default ls n);
              make_rewrite
                (#pident.nat_rect @ ??{() -> ??} @ ??{base.type.nat -> ?? -> ??} @ #?ℕ)
                (fun P O_case _ _ S_case n
                 => S_case <- @castv _ (@type.base base.type base.type.nat -> type.base P -> type.base P) S_case;
                     lret (nat_rect _ (O_case ##tt) (fun n' rec => rec <-- rec; S_case ##n' rec) n));
              make_rewrite
                (#pident.List_length @ ??{pattern.base.type.list ??})
                (fun _ xs => xs <- reflect_list_cps xs; ##(List.length xs));
              make_rewrite
                (#pident.List_combine @ ??{pattern.base.type.list ??} @ ??{pattern.base.type.list ??})
                (fun _ xs _ ys
                 => xs <- reflect_list_cps xs;
                     ys <- reflect_list_cps ys;
                     reify_list (List.map (fun '((x, y)%core) => (x, y)) (List.combine xs ys)));
              make_rewrite
                (#pident.List_update_nth @ #?ℕ @ ??{?? -> ??} @ ??{pattern.base.type.list ??})
                (fun n _ _ f A ls
                 => f <- @castv _ (A -> A) f;
                     ls <- reflect_list_cps ls;
                     lret
                       (retv <--- (update_nth
                                     n
                                     (fun x => x <-- x; f x)
                                     (List.map UnderLets.Base ls));
                          reify_list retv));
              make_rewrite (#?ℤ   + ??{ℤ}) (fun z v => v  when  Z.eqb z 0);
              make_rewrite (??{ℤ} + #?ℤ  ) (fun v z => v  when  Z.eqb z 0);
              make_rewrite (#?ℤ   + (-??{ℤ})) (fun z v => ##z - v  when  Z.gtb z 0);
              make_rewrite ((-??{ℤ}) + #?ℤ  ) (fun v z => ##z - v  when  Z.gtb z 0);
              make_rewrite (#?ℤ   + (-??{ℤ})) (fun z v => -(##((-z)%Z) + v)  when  Z.ltb z 0);
              make_rewrite ((-??{ℤ}) + #?ℤ  ) (fun v z => -(v + ##((-z)%Z))  when  Z.ltb z 0);
              make_rewrite ((-??{ℤ}) + (-??{ℤ})) (fun x y => -(x + y));
              make_rewrite ((-??{ℤ}) +   ??{ℤ} ) (fun x y => y - x);
              make_rewrite (  ??{ℤ}  + (-??{ℤ})) (fun x y => x - y);

              make_rewrite (#?ℤ   - (-??{ℤ})) (fun z v =>  v  when  Z.eqb z 0);
              make_rewrite (#?ℤ   -   ??{ℤ} ) (fun z v => -v  when  Z.eqb z 0);
              make_rewrite (??{ℤ} - #?ℤ     ) (fun v z =>  v  when  Z.eqb z 0);
              make_rewrite (#?ℤ   - (-??{ℤ})) (fun z v => ##z + v  when  Z.gtb z 0);
              make_rewrite (#?ℤ   - (-??{ℤ})) (fun z v => v - ##((-z)%Z)     when  Z.ltb z 0);
              make_rewrite (#?ℤ   -   ??{ℤ} ) (fun z v => -(##((-z)%Z) + v)  when  Z.ltb z 0);
              make_rewrite ((-??{ℤ}) - #?ℤ  ) (fun v z => -(v + ##((-z)%Z))  when  Z.gtb z 0);
              make_rewrite ((-??{ℤ}) - #?ℤ  ) (fun v z => ##((-z)%Z) - v     when  Z.ltb z 0);
              make_rewrite (  ??{ℤ}  - #?ℤ  ) (fun v z => v + ##((-z)%Z)     when  Z.ltb z 0);
              make_rewrite ((-??{ℤ}) - (-??{ℤ})) (fun x y => y - x);
              make_rewrite ((-??{ℤ}) -   ??{ℤ} ) (fun x y => -(x + y));
              make_rewrite (  ??{ℤ}  - (-??{ℤ})) (fun x y => x + y);

              make_rewrite (#?ℤ   * ??{ℤ}) (fun z v => ##0  when  Z.eqb z 0);
              make_rewrite (??{ℤ} * #?ℤ  ) (fun v z => ##0  when  Z.eqb z 0);
              make_rewrite (#?ℤ   * ??{ℤ}) (fun z v => v  when  Z.eqb z 1);
              make_rewrite (??{ℤ} * #?ℤ  ) (fun v z => v  when  Z.eqb z 1);
              make_rewrite (#?ℤ      * (-??{ℤ})) (fun z v =>  v  when  Z.eqb z (-1));
              make_rewrite ((-??{ℤ}) * #?ℤ     ) (fun v z =>  v  when  Z.eqb z (-1));
              make_rewrite (#?ℤ      *   ??{ℤ} ) (fun z v => -v  when  Z.eqb z (-1));
              make_rewrite (??{ℤ}    * #?ℤ     ) (fun v z => -v  when  Z.eqb z (-1));
              make_rewrite (#?ℤ      * ??{ℤ}   ) (fun z v => -(##((-z)%Z) * v)  when  Z.ltb z 0);
              make_rewrite (??{ℤ}    * #?ℤ     ) (fun v z => -(v * ##((-z)%Z))  when  Z.ltb z 0);
              make_rewrite ((-??{ℤ}) * (-??{ℤ})) (fun x y => x * y);
              make_rewrite ((-??{ℤ}) *   ??{ℤ} ) (fun x y => -(x * y));
              make_rewrite (  ??{ℤ}  * (-??{ℤ})) (fun x y => -(x * y));

              make_rewrite (??{ℤ} * #?ℤ) (fun x y => x << (Z.log2 y)  when  Z.eqb y (2^Z.log2 y));
              make_rewrite (#?ℤ * ??{ℤ}) (fun y x => x << (Z.log2 y)  when  Z.eqb y (2^Z.log2 y));
              make_rewrite (??{ℤ} / #?ℤ) (fun x y => x >> (Z.log2 y)  when  Z.eqb y (2^Z.log2 y));
              make_rewrite (??{ℤ} mod #?ℤ) (fun x y => #(ident.Z_land (y-1)) @ x  when  Z.eqb y (2^Z.log2 y));
              make_rewrite (-(-??{ℤ})) (fun v => v);

              (** TODO(jadep): These next two are only here for demonstration purposes; remove them once you no longer need it as a template *)
              (* if it's a concrete pair, we can opp the second value *)
              make_rewrite (#pident.Z_neg_snd @ (??{ℤ}, ??{ℤ})) (fun x y => (x, -y));
              (* if it's not a concrete pair, let-bind the pair and negate the second element *)
              make_rewrite
                (#pident.Z_neg_snd @ ??{ℤ * ℤ})
                (fun xy => lret (UnderLets.UnderLet xy (fun xyv => UnderLets.Base (#ident.fst @ $xyv, -(#ident.snd @ $xyv)))));
```


> Let's be sure to separate our gripes based on whether it is clear that an OCaml plugin could implement appropriate patterns automatically. Ideally, we want an interface in the style of Hint Rewrite, using Vernacular commands to inform the system of hints stated and proved as normal theorems.

I think the Hint Rewrite style should be almost feasible.  The main issues that I see are that I'm not sure how to deal with the difference between rewrite rules that are final (expose no new discriminees), rewrite rules that expose new potentials for rewriting (partition, flat_map), and ones that need to be called recursively (recursive list functions).  Especially, for the latter sort, I currently would need a thing that says "here is the rewrite rule for nil, and here is the rewrite rule for cons", and possibly something more special than that, to figure out which cons/nil pattern to generalize over.

> I'd like to clarify: are you talking about code that expands when we add new rewrite rules? If not, then it seems like not such a high-priority issue.

No.  This is the ~1000 lines of code which is basically independent of the number of rewrite rules

> Can you point me to one example of a useless cast and one example of a suboptimal bind?

Useless cast: Not anymore in this PR, because I removed all non-arrow types, but here's an instance in the full version of the code: https://github.com/JasonGross/fiat-crypto/blob/b270d84a5ca503be940d5c522d13b6ea4d35c010/src/Experiments/SimplyTypedArithmetic.v#L8419-L8425
```coq
| @ident.List_repeat A =>
    fun (x : expr (type.base A)) (x0 : expr (type.base (base.type.type_base base.type.nat))) =>
    match x0 with
    | #(idc)%expr_pat =>
        match
          match idc with
          | @ident.Literal t0 v =>
              match t0 as t1 return (base.base_interp t1 -> option nat) with
              | base.type.unit => fun _ : unit => None
              | base.type.Z => fun _ : Z => None
              | base.type.bool => fun _ : bool => None
              | base.type.nat => fun v0 : nat => Some v0
              end v
          | _ => None
          end
        with
        | Some args =>
            base.try_make_transport_cps (fun A0 : base.type.type => expr (type.base (base.type.list A0))) A A
              (UnderLets.UnderLets base.type.type ident var (expr (type.base (base.type.list A))))
              (fun a : option (expr (type.base (base.type.list A)) -> expr (type.base (base.type.list A))) =>
               match a with
               | Some x' => UnderLets.Base (x' (reify_list (List.repeat x args)))
               | None => UnderLets.Base (#(ident.List_repeat)%expr @ x @ x0)%expr_pat
               end)
        | None => UnderLets.Base (#(ident.List_repeat)%expr @ x @ x0)%expr_pat
        end
    | _ => UnderLets.Base (#(ident.List_repeat)%expr @ x @ x0)%expr_pat
    end
```
Suboptimal bind: https://github.com/JasonGross/fiat-crypto/blob/3820f1880bf519c0ac0cbc99db13cf816ef051ad/src/Experiments/RewriteRulesDecisionTreeNatProdListPHOASCurriedWithLets.v#L1482-L1486
```
                     (fv <-- (fxs <--- map x xs;
                              UnderLets.Base
                                (#(ListFoldRight) @ (λ ls1 ls2 : expr var ??,
                                                     $ls1 ++ $ls2) @ (λ _ : expr var ??,
                                                                      []) @ $(reify_list fxs)));
```

> Do we have a concrete use case in mind that needs a higher-order interface?

No

> Once we see some performance-benchmark results,

Each operation on 32-bit x25519 takes about 3-4 seconds to synthesize + the same in Qed.  Over half this time is spent computing bounds, because `encode (2^255-19 - 1)` takes about 0.6 seconds and shows up in 3 places (once for each of the pair of inputs, and once for the output).  (Each operation on the 64-bit x25519 takes about 2 seconds to synthesize + the same in Qed.  This is comparable to the version with hand-written rewrite rules.  If we don't pre-expand evaluation of the discrimination tree, it takes about 2x as long to synthesize/Qed.)

PR #372, opened 2018-06-12
[WIP] Experiments no cps with rewrites extraction to string

JasonGross:
Mostly for visibility.  This is the full pipeline, with the rewriter, without cps, with to-C-stringification, with extraction


davidben:
Notes from when Jason and I walked over a subset of this:


Maybe have a comment explaining motivation of the tower of coercions (what you described to me on the whiteboard).

uint64_t[5] x1 should be uint64_t x1[5].

Consider replacing Seq constructor with just talking about list expr.

Lost the /* bool */ business. (This was relevant so that we'd know the carry bits in all the carry intrinsics could be implemented assuming 0 or 1.) Could do fiat_uint1 if comments are silly. (Officially '_t' suffixes are reserved if you care, but most of the rules in here are absurd and everyone ignores them. https://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html ) Note Zstatic_cast will print interestingly here because of the & 1 needed.

Comment differencce between to_string and to_string_relaxed.


Some bugs:
```coq
            Definition to_string (t : type.primitive) (r : option zrange) : string
              := match r with
                 | Some r[ lower ~> upper ]%zrange
                   => let lg2u := Z.log2_up (upper + 1) in
                      let default := "ℤ[" ++ Z.to_hex_string lower ++ "," ++ Z.to_hex_string upper ++ "]" in
                      if (2^lg2u - 1 =? upper) && (lower =? 0)
                      then "uint" ++ Z.to_decimal_string lg2u ++ "_t"
                      else if (-2^lg2u =? lower) && (2^lg2u - 1 =? upper)
                           then "int" ++ Z.to_decimal_string (lg2u + 1) ++ "_t"
                           else default
                 | None
                   => "ℤ"
                 end ++ match t with
                        | type.Zptr => "*"
                        | type.Z => ""
                        end.

            Example to_string_int16 : to_string type.Z (Some r[-2^15 ~> 2^15 - 1]%zrange) = "int16_t" := eq_refl.
            Example to_string_uint32 : to_string type.Z (Some r[0 ~> 2^32 - 1]%zrange) = "uint32_t" := eq_refl.
            
            Definition to_string_relaxed (t : type.primitive) (r : option zrange) : string
              := match r with
                 | Some r[ lower ~> upper ]%zrange
                   => let lg2u := Z.log2_up (upper + 1) in
                      let lg2lg2u := Z.log2_up lg2u in
                      let lg2u := 2^lg2lg2u in
                      let default := "ℤ[" ++ Z.to_hex_string lower ++ "," ++ Z.to_hex_string upper ++ "]" in
                      if (0 <=? lower) && (upper <=? 2^lg2u - 1)
                      then "uint" ++ Z.to_decimal_string lg2u ++ "_t"
                      else if (-2^(lg2u - 1) <=? lower) && (upper <=? 2^(lg2u - 1) - 1)
                           then "int" ++ Z.to_decimal_string lg2u ++ "_t"
                           else default
                 | None
                   => "ℤ"
                 end ++ match t with
                        | type.Zptr => "*"
                        | type.Z => ""
                        end.

            Example to_string_relaxed_int16 : to_string_relaxed type.Z (Some r[-2^15 ~> 2^15 - 1]%zrange) = "int16_t" := eq_refl.
            Example to_string_relaxed_int32 : to_string_relaxed type.Z (Some r[-2^15 - 1 ~> 2^15 - 1]%zrange) = "int32_t". Admitted.
            Example to_string_relaxed_int32' : to_string_relaxed type.Z (Some r[-2^15 ~> 2^15]%zrange) = "int32_t". Admitted.
            Example to_string_relaxed_uint32 : to_string_relaxed type.Z (Some r[0 ~> 2^32 - 1]%zrange) = "uint32_t" := eq_refl.
            Example to_string_relaxed_uint32' : to_string_relaxed type.Z (Some r[2^30 ~> 2^32 - 100]%zrange) = "uint32_t" := eq_refl.
            Example to_string_relaxed_uint64 : to_string_relaxed type.Z (Some r[0 ~> 2^32]%zrange) = "uint64_t" := eq_refl.
```

Comment around real_C_type_promote_range seems to have been lost.

to_string_relaxed should call real_C_type_promote_range.

"I'm going to just assume they're right... maybe write unit tests?"

integer_promote_range should preferentially promote to signed int, not unsigned. C is terrible. (Add tests? ;-) )
Does it make sense to make a type signed:yes/no + rank?

Might be worth writing down (or embedding in type?) when you've rounded things.

Consider taking all the Z.abs out of upper-lower since they're presumably impossible.

Nit: unsigned_counterpart_of could just call rank.

Maybe interleave spec text from C with implementation of common_type. Also, it's identical, but, if you feel like, common_type's "signed type is bigger" case.

The unsigned_counterpart_of case is actually unreachable. It's to account for C believing that long and long long are always different types, even if the sizes match. That doesn't come up here. Maybe worth a comment (though understandable that we'd otherwise want a boring translation of the spec text).

base_cast_down_if_needed should write down which range is guaranteed vs desired or soemthing.

Jason: "zselect also has the common_type bug".

JasonGross:
Other comment: reminder to deal with aliasing issue in fesub output

PR #380, opened 2018-06-28
[WIP] Generating C files with the new pipeline

JasonGross:
The new pipeline generates C files!  (Target `c-files`.)  (And they compile with https://godbolt.org/)

(There are some minor unfortunate things about the pipeline, like that the OCaml source file is about 67 MB.)

There are still a couple of `admit`s that I haven't fixed yet (I think mostly around `freeze`).  Montgomery isn't done yet.  (And there are still not really any proofs.)

However, I'd like to work on getting the C files / fiat-crypto in a state where it can be integrated with Chrome's code and CI, in parallel with doing Montgomery and proofs.  So I'd like review comments on the generated C code, with regards to what should to be done for integration (beyond proofs).
cc @agl @davidben

As far as performance goes, running the extracted pipeline takes about 2 seconds (2.3 real, 2.0 user for 32-bit 25519; 2.0 real, 1.7 user for 64-bit 25519).   Compiling the pipeline in ocamlopt takes about 20 seconds.  The slowest part of compilation is reducing the rewriter, which takes about 200 seconds.  (Also, freeze is really slow to synthesize in the vm because it spends a lot of time encoding `2^255 - 1` multiple times in various bases.)


Makefile
```diff
@@ -487,6 +487,19 @@ standalone: standalone-haskell standalone-ocaml
 standalone-haskell: $(STANDALONE:%=src/Experiments/NewPipeline/ExtractionHaskell/%)
 standalone-ocaml: $(STANDALONE:%=src/Experiments/NewPipeline/ExtractionOCaml/%)
 
+UNSATURATED_SOLINAS_C_FILES := curve_25519_64.c curve_25519_32.c
```
davidben: Nit: usually people write "curve25519" rather than "curve_25519".

curve_25519_64.c
```diff
@@ -0,0 +1,333 @@
...
+  out1[4] = x5;
+}
+
+/*
+ * Input Bounds:
+ *   arg1: [[0x0 ~> 0x8cccccccccccc], [0x0 ~> 0x8cccccccccccc], [0x0 ~> 0x8cccccccccccc], [0x0 ~> 0x8cccccccccccc], [0x0 ~> 0x8cccccccccccc]]
+ * Output Bounds:
+ *   out1: [[0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0x7f]]
+ */
+void fiat_25519_freeze(uint8_t out1[32], uint64_t arg1[5]) {
```
davidben: Now that you've incorporated the conversion to bytes (yay!), mind calling this `tobytes` or `to_bytes` or something of the sort. Also, if it's easy, `from_bytes` would also be useful.

PR #383, opened 2018-06-30
Adapt to coq#7254

jashug:
We make all parameters implicit, and set `Uniform Inductive Parameters`.

I expect this is backwards compatible, hopefully CI backs me up.

Not to be merged until coq/coq#7254 and coq/coq#7703 have converged.


JasonGross:
> Not to be merged until

When you're ready for this to be merged, please rebase so the CI runs again, and leave a comment here saying that things have converged.

PR #390, opened 2018-07-12
Allow specifying the bounds multipliers manually

JasonGross:


JasonGross:
This is old and operates on outdated code, so I'm going to close it.  If we want to do this in the future, I'll reimplement the small change required.

PR #391, opened 2018-07-12
Allow saturated bounds on unsat solinas

JasonGross:
Possible alternative to #390


JasonGross:
This is old and operates on outdated code, so I'm going to close it.  If we want to do this in the future, I'll reimplement the small change required.

PR #393, opened 2018-07-12
Use bounds of s-1 rather than bitwidth for saturated solinas

JasonGross:
We now use bounds of [encode (s-1)] (e.g., 2^127-1) for saturated
solinas, rather than bounds of [repeat (2^bitwidth-1)].  This makes
2^127-1 come closer to bounds checking.


JasonGross:
Now on top of #395

JasonGross:
This is old and operates on outdated code, so I'm going to close it, as it caused bounds-checking to fail with the fact that computed bounds were not tight enough.

Issue #326, opened 2018-03-18
feature request: saturated solinas


0x64616E69656C:
As discussed with Jason by email, saturated solinas with equal base and bitwidth is currently not supported, but is being looked into. This issue is just opened to keep track of it's progress.

jadephilipoom:
Saturated solinas is definitely within the range of things we can support
in the near future (pending the unrelated pipeline refactoring Jason
mentioned in the email thread). We'll keep you in the loop.

jadephilipoom:
Some progress has been made with this in #357.

JasonGross:
Some more progress has been made with #373.  It is not yet proven, but we can now synthesize C code for mulmod for saturated solinas.  It only seems to work for primes of the form `2^n - a - b - ... - z`, and not work for primes that have both `+` and `-`, because the code ends up doing an add-with-get-carry where bounds analysis can't tell whether the output bit is a carry or a borrow.  I have no idea how to fix this.  (We also don't yet do the operations other than mulmod for saturated solinas, but I expect that's easier to fix.)  Proofs should come later this summer.

0x64616E69656C:
Yeah, still no success with solinas64_2e127m1_2limbs (coq 8.7,2, gotta upgrade to 8.8.0):

```
make --no-print-directory -C coqprime src/Coqprime/PrimalityTest/Zp.vo
make[1]: 'src/Coqprime/PrimalityTest/Zp.vo' is up to date.
COQC src/Specific/solinas64_2e127m1_2limbs/Synthesis.v
File "./src/Specific/solinas64_2e127m1_2limbs/Synthesis.v", line 6, characters 9-31:
Error:
In nested Ltac calls to "make_Synthesis_package", "get_Synthesis_package",
"add_Synthesis_package",
"ReificationTypesPackage.add_ReificationTypes_package",
"ReificationTypesPackage.add_feBW_tight_bounded",
"ReificationTypes.pose_feBW_tight_bounded",
"ReificationTypes.internal_pose_feBW_bounded",
"CacheTerm.cache_proof_with_type_by",
"CacheTerm.cache_term_with_type_by_gen", "abstract_tac" (bound to
fun id' => abstract (tac) using id'), "tac" (bound to
apply
 (ReificationTypes.feBW_bounded_helper sz adjusted_bitwidth' bounds wt
    wt_nonneg); cbv-[BinInt.Z.lt BinInt.Z.le]; clear; Decidable.vm_decide) and
[...]

PR #395, opened 2018-07-14
Use a better strategy in Rows.sat_reduce

JasonGross:
We now preserve the invariant that weights are in the image of the
weight function, insofar as we can, which allows a more fine-tuned
analysis of values.

With some inspiration and help from @jadephilipoom .  This should help with #326

### DIFF
--- a/src/Experiments/NewPipeline/Arithmetic.v
+++ b/src/Experiments/NewPipeline/Arithmetic.v
@@ -15,6 +15,7 @@ Require Import Crypto.Arithmetic.PrimeFieldTheorems.
 Require Import Crypto.Util.ZUtil.Tactics.PullPush.Modulo.
 Require Import Crypto.Util.Tactics.RunTacticAsConstr.
 Require Import Crypto.Util.Tactics.Head.
+Require Import Crypto.Util.Tactics.RewriteHyp.
 Require Import Crypto.Util.Option.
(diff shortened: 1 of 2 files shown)


JasonGross:
The timing diff is

```
After     | File Name                                                            | Before    || Change     | % Change
---------------------------------------------------------------------------------------------------------------------
79m58.21s | Total                                                                | 19m03.93s || +60m54.27s | +319.44%
---------------------------------------------------------------------------------------------------------------------
66m49.59s | Experiments/NewPipeline/SlowPrimeSynthesisExamples.vo                | 6m10.66s  || +60m38.93s | +981.74%
4m39.21s  | Experiments/NewPipeline/Toplevel1.vo                                 | 4m35.82s  || +0m03.38s  | +1.22%
1m34.88s  | Experiments/NewPipeline/Toplevel2.vo                                 | 1m32.56s  || +0m02.31s  | +2.50%
```

This is rather unfortunate; we should possibly dig into what's going on here before merging this.

JasonGross:
It seems that, of the 3987.84s, 3653.58 of them are taken up by `Time Print mulmod` at the end?  o.O

JasonGross:
This is presumably because the number of lines of code in the synthesized p256 mulmod went from 2253 to 7575.

PR #413, opened 2018-08-09
README: note remake_curves.sh is in Specific/CurveParameters

gtank:
The readme gives a close but not correct location for the remake_curves.sh script.


JasonGross:
Thanks!

I will note that in the near future (as soon as I finish the proofs I'm working on), we'll be removing the current pipeline, and replacing it with a better one (which currently lives in `src/Experiments/NewPipeline`).

PR #434, opened 2018-09-16
Fix up canonical-representation arithmetic logic

jadephilipoom:
Previously, Columns and Rows had a mishmash of different techniques and proofs trying to declare various properties of canonical representations. In preparation for proving some of the WBW Montgomery admits, I have done a long-overdue overhaul (I also used it to prove a couple of those).

Now, we use `partition (weight nat -> Z) (n : nat) : Z -> list Z`, which splits an integer across `n` digits according to the weight function.  Proofs of correctness for functions splitting out a canonical representation are phrased in terms of this function. We also have a proof that evaluating the result produces the input modulo `weight n`, so the div/mod properties of those other functions come "for free" with the canonicalization proof.

### DIFF
--- a/src/Experiments/NewPipeline/Arithmetic.v
+++ b/src/Experiments/NewPipeline/Arithmetic.v
@@ -2474,7 +2474,7 @@ hd 0 p).
   Hint Rewrite length_from_columns using eassumption : distr_length.
   Hint Rewrite length_sum_rows using solve [ reflexivity | eassumption | distr_length; eauto ] : distr_length.
   Hint Rewrite length_fst_extract_row length_snd_extract_row length_flatten length_flatten' length_partition length_fst_from_columns' length_snd_from_columns' : distr_length.
-  Hint Rewrite @eval_partition : push_eval.
+  Hint Rewrite @eval_partition using auto : push_eval.
 End Rows.
 
 Module BaseConversion.
@@ -3064,6 +3064,30 @@ Module UniformWeight.
     rewrite !Z.pow_succ_r by auto using Nat2Z.is_nonneg.
     ring.
   Qed.
+
+  (* Because the weight is uniform, we can start partitioning from
+  any index and end up with the same result. *)
+  Lemma uweight_recursive_partition_change_start lgr (Hr : 0 <= lgr) n :
+    forall i j x,
+      Partition.recursive_partition (uweight lgr) n i x
+      = Partition.recursive_partition (uweight lgr) n j x.
+  Proof.
+    induction n; intros; [reflexivity | ].
+    cbn [Partition.recursive_partition].
+    rewrite !uweight_eq_alt by omega.
+    autorewrite with push_Zof_nat push_Zpow.
+    rewrite <-!Z.pow_sub_r by auto using Z.pow_nonzero with omega.
+    rewrite !Z.sub_succ_l.
+    autorewrite with zsimplify_fast.
+    erewrite IHn. reflexivity.
+  Qed.
+  Lemma uweight_recursive_partition_equiv lgr (Hr : 0 < lgr) n i x:
+    Partition.partition (uweight lgr) n x =
+    Partition.recursive_partition (uweight lgr) n i x.
+  Proof.
+    rewrite Partition.recursive_partition_equiv by auto using uwprops.
+    auto using uweight_recursive_partition_change_start with omega.
+ Qed.
 End UniformWeight.
 
 Module WordByWordMontgomery.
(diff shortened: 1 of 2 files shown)


src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -1260,76 +1288,153 @@ Module Saturated.
...
-  Section DivMod.
-    Lemma mod_step a b c d: 0 < a -> 0 < b ->
-                            c mod a + a * ((c / a + d) mod b) = (a * d + c) mod (a * b).
+Module Partition.
+  Section Partition.
+    Context weight {wprops : @weight_properties weight}.
+    Definition partition n x :=
+      map (fun i => (x mod weight (S i)) / weight i) (seq 0 n).
+
+    (* TODO : move to ListUtil *)
```
JasonGross: Please actually move this to ListUtil rather than leaving a to-do

src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -1260,76 +1288,153 @@ Module Saturated.
...
+
+    Lemma length_recursive_partition n : forall i x,
+      length (recursive_partition n i x) = n.
+    Proof.
+      induction n; cbn [recursive_partition]; [reflexivity | ].
+      intros; distr_length; auto.
+    Qed.
+    Hint Rewrite length_recursive_partition : distr_length.
+
+    (* TODO : move *)
```
JasonGross: Please actually move this to NatUtil

src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -510,7 +510,7 @@ Module Positional. Section Positional.
 
   Lemma eval_snoc_S n x y : n = length x -> eval (S n) (x ++ [y]) = eval n x + weight n * y.
   Proof using Type. intros; erewrite eval_snoc; eauto. Qed.
-  Hint Rewrite eval_snoc_S : push_eval.
+  Hint Rewrite eval_snoc_S using (solve [distr_length]) : push_eval.
```
JasonGross: I think you don't need the parentheses here.

src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -918,7 +918,7 @@ Module Positional. Section Positional.
 End Positional.
 (* Hint Rewrite disappears after the end of a section *)
 Hint Rewrite length_zeros length_add_to_nth length_from_associational @length_add @length_carry_reduce @length_chained_carries @length_encode @length_sub @length_opp @length_select @length_zselect @length_select_min @length_extend_to_length @length_drop_high_to_length : distr_length.
-Hint Rewrite @eval_zeros @eval_nil @eval_snoc_S @eval_select @eval_zselect @eval_extend_to_length (*@eval_drop_high_to_length*) : push_eval.
+Hint Rewrite @eval_zeros @eval_nil @eval_snoc_S @eval_select @eval_zselect @eval_extend_to_length (*@eval_drop_high_to_length*) using (solve [auto; distr_length]): push_eval.
```
JasonGross: I think you don't need the parentheses here.

src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -1405,32 +1378,9 @@ Module Partition.
...
-               | _ => progress autorewrite with zsimplify_fast natsimplify push_skipn
-               | _ => reflexivity
-               end.
-      erewrite IHm by (auto; omega).
-      repeat (f_equal; try omega).
-    Qed.
   End Partition.
   Hint Rewrite length_partition length_recursive_partition : distr_length.
-  Hint Rewrite eval_partition using auto : push_eval.
+  Hint Rewrite eval_partition using (solve [auto; distr_length]) : push_eval.
```
JasonGross: I think you don't need the parentheses here.

JasonGross:
Can you rebase this on current master to fix the 8.7 build?

src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -1260,76 +1288,103 @@ Module Saturated.
...
   End Associational.
+End Saturated.
 
-  Section DivMod.
-    Lemma mod_step a b c d: 0 < a -> 0 < b ->
-                            c mod a + a * ((c / a + d) mod b) = (a * d + c) mod (a * b).
+Module Partition.
+  Section Partition.
+    Context weight {wprops : @weight_properties weight}.
+    Definition partition n x :=
```
andres-erbsen: This used to be called `encode`, right? `partition` is also the name of a list function that is used in `mod` implementation.

jadephilipoom:
Superseded by #436

PR #437, opened 2018-09-20
Prove admits for word-by-word Montgomery

jadephilipoom:
I also did a couple things along the way, like:
- relocate some proofs about weight functions (they were previously in `Saturated`, but had nothing to do with saturated weights)
- restructure the `eval` and `small` proofs a bit so that work wasn't duplicated--instead I proved that they were both at the same time and then made tiny proofs to say each separately
- remove some cruft -- unnecessary preconditions to a few lemmas and a commented-out lemma that ended up not being needed

### DIFF
--- a/src/Experiments/NewPipeline/Arithmetic.v
+++ b/src/Experiments/NewPipeline/Arithmetic.v
@@ -480,7 +480,76 @@ Module Associational.
   End Carries.
 End Associational.
 
-Module Positional. Section Positional.
+Module Weight.
+  Section Weight.
+    Context weight
+            (weight_0 : weight 0%nat = 1)
+            (weight_positive : forall i, 0 < weight i)
+            (weight_multiples : forall i, weight (S i) mod weight i = 0)
+                                                                              (weight_divides : forall i : nat, 0 < weight (S i) / weight i).
+
+    Hint Resolve Z.positive_is_nonzero Z.lt_gt Nat2Z.is_nonneg.
+
+    Lemma weight_multiples_full' j : forall i, weight (i+j) mod weight i = 0.
+    Proof using weight_positive weight_multiples.
+      induction j; intros;
+        repeat match goal with
+               | _ => rewrite Nat.add_succ_r
+               | _ => rewrite IHj
+               | |- context [weight (S ?x) mod weight _] =>
+                 rewrite (Z.div_mod (weight (S x)) (weight x)), weight_multiples by auto
+               | _ => progress autorewrite with push_Zmod natsimplify zsimplify_fast
+               | _ => reflexivity
+               end.
+    Qed.
+
+    Lemma weight_multiples_full j i : (i <= j)%nat -> weight j mod weight i = 0.
+    Proof using weight_positive weight_multiples.
+      intros; replace j with (i + (j - i))%nat by omega.
+      apply weight_multiples_full'.
+    Qed.
+
+    Lemma weight_divides_full j i : (i <= j)%nat -> 0 < weight j / weight i.
+    Proof using weight_positive weight_multiples. auto using Z.gt_lt, Z.div_positive_gt_0, weight_multiples_full. Qed.
+
+    Lemma weight_div_mod j i : (i <= j)%nat -> weight j = weight i * (weight j / weight i).
+    Proof using weight_positive weight_multiples. intros. apply Z.div_exact; auto using weight_multiples_full. Qed.
+
+    Lemma weight_mod_pull_div n x :
+      x mod weight (S n) / weight n =
+      (x / weight n) mod (weight (S n) / weight n).
+    Proof using weight_positive weight_multiples weight_divides.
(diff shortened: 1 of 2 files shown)


src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -480,7 +480,76 @@ Module Associational.
...
 End Associational.
 
-Module Positional. Section Positional.
+Module Weight.
+  Section Weight.
+    Context weight
+            (weight_0 : weight 0%nat = 1)
+            (weight_positive : forall i, 0 < weight i)
+            (weight_multiples : forall i, weight (S i) mod weight i = 0)
+                                                                              (weight_divides : forall i : nat, 0 < weight (S i) / weight i).
```
JasonGross: Indentation?

src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -891,20 +926,30 @@ Module Positional. Section Positional.
     Lemma length_zselect mask cond p :
       length (zselect mask cond p) = length p.
     Proof using Type. clear dependent weight. cbv [zselect Let_In]; break_match; intros; distr_length. Qed.
+
+    (* We need an explicit equality proof here, because sometimes it
```
JasonGross: Mention that the alternative to "an explicit equality proof" is "a proof of equality under eval" (or just name `eval_select` as the weaker version)

src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -1377,16 +1366,26 @@ Module Partition.
       induction n; cbn [recursive_partition]; [reflexivity | ].
       intros; distr_length; auto.
     Qed.
-    Hint Rewrite length_recursive_partition : distr_length.
+
+    Lemma drop_high_to_length_partition n m x :
+      (n <= m)%nat ->
+      Positional.drop_high_to_length n (partition m x) = partition n x.
+    Proof.
```
JasonGross: Might be nice to use `Proof using` here.  (You can get the correct annotation either by doing `Proof using Type.` and seeing how `Qed` fails, or else by doing `Set Suggest Proof Using`).  This would allow more parallelism in async mode.

src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -2178,22 +2177,64 @@ Module Rows.
...
+        length p = n -> length q = n ->
         snd (sub n p q) = (Positional.eval weight n p - Positional.eval weight n q) / weight n.
       Proof using wprops. solver. Qed.
 
+      Lemma conditional_sub_partitions n p q
+        (Hp : p = partition weight n (Positional.eval weight n p)) :
+        length q = n ->
+        0 <= Positional.eval weight n q < weight n ->
+        conditional_sub n p q = partition weight n (if Positional.eval weight n q <=? Positional.eval weight n p then Positional.eval weight n p - Positional.eval weight n q else Positional.eval weight n p).
+      Proof.
```
JasonGross: Might be nice to use `Proof using` here.

src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -2823,6 +2864,22 @@ Module UniformWeight.
     rewrite !Z.pow_succ_r by auto using Nat2Z.is_nonneg.
     ring.
   Qed.
+  Lemma uweight_S lgr (Hr : 0 <= lgr) n : uweight lgr (S n) = 2 ^ lgr * uweight lgr n.
+  Proof.
```
JasonGross: Might be nice to use `Proof using` here.

src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -2823,6 +2864,22 @@ Module UniformWeight.
...
+    rewrite !uweight_eq_alt by auto.
+    autorewrite with push_Zof_nat.
+    rewrite Z.pow_succ_r by auto using Nat2Z.is_nonneg.
+    reflexivity.
+  Qed.
+  Lemma uweight_double_le lgr (Hr : 0 < lgr) n : uweight lgr n + uweight lgr n <= uweight lgr (S n).
+  Proof.
```
JasonGross: Might be nice to use `Proof using` here.

src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -2851,6 +2908,7 @@ End UniformWeight.
 
 Module WordByWordMontgomery.
   Import Partition.
+  Hint Resolve Z.positive_is_nonzero Z.lt_gt Nat2Z.is_nonneg.
```
JasonGross: Make this `Local Hint`, or else stick it in the section.

JasonGross:
Now that #454 has been merged, if you rebase this on top of master, Travis should be happy.

PR #445, opened 2018-10-04
Compatibility with Coq PR#8457

fajb:
Based on hints from @andres-erbsen


andres-erbsen:
(Travis looks like a flake, testing on MIT infrastructure now...)

PR #448, opened 2018-10-10
Add interp-correctness condition for rewriter

JasonGross:
The NBE-rewrite rules are proven correct.  The arith and fancy rewrite
rules are reduced to lemmas about Z, most of which are inconsistent (and
therefore indicate rewrite rules which are missing side-conditions).

One bad rewrite rule was found in this process, and corrected in
0842563b23f8d780f4ff1080d7620fc3f368191f

The next step after this will be using the rewrite rule correctness to
prove the rewriter correct.

@jadephilipoom I'd like to leave the remaining admits in `src/Experiments/NewPipeline/RewriterRulesInterpGood.v` in this PR to you (see the comments before `exact admit`); there are 25 of them, and most of them are false.  If you back up to after the `all: try solve [ ... ]` lines, you can see where the goals originally come from; I'm happy to help you adjust the rewrite rules to be valid.  Are you willing to take this on your plate / on our joint plate?

@andres-erbsen You might or might not be interested in commenting on the `with_interp` section in `src/Experiments/NewPipeline/RewriterWf1.v`, which defines the interp correctness condition for rewrite rules.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -264,6 +264,7 @@ src/Experiments/NewPipeline/MiscCompilerPassesProofs.v
 src/Experiments/NewPipeline/Rewriter.v
 src/Experiments/NewPipeline/RewriterProofs.v
 src/Experiments/NewPipeline/RewriterRulesGood.v
+src/Experiments/NewPipeline/RewriterRulesInterpGood.v
 src/Experiments/NewPipeline/RewriterWf1.v
 src/Experiments/NewPipeline/RewriterWf2.v
 src/Experiments/NewPipeline/SlowPrimeSynthesisExamples.v
--- a/src/Experiments/NewPipeline/RewriterRulesInterpGood.v
+++ b/src/Experiments/NewPipeline/RewriterRulesInterpGood.v
@@ -1,0 +1,532 @@
+Require Import Coq.ZArith.ZArith.
+Require Import Coq.micromega.Lia.
+Require Import Coq.Lists.List.
+Require Import Coq.Classes.Morphisms.
+Require Import Coq.MSets.MSetPositive.
+Require Import Coq.FSets.FMapPositive.
+Require Import Crypto.Experiments.NewPipeline.Language.
+Require Import Crypto.Experiments.NewPipeline.LanguageInversion.
+Require Import Crypto.Experiments.NewPipeline.LanguageWf.
+Require Import Crypto.Experiments.NewPipeline.UnderLetsProofs.
+Require Import Crypto.Experiments.NewPipeline.GENERATEDIdentifiersWithoutTypesProofs.
+Require Import Crypto.Experiments.NewPipeline.Rewriter.
+Require Import Crypto.Experiments.NewPipeline.RewriterWf1.
+Require Import Crypto.Util.ZUtil.Tactics.LtbToLt.
+Require Import Crypto.Util.ZUtil.Tactics.DivModToQuotRem.
+Require Import Crypto.Util.ZUtil.Tactics.PullPush.Modulo.
+Require Import Crypto.Util.ZUtil.Hints.
+Require Import Crypto.Util.ZUtil.Hints.Core.
+Require Import Crypto.Util.ZUtil.ZSimplify.Core.
+Require Import Crypto.Util.ZUtil.ZSimplify.
+Require Import Crypto.Util.ZUtil.ZSimplify.Simple.
+Require Import Crypto.Util.ZUtil.Definitions.
+Require Import Crypto.Util.ZUtil.AddGetCarry.
+Require Import Crypto.Util.ZUtil.MulSplit.
+Require Import Crypto.Util.ZUtil.Zselect.
+Require Import Crypto.Util.Tactics.NormalizeCommutativeIdentifier.
+Require Import Crypto.Util.Tactics.BreakMatch.
+Require Import Crypto.Util.Tactics.SplitInContext.
+Require Import Crypto.Util.Tactics.SpecializeAllWays.
+Require Import Crypto.Util.Tactics.SpecializeBy.
+Require Import Crypto.Util.Tactics.RewriteHyp.
+Require Import Crypto.Util.Tactics.Head.
+Require Import Crypto.Util.Prod.
+Require Import Crypto.Util.Bool.
+Require Import Crypto.Util.ListUtil.
+Require Import Crypto.Util.ListUtil.ForallIn.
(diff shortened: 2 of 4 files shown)


jadephilipoom:
I'll probably want your help with the first 9 admits, since I don't really understand what they're saying.

The other 16 pretty much seem to boil down to bounds preconditions. For instance, I should only do the rewrites including `shiftl` if the thing being shifted is small enough, because the rewrite changes a normal left-shift to a truncating left-shift. Is there a way to access bounds information from the rewriter? I would want, for instance, this rule:

```
make_rewriteo
  (#pattern.ident.Z_add_get_carry @ #?ℤ @ ?? @ (#pattern.ident.Z_shiftl @ ?? @ #?ℤ))
  (fun s x y offset => #(ident.fancy_add (Z.log2 s) offset) @ (x, y)
     when  s =? 2^Z.log2 s)
```

...to have an extra condition that looks like:

```
make_rewriteo
  (#pattern.ident.Z_add_get_carry @ #?ℤ @ ?? @ (#pattern.ident.Z_shiftl @ ?? @ #?ℤ))
  (fun s x y offset => #(ident.fancy_add (Z.log2 s) offset) @ (x, y) 
     when  ((s =? 2^Z.log2 s) && (in_range y r[0~>2^(Z.log2 s - offset)]))
```

JasonGross:
The way to read the first 9 admits is as
```coq
 match
    REPLACEMENT
  with
  | Some v0 =>
      expr.interp (@ident.interp) (UnderLets.interp (@ident.interp) v0) =
      RULE
  | None => True
  end
```
so, e.g.,
```coq
 match
    (x1 <- rwhen (Some (v1, (##0)%expr)%expr_pat) (x0 =? 1);
     Some (UnderLets.Base x1))%option
  with
  | Some v0 =>
      expr.interp (@ident.interp) (UnderLets.interp (@ident.interp) v0) =
      Z.mul_split x x0 (expr.interp (@ident.interp) v1)
  | None => True
  end
```
corresponds to the rule that replaces `Z.mul_split x x0 (expr.interp (@ident.interp) v1)` with `rwhen (Some (v1, (##0)%expr)%expr_pat) (x0 =? 1)`, i.e., the rule
```coq
make_rewriteo (#pattern.ident.Z_mul_split @ #?ℤ @ #?ℤ @ ??) (fun s xx y => (y, ##0)%Z  when  xx =? 1)
```
(the arguments wrapped in `expr.interp` are wildcards; the ones not wrapped in `expr.interp` are usually literals).  This rule is false, for example, when the last argument is not bounded by the bitwidth.

JasonGross:
> Is there a way to access bounds information from the rewriter?

There is not, currently.  I see two ways to add support for this:
1. I can run the rewriter again after bounds analysis.  The rewrite rules used here must annotate all operations with `Z_cast` or `Z_cast2`, and then you will have access to the bounds for each cast node.  (Note that the rewrite rule will not match if you miss some cast nodes.)  I will have to update the interp proof to be parameterized over the cast-outside-of-range behavior, but this should be pretty easy.
2. An alternative would be to add an identifier that is `if <boolean condition of the bounds on an expression> then A else B`, and teach the bounds analysis pass to interpret this identifier.  This may be more generalizable to dealing with bounds information that needs to be analyzed / proven before the algorithm is reified, but is a great deal more work.  The only way I see to make this work is to parameterize the entire framework over a hinted non-deterministic (parameterized/"implementation-defined") choice of bounds for a given value (where the only guarantee that you get is that the value is in the bounds chosen).  Then you need to somehow pick out this constant when reifying so your reified AST can be parametric over it, and you need to prove before-hand that your algorithm does the same thing regardless of choice of this function.  Then I'd need to update all of the passes to be parameterized over this function, and I'd need to update bounds analysis to replace uses of this function with the actual bounds.

I'm inclined to pick option (1), because I expect option (2) to be somewhere between a week and a month of additional work (maybe more if I hit truly unexpected snags), so I'd rather leave (2) for a future version of fiat-crypto.

With option (1), the rewrite rule would become
```coq
make_rewriteo
  (#pattern.ident.Z_add_get_carry @ (#pattern.ident.Z_cast @ #?ℤ) @ (#pattern.ident.Z_cast @ ??) @ (#pattern.ident.Z_cast @ (#pattern.ident.Z_shiftl @ (#pattern.ident.Z_cast @ ??) @ (#pattern.ident.Z_cast @ #?ℤ))))
  (fun sr s xr x _ yr y offsetr offset => #(ident.fancy_add (Z.log2 s) offset) @ (#(ident.Z_cast xr) @ x, #(ident.Z_cast yr) @ y) 
     when  ((s =? 2^Z.log2 s) && (ZRange.is_tighter_than_bool yr r[0~>2^(Z.log2 s - offset)]))
```

@jadephilipoom @andres-erbsen Thoughts?

jadephilipoom:
My main thought is that making another rewriter to run after bounds analysis for just fancy-machine stuff is exactly what I was doing before with my specialized post-bounds-analysis pass. But option (1) is okay by me--as long as I'm not taking on the responsibility of writing this new rewriting pass.

However, I do have one other idea. Maybe if we changed the definitions of fancy-machine operations at this level of abstraction (i.e. the interpretation of `ident.fancy_addc`) so that they *don't* have bounds preconditions, then we could push the bounds preconditions into the post-bounds-analysis `of_prefancy` pass, which already has to deal with casts a bit.

JasonGross:
> But option (1) is okay by me--as long as I'm not taking on the responsibility of writing this new rewriting pass.

I'd like your help in either translating the existing rewriting rules to have all of the `Z_cast` nodes in them, or in making sure that my translation is correct.  The rewriting infrastructure will stay basically the same (in fact it already supports running after bounds analysis, I just need to insert the call in the right place and hook up the proofs).

> Maybe if we changed the definitions of fancy-machine operations at this level of abstraction (i.e. the interpretation of `ident.fancy_addc`) so that they don't have bounds preconditions,

Yeah, that could also work, and might be easier.  But it doesn't solve the problem of the nine admits about `Z.mul_split` and `Z.add_{,with_}get_carry_full`, which talk about what to do when you have zeros / ones in your addition/multiplication.  I don't see a way around having these deal with bounds...

jadephilipoom:
> I'd like your help in either translating the existing rewriting rules to have all of the Z_cast nodes in them, or in making sure that my translation is correct. 

Can do. The old `of_straightline` pass should be a good reference for this, but my brain is probably easier to read. And yeah, if you already have most of the infrastructure in place to support a post-bounds-rewriter, then let's do that.

PR #452, opened 2018-10-26
Remove duplicate abstraction

jashug:
`base_type` and `base_interp` are already abstracted by the section.
For compatibility with coq/coq#8820.


JasonGross:
Does it compile with this change?  I might have done it intentionally to allow re-instantiating those arguments with different things...  (Travis failure is current spurious, I'll restart it)

jashug:
Yes, it worked locally.

jashug:
@JasonGross ping

PR #464, opened 2018-12-18
Finish rewriter proofs modulo funext

JasonGross:
Other than the use of function extensionality in a particular place,
this completes the proofs of the rewriter.

It was rather painful to figure out the right way to phrase rewriter
correctness.  I tried a number of more lax approaches (e.g., various
variants of saying what it means for two rawexprs to be equivalent,
saying what it means for two values to be equivalent, etc), but ran into
issues with incomparable relations.  The right approach was to find the
most precise thing that could be said, and to relate each kind of thing
(rawexpr, value, expr, etc) to an interpreted value, and push that
around throughout the proof.  The only exception to this pattern is
`eval_decision_tree`, for which a fine-grained notion of rawexpr
equivalence can be stated (basically, the equivalence-closure of the
reveal-rawexpr operation), and a couple of `Proper` lemmas can be proven
about this relation.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -262,6 +262,7 @@ src/Experiments/NewPipeline/LanguageWf.v
 src/Experiments/NewPipeline/MiscCompilerPasses.v
 src/Experiments/NewPipeline/MiscCompilerPassesProofs.v
 src/Experiments/NewPipeline/Rewriter.v
+src/Experiments/NewPipeline/RewriterInterpProofs1.v
 src/Experiments/NewPipeline/RewriterProofs.v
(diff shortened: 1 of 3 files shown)


JasonGross:
cc @davidben @agl These are the last big admits in the pipeline (I still have a couple of minor cleanup things to do, but the new C files should be ready for merging before the new year)

davidben:
\o/

Thanks a bunch! The C files are the ones checked in at the top-level I assume?

JasonGross:
> The C files are the ones checked in at the top-level

Yes

PR #466, opened 2018-12-19
Add uweight_partition_unique, move weight_1 to uweight_1, add Positional.eval_map_mul

JasonGross:
uweight_partition_unique can probably be generalized to not be specific to uniform weights, but that is left to a future person.

### DIFF
--- a/src/Experiments/NewPipeline/Arithmetic.v
+++ b/src/Experiments/NewPipeline/Arithmetic.v
@@ -550,6 +550,18 @@ Module Positional. Section Positional.
   Lemma zeros_ext_map {A} n (p : list A) : length p = n -> zeros n = map (fun _ => 0) p.
   Proof using Type. cbv [zeros]; intro; subst; induction p; cbn; congruence. Qed.
 
+  Lemma eval_mul_each (n:nat) (a:Z) (p:list Z)
+        (Hn : length p = n)
+    : eval n (List.map (fun x => a*x) p) = a*eval n p.
+  Proof using Type.
+    clear -Hn.
+    transitivity (Associational.eval (map (fun t => (1 * fst t, a * snd t)) (to_associational n p))).
+    { cbv [eval to_associational]; rewrite !combine_map_r.
+      f_equal; apply map_ext; intros; f_equal; nsatz. }
+    { rewrite Associational.eval_map_mul, eval_to_associational; nsatz. }
+  Qed.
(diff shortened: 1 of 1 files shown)


src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -550,6 +550,18 @@ Module Positional. Section Positional.
   Lemma zeros_ext_map {A} n (p : list A) : length p = n -> zeros n = map (fun _ => 0) p.
   Proof using Type. cbv [zeros]; intro; subst; induction p; cbn; congruence. Qed.
 
+  Lemma eval_map_mul (n:nat) (a:Z) (p:list Z)
```
jadephilipoom: Please rename this lemma. It might look similar to `Associational.eval_map_mul`, but the context is a bit different and having them share a name is confusing.

src/Experiments/NewPipeline/Arithmetic.v
```diff
@@ -2878,7 +2896,38 @@ Module UniformWeight.
...
+  Qed.
+
+  Lemma uweight_partition_unique lgr (Hr : 0 < lgr) n ls :
+    length ls = n -> (forall x, List.In x ls -> 0 <= x <= 2^lgr - 1) ->
+    ls = Partition.partition (uweight lgr) n (Positional.eval (uweight lgr) n ls).
+  Proof using Type.
+    intro; subst n.
+    rewrite uweight_recursive_partition_equiv with (i:=0%nat) by assumption.
+    induction ls as [|x xs IHxs]; [ reflexivity | ].
+    repeat first [ progress cbn [List.length Partition.recursive_partition List.In] in *
```
jadephilipoom: I'm surprised by how complex this proof looks (no action necessary)

src/Experiments/NewPipeline/Arithmetic.v
JasonGross: Suggested name?

src/Experiments/NewPipeline/Arithmetic.v
JasonGross: It's probably possible to simplify it.  I did a decent amount of arithmetic reasoning manually.

src/Experiments/NewPipeline/Arithmetic.v
JasonGross: Updated

Issue #475, opened 2019-01-02
refactoring: new pipeline becomes default


andres-erbsen:
As @JasonGross just finished proofs for the `Experiments/NewPipeline` support for Solinas reduction and word-by-word Montgomery reduction, I am tempted to move this code out of the `Experiments` directory into the spotlight of this repository. I think this would also be a major cleanup. I would do this tomorrow, or the day after, or something like that.

This will likely interfere with work in all branches. I am discarding all of mine. If you have branches you care about maintaining, please say so here and we will figure something out. @JasonGross @jadephilipoom 

Currently, the plan is to

~~~
keep: Algebra, Curves, Util, Primitives, Spec, Demo.v, Arithmetic/*Theorems.v
delete: measurements, Specific, Compilers, Arithmetic, benchmarking stuff, expansion.md, regenerate-curves.sh, register-allocate.py, synthesis-parameters.txt, optimizations.md, failures.txt, extract-*.sh, third_party, liblow, Experiments/*.v, LegacyArithmetic
promote: Experiments/NewPipeline/* to toplevel
otherwise move: checked-in generated c code, measure.c to etc
~~~

later we would most likely to git-restore: benchmarking stuff (although BoringSSL benchmarks also work), third_party(?), 

TODO:

[ ] what to do with Arithmetic/BarrettReduction? is it superseded?
[ ] Arithmetic.Saturated?
[ ] how to deal with not-yet-ready stuff in the new pipeline -- FancyMachine and Barrett?

JasonGross:
What do we want to do with the legacy pipeline for Coq's CI?  Do we want to branch?  Or migrate to coq-community?

> Arithmetic/BarrettReduction
We `Require Import Crypto.Arithmetic.BarrettReduction.Generalized.` for the Barrett stuff in the new pipeline.  Similarly with the MontgomeryReduction stuff there.

andres-erbsen:
I'm thinking something like the current master will remain as a branch.

JasonGross:
We should make sure to update Coq's CI before we change master.  Things to do:
- make a section like https://github.com/coq/coq/blob/0a5bbf347b5bbcb579f94eb3d0166778cd92cfdb/dev/ci/ci-basic-overlay.sh#L145-L150 for fiat-crypto-legacy
- update https://github.com/coq/coq/blob/master/dev/ci/ci-fiat-crypto-legacy.sh to use the alternative branch
[...]

PR #477, opened 2019-01-08
Promote new pipeline

andres-erbsen:
As per #475


andres-erbsen:
8.8 all https://builds.sr.ht/~andres/job/20584
master all https://builds.sr.ht/~andres/job/20585

andres-erbsen:
https://github.com/mit-plv/fiat-crypto/tree/sp2019latest is the left-behind branch

andres-erbsen:
local build time for `all`: 33min

JasonGross:
Please make sure that Travis passes and that you at least submit a PR updating Coq's CI before merging this.  Also we should submit a new opam package or something

Makefile
```diff
@@ -21,145 +21,60 @@ INSTALLDEFAULTROOT := Crypto
...
-UNMADE_VOFILES :=
-UNMADE_C_FILES := \
-	src/Specific/X25519/C64/fesub.c src/Specific/X25519/C64/feadd.c src/Specific/X25519/C64/fecarry.c \
-	src/Specific/X25519/C64/fesub.h src/Specific/X25519/C64/feadd.h src/Specific/X25519/C64/fecarry.h \
-	src/Specific/X25519/C32/fesub.c src/Specific/X25519/C32/feadd.c src/Specific/X25519/C32/fecarry.c \
-	src/Specific/X25519/C32/fesub.h src/Specific/X25519/C32/feadd.h src/Specific/X25519/C32/fecarry.h
-# files that are treated specially
+# coq .vo files that are not compiled using coq_makefile
 SPECIAL_VOFILES := \
 	src/Specific/%Display.vo \
```
JasonGross: Why do you leave `Specific` here?

Makefile
```diff
@@ -21,145 +21,60 @@ INSTALLDEFAULTROOT := Crypto
...
+	src/Curves/Weierstrass/AffineProofs.vo \
 	src/Curves/Weierstrass/Jacobian.vo
-NO_CURVES_PROOFS_NON_SPECIFIC_UNMADE_VOFILES := $(filter $(NO_CURVES_PROOFS_UNMADE_VOFILES) src/Specific/%.vo,$(VOFILES))
-REAL_SPECIFIC_GENERATED_VOFILES := $(filter $(SPECIFIC_GENERATED_VOFILES),$(VOFILES))
-NEW_PIPELINE_PRE_VOFILES := $(filter $(NEW_PIPELINE_FILTER),$(REGULAR_VOFILES))
-PRE_STANDALONE_PRE_VOFILES := $(filter src/Experiments/NewPipeline/Standalone%.vo,$(REGULAR_VOFILES))
+PRE_STANDALONE_PRE_VOFILES := $(filter src/Standalone%.vo,$(REGULAR_VOFILES))
 UTIL_PRE_VOFILES := $(filter bbv/%.vo src/Algebra/%.vo src/Tactics/%.vo src/Util/%.vo,$(REGULAR_VOFILES))
-COQ_WITHOUT_NEW_PIPELINE_VOFILES := $(filter-out $(NEW_PIPELINE_FILTER),$(REGULAR_VOFILES))
-SOME_EARLY_VOFILES := \
```
JasonGross: Why remove this?  I think it's essential for getting Travis to not time out.

Makefile
```diff
@@ -169,81 +84,30 @@ UTIL_VOFILES := $(call vo_closure,$(UTIL_PRE_VOFILES))
...
 no-curves-proofs: $(NO_CURVES_PROOFS_VOFILES)
-no-curves-proofs-non-specific: $(NO_CURVES_PROOFS_NON_SPECIFIC_VOFILES)
-new-pipeline: $(NEW_PIPELINE_VOFILES)
 pre-standalone: $(PRE_STANDALONE_VOFILES)
-specific-display: $(SPECIFIC_DISPLAY_VO:.vo=.log)
-specific-c: $(filter-out $(UNMADE_C_FILES),$(SPECIFIC_DISPLAY_VO:Display.vo=.c) $(SPECIFIC_DISPLAY_VO:Display.vo=.h))
-selected-specific: $(SELECTED_SPECIFIC_VOFILES)
-selected-specific-display: $(SELECTED_SPECIFIC_DISPLAY_VO:.vo=.log)
-selected-c: $(filter-out $(UNMADE_C_FILES),$(SELECTED_SPECIFIC_DISPLAY_VO:Display.vo=.c) $(SELECTED_SPECIFIC_DISPLAY_VO:Display.vo=.h))
-some-early: $(SOME_EARLY_VOFILES)
```
JasonGross: I think this should come back, for Travis.

Makefile
```diff
@@ -623,15 +250,8 @@ p224_64.c p224_32.c : p224_%.c :
 	$(SHOW)'SYNTHESIZE > $@'
 	$(HIDE)$(TIMER_FULL) $(WORD_BY_WORD_MONTGOMERY) 'p224' '2^224' '2^96,1;1,-1' '$*' > $@
 
-CFLAGS?=
-.PHONY: only-test-c-files test-c-files
-only-test-c-files test-c-files:
```
JasonGross: Why are you removing the test that the new pipeline generates .c files that compile?

README.md
```diff
@@ -1,8 +1,14 @@
-[![Build Status](https://api.travis-ci.org/mit-plv/fiat-crypto.png?branch=master)](https://travis-ci.org/mit-plv/fiat-crypto)
-
 Fiat-Crypto: Synthesizing Correct-by-Construction Code for Cryptographic Primitives
 =====
 
+[![Build Status](https://api.travis-ci.org/mit-plv/fiat-crypto.png?branch=master)](https://travis-ci.org/mit-plv/fiat-crypto)
+
+See [src/README.md](src/README.md).
+
+OLD PIPELINE README BELOW
```
JasonGross: Uhhhhh, if the old pipeline no longer exists, maybe we should drop this?

crypto-defects.md
```diff
@@ -1,42 +0,0 @@
-Here is an incomplete list of defects in cryptographic implementations. We
```
JasonGross: Why are we removing this file?

expansion.md
```diff
@@ -1,79 +0,0 @@
-This file is to list tractable, can-be-started-now expansions of the library.
```
JasonGross: This file looks interesting, though I wasn't aware we had it.  Why is it being removed?

JasonGross:
I've made https://github.com/coq/opam-coq-archive/pull/540 for opam and https://github.com/coq/coq/pull/9322 for Coq's CI.

Makefile
andres-erbsen: I don't understand the significance again

Makefile
andres-erbsen: Actually, I do not understand what this target does.

Makefile
JasonGross: It previously contained
```
	src/Experiments/NewPipeline/Arithmetic.vo \
	src/Experiments/NewPipeline/Rewriter.vo \
	src/Experiments/SimplyTypedArithmetic.vo
```
Presumably `SimplyTypedArithmetic` went away?  But it still contained `Rewriter` and `Arithmetic`.  The issue is that Travis is slower than our local machines, so the build, even of just the new pipeline, takes over 50 minutes.  So I picked these two files as a decent chunk of the way through the new pipeline, without any of the proofs, to run early (along with all their dependencies), and then have the rest of the new pipeline in another target.

JasonGross:
Btw, the conflict in Toplevel2 is just that `2 : {` is not valid in Coq 8.7, so I pushed a commit fixing the `fix_fancy4` merge that changed it to `Focus 2. {`

Makefile
JasonGross: > Is there any reason not to just have something that depends on corresponding .o?

Because I don't know enough about implicit makefile rules?  The purpose of the target is just to ensure that the `.c` files don't contain any invalid syntax, etc.

andres-erbsen: Okay, what I am saying is that I don't understand how this target achieves this purpose. What implicit rules are you talking about?

Makefile
andres-erbsen: Okay, the idea makes sense, but how did that work?

andres-erbsen:
So, I think I am approximately at the place I can get here. I will figure something out for the prose files, but I don't understand the Makefile-magic and corresponding .travis.yml-magic that goes into chunking of the build proceess. I wasn't aware how much of this was going on until this PR. @JasonGross what do you think we should do at this point, given that you are busy with other things?

I am strongly in favor of just getting rid of (as much as we can of) it, even if this means not having continuous integration. But that does not need to be resolved for this PR.

Makefile
JasonGross: I am talking about the implicit targets where `make` knows how to compile `.o` files even if you don't tell it how to.  I forgot that `make` could do that.

Is the fact that you can only see the top line confusing?  The full target is
```makefile
CFLAGS?=
.PHONY: only-test-c-files test-c-files
only-test-c-files test-c-files:
	$(CC) -Wall -Wno-unused-function -Werror $(CFLAGS) -c $(ALL_C_FILES)

test-c-files: $(ALL_C_FILES)
```

expansion.md
andres-erbsen: This file looks badly outdated to me, both wrt the state of the repository and what looks worthwhile now. I am currently leaning towards just removing it.

Makefile
andres-erbsen: I do not understand the copy-pasted code. how about `%.o: %.c` `gcc -o %.o %.c`?

Makefile
JasonGross: Each stage in `.travis.yml` invokes a particular set of makefile targets.  The list under `stages:` is the order the stages get run in.  For example, the block
```
    - stage: some-early util
      env: COQ_VERSION="master" COQ_PACKAGE="coq"       PPA="ppa:jgross-h/coq-master-daily"
      allow_failure: true
      script: CUR=early ./etc/ci/travis.sh some-early util
```
says: The stage named "some-early util" with environment variables `COQ_VERSION="master" COQ_PACKAGE="coq"       PPA="ppa:jgross-h/coq-master-daily"` (the `allow_failure: true` is a red herring and currently doesn't work) runs the command `CUR=early ./etc/ci/travis.sh some-early util`.  The script `./etc/ci/travis.sh` passes its command line arguments along to `make`, so this stage builds the make targets `some-early` and `util`.

Makefile
andres-erbsen: Oh, are you relying on gcc -c behavior when passed multiple files, and the entire target is a phony? :D

JasonGross:
> corresponding .travis.yml-magic that goes into chunking of the build proceess.

There is actually not that much magic.  There are a bunch of different targets that build various files, and the travis magic is just sequencing the relevant makefile targets in a relevant order.  The relevant targets are
```
some-early util
printlite lite
pre-standalone
new-pipeline
no-curves-proofs-non-specific
curves-proofs
selected-specific selected-specific-display
build-selected-test build-selected-bench
standalone-ocaml c-files test-c-files CC=gcc
```
Some of these are no longer meaningful without the old pipeline.  They should be removed.  The ones that talked about the new pipeline, or unrelated things, should be preserved.

andres-erbsen:
I guess I fail to see anything other than magic in the makeup of that list -- arbitrary words that contain arbitrary files, just to make travis pass.

Makefile
JasonGross: Does this work when all functions are static?  We may want to keep the -W flags we have, one way or another.

Yes, I'm relying on the target being PHONY to always call gcc, and I'm relying on gcc without -o to deal correctly with multiple arguments to -c.

If your target works (I don't think % does the thing you think it does in the arguments to gcc), then I'm fine with your rule.  (Though `$(CC)` may be better than `gcc` to allow customization.)

JasonGross:
> I guess I fail to see anything other than magic in the makeup of that list -- arbitrary words that contain arbitrary files, just to make travis pass.

Yes, but you don't need to create magic, only do path surgery on existing magic and delete some.

andres-erbsen:
Pushed requested changes, old build still going at https://travis-ci.org/mit-plv/fiat-crypto/builds/477466751?utm_source=github_status&utm_medium=notification

README.md
```diff
@@ -1,159 +1,256 @@
...
- - `saturated_solinas`
- - `unsaturated_solinas`
- - `word_by_word_montgomery`
-
-Passing no arguments, or passing `-h` or `--help` (or any other invalid arguments) will result in a usage message being printed.  These binaries output C code on stdout.
-
-Here are some examples of ways to invoke the binaries (from the directories that they live in):
-
-    # Generate code for 2^255-19
-    ./unsaturated_solinas '25519' '5' '2^255' '1,19' '64' carry_mul carry_square carry_scmul121666 carry add sub opp selectznz to_bytes from_bytes > curve25519_64.c
```
JasonGross: Did you decide that the instructions for invoking the pipeline were not worth keeping, or did you just forget to put this back?

.travis.yml
```diff
@@ -114,104 +104,74 @@ jobs:
...
-    - stage: build-selected-test build-selected-bench
-      env: COQ_VERSION="8.8.2"  COQ_PACKAGE="coq-8.8.2" PPA="ppa:jgross-h/many-coq-versions"
-      script: PREV=selected-specific CUR=build-selected ./etc/ci/travis.sh build-selected-test build-selected-bench
-    - stage: build-selected-test build-selected-bench
-      env: COQ_VERSION="8.7.2"  COQ_PACKAGE="coq-8.7.2" PPA="ppa:jgross-h/many-coq-versions"
-      script: PREV=selected-specific CUR=build-selected ./etc/ci/travis.sh build-selected-test build-selected-bench
+      script: PREV=no-curves-proofs CUR=curves-proofs ./etc/ci/travis.sh curves-proofs
 
     - stage: standalone-ocaml
       env: COQ_VERSION="8.8.2"  COQ_PACKAGE="coq-8.8.2" PPA="ppa:jgross-h/many-coq-versions"
```
JasonGross: This stage needs it's `PREV` updated

Makefile
```diff
@@ -21,145 +23,62 @@ INSTALLDEFAULTROOT := Crypto
...
-	src/Experiments/NewPipeline/Toplevel2.vo \
-	src/Experiments/NewPipeline/SlowPrimeSynthesisExamples.vo \
-	src/Experiments/SimplyTypedArithmetic.vo \
-	$(SPECIFIC_GENERATED_VOFILES)
+	src/RewriterWf1.vo \
+	src/RewriterWf2.vo \
+	src/RewriterRulesGood.vo \
+	src/RewriterProofs.vo \
+	src/Toplevel2.vo \
+	src/SlowPrimeSynthesisExamples.vo
```
JasonGross: Either this line should be removed....

Makefile
```diff
@@ -21,145 +23,62 @@ INSTALLDEFAULTROOT := Crypto
...
-REAL_SPECIFIC_GENERATED_VOFILES := $(filter $(SPECIFIC_GENERATED_VOFILES),$(VOFILES))
-NEW_PIPELINE_PRE_VOFILES := $(filter $(NEW_PIPELINE_FILTER),$(REGULAR_VOFILES))
-PRE_STANDALONE_PRE_VOFILES := $(filter src/Experiments/NewPipeline/Standalone%.vo,$(REGULAR_VOFILES))
+PRE_STANDALONE_PRE_VOFILES := $(filter src/Standalone%.vo,$(REGULAR_VOFILES))
 UTIL_PRE_VOFILES := $(filter bbv/%.vo src/Algebra/%.vo src/Tactics/%.vo src/Util/%.vo,$(REGULAR_VOFILES))
-COQ_WITHOUT_NEW_PIPELINE_VOFILES := $(filter-out $(NEW_PIPELINE_FILTER),$(REGULAR_VOFILES))
 SOME_EARLY_VOFILES := \
-	src/Experiments/NewPipeline/Arithmetic.vo \
-	src/Experiments/NewPipeline/Rewriter.vo \
-	src/Experiments/SimplyTypedArithmetic.vo
```
JasonGross: ..... or this line should not be.

README.md
andres-erbsen: just forgot, will fix once laptop

andres-erbsen:
This probably isn't the cleanest it could be yet, but it builds and carrying a 430K line patch is a pain. I'll be available to fix any issues yet to be found ASAP.

PR #487, opened 2019-01-13
Autocompute s and c in WBW Montgomery

### LINKED ISSUE #342: feature request: primes which are != pow2
As discussed with Jason by email, it might be possible to adjust the montgomery reduction to support primes which are not powers of two (e.g. 2^216*3^137-1).

### LINKED ISSUE #486: Have word_by_word_montgonery compute s and c from the prime
https://github.com/mit-plv/fiat-crypto/issues/342#issuecomment-453802799

JasonGross:
This allows us to support primes which are not a power of 2.

We also add p484 as an example.

To support this, I added a small parser combinator library which can
parse arithmetic expressions involving `()^*/+-` and decimal digits.

Closes #486
Closes #342

N.B. I had to move the definition of the c-files variables up above the `c-files` target to make it build anything at all.

Timing diff coming soon


src/PushButtonSynthesis.v
```diff
@@ -1519,7 +1519,7 @@ Local Opaque reified_%s_gen. (* needed for making [autorewrite] not take a very
     Local Notation saturated_bounds := (saturated_bounds n machine_wordsize).
 
     Definition m_enc : list Z
-      := encode (UniformWeight.uweight machine_wordsize) n s c (s-Associational.eval c).
+      := encode (UniformWeight.uweight machine_wordsize) n s c m.
```
JasonGross: @andres-erbsen this is the one place where `c` is used outside of `Associational.eval`.  Should it be replaced by something else, like `partition `, so that we have `c : Z`?

andres-erbsen:
overall change lgtm. could even further inline c and I would expect it would wind up mostly unused. but not important so not blocking.

I'm feeling uneasy about checking in tens of thousands of lines of generated c code, but I guess that is already happening, so we might as well let this one slip...

PR #490, opened 2019-01-15
Don't cast signed to unsigned before shifting

### LINKED ISSUE #489: fiat_25519_subborrowx_u25 and fiat_25519_subborrowx_u26 aren't quite right
The synthesized `fiat_25519_subborrowx_u25` casts to `uint32_t` before right-shifting (see `x2`).
https://github.com/mit-plv/fiat-crypto/blob/8faf6852f5bb36f5c663386f7dfbd0ae258445f9/curve25519_32.c#L40-L46

This doesn't look right. (And it causes BoringSSL to give the wrong answer.) Shifting as `uint32_t` means it no longer sign extends as that code seems to want. The u26 version has the same issue. The u51 version in the 64-bit curve25519 also has this issue, but for some reason it's not causing the wrong answer in BoringSSL.

JasonGross:
Unfortunately, signed->unsigned casts do not commute with shifts.  We
take care to only extend the range when it needs extending, now.  This
was previously causing issues with subborrow.

We should really get proofs about casts in C semantics at some point
soon.

Fixes #489

@davidben Does this fix your issue?

### DIFF
--- a/p256_32.c
+++ b/p256_32.c
@@ -14,8 +14,6 @@
 #include <stdint.h>
 typedef unsigned char fiat_p256_uint1;
 typedef signed char fiat_p256_int1;
-typedef signed __int128 fiat_p256_int128;
-typedef unsigned __int128 fiat_p256_uint128;
 
 
 /*
@@ -46,7 +44,7 @@ static void fiat_p256_addcarryx_u32(uint32_t* out1, fiat_p256_uint1* out2, fiat_
  */
 static void fiat_p256_subborrowx_u32(uint32_t* out1, fiat_p256_uint1* out2, fiat_p256_uint1 arg1, uint32_t arg2, uint32_t arg3) {
   int64_t x1 = ((arg2 - (int64_t)arg1) - arg3);
-  fiat_p256_int1 x2 = (fiat_p256_int1)((fiat_p256_int128)x1 >> 32);
+  fiat_p256_int1 x2 = (fiat_p256_int1)(x1 >> 32);
   uint32_t x3 = (uint32_t)(x1 & UINT32_C(0xffffffff));
   *out1 = x3;
   *out2 = (fiat_p256_uint1)(0x0 - x2);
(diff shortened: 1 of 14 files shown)


p256_32.c
```diff
@@ -44,7 +46,7 @@ static void fiat_p256_addcarryx_u32(uint32_t* out1, fiat_p256_uint1* out2, fiat_
  */
 static void fiat_p256_subborrowx_u32(uint32_t* out1, fiat_p256_uint1* out2, fiat_p256_uint1 arg1, uint32_t arg2, uint32_t arg3) {
   int64_t x1 = ((arg2 - (int64_t)arg1) - arg3);
-  fiat_p256_int1 x2 = (fiat_p256_int1)((uint64_t)x1 >> 32);
+  fiat_p256_int1 x2 = (fiat_p256_int1)((fiat_p256_int128)x1 >> 32);
```
davidben: Now I have a different problem, which is that 32-bit targets usually doing have `__int128`.

JasonGross: Should be fixed now

p384_32.c
```diff
@@ -1,7565 +0,0 @@
-/* Autogenerated */
```
davidben: This file seems to have gotten swallowed.

src/CStringification.v
```diff
@@ -1117,11 +1124,14 @@ Module Compilers.
                               behavior to shift >= width of the type.
                               We should probably figure out how to not
                               generate these things in the first
-                              place... *)
-                       let rpre_out := int.of_zrange_relaxed r[0~>2^offset]%zrange in
+                              place...
+
+                              N.B. We must preserve signedness of the
+                              value being shifted, because shift does
+                              not commute with mod. *)
```
davidben: I think it does commute with left-shift. And indeed left-shifts are tricky because it's UB if you ever shift into the sign bit (if I recall, left-shifting a negative value is not defined), so I suspect we want to only ever left-shift unsigned types.

In contrast, right-shifts are relatively sane and the signedness matters for sign extension.

p384_32.c
JasonGross: Kind-of fixed now.  I think the binary is unhappy with this generation on my windows machine, and for some reason, in-flight wifi is now forbidding ssh, and I can't figure out how to get athena.dialup.mit.edu paste to work, so I just checked out the previous version of this file.

src/CStringification.v
JasonGross: Urgh, so the thing to do then is to take the unsigned counterpart of whatever type can fit the input?

JasonGross: Updated

JasonGross:
Yeah, we should get a spec of the semantics and a proof at some point.

PR #494, opened 2019-01-15
Add a rewrite rule to collapse constant casts

### LINKED ISSUE #493: MSVC warns on uint8_t >> 8.
When building fiat-generated code with MSVC, lines like the following trigger MSVC's warning C4333,
https://github.com/mit-plv/fiat-crypto/blob/master/p256_32.c#L3186
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4333?view=vs-2017

How hard would it be to add a pass to replace those lines with zero? It would probably trim the subsequent code too, if constant propagation and such is applied.

Aside: normally right-shifting an integer by all its bits is not only redundant but also UB because one cannot shift `(u)intN_t` by `N` bits, only up to `N-1`. However, integer promotion rules apply to shifts too, so this is merely redundant, not UB:

> The integer promotions are performed on each of the operands. The type of the result is
> that of the promoted left operand. If the value of the right operand is negative or is
> greater than or equal to the width of the promoted left operand, the behavior is undefined.

JasonGross:
If, e.g., we know from bounds analysis that the result of an operation
fits in the range r[0~>0], we now just replace it with the literal
constant.

Fixes #493


p224_32.c
```diff
@@ -2100,475 +2100,472 @@ static void fiat_p224_from_montgomery(uint32_t out1[7], const uint32_t arg1[7])
   fiat_p224_addcarryx_u32(&x38, &x39, x37, x26, x29);
   uint32_t x40;
   fiat_p224_uint1 x41;
-  fiat_p224_addcarryx_u32(&x40, &x41, x23, 0x0, 0x0);
+  fiat_p224_addcarryx_u32(&x40, &x41, 0x0, x32, x12);
```
davidben: Mostly out of curiosity (I haven't tried this code yet, so I have no opinions either way on which is correct), do you know what's going on with these changes? It seems more substantially different than the other ones.

curve25519_32.c
```diff
@@ -748,43 +748,42 @@ static void fiat_25519_to_bytes(uint8_t out1[32], const uint32_t arg1[10]) {
...
-  uint8_t x113 = (uint8_t)(x110 & UINT8_C(0xff));
-  uint32_t x114 = (x112 + x42);
-  uint32_t x115 = (x114 >> 8);
-  uint8_t x116 = (uint8_t)(x114 & UINT8_C(0xff));
-  uint32_t x117 = (x115 >> 8);
-  uint8_t x118 = (uint8_t)(x115 & UINT8_C(0xff));
-  uint8_t x119 = (uint8_t)(x117 >> 8);
-  uint8_t x120 = (uint8_t)(x117 & UINT8_C(0xff));
+  uint8_t x84 = (uint8_t)(x82 & UINT8_C(0xff));
+  uint32_t x85 = (0x0 + x32);
```
davidben: This is kinda silly, though one hopes the C compiler can figure it out.

JasonGross: Urgh, really?  Thanks for catching this. Let me see if I can do something about it

p224_32.c
JasonGross: It only seems more different because of all the renaming and the variable declarations that stayed the same.  I think previously only the carry bit from this line was used, and now we can see that the carry bit must be 0, and that gets propagated away.  So the `fiat_p224_addcarryx_u32(&x48, &x49, x47, x32, x12);` line below gets moved up to here and we end up knowing that `x47` is 0.

curve25519_32.c
JasonGross: This should be pretty easy, I just need to add the 0-propagation rules to the after-casts code.  I'll do it later tonight or tomorrow.

davidben:
Thanks! (New version still passes our CI.)

PR #495, opened 2019-01-15
[WIP] Add partial semantics for the C AST

JasonGross:
This is definitely not ready for merge (though I believe it does build)

I've left a bunch of `TODO`s for @davidben (or possible for me-in-the-future) to fill in the actual semantics of C.

I've run out of time at the moment to fill in the semantics of statements (rather than just single expressions), but I'll be doing that pretty soon.


JasonGross:
I've finished filling the skeleton for the interpretation.  There are many `TODO`s left about semantics of C.

PR #500, opened 2019-01-17
 Split up PushButtonSynthesis.v

### LINKED ISSUE #497: Split up PushButtonSynthesis.v
It's too big and slow.  At the very least, each top-level module should get its own file.  I'd like to split it up further, and give each top-level module two files: one for the reification / reification cache, and the other for the pipeline composition / lemmas.  This is because the reification is extremely slow, but also doesn't depend on, e.g., the rewriter, so having it built earlier will significantly speed up any further development that needs to happen with the rewriter, semantics of the C AST, etc.

JasonGross:
Closes #497 

@andres-erbsen @jadephilipoom Please comment on file organization.  Almost every change here is cut/paste/reindent.


andres-erbsen:
lgtm

it's definitely good to split these up. I'd like to eventually reduce the pushbuttonsynthesis parts to just boilerplate and have an interesting parts categorized under arithmetic, but that is a separate, larger refactoring. unsaturated solinas is almost there already.

PR #499, opened 2019-01-17
Don't allow r[0~>0] ranges in fancy synthesis

JasonGross:
This disables the optimization that was recently introduced for
replacing things known to be 0 from the bounds analyzer with the
constant 0.  It was messing up some of the other fancy synthesis stages.

Fixes #498

PR #501, opened 2019-01-17
Be stricter about rejecting command line arguments

### LINKED ISSUE #488: Better error resilience to bitwidth 1
See https://github.com/mit-plv/fiat-crypto/issues/486#issuecomment-453861606

It would be nice to give a nice error message and not stack overflow

JasonGross:
We no longer accept `1,1` as a valid bitwidth.

This closes #488.  It does not quite fix the issue as stated, but it
resolves the underlying confusion that generated it (the difference
between passing `1,1` as a `c` vs passing `1,1` as a bitwidth).  It
still might be nice to someday emit warnings as suggested in the bug
report.

I've also factored out a bit of code to make it easier to report parse
failures, by folding over a list of parses.

On top of #499, so that travis accepts the build.

### DIFF
--- a/src/CLI.v
+++ b/src/CLI.v
@@ -27,17 +27,24 @@ Module ForExtraction.
                  else (s, 1)
        | _ => (s, 1)
        end.
-  Definition parse_N (s : string) : N
-    := DecimalHelpers.N.of_uint (DecimalHelpers.String.to_uint s).
-  Definition parse_Z (s : string) : Z
-    := let '(s, sgn) := parse_neg s in
-       sgn * Z.of_N (parse_N s).
-  Definition parse_nat (s : string) : nat
-    := N.to_nat (parse_N s).
+  Definition parse_Z (s : string) : option Z
+    := z <- ParseArithmetic.parse_Z s;
+         match snd z with
+         | EmptyString => Some (fst z)
+         | _ => None
+         end.
+  Definition parse_N (s : string) : option N
+    := match parse_Z s with
+       | Some Z0 => Some N0
+       | Some (Zpos p) => Some (Npos p)
+       | _ => None
+       end.
+  Definition parse_nat (s : string) : option nat
+    := option_map N.to_nat (parse_N s).
   Definition parse_Q (s : string) : option Q
     := match List.map parse_Z (String.split "/" s) with
-       | [num;Zpos den] => Some (Qmake num den)
-       | [num] => Some (Qmake num 1)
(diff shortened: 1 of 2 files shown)


andres-erbsen:
I approve all changes before `Dyn`, and I do not understand how or why `Dyn` is used.

JasonGross:
The design considerations leading up to the reason I use `Dyn`:
1. I want the error message to display all command line arguments that failed to parse in the error message, rather than just the first one that failed to parse.  (Similarly, I have the code display error messages across all functions that fail to synthesize, not just the first one.)
2. When there were only two command line arguments that could fail to parse, I just wrote down the three possible ways they could fail to parse (one, the other, or both).  Now that there are four command line arguments that can fail to parse, and I don't want to hand-code 15 cases for error messages, I instead wrote a thing that takes a list of name * string input * option parse output, and returns either the successful parse outputs (as a tuple indexed over the input list), or else the error message from all cases that failed, combined.
3. Different parses land in different types, so I had to make (2) generic over the type of parse output in each element of the list.  The simplest way to do this was with `Dyn`.

src/CLI.v
```diff
@@ -150,20 +210,16 @@ Module ForExtraction.
...
-           => inr ["Could not parse s (" ++ s ++ ")"]
-         | _, None
-           => inr ["Could not parse c (" ++ c ++ ")"]
-         | Some s, Some c
+         match parse_many [("n", n, parse_n n:Dyn)
+                           ; ("machine_wordsize", machine_wordsize, parse_machine_wordsize machine_wordsize:Dyn)
+                           ; ("s", s, parse_s s:Dyn)
+                           ; ("c", c, parse_c c:Dyn)] with
+         | inr errs => inr errs
+         | inl (n, machine_wordsize, s, c)
```
andres-erbsen: What is the type of `c` bound here, and how is it determined? `parse_resultL` returns `Dyn`, but it seems to be used as something else?

JasonGross: It's type is ~ that of `invert_Some (parse_c c)`.  `parse_resultL` does not return `Dyn`; it takes in a `Dyn` and unpacks the type.  If the nth element of the input list is `(_, _, dyn T (Some v))`, then the nth element of the output tuple has type `T`.

JasonGross:
It's not too insanely dependently typed?  (I could always give up on reporting all parse errors at once, for the sake of code simplicity.)

JasonGross:
I will merge this now, since I'll be away for most of the weekend.  If it is too complicated, we can always revert and pick a simpler version.

PR #508, opened 2019-01-23
Add remove_duplicates

JasonGross:


andres-erbsen:
lgtm, I wish we had a process for sending this kind of stuff towards the standard library...

JasonGross:
Merging after "ok" from @andres-erbsen on signal

PR #509, opened 2019-01-23
Add better computation of carry chain

### LINKED ISSUE #507: Synthesis fails on p448
Quoting @0x64616E69656C over email:

> Hi Jason,
> 
> I'm not sure if I'm doing something wrong or if it is a bug, but I'm 
> trying to generate carry_mul / carry_square C-code for the "goldilocks" 
> prime 2^448 - 2^224 - 1 using unsaturated_solinas:
> 
> The param |c| states:
> 
> "The semi-colon separated list of taps, each of which is specified as 
> weight,value (no parentheses) (e.g., '2^96,1;1,-1' in '2^224 - 2^96 + 1')"
> 
> (The above sample does not work either, btw, but p224_64.c is generated 
> using word_by_word_montgomery anways)
> 
> So I tried:
> 
> ./src/ExtractionOCaml/unsaturated_solinas 'p448' '8' '2^448' 
> '2^224,1;1,1' '64' carry_mul
> 
> 
> which gives:
> 
> In fiat_p448_carry_mul:
> Computed bounds (Some [Some [0x0 ~> 0xffffffffffffff], Some [0x0 ~> 
> 0xffffffffffffff], Some [0x0 ~> 0xae3d70a3d70a3e1f0a3d70a3d70af], Some 
> [0x0 ~> 0xffffffffffffff], Some [0x0 ~> 
> 0xae3d70a3d70a3e0f0a3d70a3d70affd70a3d70a3dd728f5c28f5c28fd], Some [0x0 
> ~> 0xffffffffffffff], Some [0x0 ~> 0xffffffffffffff], Some [0x0 ~> 
> 0xffffffffffffff]]) are not tight enough (expected bounds not looser 
> than (Some [Some [0x0 ~> 0x119999999999999], Some [0x0 ~> 
> 0x119999999999999], Some [0x0 ~> 0x119999999999999], Some [0x0 ~> 
> 0x119999999999999], Some [0x0 ~> 0x119999999999999], Some [0x0 ~> 
> 0x119999999999999], Some [0x0 ~> 0x119999999999999], Some [0x0 ~> 
> 0x119999999999999]])).
> The bounds [0x0 ~> 0xae3d70a3d70a3e1f0a3d70a3d70af] are looser than the 
> expected bounds [0x0 ~> 0x119999999999999]
> The bounds [0x0 ~> 
> 0xae3d70a3d70a3e0f0a3d70a3d70affd70a3d70a3dd728f5c28f5c28fd] are looser 
> than the expected bounds [0x0 ~> 0x119999999999999]
> 
> I tried switching weight and value like it's documented for curve25519 
> but no luck.
> 
> All other synthesized functions (add, sub, from_bytes, to_bytes) work 
> just fine with X448/Ed448.
> 
> I'm currently using mul and square from 
> https://github.com/mit-plv/fiat-crypto/tree/00be3de14ee27a79f2b9f5d2dcb0a9e48491ad7b/src/Specific/solinas64_2e448m2e224m1_8limbs.
> 
> WBWM mul does not work, it generates code using 7 limbs with 
> machine_wordsize = 64 and non-working mul with 8 limbs using wordsize = 56.

Part of the issue is that we don't currently support goldilocks nor karatsuba optimizations.  The blocking issue is that automatic bounds analysis is insufficient for both of these multiplication algorithms, and the new pipeline does not support leaving over algebraic side conditions during bounds analysis.  We plan to move bounds analysis earlier in the pipeline, which should allow us to handle karatauba and goldilocks.

However, @andres-erbsen tells me that the unoptimized code should still work fine.  Perhaps it is an issue with the carry chain?  The curve parameters from the old pipeline, at https://github.com/mit-plv/fiat-crypto/blob/sp2019latest/src/Specific/solinas64_2e448m2e224m1_8limbs/CurveParameters.v, are
https://github.com/mit-plv/fiat-crypto/blob/c60b1d2556a72c37f4bc7444204e9ddc0791ce4f/src/Specific/solinas64_2e448m2e224m1_8limbs/CurveParameters.v#L11-L35
and in particular the carry chain there is `[[3; 7]; [4; 0; 5; 1; 6; 2; 7; 3]; [4; 0]]`, which is different from the current `seq 0 n ++ [0; 1]`.

This is generated by the code at https://github.com/mit-plv/fiat-crypto/blob/c60b1d2556a72c37f4bc7444204e9ddc0791ce4f/generate_parameters.py#L286-L297

Perhaps we should be doing this for unsaturated solinas at least?  (Also saturated solinas? @andres-erbsen ? @jadephilipoom ?)

JasonGross:
We port the computation of the carry chain from generate_parameters.py to Coq, for unsaturated solinas.

Note that while we now bounds-check p448, we do not yet support goldilocks nor karatsuba.

<strike>However, there is still an issue with the synthesized p448 code, which is that on 64-bit, it tries to use 256-bit and 512-bit integers.  I'm not sure what's up with that. (@jadephilipoom, do you know?)
Note that this is the cause of travis failing.</strike>

Closes #507


andres-erbsen:
@jadephilipoom can you take a look at this as you wrote the original code? I don't know whether this synthesis should get a nice result without karatsuba.

jadephilipoom:
I can check this out tomorrow in more detail, but here are my initial thoughts:

1. There's a right shift by 168 that definitely shouldn't be happening; I suspect something has been lost in translation here, or maybe there was a bug in the original code.
2. Usaturated solinas isn't a fantastic fit for p448 without karatsuba on 64-bit. The prime is 2^448 - 2^224 - 1, and 224 is not a multiple of 64.

Jason, did you make sure the "starts" given by the Python and Coq scripts match?

andres-erbsen:
448=56*8 so unsaturated solinas should work fine on 64-bit.

jadephilipoom:
Ah, good point Andres. I was thinking saturated for some reason.

JasonGross:
When I set up the context as
```coq
  Context (n : nat := 8%nat)
          (s : Z := 2^448)
          (c : list (Z * Z) := [(2^224,1);(1,1)])
          (machine_wordsize : Z := 64).
```
The result of `Open Scope nat_scope. Compute idxs.` is
`[3; 7; 4; 0; 5; 1; 6; 2; 7; 3; 4; 0]`.
The old carry chain is
`[[3; 7]; [4; 0; 5; 1; 6; 2; 7; 3]; [4; 0]]`.

Those look the same to me, the diff is
```diff
-[[3; 7]; [4; 0; 5; 1; 6; 2; 7; 3]; [4; 0]]
+[ 3; 7 ;  4; 0; 5; 1; 6; 2; 7; 3 ;  4; 0 ]
```

andres-erbsen:
@JasonGross I believe the list-of-lists notation meant that `reduce` gets called only at the boundaries of intermediate lists. I am not sure what the current convention is, but AFAIK it is necessary to call reduce at these points.

p448_solinas_64.c
```diff
@@ -0,0 +1,816 @@
...
+  fiat_p448_uint128 x20 = ((fiat_p448_uint128)(arg1[4]) * (arg2[6]));
+  fiat_p448_uint128 x21 = ((fiat_p448_uint128)(arg1[4]) * (arg2[5]));
+  fiat_p448_uint128 x22 = ((fiat_p448_uint128)(arg1[4]) * (arg2[4]));
+  fiat_p448_uint128 x23 = ((fiat_p448_uint128)(arg1[3]) * (arg2[7]));
+  fiat_p448_uint128 x24 = ((fiat_p448_uint128)(arg1[3]) * (arg2[6]));
+  fiat_p448_uint128 x25 = ((fiat_p448_uint128)(arg1[3]) * (arg2[5]));
+  fiat_p448_uint128 x26 = ((fiat_p448_uint128)(arg1[2]) * (arg2[7]));
+  fiat_p448_uint128 x27 = ((fiat_p448_uint128)(arg1[2]) * (arg2[6]));
+  fiat_p448_uint128 x28 = ((fiat_p448_uint128)(arg1[1]) * (arg2[7]));
+  fiat_p448_uint512 x29 = ((fiat_p448_uint512)((fiat_p448_uint128)(arg1[7]) * (arg2[7])) << 168);
```
andres-erbsen: this line is definitely wrong, we should never left-shift by that much (or at all?). I am guessing this is due to a missing reduce in the midddle of a carry chain.

JasonGross:
@andres-erbsen I don't think so.  This is the pipeline run on `carry_mulmod` which is
```coq
carry_mulmod = 
fun (limbwidth_num limbwidth_den s : Z) (c : list (Z * Z)) (n : nat) (idxs : list nat)
  (f g : list Z) =>
chained_carries (weight limbwidth_num limbwidth_den) n s c
  (mulmod (weight limbwidth_num limbwidth_den) s c n f g) idxs
     : Z -> Z -> Z -> list (Z * Z) -> nat -> list nat -> list Z -> list Z -> list Z
```
and `chained_carries` is
```coq
fun (weight : nat -> Z) (n : nat) (s : Z) (c : list (Z * Z)) (p : list Z) (idxs : list nat) =>
List.fold_right (fun (a : nat) (b : list Z) => carry_reduce weight n s c a b) p (List.rev idxs)
     : (nat -> Z) -> nat -> Z -> list (Z * Z) -> list Z -> list nat -> list Z
```
and `carry_reduce` is
```coq
carry_reduce = 
fun (weight : nat -> Z) (n : nat) (s : Z) (c : list (Z * Z)) (index : nat) (p : list Z) =>
from_associational weight n
  (reduce s c (to_associational weight (S n) (carry weight n (S n) index p)))
     : (nat -> Z) -> nat -> Z -> list (Z * Z) -> nat -> list Z -> list Z
```

So I think we're running reduce after every single carry.

JasonGross:
Issue tracked down!  For primes with more than one TAP, we need more than one `reduce` before converting back to positional in `mulmod`.

jadephilipoom:
lgtm!

PR #510, opened 2019-01-24
Make trivial instances explicit

maximedenes:
This is in preparation for coq/coq#9274.

Should backward compatible but that remains to be tested.


JasonGross:
Note that you may also need a separate PR for this change on the sp2019latest branch, which is where legacy-fiat-crypto lives.

LGTM if Travis okays it

PR #511, opened 2019-01-24
Make trivial instances explicit

maximedenes:
This is in preparation for coq/coq#9274.


maximedenes:
@JasonGross I think there was just a spurious failure, right?

JasonGross:
This is the ever-present problem of "Coq merged a commit into master in the middle of our build and we're not robust to that".  I've restarted the relevant jobs.

PR #513, opened 2019-01-31
Add autogenerated docstrings to synthesized code

### LINKED ISSUE #512: Docstrings on synthesized code.
This is just a "nice to have" thing, but if the generated comments had documentation strings like "fiat_p256_mul sets out1 to the product of arg1 and arg2 as P-256 field elements", so callers know what the function does, that would be handy.

JasonGross:
We now stringify the correctness conditions to generate docstrings for
the synthesized code.

@davidben Please review the generated docstrings and request any changes you want to see in parenthesization, formatting, notations, etc.  (Also please suggest docstring headers that are better than `Function .... does stuff.`.)

@andres-erbsen Please review the Gallina changes for whether or not you think they clutter the glue / pipeline too much, and the Ltac usage for whether or not the things are below the threshold of "too insane to include".

Closes #512

Time commitment: about 6 hours


curve25519_32.c
```diff
@@ -32,6 +35,9 @@ static void fiat_25519_addcarryx_u26(uint32_t* out1, fiat_25519_uint1* out2, fia
 }
 
 /*
+ * The function fiat_25519_subborrowx_u26 is a sub with borrow.
```
davidben: Nit: subtraction with borrow, maybe?

curve25519_32.c
```diff
@@ -98,6 +112,8 @@ static void fiat_25519_cmovznz_u32(uint32_t* out1, fiat_25519_uint1 arg1, uint32
 }
 
 /*
+ * The function fiat_25519_carry_mul does stuff.
```
davidben: multiplies two field elements and reduces the result

curve25519_32.c
```diff
@@ -265,6 +281,8 @@ static void fiat_25519_carry_mul(uint32_t out1[10], const uint32_t arg1[10], con
 }
 
 /*
+ * The function fiat_25519_carry_square does stuff.
```
davidben: squares a field element and reduces the result

curve25519_32.c
```diff
@@ -470,6 +490,8 @@ static void fiat_25519_carry_scmul_121666(uint32_t out1[10], const uint32_t arg1
 }
 
 /*
+ * The function fiat_25519_carry does stuff.
```
davidben: reduces a field element

curve25519_32.c
```diff
@@ -404,6 +422,8 @@ static void fiat_25519_carry_square(uint32_t out1[10], const uint32_t arg1[10])
 }
 
 /*
+ * The function fiat_25519_carry_scmul_121666 does stuff.
```
davidben: multiplies a field element by 121666 and reduces the result

curve25519_32.c
```diff
@@ -541,6 +565,8 @@ static void fiat_25519_add(uint32_t out1[10], const uint32_t arg1[10], const uin
 }
 
 /*
+ * The function fiat_25519_sub does stuff.
```
davidben: subtracts two field elements

curve25519_32.c
```diff
@@ -511,6 +533,8 @@ static void fiat_25519_carry(uint32_t out1[10], const uint32_t arg1[10]) {
 }
 
 /*
+ * The function fiat_25519_add does stuff.
```
davidben: adds two field elements

curve25519_32.c
```diff
@@ -571,6 +597,8 @@ static void fiat_25519_sub(uint32_t out1[10], const uint32_t arg1[10], const uin
 }
 
 /*
+ * The function fiat_25519_opp does stuff.
```
davidben: negates a field element

curve25519_32.c
```diff
@@ -641,6 +671,8 @@ static void fiat_25519_selectznz(uint32_t out1[10], fiat_25519_uint1 arg1, const
 }
 
 /*
+ * The function fiat_25519_to_bytes does stuff.
```
davidben: serializes a field element to bytes in little-endian order

curve25519_32.c
```diff
@@ -821,6 +853,8 @@ static void fiat_25519_to_bytes(uint8_t out1[32], const uint32_t arg1[10]) {
 }
 
 /*
+ * The function fiat_25519_from_bytes does stuff.
```
davidben: deserializes a field element from bytes in little-endian order

curve25519_32.c
```diff
@@ -15,6 +15,9 @@ typedef signed char fiat_25519_int1;
 
 
 /*
+ * The function fiat_25519_addcarryx_u26 is an add with carry.
+ * out1 = (arg1 + arg2 + arg3) mod 2^26
+ * ∧ out2 = ⌊(arg1 + arg2 + arg3) / 2^26⌋
```
davidben: Nit: The ∧ is mildly odd to see in a comment like this. (But if it's hard to remove, that's fine. I'm assuming you're just doing things with Coq notations or something?)

Edit: Ah yeah, I see from later files this is actually a whole correctness statement. Not sure how removing the ∧ meshes with the mess that is the Montgomery form statements. (Then again, the Montgomery form state is... a little difficult to read. :-) )

p224_32.c
```diff
@@ -82,6 +93,12 @@ static void fiat_p224_cmovznz_u32(uint32_t* out1, fiat_p224_uint1 arg1, uint32_t
 }
 
 /*
+ * The function fiat_p224_mul does stuff.
```
davidben: multiplies two field elements in Montgomery form

p224_32.c
```diff
@@ -82,6 +93,12 @@ static void fiat_p224_cmovznz_u32(uint32_t* out1, fiat_p224_uint1 arg1, uint32_t
...
+ * arg1 = map (λ x, ⌊eval arg1 mod 2^(32 * (x + 1)) / 2^(32 * x)⌋) [0..6] ∧ 0 ≤ eval arg1 < m →
+ * arg2 = map (λ x, ⌊eval arg2 mod 2^(32 * (x + 1)) / 2^(32 * x)⌋) [0..6] ∧ 0 ≤ eval arg2 < m →
+ * eval (fiat_p224_from_montgomery out1) mod m = (eval (fiat_p224_from_montgomery arg1) * eval (fiat_p224_from_montgomery arg2)) mod m
+ * ∧ out1 = map (λ x, ⌊eval out1 mod 2^(32 * (x + 1)) / 2^(32 * x)⌋) [0..6]
+ * ∧ 0 ≤ eval out1 < m
```
davidben: This is a bit of a mouthful. :-)

What's going on with the `map` lambda bits?

p224_32.c
```diff
@@ -2034,6 +2073,11 @@ static void fiat_p224_opp(uint32_t out1[7], const uint32_t arg1[7]) {
 }
 
 /*
+ * The function fiat_p224_from_montgomery does stuff.
+ * arg1 = map (λ x, ⌊eval arg1 mod 2^(32 * (x + 1)) / 2^(32 * x)⌋) [0..6] ∧ 0 ≤ eval arg1 < m →
+ * eval out1 mod m = (eval arg1 * 26959946660873538059280334323183841250350249843942399443119741337601^7) mod m
```
davidben: I'm actually a little confused what's going on here. I see 26959946660873538059280334323183841250350249843942399443119741337601 is 0xfffffffeffffffffffffffffffffffff000000010000000000000001 which I assume is 2^256 - m or  something. Where is the `^7` coming from?

(Expressing it in terms of 2^256 would be a bit more useful, if that's possible.)

p224_32.c
```diff
@@ -2034,6 +2073,11 @@ static void fiat_p224_opp(uint32_t out1[7], const uint32_t arg1[7]) {
 }
 
 /*
+ * The function fiat_p224_from_montgomery does stuff.
```
davidben: translates a field element out of the Montgomery domain?

I dunno, I tried to say "converts a field element from Montgomery to ???" but wasn't sure how to say the ???. I see AGL used a wording like this for our `BN_from_montgomery` so I copied that. But I guess that means the above suggestions should say "in the Montgomery domain" rather than "in Montgomery form"?

p224_32.c
```diff
@@ -2569,6 +2613,9 @@ static void fiat_p224_from_montgomery(uint32_t out1[7], const uint32_t arg1[7])
 }
 
 /*
+ * The function fiat_p224_nonzero does stuff.
```
davidben: outputs a single non-zero word if the input is non-zero and zero otherwise

Aside: I noticed when integrating the P-256 code that `fiat_*_nonzero` and `fiat_*_select` don't play together well. I needed to add a `!!` to bridge the two.

p224_32.c
JasonGross: The base is `r'`, which is defined as
```coq
  Let r := 2^machine_wordsize.
  Let r' := match Z.modinv r m with
            | Some r' => r'
            | None => 0
            end.
```
i.e., the modular inverse of `2^bitwidth `, modulo the prime.  Raising it to the 7 is to get the modular inverse of 2^(width of the entire field element).  It's computed this way for ease of proofs, iirc.  How do you want this rendered?

p224_32.c
JasonGross: > don't play together well

Do you want me to fix nonzero to always do `!!`?  Augment `selectznz` to start with `!!`?  (I think if I start `selectznz` with `!!`, gcc might compile it to a conditional jump + mov...)

p224_32.c
JasonGross: It is possible that we want to drop the `map` λ bits, because I think they are redundant with the bounds in the saturated case.   These bits say that the arguments must be in fully reduced form, and the output is guaranteed to be in fully reduced form.  Where the reduced form / canonical representation is defined by saying that digit `i` gets to represent as much of the total value (mod (weight (i+1)) as it can.

p224_32.c
JasonGross: Updated

p224_32.c
JasonGross: It now renders as `eval out1 mod m = (eval arg1 * ((2^32)⁻¹ mod m)^7) mod m`  Is this want you want, or do you want me to make it be `eval out1 mod m = (eval arg1 * ((2^224)⁻¹ mod m)) mod m`, or something else?

p224_32.c
```diff
@@ -51,6 +57,9 @@ static void fiat_p224_subborrowx_u32(uint32_t* out1, fiat_p224_uint1* out2, fiat
 }
 
 /*
+ * The function fiat_p224_mulx_u32 is an extended multiplication.
```
JasonGross: @davidben Are you okay with this docstring, or do you have a suggestion for a better one?

p224_32.c
JasonGross: I've dropped the `map λ` bits and dropped the `fiat_p224` prefix from `from_montgomery` in most of the docstrings.

Also, do you think it's more readable to have `x mod m = y mod m`, or to write `x ≡ y  (mod m)`?

p224_32.c
JasonGross: Oh, it looks like `selectznz` goes through `cmovznz` which starts with `!!`.  So the only issue is the types / bounds.  Do you want me to extend `cmovznz` and `selectznz` so that they accept larger arguments for the condition?  Or stick a `!!` at the end of `nonzero` so that it fits in smaller bounds?

andres-erbsen:
I am not a fan of the Gallina boilerplate. Could we talk over it for like 5 minutes to see if we can do better next time we are in the office at the same time?

curve25519_32.c
JasonGross: Done.  I arbitrarily decided to keep the plural form even when there is only one element in the list of pre/post conditions.  Let me know if you want other formatting for when the list is singleton

PR #533, opened 2019-03-12
Generalize barrett

jadephilipoom:
Same old code is generated for the fancymachine, but using proofs that aren't specialized to limb size. I'll merge this once the tests pass.

### DIFF
--- a/src/Arithmetic.v
+++ b/src/Arithmetic.v
@@ -1889,6 +1889,14 @@ Module Partition.
       reflexivity.
     Qed.
 
+    Lemma partition_0 n : partition n 0 = Positional.zeros n.
+    Proof.
+      cbv [partition].
+      erewrite Positional.zeros_ext_map with (p:=seq 0 n) by distr_length.
+      apply map_ext; intros.
+      autorewrite with zsimplify; reflexivity.
+    Qed.
+
   End Partition.
   Hint Rewrite length_partition length_recursive_partition : distr_length.
   Hint Rewrite eval_partition using (solve [auto; distr_length]) : push_eval.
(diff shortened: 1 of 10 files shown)


src/Arithmetic.v
```diff
@@ -4820,497 +4927,578 @@ Module WordByWordMontgomery.
...
-             SuchThat (barrett_reduce = x mod M)
-             As barrett_reduce_correct.
-      Proof.
-        erewrite <-reduce_correct with (rep:=represents) (muSelect:=muSelect) (k0:=k) (mut:=[muLow;1]) (xt0:=xt)
-          by (auto using x_bounds, muSelect_correct, x_rep, mu_rep; omega).
-        subst barrett_reduce. reflexivity.
-      Qed.
-    End Defn.
-  End BarrettReduction.
+        Axiom admit : forall {T}, T.
```
JasonGross: Is this cruft?

src/Arithmetic.v
```diff
@@ -4820,497 +4927,578 @@ Module WordByWordMontgomery.
...
+                   fancy_reduce xLow xHigh = (xLow + 2^k * xHigh) mod M)
+               As fancy_reduce_correct.
+        Proof.
+          intros. pose proof w_eq_22k. assert (k = width) as width_eq_k by nia.
+          assert (0 < 2 ^ (width - 1)) by Z.zero_bounds.
+          pose proof (Z.mod_pos_bound (xLow + 2^width * xHigh) M ltac:(lia)).
+          transitivity (hd 0 (partition w sz ((xLow + 2^k * xHigh) mod M)));
+            [ | rewrite sz_eq_1; rewrite width_eq_k in *; cbv [Partition.partition map seq hd];
+                rewrite !UniformWeight.uweight_S, !weight_0 by auto with zarith lia;
+                autorewrite with zsimplify; reflexivity ].
```
JasonGross: Coq <= 8.7 is anomalying on this line (rather on the `transitivity` line chained with this one), presumably because something here is using `abstract` and `Derive` in Coq <= 8.7 does not support `abstract`.

jadephilipoom: I've just rewritten this proof to remove `Derive` entirely, since it's not actually that much more work and it's annoying to debug a coq version I don't use. Hopefully the latest build fixes the issue.

jadephilipoom:
Superceded by #537

PR #540, opened 2019-04-03
Split Arithmetic.v

### LINKED ISSUE #485: split up Arithmetic.v

jadephilipoom:
Fixes #485. Mostly following existing section structures, I have replaced the old monolithic `Arithmetic.v` with 11 new files under `src/Arithmetic/` (shown here in approximately their original order):

1. `Core` - Associational/Positional definitions and proofs and required properties of weight functions
2. `ModOps` - definition of a weight function given the numerator/denominator of a limb width, as well as core modular arithmetic using `Positional` operations
3. `Saturated` - specialized operations that will only work on saturated limbs, row-wise addition/subtraction, and column-wise addition/subtraction
4. `Partition` - definitions and proofs relevant to splitting a number across limbs in saturated form
5. `BaseConversion` - transforming a number from one representation (limb-weight function) to another, using saturated or unsaturated weights
6. `Freeze` - full modular reduction and converting limbs to/from bytes
7. `Primitives` - definitions/proofs of `mulx`, `addcarryx`, `subborrowx`, and `cmovnz`
8. `UniformWeight` - facts about representations with equal-size limbs
9. `WordByWordMontgomery` - non-Fancy wbw Montgomery modular reduction code
10. `BarrettReduction` - fancy machine Barrett modular reduction
11. `FancyMontgomeryReduction` - fancy machine Montgomery modular reduction

Note: In the Makefile, the `Arithmetic.v` file was hardcoded as a target for `make some-early` to make Travis timing work out. I've replaced it with `Arithmetic/Core.v`. Jason said (in an offline conversation) that he'd prefer I add all the new files so that the timing juggling is stable, but the timing on Travis still seems to work and I'd really rather not have to go through the "I'm getting mysterious build failures and need to figure out where this file name is hardcoded" dance every time anyone removes or changes one of these file names. So, barring vehement objection, I'm leaving it out.


src/Arithmetic/Primitives.v
```diff
@@ -0,0 +1,119 @@
+
+(* TODO: prune these *)
```
JasonGross: I'm amused that the requires are almost all long as the contents.  You can prune them automatically with https://github.com/JasonGross/coq-tools/blob/master/minimize-requires.py

src/PushButtonSynthesis/WordByWordMontgomery.v
```diff
@@ -653,15 +657,15 @@ Section __.
...
-        Arithmetic.WordByWordMontgomery.onemod
-        Arithmetic.WordByWordMontgomery.from_montgomerymod
-        Arithmetic.WordByWordMontgomery.mulmod
-        Arithmetic.WordByWordMontgomery.squaremod
-        Arithmetic.WordByWordMontgomery.encodemod
-        Arithmetic.WordByWordMontgomery.addmod
-        Arithmetic.WordByWordMontgomery.submod
-        Arithmetic.WordByWordMontgomery.oppmod
-        Arithmetic.WordByWordMontgomery.to_bytesmod.
+        WordByWordMontgomery.WordByWordMontgomery.onemod
```
JasonGross: Do you need to double-qualify here?  I think it would be nicer too import `WordByWordMontgomery` and only single qualify.

src/PushButtonSynthesis/FancyMontgomeryReductionReificationCache.v
```diff
@@ -1,11 +1,13 @@
-(** * Push-Button Synthesis of Saturated Solinas: Reification Cache *)
+(** * Push-Button Snthesis of fancy mongomery reduction : Reification Cache *)
```
JasonGross: Typo: missing "y" in snthesis

src/PushButtonSynthesis/WordByWordMontgomery.v
jadephilipoom: There are multiple modules called `WordByWordMontgomery` in different places, so I'm double-qualifying to make it clear where this comes from.

src/PushButtonSynthesis/WordByWordMontgomery.v
```diff
@@ -98,9 +103,9 @@ Section __.
             end.
   Let n_bytes := bytes_n machine_wordsize 1 n.
   Let prime_upperbound_list : list Z
-    := Partition.partition (UniformWeight.uweight machine_wordsize) n (s-1).
+    := partition (uweight machine_wordsize) n (s-1).
   Let prime_bytes_upperbound_list : list Z
-    := Partition.partition (weight 8 1) n_bytes (s-1).
+    := partition (weight 8 1) n_bytes (s-1).
```
JasonGross: Please don't.  I don't want this becoming `List.partition` if I change the order of imports

src/Arithmetic/Primitives.v
jadephilipoom: Thanks for the tool, way better than my previous strategy :)

src/PushButtonSynthesis/WordByWordMontgomery.v
jadephilipoom: I put `partition` in its own module all by itself, so now it's forced to be qualified.

PR #542, opened 2019-04-06
Automate even more of the rewriter

JasonGross:
On top of #541


README.md
```diff
@@ -194,6 +195,94 @@ The files contain:
...
+  taking in wf-goodness of rewrite rules as a hypothesis.
+
+- RewriterInterpProofs1.v: Proves interp-correctness of the generic
+  rewriter, taking in interp-goodness of rewrite rules as a
+  hypothesis.
+
+- RewriterProofsTactics.v: Defines the tactic
+  `RewriteRules.Tactic.make_rewriter` (and a similar tactic notation)
+  which build the entire `VerifiedRewriter`.  They take in the
+  `include_interp` as in Rewriter.v tactics, as well as an hlist of
```
andres-erbsen: could this list of proofs just be `fold_right conj . map fst`?

README.md
```diff
@@ -194,6 +195,94 @@ The files contain:
...
+  rewrite rule types.  This is the primary interface for defining a
+  rewriter from a list of rewrite rules.
+
+- Rewriter/{NBE, Arith, ArithWithCasts, StripLiteralCasts, ToFancy,
+  ToFancyWithCasts}.v: Use the tactic from RewriterProofsTactics.v
+  together with the proven list of rewrite rules from
+  RewriterRulesProofs.v to reify and reduce the corresponding pass and
+  generate a rewriter.
+
+- RewriterProofs.v: `Definition`less file that `Export`s the rewriters
```
andres-erbsen: if it exports stuff other than proofs, perhaps `RewriterAll.v`?

README.md
JasonGross: Do you mean `fold_right and True . map snd`?  (That is effectively what `hlist` does, except it uses `prod`.  Is there a reason to prefer the `fold` version over `hlist`?)

README.md
JasonGross: Done

README.md
andres-erbsen: Perhaps it doesn't matter indeed, just seeing "hlist" made me think of term encodings...

PR #546, opened 2019-04-26
Remove bbv dependency

bbarenblat:
Remove bbv and everything that depends on it. Make a pass over Notations.v to address some TODOs affected by the removal.


src/Util/Notations.v
```diff
@@ -171,15 +171,10 @@ Reserved Notation "'return' x" (at level 70, format "'return'  x").
 Reserved Notation "f x" (only printing, at level 10, left associativity).
 (* TODO: Remove this next one if and when we drop support for Coq 8.7; it's present in the stdlib in 8.8 *)
 Reserved Notation "{ x  &  P }" (at level 0, x at level 99).
-(* TODO: Uncomment this block when https://github.com/mit-plv/bbv/pull/13 is merged
 Reserved Notation "$ x" (at level 9, x at level 9, format "$ x").
-Reserved Notation "# x" (at level 55, x at level 9, format "# x"). (* 55, so we don't clash with [_ # _] *)
- *)
+Reserved Notation "# x" (at level 9, x at level 9, format "# x").
```
JasonGross: Why `at level 9` rather than `at level 55`?

bbarenblat: For consistency with `## x`, `### x`, etc., all of which are at level 9. (This required one extra set of parentheses, in src/Rewriter.v.)

andres-erbsen:
Thanks! I'm sorry to slipped me for so long.

src/Util/Notations.v
JasonGross: I am not sure that consistency is worth making first-crypto incompatible with the notations for `Q` in the standard library.  I may decide to change this to 55.

andres-erbsen: sounds good, feel free to just do that

PR #551, opened 2019-05-29
Add CPS'd versions of arithmetic operations

JasonGross:
This is primarily for perf testing purposes, so we can evaluate the time
spent in the rewriter vs cbv/vm/native/etc for the cps'd versions.

We evaluate vm_compute and native_compute by parameterizing everything
over a module of runtime definitions, which can be instantiated with
module-locked opaque constants, which are not unfolded by the vm.


andres-erbsen:
uh I totally don't have the time to review this one. if it's just for comparison and Ray targets don't depends on it, I don't think I need to either

JasonGross:
It's just for perf testing and the C targets do not and will not depend on it.  Mostly I just wanted your OK for including it at all.  (And possibly any comments you have on me leaving commented out reduction debugging at the bottom of ModOps and WordByWordMontgomery.)

PR #552, opened 2019-06-02
In solinas binaries, parse the taps from the prime

JasonGross:
Thus we no longer give the list of taps in a special format, and can now
invoke it as
```bash
./src/ExtractionOCaml/unsaturated_solinas 25519 5 '2^255 - 19' 64
```


JasonGross:
@andres-erbsen Do the changes to the help strings and the changes to the comments in the .c files look okay?  (Feel free to also review the rest, but I'm especially interested in your review on those bits.)

andres-erbsen:
would it be possible to print `2^logs` instead of hex `s`? or 2^(64*4) even if thats easier.

perhaps "the prime, which must be expressed as a difference of a power of two and a small field element" followed by a 2^255-19 and goldilocks as examples?

JasonGross:
`2^logs` should be quite easy, I'll update it soon.  Do you mean `2^(64*4-1)`?  Is that preferred over `2^logs`? If so, what's the algorithm for generating it?

JasonGross:
Also, do you have preferences for how to display `m` in montgomery .c files?

JasonGross:
And do you want the weights in `c` to show up as powers of 2, also?

JasonGross:
Code updated

PR #553, opened 2019-06-03
Add more heuristic definitions for unsat solinas

JasonGross:
We now allow `(auto)` and `(auto0)` and `(auto1)`, etc, to be used to automatically guess the number of limbs for unsaturated solinas.

Inspired by
https://github.com/mit-plv/fiat-crypto/blob/c60b1d2556a72c37f4bc7444204e9ddc0791ce4f/generate_parameters.py

<strike>I have copied the `< 2 * bitwidth` from
https://github.com/mit-plv/fiat-crypto/blob/c60b1d2556a72c37f4bc7444204e9ddc0791ce4f/generate_parameters.py#L233,
but I have no idea why we check against `2 * bitwidth` rather than
against `bitwidth`. (@jadephilipoom, can you explain?)</strike>
Edit: It's because we support double-width operations.

I have added these in large part so that I can get scaffolding for benching.


JasonGross:
Also, @jadephilipoom (or @andres-erbsen), can you sanity-check my definition of `overflow_free`?  It is rather different from the version in the python script, but I think it does the same thing.

JasonGross:
Coq master is broken; see https://github.com/coq/coq/issues/10352

Merging based on signal conversation with Jade: "the spec of that routine is no false negatives, and enough true negatives that we're not wasting time on things that are obviously going to overflow. As long as it still fits the spec, it's okay."

PR #557, opened 2019-07-10
Add scripts for testing rewriting performance

JasonGross:
These are the test scripts used for `make perf` and `make perf-csv` used
to generate the data for the 2020 POPL rewriting paper submission.

@andres-erbsen Any objection to adding these to master?


JasonGross:
Got ack from Andres on signal, so merging

PR #558, opened 2019-07-10
Add a package-rewriter.sh script

JasonGross:
For the rewriter paper

Any objection to having this in master, @andres-erbsen ?


JasonGross:
Got ack from Andres on signal, so merging

PR #561, opened 2019-07-16
Refactor the CLI to make modifications easier

JasonGross:
There is now a Class which defines what the pipeline-specific parts of
the CLI are (pipeline-specific argument parsing, usage/help string,
display of arguments, synthesize, check_args), so that we can define
these separately for each pipeline, but not have to copy most of the
boilerplate.

It's plausible that argument parsing, help strings, etc, should be
factored through Crypto.Util.Arg, which implements roughly the API for
OCaml's Arg module for command line parsing.  This seems like a very
heavy hammer to use, though, so for now I'm leaving the argument parsing
as specialized (and therefore slightly duplicated).

This will hopefully make it a bit easier to only have to add
output-language-command-line logic in one place.

Alternative to #559

Closes #559

@zoep Does this refactoring give you everything you're wanting from #559, or did I miss some opportunities for de-duplication that are important to you?


zoep:
@JasonGross, 

Thank you, that takes care of the need to make the same edits x3. I'll be adding some comments that will hopefully provide some compiler design feedback for future iterations (and ofc need not addressed by this PR). 

Cheers,

src/CLI.v
```diff
@@ -193,300 +198,291 @@ Module ForExtraction.
...
+      ArgsT : Type;
+
+      (** Takes in argv except without the binary name and without any
+          general options (such as language), and parses the
+          curve-specific arguments, returning either [Some (inl
+          value)], [Some (inr errors)], or [None] if there are not
+          enough arguments and the usage string should be displayed. *)
+      parse_args : list string -> option (ArgsT + list string);
+
+      (** Renders a header at the top displaying the command line
```
zoep: Why is comment wrapping something that needs to be done by the CLI? Does this need to be exposed?

src/CLI.v
```diff
@@ -193,300 +198,291 @@ Module ForExtraction.
...
+      show_lines_args : ArgsT -> list string;
+
+      (** The part of the usage-string for arguments after curve_description *)
+      pipeline_usage_string : string;
+      (** The list of help strings (to be joined by newlines) *)
+      help_lines : list string;
+
+      (** The Synthesize function from the pipeline *)
+      (* TODO: make handling of commenting more uniform *)
+      Synthesize : forall (args : ArgsT) (function_name_prefix : string),
```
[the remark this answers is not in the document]
JasonGross: I agree.  I'm not sure yet what the structure of the IR should be, though, so I'm tempted to hold off on this change for now.

src/CLI.v
```diff
@@ -193,300 +198,291 @@ Module ForExtraction.
...
+          list string (* extra comment header (already wrapped in a comment block) *)
+          * list (string * Pipeline.ErrorT (list string)) (* result of synthesis *)
+          * MSetPositive.PositiveSet.t (* types used *);
+
+      (** The check_args function makes sure the arguments satisfy the preconditions *)
+      check_args : forall (args : ArgsT) {T} (res : Pipeline.ErrorT T),
```
zoep: AFAICT, this is exposed only because `show_line_args` is exposed, and could also be something internal to the synthesis code.

src/CLI.v
```diff
@@ -193,300 +198,291 @@ Module ForExtraction.
...
+      (** Takes in argv except without the binary name and without any
+          general options (such as language), and parses the
+          curve-specific arguments, returning either [Some (inl
+          value)], [Some (inr errors)], or [None] if there are not
+          enough arguments and the usage string should be displayed. *)
+      parse_args : list string -> option (ArgsT + list string);
+
+      (** Renders a header at the top displaying the command line
+          arguments.  Will be wrapped in a comment block *)
+      show_lines_args : ArgsT -> list string;
```
zoep: Why is comment wrapping something that needs to be done by the CLI? Does this need to be exposed?

JasonGross: Comment-wrapping is language specific (C uses `/* ... */`, Python uses `#`, OCaml uses `(* ... *)`, Haskell uses `--`, Agda uses `{- ... -}`) but pipeline independent.  Rendering the display of the command-line arguments is language inspecific but pipeline dependent.  Either the CLI needs to tell the function the language choice just so it can do comment wrapping, or the CLI needs to do the comment-wrapping.  I figured the latter was a better factoring.

src/CLI.v
JasonGross: This is only sort-of true.

I exposed `show_line_args` because that is a thing that is specific to each pipeline *and also* specific to parsing arguments from the command line / from a list of strings; it uses the string that was passed on the command line as part of the rendering.  It is in fact the only part that uses the string version of things passed on the command-line, and `Synthesize` throws away this information.  Should I make this more explicit by splitting up `ArgsT` into `ParsedArgsT` and `StringArgsT`?  (Then `show_line_args` would take both, and everything else would take only `ParsedArgsT`.)

When `check_args` fails, I think it's important to display the way the command-line arguments were parsed (unlike most of the other failures where I think including the header is less essential).  This means that we'd need to pass the header into `Synthesize` if we wanted to fold `check_args` into `Synthesize`.  We could do that, and I suspect it will make sense to do that in the world where we return an IR of the entire document before turning it into a specific language.

JasonGross: I've added a commit (8df01cf0c9c7b26ab9dd2243f971dabfaaf3d5f0) to #562 so that `check_args` is no longer exposed.  I've also split `ArgsT` into two types.

src/CLI.v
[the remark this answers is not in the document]
zoep: Sure, that makes sense.

src/CLI.v
zoep: With your new PR however, 

> Either the CLI needs to tell the function the language choice
Yes but that's needed anyway. CLI will need to tell the synthesis function the language, since the backend is called from the frontend and not the CLI. Also, in your new PR the frontend is parameterized by the` OutputLanguageAPI` so it has access to the commenting style.

zoep: i.e. that would need to discriminations on the backend language (1. to pass the right API parameter to the frontend and 2. to chose the correct comment style in the CLI).

Also in the ideal pipeline, where there is a clear separation between frontend-backend, generating comments would be internal to the frontend and that would be wrapped in some language-independent comment IR node.

JasonGross: `Synthesize` knows the language choice but not that its arguments were parsed from strings (nor what strings they were parsed from), and `show_lines_args` knows about parsed-from-strings but doesn't know about the language choice (and I think that's the correct decision about `show_lines_args`).  In #562, I've made it the job of `Synthesize` to add a comment block around the `list string` it gets from `show_lines_args`.

The data-flow is "CLI parses `curve_description` and the language choice, then asks pipeline-CLI to parse remaining command line arguments, then asks pipeline-CLI to describe it's parsing choices, then asks `Synthesize` to output code or errors for all the functions based on language choice and parsed arguments and to please include the text describing the parsing choices in a comment, and then the CLI either displays all the errors or concatenates all of the code and displays it"

JasonGross: > Also in the ideal pipeline, where there is a clear separation between frontend-backend, generating comments would be internal to the frontend and that would be wrapped in some language-independent comment IR node.

Yes.  In this world, the frontend would still need to tell the backend "here is some text in a language-independent comment IR node that I want you insert into the document header" and then separately would need to turn the IR into a language-specific thing.  `show_lines_args` is about the first part.

zoep: As I see it, the only reason `show_lines_args` needs to be exposed is because the following construction takes place in the CLI:
```
           let header :=
               (comment_block
                  ((["Autogenerated";
                       "curve description: " ++ curve_description]%string)
                     ++ show_lines_args args)
                  ++ [""]
                  ++ extra_comment_header
                  ++ ToString.C.String.typedef_header prefix types_used
                  ++ [""])%list in
```
And what I'm questioning here is why this header needs to be constructed at the CLI.v file (which ideally should know nothing about the assembly of the final document). The only reason I see is because `curve_desciption` is not known to the synthesis function, but I don't think that's fundamentally necessary. Arguably, it's preferable to pass `curve_desciption` than the `function_name_prefix ` (the only reason that the latter is constructed in the CLI is because the synthesis function don't know the `curve_desciption`).

zoep: > Yes. In this world, the frontend would still need to tell the backend "here is some text in a language-independent comment IR node that I want you insert into the document header"

Not sure I fully understand that. What I meant is that the header comment will be in the IR representation of the full document as the comment-header of the document. So the frontend will not explicitly need to tell the backend "here is some text in a language-independent comment IR node that I want you insert into the document header", but "here's my document traverse it's structure and generate some other representation of if". Maybe we mean the same thing though.

JasonGross: > What I meant is that the header comment will be in the IR representation of the full document as the comment-header of the document.

Right.  The question is, "who is responsible for generating the text of the header comment, and who is responsible for inserting the header comment into the IR?"

It's possible I'm mixing up "front-end" and "backend", so instead I'm going to use "CLI" for the code in CLI.v and "Pipeline" for the code in PushButtonSynthesis/* (i.e., the Synthesize definition).  Which is which under your terminology?  (Or are we confused because there are actually three parts?  The part that does I/O and command line parsing and error printing (main?), the part that generates the IR (front-end?), and the part that turns the IR into C code (backend?))

The Pipeline is not capable of generating the header comment because the Pipeline never sees the unparsed command line arguments.  Only the main function / CLI sees command line arguments.  That is, the p256_32.c contains
```c

/* Autogenerated */
/* curve description: p256 */
/* requested operations: (all) */
/* m = 0xffffffff00000001000000000000000000000000ffffffffffffffffffffffff (from "2^256 - 2^224 + 2^192 + 2^96 - 1") */
/* machine_wordsize = 32 (from "32") */
```
but the Pipeline never sees the string `2^256 - 2^224 + 2^192 + 2^96 - 1` (it only sees the `Z` that string parses to) so it's incapable of generating the text of this comment.

JasonGross: Also, I want to note, the CLI no longer does comment wrapping.

JasonGross:
Closed by #562

src/CLI.v
zoep: > It's possible I'm mixing up "front-end" and "backend", so instead I'm going to use "CLI" for the code in CLI.v and "Pipeline" for the code in PushButtonSynthesis/* (i.e., the Synthesize definition). Which is which under your terminology?

The way I'm using them: 
frontend: The synthesis functionality up until HOAS
backend: Everything after PHOAS (including the PHOAS-> IR translation) that's common for all synthesis modes.
CLI: Command line interface, i.e. argument parsing and dispatching of calls. 

> but the Pipeline never sees the string 2^256 - 2^224 + 2^192 + 2^96 - 1 (it only sees the Z that string parses to)

`ArgsT` carry the string representation too, no? If `show_lines_args` can generate the header based on `argsT` so can `Sythesize`.

> Right. The question is, "who is responsible for generating the text of the header comment, and who is responsible for inserting the header comment into the IR?"

What I mean is that `Synthesize` should be responsible for generating the comment text and attaching is to the header comment that already returns (I don't think it's a good idea to generate the header in two different places in the code).  `Synthesize` has access to both the `argsT` and their string representation (which is actually part of `argsT`) so  I don't see the problem here. The only extra bit of information that it need is the `curve_description` that can be easily passed as an argument (instead of the function header, that can be constructed later on using the `curve_description`).

JasonGross: I've gone the other way: the string versions of the arguments are not used for anything in Synthesize, so I now no longer give Synthesize access to them, and instead just pass Synthesize a `list string` which is a summary of all the arguments.

zoep: Ah, that's a commit that's not part of this PR. I still think than it makes more sense for synthesize to generate the text of the comment, and that's less heavy to pass string representations of args than the header itself, but of course this design is up to you and doesn't interfere with our Rust stuff.

PR #562, opened 2019-07-17
Add support for --lang=C

JasonGross:
On top of #561

This PR factors out the specific-to-C code, and make it a typeclass, and add support for `--lang=C` (which is also the default, so the option is currently a no-op).

This should hopefully make it quite easy to pipe a different output language through the pipeline.

@zoep, what do you think?  Does this work well for you as an entry-point for adding Rust?  (The most relevant bits to look at are the `OutputLanguageAPI` class at the bottom of `src/LanguageStringification.v`, and the definition `supported_languages` in `src/CLI.v`.  To add Rust, you should only have to make a Rust-version of the now-smaller CStringification.v ("only" 1565 lines rather than 2367 lines) and provide a Rust definition of type `OutputLanguageAPI` (feel free to factor common things out of CStringification and the Rust version, if that's helpful), and add one line to the definition `supported_languages` in `src/CLI.v`; everything else should be automatic.)


zoep:
@JasonGross, 

Yay, it's great that I don't have to modify >=30 call sites to `FromPipelineToString` to propagate the rust extraction option! As for the CLI, I've already added the option in my Rust branch, I can rebase if needed, but it would be great if we can coordinate to avoid duplicate work.

That doesn't take care of decoupling the backend from the frontend and having a clear frontend -> backend pipeline (which has it's own merit), but it suffices for us since at least the frontend is parameterized by the backend, and it doesn't need to be modified to add a new backend.

Tagging @huitseeker to keep him up to date.

JasonGross:
> I can rebase if needed, but it would be great if we can coordinate to avoid duplicate work.

I don't currently have any further changes planned until September (unless you make comments where you want me to change or refactor things and it wouldn't take me too much time), so I'd appreciate you rebasing on this (or rebasing once it's merged).  If you are happy with the state of this PR, I'll go ahead and merge it; is there anything you want changed or want to comment on before I merge this and #561?

FWIW, I expect the changes you need to make to `CLI.v` will be adding
```coq
Require Import Crypto.RustStringification.
```
and changing
```diff
Import
  LanguageStringification.Compilers
-  CStringification.Compilers.
+  CStringification.Compilers
+  RustStringification.Compilers.
```
```diff
  Definition supported_languages : list (string * ToString.OutputLanguageAPI)
-    := [("C", ToString.OutputCAPI)].
+    := [("C", ToString.OutputCAPI);
+        ("Rust", ToString.OutputRustAPI)].
```

zoep:
Yes that looks good, since it allows us to add a pretty much independent Rust extraction pipeline (modulo the PHOAS -> IR translation which will need some language specific parameterization). It would be great if you can merge this.

JasonGross:
Merged!

huitseeker:
Thanks @JasonGross  for the quick iterations on this! It's so awesome to see this move forward! :fireworks:

zoep:
Hi @JasonGross ,

I was just reading this more carefully, and I noticed that the PHOAS -> IR (which I plan to reuse modulo  some parametrization is still in CStringification that is a C-specific file.

I plan to split that in in two files (1. PHOAS -> IR 2. IR -> C). (Note that I'm using the same IR for the Rust translation). Let me know if there's any opposition to that.

Cheers,

JasonGross:
Sure, that seems fine.  It seems a bit unfortunate to me if the IR representation of the code depends on what language you're eventually targeting (from passing different parameters for C vs Rust), but I don't strongly object to it.

(I do think that the computation of common type and of which casts are needed in C should continue to live in the C file, which suggests that the correct parameterization might be over a "choose which casts to include" function or something.  What are your thoughts on parameterization?)

zoep:
> It seems a bit unfortunate to me if the IR representation of the code depends on what language you're eventually targeting

I completely agree, but I'd very much do that rather than change the IR or the translation, so that this can be handled in the next stage (IR -> C/Rust).

> (I do think that the computation of common type and of which casts are needed in C should continue to live in the C file, which suggests that the correct parameterization might be over a "choose which casts to include" function or something. What are your thoughts on parameterization?)

I'm thinking of parameterizing over two functions that define what cast should be inserted in the operands of binary/unary operations. I was thinking of adding Zcast_up and Zcast_down too (that as well could be language specific) but currently for Rust it's exactly the same as C, so I'm not sure yet.

PR #563, opened 2019-07-18
Thread the static arg all the way through the pipeline

JasonGross:
Now in order to get static functions out of the pipeline, you need to
pass --static.

This should hopefully allow for more flexible backend choices.

@zoep I noticed you decided to ignore the `static` flag at https://github.com/zoep/fiat-crypto/commit/1087360372d0482cc2041561fa1d7b62b68b4a16#r34349304 .  I'm hoping this is a better solution, and will mean that you can actually respect the `static` flag in the Rust backend.  What do you think?


zoep:
Seems great @JasonGross, thank you!

PR #564, opened 2019-07-18
Coq PR# 9918 overlay: add -no-template-check

mattam82:
Same story as mit-plv/fiat#28 adapt to coq/coq#9918

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -1,3 +1,4 @@
+-arg -no-template-check
 -R src Crypto
 -R bbv/theories bbv
 bbv/theories/BinNotation.v


JasonGross:
Is it possible to fix fiat-crypto without relying on unsound flags?  Unlike fiat, fiat-crypto shouldn't be relying on being able to instantiate fancy inductives with Prop/Set.  I'd be fine with a fix that, e.g., replaces all instances of Set with Type or something.

mattam82:
It's certainly doable but requires moving the inductive declarations that you want to be template-polymorphic outside of sections: i.e. make the parameterization explicit when you define them. It will be backward compatible.

JasonGross:
That should be fine.  Which inductives are problematic?

mattam82:
I don't know, for sure things in Parsing rely on polymorphism. It's easy to check, just remove the flag and see where you get typing errors, they will involved inductives that are no longer seen as template polymorphic with the new criterion.

JasonGross:
It looks like only a small change was needed for fiat-crypto: #568.  I will merge that as soon as the CI passes

JasonGross:
I've now merged #568, so this is no longer needed

PR #565, opened 2019-07-22
Moved PHOAS->IR translation out of CStringification.v

zoep:
This splits CStringification.v into two separate files, namely IR.v (that contains the IR definition and the translation from PHOAS to IR) and CStringification.v that contains the printer prints C from the IR representation. 

There are two other PR's that depend on this one: 
1. https://github.com/zoep/fiat-crypto/pull/1 that parameterizes the translation over language-specific casts.
2.  https://github.com/zoep/fiat-crypto/pull/2 that adds the Rust extractor and involves a minor fix to the PHOAS -> IR translation.


src/IR.v
```diff
@@ -0,0 +1,1206 @@
...
+               => base.interp t
+             end.
+        Definition arith_expr_for (t : Compilers.type.type base.type) : Set
+          := match t with
+             | type.base t => arith_expr_for_base t
+             | type.arrow s d => Empty_set
+             end.
+
+        (** Quoting
+            http://en.cppreference.com/w/c/language/conversion:
```
JasonGross: Please keep `integer_promote_type` and `common_type` in `CStringification`, as they cite the C spec and don't make much sense for other languages.  Feel free to parameterize the code in IR over these functions (or over a function that is "for a binop with these input types, what will the output type be if I don't do any casting or something).

zoep: This is used in the PHOAS -> IR translation itself, and therefore it cannot be move right now as this will create a circular dependency. Please refer to the followup PR zoep#1 that parameterizes the translation, and therefore such things are move to the files they belong to.

huitseeker:
@JasonGross It's not completely obvious, but https://github.com/zoep/fiat-crypto/pull/1 is opened against the branch that forms this diff, so should you want @zoep to merge that PR within the present one because they form a more coherent whole, it's a click away.

JasonGross:
Eh, I'm okay with merging this and then merging https://github.com/zoep/fiat-crypto/pull/1 in the near future.

zoep:
Hi @jason, thanks for merging! I created a new one (#567) from zoep#1, that addresses most of the comments, let us know if there's something else you'd like us to address.

PR #568, opened 2019-07-25
[sp2019latest] Move a template inductive outside a section

JasonGross:
For compatibility with coq/coq#9918

Closes #564

We could instead have just removed all of the Compilers/InSet stuff, as
nothing depends on it (and the issue it was meant to work around in
reification by parametericity is (a) fixed by using `type of` in Ltac
and (b) doesn't even seem to be used in the reification pass).

Instead, I chose to leave the existing code mostly as-is, and insead
move the one inductive that needs to be template-polymorphic out of a
section.  I chose this because it seems to be the decision that is
marginally in the direction of stress-testing Coq more.

### DIFF
--- a/src/Compilers/Syntax.v
+++ b/src/Compilers/Syntax.v
@@ -10,18 +10,23 @@ Delimit Scope ctype_scope with ctype.
 Local Open Scope ctype_scope.
 Delimit Scope expr_scope with expr.
 Local Open Scope expr_scope.
+Inductive flat_type (base_type_code : Type)
+  := Tbase (T : base_type_code) | Unit | Prod (A B : flat_type).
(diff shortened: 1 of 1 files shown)


JasonGross:
Apparently this works in 8.9 and master but not 8.8 nor 8.7, I'm looking into it

JasonGross:
@mattam82 this is unfortunately blocking on [Travis fixing a bug in their setup that is making every single build of ours fail](https://travis-ci.community/t/all-our-travis-builds-are-failing-because-apt-has-bad-keys-for-chrome/4497).  If Travis doesn't get their act together, I'll probably try to test this myself sometime next week, and/or move to a different CI setup.

PR #570, opened 2019-07-29
Rust code generation

zoep:
Added additional code generation backend to generate Rust programs.


src/IR.v
```diff
@@ -439,8 +439,65 @@ Module Compilers.
                  | ident.pair A B
                    => fun _ _ _ => inr ["Invalid identifier in arithmetic expression " ++ show true idc]%string
                  | ident.Z_opp (* we pretend this is [0 - _] *)
-                   => fun r y => let zero := (literal 0 @@ TT, Some (int.of_zrange_relaxed r[0~>0])) in
-                                 ret (arith_bin_arith_expr_of_PHOAS_ident Z_sub r (zero, y))
```
zoep: This changes the way arithmetic negation was handled by changing the upcast from `desired_type` to `signed_counterpart desired_type`. This has no effect for C generated code.

fiat-rust/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+pub mod curve25519_32;
+pub mod curve25519_64;
+pub mod p448_solinas_64;
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
+        assert_eq!(2 + 2, 4);
```
JasonGross: Interesting test.  What's it here for?

src/RustStringification.v
```diff
@@ -0,0 +1,391 @@
...
+Module IR := IR.Compilers.ToString.IR.
+Module ToString := LanguageStringification.Compilers.ToString.
+
+Module Rust.
+
+  (* Header imports *)
+  Definition imports : string := "#![allow(unused_parens)]" ++ String.NewLine ++ "use std::prelude;".
+
+  (* Header imports and type defs *)
+  (* TODO thread static flag in order to append pub *)
```
JasonGross: Oops, I can fix this

src/RustStringification.v
```diff
@@ -0,0 +1,391 @@
...
+     non-void functions in Rust and just return the result *)
+
+  Definition primitive_type_to_string (prefix : string) (t : IR.type.primitive)
+             (r : option ToString.int.type) : string :=
+    match t with
+    | IR.type.Zptr => "&mut "
+    | IR.type.Z => ""
+    end ++ match r with
+           | Some int_t => int_type_to_string prefix int_t
+           | None => "ℤ" (* XXX what is this unicode symbol?? *)
```
JasonGross: “ℤ” U+2124 Double-Struck Capital Z Unicode Character

It's the name I used for "bigint; please syntax/name error" because I thought it was cute.  Also because it was useful for error messages to be able to render code that had badly-sized bounds, sometimes, I think.

src/RustStringification.v
```diff
@@ -0,0 +1,391 @@
...
+       (* arithmetic operations *)
+       | (IR.Z_add @@ (x1, x2)) =>
+         "(" ++ arith_to_string prefix x1 ++ " + " ++ arith_to_string prefix x2 ++ ")"
+       | (IR.Z_mul @@ (x1, x2)) =>
+         "(" ++ arith_to_string prefix x1 ++ " * " ++ arith_to_string prefix x2 ++ ")"
+       | (IR.Z_sub @@ (x1, x2)) =>
+         "(" ++ arith_to_string prefix x1 ++ " - " ++ arith_to_string prefix x2 ++ ")"
+       | (IR.Z_opp @@ e) => "(-" ++ arith_to_string prefix e ++ ")" (* arithmetic negation *)
+       | (IR.Z_bneg @@ e) => "(!" ++ arith_to_string prefix e ++ ")" (* logical negation. XXX this has different semantics for numbers <>
+                                                                        0 or 1 than it did before *)
```
JasonGross: You should check that this is okay; we use `!!` to convert to a Boolean in some places, I think.

src/RustStringification.v
```diff
@@ -0,0 +1,391 @@
...
+
+  Definition to_strings (prefix : string) (e : IR.expr) : list string :=
+    List.map (stmt_to_string prefix) e.
+
+  Import Crypto.Language.Compilers Crypto.Language.Compilers.defaults IR.OfPHOAS.
+
+  Inductive Mode := In | Out.
+
+
+  (* This would have been nice, but coercions don't work *)
```
JasonGross: Which coercions do you want?

fiat-rust/src/curve25519_32.rs
```diff
@@ -0,0 +1,902 @@
...
+ * Postconditions:
+ *   eval out1 mod m = bytes_eval arg1 mod m
+ *
+ * Input Bounds:
+ *   arg1: [[0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0x7f]]
+ * Output Bounds:
+ *   out1: [[0x0 ~> 0x4666666], [0x0 ~> 0x2333333], [0x0 ~> 0x4666666], [0x0 ~> 0x2333333], [0x0 ~> 0x4666666], [0x0 ~> 0x2333333], [0x0 ~> 0x4666666], [0x0 ~> 0x2333333], [0x0 ~> 0x4666666], [0x0 ~> 0x2333333]]
+ */
+pub fn fiat_25519_from_bytes(out1: &mut [u32; 10], arg1: &[u8; 32]) -> () {
+  let x1: u32 = (((arg1[31]) as u32) << 18);
```
briansmith: It would be nice if these used `u32::from()` instead of `as u32` to be clear that the cast is guaranteed to be lossless..

src/RustStringification.v
```diff
@@ -0,0 +1,391 @@
...
+  Fixpoint to_arg_list_for_each_lhs_of_arrow (prefix : string) {t} : type.for_each_lhs_of_arrow var_data t -> list string
+    := match t return type.for_each_lhs_of_arrow var_data t -> _ with
+       | type.base t => fun _ => nil
+       | type.arrow s d
+         => fun '(x, xs)
+            => to_arg_list prefix In x ++ @to_arg_list_for_each_lhs_of_arrow prefix d xs
+       end%list.
+
+  (* Used in comment *)
+  Fixpoint bound_to_string {t : base.type} : var_data t -> Compilers.ZRange.type.base.option.interp t -> list string :=
```
JasonGross: Maybe this belongs in IR.v?

src/RustStringification.v
```diff
@@ -0,0 +1,391 @@
...
+                  end]%string
+    | base.type.option _
+    | base.type.unit
+    | base.type.bool
+    | base.type.nat
+    | base.type.zrange => fun _ _ => []
+    end%list.
+
+  (* Used in comment *)
+  Fixpoint input_bounds_to_string {t}
```
JasonGross: Maybe this belongs in IR.v (it's shared between Rust and C, right?)

fiat-rust/src/curve25519_32.rs
```diff
@@ -0,0 +1,902 @@
...
+ * The function fiat_25519_to_bytes serializes a field element to bytes in little-endian order.
+ * Postconditions:
+ *   out1 = map (λ x, ⌊((eval arg1 mod m) mod 2^(8 * (x + 1))) / 2^(8 * x)⌋) [0..31]
+ *
+ * Input Bounds:
+ *   arg1: [[0x0 ~> 0x4666666], [0x0 ~> 0x2333333], [0x0 ~> 0x4666666], [0x0 ~> 0x2333333], [0x0 ~> 0x4666666], [0x0 ~> 0x2333333], [0x0 ~> 0x4666666], [0x0 ~> 0x2333333], [0x0 ~> 0x4666666], [0x0 ~> 0x2333333]]
+ * Output Bounds:
+ *   out1: [[0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0x7f]]
+ */
+pub fn fiat_25519_to_bytes(out1: &mut [u8; 32], arg1: &[u32; 10]) -> () {
```
briansmith: It would be better for this to return a `[u32; 10]` array as a return value rather than take an out parameter.

fiat-rust/src/curve25519_32.rs
```diff
@@ -0,0 +1,902 @@
...
+ *   eval out1 mod m = (eval arg1 * eval arg2) mod m
+ *
+ * Input Bounds:
+ *   arg1: [[0x0 ~> 0xd333332], [0x0 ~> 0x6999999], [0x0 ~> 0xd333332], [0x0 ~> 0x6999999], [0x0 ~> 0xd333332], [0x0 ~> 0x6999999], [0x0 ~> 0xd333332], [0x0 ~> 0x6999999], [0x0 ~> 0xd333332], [0x0 ~> 0x6999999]]
+ *   arg2: [[0x0 ~> 0xd333332], [0x0 ~> 0x6999999], [0x0 ~> 0xd333332], [0x0 ~> 0x6999999], [0x0 ~> 0xd333332], [0x0 ~> 0x6999999], [0x0 ~> 0xd333332], [0x0 ~> 0x6999999], [0x0 ~> 0xd333332], [0x0 ~> 0x6999999]]
+ * Output Bounds:
+ *   out1: [[0x0 ~> 0x4666666], [0x0 ~> 0x2333333], [0x0 ~> 0x4666666], [0x0 ~> 0x2333333], [0x0 ~> 0x4666666], [0x0 ~> 0x2333333], [0x0 ~> 0x4666666], [0x0 ~> 0x2333333], [0x0 ~> 0x4666666], [0x0 ~> 0x2333333]]
+ */
+pub fn fiat_25519_carry_mul(out1: &mut [u32; 10], arg1: &[u32; 10], arg2: &[u32; 10]) -> () {
+  let x1: u64 = (((arg1[9]) as u64) * (((arg2[9]) * ((0x2 as u32) * (0x13 as u32))) as u64));
```
briansmith: Can you write `0x2 as u32` as `0x2_u32`?

src/RustStringification.v
```diff
@@ -0,0 +1,391 @@
...
+              ++ (List.map (fun s => if (String.length s =? 0)%nat then " *" else (" * " ++ s))%string (comment indata outdata))
+              ++ [" * Input Bounds:"%string]
+              ++ List.map (fun v => " *   "%string ++ v)%string (input_bounds_to_string indata inbounds)
+              ++ [" * Output Bounds:"%string]
+              ++ List.map (fun v => " *   "%string ++ v)%string (bound_to_string outdata outbounds)
+              ++ [" */"%string]
+              ++ to_function_lines static prefix name (indata, outdata, f))%list,
+           (* Zoe: This is a function in the CStringification file,
+              consider moving this elsewhere? *)
+           Compilers.ToString.C.collect_infos f)
```
JasonGross: Should probably live in IR.v

src/CLI.v
```diff
@@ -130,7 +131,7 @@ Module ForExtraction.
        end.
 
   Definition supported_languages : list (string * ToString.OutputLanguageAPI)
-    := [("C", ToString.OutputCAPI)].
+    := [("C", ToString.OutputCAPI); ("Rust", Rust.OutputRustAPI)].
```
JasonGross: Slight preference for these to be on separate lines so that if we get more languages it's easier to have fewer merge conflicts
```suggestion
    := [("C", ToString.OutputCAPI)
       ; ("Rust", Rust.OutputRustAPI)].
```

src/IR.v
```diff
@@ -439,8 +439,65 @@ Module Compilers.
...
+                                casted to the desired type, but to it's
+                                signed counterpart, then once the
+                                operation is performed the result is
+                                casted to the desired type. AFAICT this
+                                does't affect at all the C generated code
+                                but it fixes the problem we have with
+                                Rust. *)
+                        let zero := (literal 0 @@ TT, Some (int.of_zrange_relaxed r[0~>0])) in
+                        let '((e1, t1), (e2, t2)) :=
+                            bin_op_conversion (option_map int.signed_counterpart_of r)
```
JasonGross: This is wrong.  The signed counterpart cannot hold all of the values that the unsigned version can.  Does it change any of the output code if you union `r` with its signed counterpart?  (I'm also a bit confused what the issue you're having is; either `r` should already be signed, or else `t` should be signed, and then v the bin_op_conversion should keep them both signed...)

Ah, okay, I see what the issue is.  There's an implicit precondition that I forgot to write down of `Zcast_up_if_needed` that one of the ranges fits entirely inside the other one.  The correct fix here is to modify `Zcast_up_if_needed` so that if neither range is tighter than the other one, it will cast up to the *union* of the two ranges.  (In fact, you can just change the `else ` case to always cast to the union, because in the case where the existing range is tighter, the union will be the same as the desired range.)  This should fix your problem without needing to modify the handling of opp.

fiat-rust/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "fiat-rust"
+version = "0.1.0"
+authors = ["Zoe Paraskevopoulou <zoopar@fb.com>"]
+edition = "2018"
```
JasonGross:
```suggestion
edition = "2019"
```
?

src/RustStringification.v
```diff
@@ -0,0 +1,391 @@
+From Coq Require Import ZArith.ZArith MSets.MSetPositive FSets.FMapPositive
+     Strings.String Strings.Ascii Bool.Bool Lists.List.
+From Crypto.Util Require Import
```
JasonGross: I think this might result in warnings from coqdep; does it?

src/RustStringification.v
zoep: Yes, I agree. It would be convenient if `ToFunctionLines` (that's also shared and it's the one that calls the shared comment functions) was part of IR.v and the exposed function in `OutputLanguageAPI` was `to_function_lines` instead.

src/IR.v
zoep: @JasonGross  I'm not sure if the code before was capturing correctly what you actually want to have. Does your generated code make use of the wraparound semantics of C? Do you actually want to generate (0 - (something of unsigned type))? I remember from some previous conversation that all operations are guaranteed to fit in their desired type, which is clearly not the case here for types other than `uint_1` and `int1`.

zoep: The reason I'm saying for types other than `uint_1` and `int1` is because integer promotion happens in this case and so the operand is never casted to an unsigned type.

zoep: You can also have a look at the comment in the code of a concrete example where this happens.

src/RustStringification.v
zoep: `!` is overloaded with both boolean negation and bitwise negation in Rust. You actually do that:
```fiat_25519_uint1 x1 = (!(!arg1));```
where arg1 is `fiat_25519_uint1`. 

This works just fine in Rust, although I'm not sure what's the purpose is. To make sure that arg1 \in [0,1]? Then why not do it for other functions/inputs as well (e.g. in `
static void fiat_25519_subborrowx_u51` all args can take values greater than the expected ones)?

File: fiat-rust/src/lib.rs (hunk 2 above)
Comment (zoep) [reply fiat-rust/src/lib.rs]: This is placeholder code generated by cargo when a new Rust crate is created and therefore con be removed.

File: src/IR.v (hunk 14 above)
Comment (zoep) [reply src/IR.v]: > either `r` should already be signed, or else `t` should be signed, and then v the bin_op_conversion should keep them both signed...)

No that's not true. If `r` is unsigned and one operand has a signed type `t`, then `bin_op_conversion` might coerce `t` to `r`.

File: fiat-rust/src/curve25519_32.rs (hunk 10 above)
Comment (zoep) [reply fiat-rust/src/curve25519_32.rs]: I would like that too, but right now this decision is being made in a transformation prior to the Rust translation. In the future it would be nice to make this transformation more flexible so that the C and the Rust translation can decide how to output the result.

File: fiat-rust/src/curve25519_32.rs (hunk 7 above)
Comment (zoep) [reply fiat-rust/src/curve25519_32.rs]: This sounds like something that's definitely worth investigating.

File: fiat-rust/src/curve25519_32.rs (hunk 11 above)
Comment (zoep) [reply fiat-rust/src/curve25519_32.rs]: Actually no, because at the time the literals are printed their desired type is not known. I am considering removing casts from  integer literals all together, but I'm also thinking that casting a literal should be noop when the literal fits into the desired type.

File: src/RustStringification.v (hunk 3 above)
Comment (zoep) [reply src/RustStringification.v]: That'd be great!

File: fiat-rust/Cargo.toml (hunk 15 above)
Comment (huitseeker) [reply fiat-rust/Cargo.toml]: In that case, 2018 is appropriate. They occur on a ~3-year cycle.

File: fiat-rust/src/curve25519_32.rs (hunk 11 above)
Comment (huitseeker) [reply fiat-rust/src/curve25519_32.rs]: {Numeric casts](https://doc.rust-lang.org/nomicon/casts.html) between two integers of the same size is a no-op.

Comment (JasonGross) [reply fiat-rust/src/curve25519_32.rs]: Huh?  Can't you just append the string "_u32" rather than appending the string " as u32"?  I did have a specific function used to wrap literals in the C cast macros, and I imagine you can have that function do this instead.

File: fiat-rust/src/curve25519_32.rs (hunk 7 above)
Comment (huitseeker) [reply fiat-rust/src/curve25519_32.rs]: Yep, and the `impl` resolves to the same thing when casting to a smaller integer.

File: src/RustStringification.v (hunk 5 above)
Comment (JasonGross) [reply src/RustStringification.v]: Oh, maybe I was thinking of `nonzero` where we do this...  I think for the Montgomery code it's used to get gcc to recognize that it can transform some code into an addc insurrection.  And possibly I was just reusing the same code here.

File: src/RustStringification.v (hunk 9 above)
Comment (JasonGross) [reply src/RustStringification.v]: Feel free to make that change

File: fiat-rust/src/curve25519_32.rs (hunk 11 above)
Comment (zoep) [reply fiat-rust/src/curve25519_32.rs]: @JasonGross thats the printing of a cast that’s generated in PHOAS->IR. Yes I can print “_u32” but it will not get rid of the cast itself.

File: src/IR.v (hunk 14 above)
Comment (JasonGross) [reply src/IR.v]: > Does your generated code make use of the wraparound semantics of C? Do you actually want to generate (0 - (something of unsigned type))?

No.  The only case where we use something remotely like wraparound is `Z_lnot_modulo`.  We should never generate `0 - something of unsigned type`, that we would was a bug in `Zcast_up_if_needed`.  Just because the output of `0 - x` fits in an unsigned type does not mean that `x` fits in an unsigned type.  In particular, for `x = -2^31`, I believe we need to cast `x` to `int64` if we want to not take advantage of overflow behavior.

> I remember from some previous conversation that all operations are guaranteed to fit in their desired type, 

That's correct.

> which is clearly not the case here for types other than uint_1 and int1.

Again, this is a bug in `Zcast_up_if_needed`, right?

Comment (JasonGross) [reply src/IR.v]: > a concrete example where this happens which I think violates your precondition (one type is `int1` the other type is `uint1` and neither fits into the other).

Yes, the correct thing to do here is to cast up to `int8`.

Comment (zoep) [reply src/IR.v]: Yes that was a bug in ‘Zcast_up’ no disagreement here. I can update the PR with your suggested bug fix, if that works for you, and remove the special handling of this particular operation where the bug was manifested.

Comment (JasonGross) [reply src/IR.v]: > No that's not true. If r is unsigned and one operand has a signed type `t`, then `bin_op_conversion` might coerce `t` to `r`.

It would be a bug for `bin_op_conversion` to do this, I think the same bug as in `Zcast_up_if_needed`.  The spec of `bin_op_conversion` is "please ensure that my operands are big enough to not result in overflow when operating, given that the output is guaranteed to fit in this other type.

I think the correct operation to expose is something like `(option int.type)^3 -> (option int.type)^3` whose spec is "if the inputs are guaranteed to fit in these two types and the output is guaranteed to fit in this type, please tell me which casts I need to insert where".  Then C can do its fancy common-type cast-minimization thing, and Rust can just say "cast any types which are not already the union of the three types to the union""?

Comment (JasonGross) [reply src/IR.v]: > Yes that was a bug in ‘Zcast_up’ no disagreement here. I can update the PR with your suggested bug fix, if that works for you, and remove the special handling of this particular operation where the bug was manifested.

Yes, please

File: fiat-rust/src/curve25519_32.rs (hunk 11 above)
Comment (JasonGross) [reply fiat-rust/src/curve25519_32.rs]: Ah, I was confused because C does not cast the literals, but that's because C does more clever things to elide casts

Comment (zoep) [reply fiat-rust/src/curve25519_32.rs]: AFAICT both the macro and the cast will be generated in the C code here, but if there’s a smart way to get rid of such cast we’ll gladly adopt it for the Rust translation as well

Comment (JasonGross) [reply fiat-rust/src/curve25519_32.rs]: Ah, you are right, I was misremembering, sorry

File: src/RustStringification.v (hunk 5 above)
Comment (zoep) [reply src/RustStringification.v]: Hi @JasonGross, I had a look at all of the C generated files and I found this pattern only being used with the above way and in the context of negating `uint1`, so the Rust code should be OK. If in the future this changes though it will be problematic for the Rust code that should be edited to use a (perhaps handwritten?) version of negation with the same semantics as in C. What do you think?

I'm also tempted to edit the file by hand to remove the double negation and rerun benchmarks to see what's the effect of this.

File: src/RustStringification.v (hunk 9 above)
Comment (zoep) [reply src/RustStringification.v]: I suggest that I do a second iteration (perhaps next week though) and a separate PR that takes care of this (and other similar comments with functions that should live in IR.v.

**Comment by zoep** [comment]
> 



> Please also add targets to the `Makefile` to generate the Rust code, so that it's easy for me to regenerate as needed. You can copy the code for `c-files` and make a similar `rust-files` target; let me know if you have any questions or concerns about the current code or what I'm imagining for the Rust targets.

Hi @JasonGross, in the process of doing this I noticed than in quite a few places there are type errors in Rust files (which are ignored in C, but I'm not sure if these are indeed the desired types for the C code as well). In particular for the function `fiat_p256_addcarryx_uX` and `fiat_p384_subborrowx_uX` (where X \in {32, 64}) expect their second to last argument to be `uintX` however the arguments being passed are of type `uint8` in many places in the code.  

The affected files are: p384_64.rs p384_32.rs secp256k1_64.rs secp256k1_32.rs p224_64.rs p224_32.rs p434_64.rs

The pattern that's causing the problem (examples from p434_64, lines 304 and 374 respectively in the .c file) is calls of the form: 
```fiat_p434_addcarryx_u64(&x124, &x125, x123, (fiat_p434_uint1)x80, x108);```
where x80 is `uint64` (and has been used as such) but it's being downcasted to `uint8` although the function expects `u64`, and
```fiat_p434_addcarryx_u64(&mut x170, &mut x171, x169, x125, 0x0);```
where `x169` is expected to be `u64` but instead has type `u8`.

I was wondering if you could give us some context on these, and explain if that's indeed the desired behavior or some bug in e.g. propagating the desired types. I'm particularly worried about the first one since it involves a potentially unsafe downcast.

File: src/RustStringification.v (hunk 5 above)
Comment (JasonGross) [reply src/RustStringification.v]: I mean, it's already sort-of sketchy that we're implicitly relying on compiler magic to compile things into the right instructions for both performance and security.  (Did you know that gcc will sometimes emit conditional jumps for code with no if statements, but only if you preprocess your variable with `!!` (using `& 1` is insufficient to get a conditional jump)?

I think for now it's probably fine, as long as we remain aware that we are trusting the C / Rust compiler and the output needs to be benchmarked and checked to ensure that it's not doing inane things.

**Comment by JasonGross** [comment]
The relevant code to change is `recognize_3arg_2ref_ident`.  Currently on inputs it only ensures that the bounds are loose enough.  You'll want to modify it to be based on a language specific function of "cast up if needed for function argument" which in C never inserts casts but in Rust is `Zcast_up_if_needed`, I believe.  I designed it with the assumption that upcasts are only needed to ensure common type calculations pick up the right thing for arithmetic operators, which is true in C but not Rust.

The explicit downcast to uint1 is safe, note that previously this variable was created as the output of addc on a uint1, 0, and 0.  The fact that we don't remove this call suggests that we're missing a case in the rewriter to minimize code.  I'll take a look at it when I get a chance, but it's nothing to worry about (except for very minor efficiency gains), and it will be fixed (by adding an additional cast on top of the uint1 cast in Rust) if you make the change I suggest with `recognize_3arg_2ref_ident`.

File: Makefile [hunk 17]
```diff
@@ -290,6 +296,90 @@ p434_64.c p434_32.c : p434_%.c :
...
+
+# 2^216 * 3^137 - 1
+p434_64.rs p434_32.rs : p434_%.rs :
+	$(SHOW)'SYNTHESIZE > $@'
+	$(HIDE)rm -f $@.ok
+	$(HIDE)($(TIMER_FULL) $(WORD_BY_WORD_MONTGOMERY) --lang=Rust 'p434' '2^216 * 3^137 - 1' '$*' && touch $@.ok) > $@.tmp
+	$(HIDE)rm $@.ok && mv $@.tmp fiat-rust/src/$@
+
+test-rust-files: $(ALL_RUST_FILES)
+	$(shell cd fiat-rust; ${CARGO_BUILD})
```
JasonGross: Why not
```suggestion
	cd fiat-rust; $(CARGO_BUILD)
```
?

zoep: Sure if that works, I have no strong preference for the first one. My strategy when writing shell is whatever works first :).

src/IR.v
zoep: Hi again @JasonGross, 
I was just trying the suggested bug fix, and just changing `Zcast_up` to cast to the union of the current and the desired type does not immediately work because 1.) this might cast the two operands to different types (which doesn't work in Rust) 2.) generates casts to the type `fiat_25519_int2` which doesn't exist.

To address (1) I instead edited the `Rust_bin_op_conversion` to call `Zcast_up` with the desired type being the union of the desired type and the common type of the operands (which should make the implicit assumption about `Zcast_up` true). That is: 
```
        let ct := Rust_common_type t1' t2' in                                                                                                          
       (Zcast_up_if_needed (Some (ToString.int.union desired_type ct)) (e1, t1),                                                                              
        Zcast_up_if_needed (Some (ToString.int.union desired_type ct)) (e2, t2))                                                                                         
```

However, (2) remains and I have no clear understanding why it happens. Any thoughts on that?

zoep: @JasonGross actually let me entirely rephrase the previous comment as there was some confusion on my side: 

I did try to cast operands of binary operations to the union of the three types, as you suggested, but this ended up in getting weird 2-bit integer types (`fiat_25519_int2`) and I don't understand how to handle this. (I guess one option would be to edit `int.of_Zrange_relaxed` to avoid bitwidths that don't exist, but again I'm not entirely sure.)

zoep: And I guess yet an other option is to add custom 2-bit integer types (represented by 8bit ints as in the 1-bit case).

zoep:
> You'll want to modify it to be based on a language specific function of "cast up if needed for function argument" which in C never inserts casts but in Rust is `Zcast_up_if_needed`, I believe.

Yes that makes a lot of sense, and something similar is needed for assignments (Rust will need to cast them up).

src/IR.v
JasonGross: I think you want to add a function "widen to real type"; the "please tell me what casts to insert" function should widen things appropriately.

Makefile
JasonGross: I prefer the one I suggested; I think the version you have doesn't even work because make tries to interpret `${` and fails with "no variable `{`"?  (Am I wrong?)

zoep: The one I wrote works on my Mac, but I haven’t checked yours yet.

JasonGross: Ah, I see.  According to [stack overflow](https://stackoverflow.com/questions/25185607/whats-the-difference-between-parenthesis-and-curly-bracket-syntax-in-ma), `${...}` is the same as `$(...)`, just unconventional.  I'd still prefer the version I suggested.

JasonGross:
Travis says
```
File "./src/RustStringification.v", line 289, characters 28-50:
Error: The reference Rust_bin_op_conversion was not found in the current
environment.
```

huitseeker:
@JasonGross  It seems we have a build success & [the suggested fix](https://github.com/mit-plv/fiat-crypto/pull/570/commits/747d5c488f96a022f42b97bd97a8ee99342d13c0#diff-3e826a81d4d3c0ec365208a95e941178R291-R293) to [your comment](https://github.com/mit-plv/fiat-crypto/pull/570#discussion_r308473680)

PR #571, opened 2019-07-30
Avoid relying on `Export` bugs

maximedenes:
See https://github.com/coq/coq/issues/10474 and
https://github.com/coq/coq/issues/10480

This should be backward compatible, but it remains to be tested.


maximedenes:
@JasonGross The CI failure seems spurious. What should I do?

JasonGross:
I've restarted the build, let's see if it succeeds

PR #572, opened 2019-07-30
Bind Coqprime in _CoqProject

maximedenes:
Otherwise editing some fiat crypto files using CoqIDE or VsCoq will
fail.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -1,4 +1,5 @@
 -R src Crypto
+-R coqprime/src/Coqprime Coqprime
 src/AbstractInterpretation.v
 src/AbstractInterpretationProofs.v
 src/AbstractInterpretationWf.v


maximedenes:
> to get fiat-crypto to use a system-wide version of coqprime

Such a setup makes much more sense than duplicating the code of coqprime, indeed. Also it makes maintenance much easier. So what I'm wondering now is: why is there still a local copy of coqprime in fiat crypto?

JasonGross:
We do not use opam on our CI, and I personally do not use opam to manage Coq.  We also guarantee compatibility only with the pinned version of coqprime.  (Note that we do not duplicate the code, but instead have it as a git submodule.  So I don't think that there's that much of a maintenance burden; you merely have to update which commit the submodule points to.)

maximedenes:
> Note that we do not duplicate the code, but instead have it as a git submodule.

Ah ok, I missed that, sorry.

Honestly, I still find the user experience not very good (opening an IDE on fiat-crypto and trying to play scripts fails), and a bit strange to have two ways to build fiat crypto (with external coqprime or not), with only one of them being guaranteed to work.

PR #574, opened 2019-07-31
Test Rust on travis

JasonGross:
@zoep is "nightly" the right version of the Rust compiler to use?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,6 @@
-language: c
-
 sudo: required
 
 dist: trusty
-
-compiler:
-  - gcc
 
 cache:
   directories:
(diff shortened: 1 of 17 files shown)


zoep:
Should build with stable too since we’re not relying on nightly features, but nightly is fine too (and the one I used to build/test them).

JasonGross:
You don't need nightly for 128-bit integers?  I thought I saw a comment to that effect...

zoep:
> You don't need nightly for 128-bit integers? I thought I saw a comment to that effect...

Ah sorry this comment is outdated and I'll be removing it, they've been integrated to stable since 1.26.

JasonGross:
Ah, great, I'll test on stable then

PR #579, opened 2019-08-09
Avoid relying on `Export` bugs

maximedenes:
See https://github.com/coq/coq/issues/10474 and
https://github.com/coq/coq/issues/10480


maximedenes:
Same idea as https://github.com/mit-plv/fiat-crypto/pull/571, but for the sp2019latest branch.

Cc @JasonGross

PR #587, opened 2019-09-16
Add bedrock2 as a submodule

jadephilipoom:
This is so that we can make fiat-crypto output code in a format that can be consumed by the bedrock2 compiler.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ INSTALLDEFAULTROOT := Crypto
 
 .PHONY: coq clean update-_CoqProject cleanall install \
 	install-coqprime clean-coqprime coqprime coqprime-all \
+	bedrock2 clean-bedrock2 coqutil clean-coqutil \
 	install-standalone install-standalone-ocaml install-standalone-haskell \
 	util c-files rust-files \
 	nobigmem print-nobigmem \
@@ -38,9 +39,17 @@ include etc/coq-scripts/Makefile.vo_closure
 .DEFAULT_GOAL := all
 
 SORT_COQPROJECT = sed 's,[^/]*/,~&,g' | env LC_COLLATE=C sort | sed 's,~,,g' | uniq
+BEDROCK2_FOLDER := bedrock2/bedrock2
+BEDROCK2_SRC := $(BEDROCK2_FOLDER)/src
+BEDROCK2_NAME := bedrock2
+COQUTIL_FOLDER := bedrock2/deps/coqutil
+COQUTIL_SRC := $(COQUTIL_FOLDER)/src
+COQUTIL_NAME := coqutil
 update-_CoqProject::
 	$(SHOW)'ECHO > _CoqProject'
-	$(HIDE)(echo '-R $(SRC_DIR) $(MOD_NAME)'; (git ls-files 'src/*.v' | $(GREP_EXCLUDE_SPECIAL_VOFILES) | $(SORT_COQPROJECT))) > _CoqProject
+	$(HIDE)(echo '-R $(SRC_DIR) $(MOD_NAME)'; echo '-R $(BEDROCK2_SRC) $(BEDROCK2_NAME)'; \
+        echo '-R $(COQUTIL_SRC) $(COQUTIL_NAME)'; \
+        (git ls-files 'src/*.v' | $(GREP_EXCLUDE_SPECIAL_VOFILES) | $(SORT_COQPROJECT))) > _CoqProject
 
 # coq .vo files that are not compiled using coq_makefile
 SPECIAL_VOFILES := \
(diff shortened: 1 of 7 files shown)


JasonGross:
> make: *** No rule to make target `bedrock2/deps/coqutil/src/Z/Lia.vo', needed by `src/Util/ListUtil.vo'.  Stop.

I thought we had been pretty good about qualifying our `Require`s, but I guess we missed some.  One solution is to qualify the `Require` here, but I think a better solution is to use `-Q` or `COQPATH`.

Makefile
```diff
@@ -37,9 +38,17 @@ include etc/coq-scripts/Makefile.vo_closure
...
+BEDROCK2_FOLDER := bedrock2/bedrock2
+BEDROCK2_SRC := $(BEDROCK2_FOLDER)/src
+BEDROCK2_NAME := bedrock2
+COQUTIL_FOLDER := bedrock2/deps/coqutil
+COQUTIL_SRC := $(COQUTIL_FOLDER)/src
+COQUTIL_NAME := coqutil
 update-_CoqProject::
 	$(SHOW)'ECHO > _CoqProject'
-	$(HIDE)(echo '-R $(SRC_DIR) $(MOD_NAME)'; (git ls-files 'src/*.v' | $(GREP_EXCLUDE_SPECIAL_VOFILES) | $(SORT_COQPROJECT))) > _CoqProject
+	$(HIDE)(echo '-R $(SRC_DIR) $(MOD_NAME)'; echo '-R $(BEDROCK2_SRC) $(BEDROCK2_NAME)'; \
```
JasonGross:
```suggestion
	$(HIDE)(echo '-R $(SRC_DIR) $(MOD_NAME)'; echo '-Q $(BEDROCK2_SRC) $(BEDROCK2_NAME)'; \
```

Makefile
```diff
@@ -37,9 +38,17 @@ include etc/coq-scripts/Makefile.vo_closure
...
+        echo '-R $(COQUTIL_SRC) $(COQUTIL_NAME)'; \
```
JasonGross:
```suggestion
        echo '-Q $(COQUTIL_SRC) $(COQUTIL_NAME)'; \
```

jadephilipoom:
> Thoughts? (Perhaps we should all discuss in person?)

I think that's a good idea -- Andres was saying I should use `-Q` instead of `COQPATH`. Maybe the solution is to use `-Q` for coqprime as well, so we're still consistent.

JasonGross:
> Maybe the solution is to use `-Q` for coqprime as well, so we're still consistent.

No, this breaks the opam setup of fiat-crypto (which depends on coqprime as an opam dependency).

JasonGross:
> > Thoughts? (Perhaps we should all discuss in person?)
> 
> I think that's a good idea

I'm around my office for the next hour, and will be around tomorrow ~1:30--5pm.

The desiderata I'm tracking are:
1. it should be possible to install fiat-crypto via opam (required to bench fiat-crypto)
2. installing fiat-crypto via opam should result in a usable/working fiat-crypto state
3. it should be possible for a user to install both fiat-crypto and another project depending on one of fiat-crypto's dependencies, both via opam, without breaking anything

andres-erbsen:
@samuelgruetter do you have opinions on usage of `-Q` vs `COQPATH`?

samuelgruetter:
I've never created opam packages myself, and I've never used `COQPATH` before, so I don't have much to say here

jadephilipoom:
I updated bedrock to the newest version and then tried to copy the `COQPATH` logic from Coqprime, but have been unsuccessful (for some reason the `_CoqProject` in `coqutil` is empty, so it thinks nothing needs to be done and then `coqutil` doesn't build and therefore `bedrock2` doesn't build). @JasonGross, are you willing to either take this on or help me figure it out? I'd like to be able to merge some of my work on the fiat-crypto-to-bedrock compiler.

_CoqProject
```diff
@@ -1,4 +1,6 @@
 -R src Crypto
+-R bedrock2/bedrock2/src bedrock2
+-R bedrock2/deps/coqutil/src coqutil
```
samuelgruetter: I see two issues with the above two lines:
- Since @JasonGross's changes for `COQPATH` support were merged, it should now be `-R bedrock2/bedrock2/src/bedrock2 bedrock2` and `-R bedrock2/deps/coqutil/src/coqutil coqutil`.
- But if I understood correctly, these two lines should not be here at all, because @JasonGross wanted to use `COQPATH` to get these dependencies.

JasonGross:
@jadephilipoom I've rebased and updated the logic to work with COQPATH.  Let's see if this fixes things...

JasonGross:
coqutil does not build with Coq 8.8.  What should we do?  (Also, we'll want to have a bedrock2 opam package ready before this is merged.)

jadephilipoom:
Looks like Debian has updated to 8.9 and Ubuntu has either 8.6 or 8.9 -- I think we should just drop 8.8.

jadephilipoom:
Build is working for me -- I can import bedrock2 files. Thanks, Jason!

Note: as written, `make bedrock2` builds the stuff in `bedrock2/Examples`, which we probably don't want to worry about building. One solution is to add a `no-examples` make target to bedrock2.

JasonGross:
Great!  Yeah, feel free to add such a target or whatever.  I want to wait to actually merge this until https://github.com/coq/opam-coq-archive/pull/928 and https://github.com/coq/opam-coq-archive/pull/926 get merged, so that I can make a quick pull request to add bedrock2 as a dep to the fiat-crypto package.  (I also need to remember to add bedrock2 to the bench...)

PR #592, opened 2019-10-01
Hard-code overflow behavior to be wrapping

JasonGross:
This is required for dropping the carry in the mul-split addition
rewrite rule.

We also insert a compile-time test into the C code for twos-complement
wrapping behavior.

Question/TODO: Should we insert a test or use, e.g., `Wrapping`, in the
Rust code?

Note that this weakens the guarantee we get from bounds analysis; now
our guarantee is that we perform correctly when doing twos-complement
wrapping on overflow, rather than that we perform correctly regardless
of overflow behavior (as long as it is well-defined).  (@achlipala do you have thoughts/objections?)

Note also that we leave everything parameterized over the now-useless
`cast_outside_of_range` so that we don't have to change large amounts of
code just to support targeting bedrock2.

### DIFF
--- a/curve25519_32.c
+++ b/curve25519_32.c
@@ -12,6 +12,10 @@
 typedef unsigned char fiat_25519_uint1;
 typedef signed char fiat_25519_int1;
 
+#if (-1 & 3) != 3
+#error "This code only works on a two's complement system"
+#endif
+
 
 /*
  * The function fiat_25519_addcarryx_u26 is an addition with carry.
--- a/curve25519_64.c
+++ b/curve25519_64.c
@@ -14,6 +14,10 @@ typedef signed char fiat_25519_int1;
 typedef signed __int128 fiat_25519_int128;
 typedef unsigned __int128 fiat_25519_uint128;
 
+#if (-1 & 3) != 3
+#error "This code only works on a two's complement system"
+#endif
+
 
 /*
  * The function fiat_25519_addcarryx_u51 is an addition with carry.
--- a/p224_32.c
+++ b/p224_32.c
@@ -15,6 +15,10 @@
 typedef unsigned char fiat_p224_uint1;
 typedef signed char fiat_p224_int1;
 
+#if (-1 & 3) != 3
+#error "This code only works on a two's complement system"
+#endif
+
 
 /*
  * The function fiat_p224_addcarryx_u32 is an addition with carry.
--- a/p224_64.c
+++ b/p224_64.c
@@ -17,6 +17,10 @@ typedef signed char fiat_p224_int1;
 typedef signed __int128 fiat_p224_int128;
 typedef unsigned __int128 fiat_p224_uint128;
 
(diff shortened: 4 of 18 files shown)


achlipala:
I don't have a strong opinion here.  Does the C standard force two's-complement behavior?  Are there popular implementations that deviate?  What about Rust?

JasonGross:
I think basically every machine still in use uses two's-complement.  [This proposal from March 2018 for C++20](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2218.htm) proposes to encode two's-complement in the standard, at least for C++.  [This Stack Overflow answer](https://stackoverflow.com/a/12277974/377022) says:
> The most recent example I can find is the UNISYS 2200 series, based on UNIVAC, with ones-complement arithmetic. The various models were produced between 1986 and 1997 but the OS was still in active development as late as 2015. They also had a C compiler, as seen here.
>
> It seems likely that they may still be in use today.

[This Rust page](https://internals.rust-lang.org/t/a-tale-of-twos-complement/1062) suggests that Rust is specified to use wrapping two's-complement in production builds, and to throw assertion failures in debug/testing builts for overflow in both signed and unsigned arithmetic (note that overflow on unsigned integers in C has always (AFAIK) been defined to wrap), though they seem to provide a `Wrapped<T>` type family for explicit wrapping semantics.  I'm not sure if using this type family results in any overhead.

jadephilipoom:
So the state before was that one of the rewrite rules matched on an addition of two double-words (meaning two single-width words representing a double-width one for mul_split) and that was casted to a double-word, for which didn't fit the bounds because of the carry.

Now, if I understand correctly, none of that has changed, but now I could, instead of removing the cast, prove that if there was a carry then the expression would have gotten two's-complement wrapped. Doesn't that make the proof of the rewrite rule fail anyway, or would there be wrapping on both sides?

> just to support targeting bedrock2.

To be clear, this change is about bedrock2 only becsuse bedrock2 needs mul_split, right? Do we not expect anything else to need that behavior?

achlipala:
I'm satisfied that it's OK to add a requirement for the target language to expose two's-complement arithmetic.

JasonGross:
> Now, if I understand correctly, none of that has changed, but now I could, instead of removing the cast, prove that if there was a carry then the expression would have gotten two's-complement wrapped. Doesn't that make the proof of the rewrite rule fail anyway, or would there be wrapping on both sides?

There is wrapping on both sides.  The difference is that the LHS adds `c * 2^(2 * bitwidth)` and then takes it `mod` `2^(2*bitwidth)`, while the RHS just takes things `mod` `2^(2*bitwidth)`.

> To be clear, this change is about bedrock2 only becsuse bedrock2 needs mul_split, right? Do we not expect anything else to need that behavior?

That's correct, though I think we rely on twos complement representation (though not overflow behavior) when we do bit operations on signed integers in things like cmov.

PR #597, opened 2019-10-03
Generate most identifier definitions via tactics

JasonGross:
This results in minor churn in API.v and Fancy/Compiler.v.

Timing diff coming soon.

Next up will probably be removing `cast_out_of_range`, then seeing if I can do reification in a less hand-written way, and then packaging up the tactics for identifiers separately from running them.  After that, I'll probably move all of the identifier inductives to a single file, and then each of the other GENERATED files will just be a single definition with a single tactic call (except I guess there will still be a couple of `Hint Extern`s in the `Identifier` one).  After that is the code to scrape identifiers from a list of rewrite lemmas, and then finally writing the OCaml code.

### DIFF
--- a/src/AbstractInterpretation/Wf.v
+++ b/src/AbstractInterpretation/Wf.v
@@ -411,7 +411,7 @@ Module Compilers.
                            | break_innermost_match_step
                            | progress subst
                            | progress cbv [type_base ident.smart_Literal] in *
-                           | progress cbn [invert_Literal ident.invert_Literal] in *
+                           | progress cbn [invert_Literal] in *
                            | discriminate
                            | progress destruct_head' False
                            | progress expr.invert_subst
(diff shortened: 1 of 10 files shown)


JasonGross:
Hmm, it seems this makes things somewhat slower:
```
After     | File Name                                                       | Before    || Change    | % Change
---------------------------------------------------------------------------------------------------------------
52m00.91s | Total                                                           | 50m33.24s || +1m27.66s | +2.88%
---------------------------------------------------------------------------------------------------------------
0m23.36s  | Language/Identifier.vo                                          | 0m03.12s  || +0m20.23s | +648.71%
2m18.64s  | Rewriter/Passes/NBE.vo                                          | 2m01.34s  || +0m17.29s | +14.25%
1m44.30s  | Rewriter/Passes/ToFancyWithCasts.vo                             | 1m30.04s  || +0m14.26s | +15.83%
1m12.76s  | Rewriter/Passes/ArithWithCasts.vo                               | 0m59.40s  || +0m13.36s | +22.49%
1m03.62s  | Rewriter/Examples.vo                                            | 0m58.50s  || +0m05.11s | +8.75%
0m45.79s  | p521_32.c                                                       | 0m41.44s  || +0m04.35s | +10.49%
```

I'm inclined to just not care about this part of the performance at the moment, though my guess is that it's because my representation of the identifier packages is slightly different (the extra info one has an unfolded version of the normal info one internally, now).

Thoughts @achlipala @jadephilipoom @andres-erbsen?

PR #598, opened 2019-10-04
Prove mul_split admits

### LINKED ISSUE #590: Final pipeline currently relies on axioms
It relies on funext (which is not ideal, but which I can live with), but it also relies on `RewriterRulesProofs.mul_split_rewrite_rules_proofs`, which is the proof that the word-splitting rerwrite rules are correct.  @jadephilipoom, can you take charge of doing these proofs?  (I'm happy to do any part of them that is not just modular/bounded arithmetic reasoning on various operations.)

jadephilipoom:
Closes #590

### DIFF
--- a/src/Rewriter/Rules.v
+++ b/src/Rewriter/Rules.v
@@ -852,30 +852,37 @@ Section with_bitwidth.
              dont_do_again
              [(forall r x, cstZ r (cstZ r x) = cstZ r x) (* when inlining Z.combine_at_bitwidth, casts will sometimes get doubled up, so we need to strip the extra casts *)
               ; (forall x y,
-                    doublewidth (singlewidth x * singlewidth y)
-                    = (dlet lh := pairsinglewidth (Z.mul_split ('(2^bitwidth)) (singlewidth x) (singlewidth y)) in
-                           cstZsingle_to_double_pair lh))
+                    0 <= bitwidth
+                    -> doublewidth (singlewidth x * singlewidth y)
+                       = (dlet lh := pairsinglewidth (Z.mul_split ('(2^bitwidth)) (singlewidth x) (singlewidth y)) in
+                              cstZsingle_to_double_pair lh))
               ; (forall xl xh yl yh,
-                    doublewidth (cstZsingle_to_double xl xh + cstZsingle_to_double yl yh)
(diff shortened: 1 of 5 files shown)


src/Rewriter/Rules.v
```diff
@@ -852,30 +852,37 @@ Section with_bitwidth.
...
-                           cstZsingle_to_double (fst lc) h))
+                    0 <= bitwidth
+                    -> 1 <= lgcarrymax
+                    -> doublewidth (cstZsingle_to_double xl xh + singlewidth y)
+                       = (dlet lc := singlewidth_carry (Z.add_get_carry_full ('(2^bitwidth)) (singlewidth xl) (singlewidth y)) in
+                              dlet h := singlewidth (Z.add(*_with_carry*) (carrywidth (snd lc)) (singlewidth xh) (*(singlewidth ('0))*)) in
+                              cstZsingle_to_double (fst lc) h))
               ; (forall xl xh mask,
-                    0 <= mask < 2^bitwidth
+                    0 < mask < 2^bitwidth
```
JasonGross: This change shouldn't be necessary; if `mask` is `0`, the expression should just evaluate to `0` on both sides

src/Rewriter/Rules.v
```diff
@@ -852,30 +852,37 @@ Section with_bitwidth.
...
                     -> singlewidth (cstZsingle_to_double xl xh &' singlewidth ('mask))
                        = singlewidth (singlewidth xl &' singlewidth ('mask)))
               ; (forall xl xh mask,
-                    0 <= mask < 2^bitwidth
+                    0 < mask < 2^bitwidth
```
JasonGross: This change shouldn't be necessary; if `mask` is `0`, the expression should just evaluate to `0` on both sides

src/Rewriter/Rules.v
```diff
@@ -887,6 +894,17 @@ Section with_bitwidth.
                                                   (singlewidth ('bitwidth))
                                                   (singlewidth xh)
                                                   (singlewidth ('(bitwidth - offset)))))))
+              ; (forall xl xh offset,
+                    0 < offset < bitwidth
```
JasonGross: This should work even if `offset = 0`, right?

JasonGross:
Please also rebase this on top of master to get travis to accept it

Issue #524, opened 2019-02-18
bedrock2 extraction tracker issue


andres-erbsen:
It would be cool to generate bedrock2 code from fiat-crypto. I think it might go roughly as follows:

- [x] generate fiat-crypto low-level AST with only word sizes 32 and 64
- [x] break each 64-bit word into two 32-bit words (e.g., using mul_split)
- [x] translate that AST to `bedrock2.Syntax.cmd`

JasonGross:
The second pass should be possible as a rewriter pass.  The third pass probably wants something like CStringification / the translation to fancy

andres-erbsen:
I totally forgot about to today, but I'd like to brainstorm how to do the second pass (perhaps using the new and glorious lemma reification interface once it is working enough).

JasonGross:
We probably need a special identity identifier which is `inline`, and then we have definitions and lemmas like
```coq
Definition assemble (x y : Z) : Z := ((x << 32) + y).
Definition break (x : Z) : Z * Z := (ident.cast word32 (x >> 32), ident.cast word32 (x mod 2^32)).

ident.cast word64 x = dlet yz := break x in inline (assemble (ident.cast word32 (fst yz)) (ident.cast word32 (snd yz))
break (inline (assemble (ident.cast word32 x) (ident.cast word32 y))) = (ident.cast word32 x, ident.cast word32 y)
(inline (assemble (ident.cast word32 x1) (ident.cast word32 x2)) + inline (assemble (ident.cast word32 y1) (ident.cast word32 y2)) = stuff involving addc
```
Basically, the idea I have is that we first replace all of the things known to be word64 with `assemble (break x)`, and then we push `break` inwards until it hits an `assemble ` which it then cancels with.

JasonGross:
Note that there is some work-in-progress on this at https://github.com/mit-plv/fiat-crypto/compare/master...JasonGross:split-mul

It is mostly untested, and I probably won't get much of a chance to work on it before the end of the summer.

JasonGross:
I have now put a bit more work into it and tested it, and it seems to mostly work well.  For 64-bit 25519 mulmod, we get code like
https://github.com/mit-plv/fiat-crypto/blob/79356b0c7189cdb636bda23580dc213e55c5d029/src/SlowPrimeSynthesisExamples.v#L79-L321

[...]

PR #606, opened 2019-10-08
bedrock2 compiler (code, no proofs)

jadephilipoom:
Currently correctly converts curve25519 64-bit and fails on 32-bit because of the problem discussed here [0] (basically, the input is supposed to have only 32-bit words, and it in fact has 64-bit ones and therefore produces errors).

I'd like to get some feedback on the general design before I merge this and start proofs. Most of the diff is huge printouts in the `CompilerTest.v` file, which prints out the input and output for both the 64-bit and 32-bit test cases. Reviews should focus on `Compiler.v`, which is a more manageable size.

[0] https://github.com/mit-plv/fiat-crypto/issues/524#issuecomment-536057711


src/Bedrock/CompilerTest.v
```diff
@@ -0,0 +1,3662 @@
...
+      | ErrorT.Success e => snd (of_expr (e Compiler.var)
+                                         (Syntax.expr.var "y0", (Syntax.expr.var "y1", tt))
+                                         (fun _ _ => Syntax.expr.var "ret"))
+      | ErrorT.Error _ => Syntax.cmd.skip
+      end.
+
+    Import NotationsCustomEntry.
+    Eval compute in mulmod_bedrock.
+  (* using format_bedrock.sh:
+= bedrock_func_body
```
andres-erbsen: The local structure of output code looks reasonable here, probably worth checking against duplicative computation though.

src/Bedrock/Compiler.v
```diff
@@ -0,0 +1,576 @@
...
+    Fixpoint make_error t : value t :=
+      match t with
+      | type.base a => base_make_error a
+      | type.arrow a b => fun _ => make_error b
+      end.
+
+    (* given the next variable name to use and the type of a let
+       binder, generate a correctly-typed structure of variable names to
+       assign *)
+    Fixpoint get_retnames (t : base.type) (startname : Syntax.varname)
```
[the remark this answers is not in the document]
andres-erbsen: `translate_assinment`?

src/Bedrock/Compiler.v
```diff
@@ -0,0 +1,576 @@
...
+             (expr.App type_Z (type.arrow type_Z type_Z)
+                       (expr.Ident _ ident.Z_lor) x) y) =>
+          Syntax.expr.op Syntax.bopname.or (of_inner_expr true x) (of_inner_expr true y)
+        (* Z_shiftr -> bopname.sru *)
+        | (expr.App
+             type_Z type_Z
+             (expr.App type_Z (type.arrow type_Z type_Z)
+                       (expr.Ident _ ident.Z_shiftr) x) y) =>
+          Syntax.expr.op Syntax.bopname.sru (of_inner_expr true x) (of_inner_expr true y)
+        (* Z_truncating_shiftl : convert to bopname.slu if the truncation matches *)
```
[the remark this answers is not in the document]
JasonGross: I think that's accurate

src/Bedrock/Compiler.v
```diff
@@ -0,0 +1,576 @@
...
+    Local Notation maxint := (2 ^ Semantics.width).
+
+    (* Notations for commonly-used types *)
+    Local Notation type_nat := (type.base (base.type.type_base base.type.nat)).
+    Local Notation type_Z := (type.base (base.type.type_base base.type.Z)).
+    Local Notation type_ZZ :=
+      (type.base (base.type.prod (base.type.type_base base.type.Z)
+                                 (base.type.type_base base.type.Z))).
+
+    (* TODO : fill these in *)
```
JasonGross: Should these axioms be removed before or after this PR gets in?

src/Bedrock/Compiler.v
```diff
@@ -0,0 +1,576 @@
...
+        var_to_list rets = Some outnames ->
+        args_to_list argnames = Some innames ->
+        let of_expre : Syntax.cmd.cmd :=
+            snd (of_expr next_varname error (e var) nextname argnames rets) in
+        let interpe : type.interp base.interp t := API.interp (e _) in
+        In (fname, (innames, outnames, of_expre)) funnames ->
+        WeakestPrecondition.call
+          funnames fname trace mem args (fun _ => results_equivalent _ interpe).
+    Proof.
+      (* TODO : look at Jason's final correctness theorem, make sure this short-circuits correctly. *)
```
JasonGross: My final correctness theorem says that if the inputs are within bounds, then the interpretation after the bounds pipeline matches the interpretation before the bounds pipeline.
https://github.com/mit-plv/fiat-crypto/blob/5ebbd8e2a67142469cbb476312afda30163c8edb/src/BoundsPipeline.v#L526-L551
also cf
https://github.com/mit-plv/fiat-crypto/blob/5ebbd8e2a67142469cbb476312afda30163c8edb/src/Fancy/Montgomery256.v#L32-L34
and
the various correctness statements in https://github.com/mit-plv/fiat-crypto/blob/master/src/COperationSpecifications.v

(Let me know if you want me to say more things about the form of the correctness lemmas.)

src/Bedrock/Compiler.v
```diff
@@ -0,0 +1,576 @@
...
+      | base.type.prod a b => base_var a * base_var b
+      | base.type.list (base.type.type_base base.type.Z) =>
+        Syntax.expr.expr (* N.B. for lists, represents the list's *location in memory* *)
+      | _ => Syntax.varname
+      end.
+    (* interpretation of type.type base.type *)
+    Fixpoint var (t : type.type base.type) : Type :=
+      match t with
+      | type.base t => base_var t
+      | type.arrow s d => var s -> var d
```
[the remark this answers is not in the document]
jadephilipoom: Yes, when the input to the compiler is a function that isn't fully applied. The function's return variable names are allowed to be a function of the argument variable names. If we wanted to remove that ability, it could just be `var d`; that might be simpler.

src/Bedrock/Compiler.v
```diff
@@ -0,0 +1,576 @@
...
+      end.
+
+    (* the type of *values* of variables in terms of Syntax.expr.expr *)
```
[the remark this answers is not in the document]
jadephilipoom: I like calling `var` `ltype` and this `rtype`, so it helps keep the assignment logic clear.

src/Bedrock/Compiler.v
```diff
@@ -0,0 +1,576 @@
...
+        (* lists get a pass *)
+        true
+      | _ => false
+      end.
+
+    (* Used to interpret expressions that are not allowed to contain let statements *)
+    Fixpoint of_inner_expr
+             (require_cast : bool)
+             {t} (e : @API.expr var t) : value t :=
+      if (require_cast && negb (has_casts e))%bool
```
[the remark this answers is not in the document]
jadephilipoom: I'm not sure what you mean here. Why should the check be moved out, and what do you mean by "the error is never caught"?

src/Bedrock/Compiler.v
```diff
@@ -0,0 +1,576 @@
...
+      | base.type.type_base _ => Syntax.cmd.set
+      end.
+
+    Definition max_range : zrange := {| lower := 0; upper := 2 ^ Semantics.width |}.
+    Definition range_good (r : zrange) : bool := is_tighter_than_bool r max_range.
+
+    (* checks that the expression is either a) a literal nat or Z that
+    falls within the allowed range or b) an expression surrounded by
+    casts that fall within the allowed range *)
+    Definition has_casts {t} (e : @API.expr var t) : bool :=
```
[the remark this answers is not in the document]
jadephilipoom: Yes, that would follow the convention of the `invert_` lemmas in the fiat-crypto compiler. I chose this method because I've found that working with those lemmas and their complicated type statements that can be a pain in proofs, although I'm not sure that this will be less of a pain. My going plan is to try this and change it if it's causing problems.

src/Bedrock/Compiler.v
```diff
@@ -0,0 +1,576 @@
...
+    Local Notation type_ZZ :=
+      (type.base (base.type.prod (base.type.type_base base.type.Z)
+                                 (base.type.type_base base.type.Z))).
+
+    (* TODO : fill these in *)
+    Axiom valid_carry_expr : forall {t}, @API.expr var t -> Prop.
+    Axiom valid_inner_expr : forall {t}, @API.expr var t -> Prop.
+
+    (* states whether the expression is acceptable input for [of_expr] *)
+    Inductive valid_expr : forall {t}, @API.expr var t -> Prop :=
```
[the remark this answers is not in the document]
jadephilipoom: Yes, it could. That would make proofs using `valid_expr` as a precondition more annoying, I think, but would make it easier to prove that an expression fit the precondition. Probably worth it to switch.

src/Bedrock/Compiler.v
jadephilipoom: I was thinking after, it's a big enough PR already.

JasonGross:
> [0] (basically, the input is supposed to have only 32-bit words, and it in fact has 64-bit ones and therefore produces errors).

Does https://github.com/mit-plv/fiat-crypto/pull/614 fix things?

src/Bedrock/CompilerTest.v
```diff
@@ -0,0 +1,2672 @@
...
+    Definition mulmod_ : Pipeline.ErrorT (Expr _) :=
+        Pipeline.BoundsPipeline
+        false (* subst01 *)
+        None (* fancy *)
+        possible_values
+        ltac:(let r := Reify ((carry_mulmod limbwidth_num limbwidth_den s c n idxs)) in
+              exact r)
+               (Some loose_bounds, (Some loose_bounds, tt))
+               (Some tight_bounds).
+    Definition mulmod := Eval compute in mulmod_.
```
JasonGross:
```suggestion
    Definition mulmod := Eval vm_compute in mulmod_.
```
Note that `Eval compute` is `cbv` which is significantly slower than `vm_compute` (note that `Compute` is `Eval vm_compute`)

src/Bedrock/CompilerTest.v
```diff
@@ -0,0 +1,2672 @@
...
+    Definition mulmod_bedrock : Syntax.cmd.cmd :=
+      match mulmod with
+      | ErrorT.Success e => snd (translate_expr (e Compiler.ltype)
+                                                (Syntax.expr.var "y0", (Syntax.expr.var "y1", tt))
+                                                (Syntax.expr.var "ret"))
+      | ErrorT.Error _ => Syntax.cmd.skip
+      end.
+
+    Import NotationsCustomEntry.
+    Eval compute in mulmod_bedrock.
```
JasonGross:
```suggestion
    Compute mulmod_bedrock.
```
See above comment

src/Bedrock/CompilerTest.v
```diff
@@ -0,0 +1,2672 @@
...
+    Definition mulmod_ : Pipeline.ErrorT (Expr _) :=
+        Pipeline.BoundsPipeline
+        false (* subst01 *)
+        None (* fancy *)
+        possible_values
+        ltac:(let r := Reify ((carry_mulmod limbwidth_num limbwidth_den s c n idxs)) in
+              exact r)
+               (Some loose_bounds, (Some loose_bounds, tt))
+               (Some tight_bounds).
+    Definition mulmod := Eval compute in mulmod_.
```
JasonGross:
```suggestion
    Definition mulmod := Eval vm_compute in mulmod_.
```

src/Bedrock/CompilerTest.v
```diff
@@ -0,0 +1,2672 @@
...
+                     ((#Compilers.ident_Z_cast)%expr @ (#uint32)%expr @ ($x [[7]])) @
+                     ((#Compilers.ident_Z_cast)%expr @ (#uint32)%expr @
+                      (((#Compilers.ident_Z_cast)%expr @ (#uint32)%expr @ ($x0 [[3]]))%expr_pat *
+                       ((#Compilers.ident_Z_cast)%expr @ (#uint32)%expr @ (###2 * ###19))%expr_pat))))%expr_pat in
+   expr_let v23 := ((#Compilers.ident_Z_cast2)%expr @ (#uint32, #uint32)%expr @
+                    ((#Compilers.ident_Z_mul_split)%expr @ (###4294967296)%expr @
+                     ((#Compilers.ident_Z_cast)%expr @ (#uint32)%expr @ ($x [[6]])) @
+                     ((#Compilers.ident_Z_cast)%expr @ (#uint32)%expr @
+                      (((#Compilers.ident_Z_cast)%expr @ (#uint32)%expr @ ($x0 [[9]]))%expr_pat * ###19))))%expr_pat in
+   expr_let v24 := ((#Compilers.ident_Z_cast2)%expr @ (#uint32, #uint
```
JasonGross:
```suggestion
    Compute mulmod_bedrock.
```

jadephilipoom:
I think I've now addressed all of the comments here (I've left a TODO for the expression duplication in multiplication). I'll probably merge it once Travis is happy, unless anyone objects.

jadephilipoom:
Merged as #618

PR #607, opened 2019-10-08
Bump submodule and only make bedrock2 noex

JasonGross:
This should reduce the build noise a bit more, and make things faster


jadephilipoom:
I *just* put this in #606

JasonGross:
Do you mind if we merge this while you're getting feedback on #606?  I'd like to not need to keep waiting on examples when I switch versions of Coq.

jadephilipoom:
Sure

PR #614, opened 2019-10-11
Better conversion to 32-bit for bedrock2

JasonGross:
c.f.
https://github.com/mit-plv/fiat-crypto/issues/524#issuecomment-540059615

@jadephilipoom Does this fix the 32-bit bedrock2 compilation?

### DIFF
--- a/src/Rewriter/Rules.v
+++ b/src/Rewriter/Rules.v
@@ -851,14 +851,16 @@ Section with_bitwidth.
               ; (forall xl xh offset,
                     0 < offset < bitwidth
                     -> doublewidth (cstZsingle_to_double xl xh >> singlewidth ('offset))
-                       = cstZsingle_to_double
-                           (Z.lor (singlewidth (singlewidth xl >> singlewidth ('offset)))
-                                  (singlewidth
-                                     (Z.truncating_shiftl
-                                        (singlewidth ('bitwidth))
-                                        (singlewidth xh)
-                                        (singlewidth ('(bitwidth - offset))))))
-                           (singlewidth xh >> singlewidth ('offset)))
+                       = (dlet l := singlewidth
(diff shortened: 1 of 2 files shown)


src/Rewriter/Rules.v
```diff
@@ -851,14 +851,16 @@ Section with_bitwidth.
...
+                       = (dlet l := singlewidth
+                                      (Z.lor (singlewidth (singlewidth xl >> singlewidth ('offset)))
+                                             (singlewidth
+                                                (Z.truncating_shiftl
+                                                   (singlewidth ('bitwidth))
+                                                   (singlewidth xh)
+                                                   (singlewidth ('(bitwidth - offset)))))) in
+                              dlet h := singlewidth
+                                          (singlewidth xh >> singlewidth ('offset)) in
+                              cstZsingle_to_double l h))
```
jadephilipoom: Isn't this `cstZsingle_to_double` still going to produce a double-width word?

JasonGross: Yes, but it will get inlined into the use site and picked up as a pattern for further rewriting (e.g., if you add it to another double width word, it'll get eliminated there).  The overall pattern is to push the doublewidths further and further down the program, until they disappear at the end.

jadephilipoom: Well, then it depends whether it actually gets eliminated. Maybe we should merge it but not bother proving it until we run the stuff in `CompilerTest.v` from #606 and see the errors go away.

JasonGross: Travis should confirm that we don't need to adjust the proof at all.  (Such is the magic of automated proof scripts.)

JasonGross:
I've added a new rewrite rule that definitely removes the last of the combine_at_bitwidths and should remove the last of the ERRORs (I'm still checking that)

PR #615, opened 2019-10-11
Better conversion to 32-bit for bedrock2

JasonGross:
c.f.
https://github.com/mit-plv/fiat-crypto/issues/524#issuecomment-540059615

This PR is against bedrock2-backend


jadephilipoom:
Changes look good, but it seems like Travis times out.

JasonGross:
Indeed, I think `cast_to_arith r v; [ solve [bounds_arith_hammer] .. | ]` is taking forever to fail on 8.9 in some cases with these new rules.

JasonGross:
I've updated `bounds_arith_hammer` with a bit of a kludge; hopefully this will all work now

PR #623, opened 2019-12-13
Adapt to Coq's https://github.com/coq/coq/pull/7825

maximedenes:
To be merged after https://github.com/coq/coq/pull/7825 is merged.

### DIFF
--- a/src/Compilers/Z/Bounds/Pipeline/ReflectiveTactics.v
+++ b/src/Compilers/Z/Bounds/Pipeline/ReflectiveTactics.v
@@ -279,7 +279,9 @@ End with_round_up_list.
 Ltac refine_PipelineSideConditions_constructor :=
   lazymatch goal with
   | [ |- PipelineSideConditions ?opts ?evars _ ]
-    => simple refine {| Hevar := _ |}; [ shelve | .. ];
+    => simple refine {| Hevar := _ |};
+       (* for compatibility pre/post https://github.com/coq/coq/pull/7825 *)
+       [ shelve .. | | | | | | | | | | | | ];
        cbv [b e' e_final_newtype e e_pre_pkg e_pkg proj1_sig]
   end.
 Ltac refine_with_pipeline_correct opts :=
--- a/src/Compilers/Z/Bounds/Pipeline/ReflectiveTactics.v
+++ b/src/Compilers/Z/Bounds/Pipeline/ReflectiveTactics.v
@@ -279,7 +279,7 @@ End with_round_up_list.
 Ltac refine_PipelineSideConditions_constructor :=
   lazymatch goal with
   | [ |- PipelineSideConditions ?opts ?evars _ ]
-    => simple refine {| Hevar := _ |};
+    => simple refine {| Hevar := _ |}; [ shelve | .. ];
        cbv [b e' e_final_newtype e e_pre_pkg e_pkg proj1_sig]
   end.
 Ltac refine_with_pipeline_correct opts :=


maximedenes:
Sorry, but the build time of fiat-crypto-legacy is too much for me, please improve the fix if it needs to be done. The Coq PR is critical for unifall, and have been delayed by more than 6-months because of fiat-crypto-legacy problems already.

> I thought `simple refine` was supposed to never shelve things

Before the Coq PR, holes in arguments of `simple refine` would be shelved if they were unified with already shelved variables (though not always, it used to depend on the orientation of evar-evar unification problems).

JasonGross:
Okay, I will take a stab at fixing this within the next couple of days.  I suspect the fix I'll settle on is to use `[ shelve .. | | | | ]` with as many blank cases as there are expected goals.  (The issue is that there's one extra goal after this PR, right?) You don't happen to know the correct number of goals from your debugging session, do you?

JasonGross:
Also, sorry fiat-crypto legacy is such a hassle.

maximedenes:
Unfortunately, no, I could not determine if the number of goals was always the same. I studied only the first example where traces diverged.

> Also, sorry fiat-crypto legacy is such a hassle.

No worries, sorry to put the burden back on your shoulders. Shouldn't we consider tracking only the non-legacy version of fiat-crypto in our CI ? It could be better for everybody, I guess, no?

JasonGross:
> Unfortunately, no, I could not determine if the number of goals was always the same. I studied only the first example where traces diverged.

The number of goals should always be the same.  It should be equal to the number of fields of the record being refined with.  My first attempt is going to be just blindly using that as the number of expected goals, and seeing if that builds.

> Shouldn't we consider tracking only the non-legacy version of fiat-crypto in our CI ? It could be better for everybody, I guess, no?

The legacy version is the only heavy test of the performance of printing that I'm aware of.  It is also a heavy tester of both the VM and unification, and I think these things are important to continue testing.  Speaking of testing, does the test-suite have a case that tests this change in which things are shelved vs not shelved in `simple refine`?

maximedenes:
> The number of goals should always be the same. It should be equal to the number of fields of the record being refined with. My first attempt is going to be just blindly using that as the number of expected goals, and seeing if that builds.

Is the number of subgoals printed in ltac traces?

maximedenes:
> The legacy version is the only heavy test of the performance of printing that I'm aware of. It is also a heavy tester of both the VM and unification, and I think these things are important to continue testing.

Sure, but unification is well tested by other developments (it would still benefit a lot from having a spec and a suite of functional tests, but that's another story). For the VM, I'd need to check. Has the non-legacy version of fiat-crypto stopped using it?

For printing, I understand the idea, but it seems to me that having to explore millions of lines of ltac traces just to fix a tactic bug, finally realizing that the original script was relying on the bug is not a good trade-off. There must be a more optimal way of testing the performance of printing.

JasonGross:
> Has the non-legacy version of fiat-crypto stopped using it?

Mostly, yes.  The new version mostly relies on extraction + ocamlopt, though we do still use the VM in a couple of places for smaller tests.

> For printing, I understand the idea, but it seems to me that having to explore millions of lines of ltac traces just to fix a tactic bug, finally realizing that the original script was relying on the bug is not a good trade-off. There must be a more optimal way of testing the performance of printing.

Indeed.  Though do you really need to explore the millions of lines?  If you `Set Ltac Debug. Set Ltac Batch Debug.` and pipe the output into a file with Coq just before and just after the relevant PR, and then run `vimdiff` on the resulting files, shouldn't that highlight just the location where the traces diverge?

One possibility for testing printing is to copy over the inductive declarations and notations from fiat-crypto legacy, and then copy-paste the `Set Printing All` AST from a bunch of the tests.

JasonGross:
I'll merge once the CI passes

JasonGross:
https://github.com/coq/coq/pull/7825 was merged without needing this overlay, so I will close without merging

PR #626, opened 2020-01-06
Don't widen carry bits

JasonGross:
It is probably not necessary, and was messing up some of the rules for
emitting Go code.

@andres-erbsen (and maybe @davidben ) Can you review the changes to the C code here?  (The changes are that we replace some `uint32` with `uint8`, basically.)


andres-erbsen:
The changed widths do look like carry bits. I didn't try to re-run bounds analysis in my head, though. Is there anything else I could/should check?

JasonGross:
I seem to recall you mentioning that using uint8 rather than uint32 could maybe have performance impacts in some cases, and wanted to make sure these were not those cases.

andres-erbsen:
I don't know to tell for sure, but I think we are fine. I have seen it of memory use (not here) and compiler inserting explicit truncations (probably not here either).

davidben:
Looks reasonable as far as C integer rules go, though I too cannot check the bounds myself. :-) I stuffed it into BoringSSL and there wasn't a noticeable perf difference.

One observation about unsaturated limbs: the `curve25519_64.c` instances look like this, with `x49` a `uint64_t`.
```c
 fiat_25519_uint1 x50 = (fiat_25519_uint1)(x49 >> 51);
```

64 - 51 = 13 > 8, so the compiler may need to do a bit of work to prove it doesn't matter. In particular, the compiler doesn't know that the input bounds are tighter than `uint64_t`, so it's possible this would insert some bitmasks that otherwise wouldn't be needed. Perhaps this was the reason?

JasonGross:
> so it's possible this would insert some bitmasks that otherwise wouldn't be needed. Perhaps this was the reason?

Possibly.  I think I was following @andres-erbsen 's guidelines here, this might have been his reason.

In any case, I'll take this as an ack and merge.

PR #633, opened 2020-01-07
Add a Barrett reduction binary

JasonGross:
This is how we get a binary for Barrett reduction.  Unfortunately it doesn't really work well because we require 2 ^ (machine_wordsize - 1) < M < 2 ^ machine_wordsize.  So there's some work left to do, that I'd like to delegate to @jadephilipoom .  Leaving this here as a draft PR so that there's a reference for what adding it to the C output pipeline takes.


jadephilipoom:
We should *not* be using the existing Barrett template for non-Fancy code -- the problem is not that the precondition is wrong, it's that the precondition (like the entire template) is specialized to Fancy. We need to make a new template for this, and although it would be nice to have it's not at the top of my priority list and I don't know if I'll get around to it.

paola1-png:
Need access to synthesis/Barrettstsndalonehaskell0CamMain.v

PR #640, opened 2020-01-12
Add convert_bases_and_reduce to convert_bases

JasonGross:
As requested by @zx2c4, this version now does reduction

Questions for @jadephilipoom / @andres-erbsen :
1. Should we name the functions `convert_bases_no_reduce` and `convert_bases` to align with `chained_carries_no_reduce`, or should we call them `convert_bases` and `convert_bases_and_reduce`?
2. Is my carry-chain of `seq 0 dn ++ [0; 1]` correct?

```
USAGE: src/ExtractionOCaml/base_conversion [--lang=LANGUAGE] [--static] [--no-wide-int] [--widen-carry] [--widen-bytes] [--no-primitives] [--cmovznz-by-mul] curve_description src_n s-c src_limbwidth dst_limbwidth machine_wordsize [--inbounds-multipler=inbounds_multiplier] [--outbounds-multiplier=outbounds_multiplier] [--use-bitwidth-in] [--use-bitwidth-out] [function_to_synthesize*]

  LANGUAGE                The output language code should be emitted in.  Defaults to C if no language is given.  Case-sensitive.
                            Valid options are: C, Rust
  --static                Declare the functions as static, i.e., local to the file.
  --no-wide-int           Don't use integers wider than the bitwidth.
  --widen-carry           Widen carry bit integer types to either the byte type, or to the full bitwidth if --widen-bytes is also passed.
  --widen-bytes           Widen byte types to the full bitwidth.
  --no-primitives         Suppress the generation of the bodies of primitive operations such as addcarryx, subborrowx, cmovznz, mulx, etc.
  --cmovznz-by-mul        Use an alternative implementation of cmovznz using multiplication rather than bitwise-and with -1.
  curve_description       A string which will be prefixed to every function name generated
  src_n                   The number of limbs in the input
  src_limbwidth           The limbwidth of the input field element
  dst_limbwidth           The limbwidth of the field element to be returned
  machine_wordsize        The machine bitwidth (e.g., 32 or 64)
  inbounds_multiplier     The (improper) fraction by which the bounds of each input limb are scaled (default: 1)
  outbounds_multiplier    The (improper) fraction by which the bounds of each output limb are scaled (default: 1)
  --use-bitwidth-in       Instead of using an upper bound of s-1 on the input, use the maximum number that can be properly represented with the given base
  --use-bitwidth-out      Instead of using an upper bound of s-1 on the output, use the maximum number that can be properly represented with the given base
  function_to_synthesize  A space-separated list of functions that should be synthesized.  If no functions are given, all functions are synthesized.
                            Valid options are convert_bases, convert_bases_and_reduce
```

Example usage:
```c
$ src/ExtractionOCaml/base_conversion poly1305 6 '2^130-5' 25.5 '64' 64 --use-bitwidth-in --use-bitwidth-out convert_bases_and_reduce
/* Autogenerated: src/ExtractionOCaml/base_conversion poly1305 6 2^130-5 25.5 64 64 --use-bitwidth-in --use-bitwidth-out convert_bases_and_reduce */
/* curve description: poly1305 */
/* requested operations: convert_bases_and_reduce */
/* src_n = 6 (from "6") */
/* s-c = 2^130 - [(1, 5)] (from "2^130-5") */
/* src_limbwidth = 25.5 (from "25.5") */
/* dst_limbwidth = 64 (from "64") */
/* inbounds_multiplier = None (from "") */
/* outbounds_multiplier = None (from "") */
/* use_bithwidth_in = true */
/* use_bithwidth_out = true */

/* Computed values: */
/* dst_n = 3 */
/* max = 2^153 */

typedef signed __int128 fiat_poly1305_int128;
typedef unsigned __int128 fiat_poly1305_uint128;

/*
 * The function fiat_poly1305_convert_bases_and_reduce converts a field element from base 25.5 to base 64 in little-endian order, with reduction.
 * Postconditions:
 *   dst_eval out1 mod (s - c) = src_eval arg1 mod (s - c)
 *
 * Input Bounds:
 *   arg1: [[0x0 ~> 0x3ffffff], [0x0 ~> 0x1ffffff], [0x0 ~> 0x3ffffff], [0x0 ~> 0x1ffffff], [0x0 ~> 0x3ffffff], [0x0 ~> 0x1ffffff]]
 * Output Bounds:
 *   out1: [[0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0x1ffffff]]
 */
void fiat_poly1305_convert_bases_and_reduce(uint64_t out1[3], const uint64_t arg1[6]) {
  uint64_t x1 = (arg1[5]);
  uint64_t x2 = ((arg1[4]) << 38);
  uint64_t x3 = ((arg1[3]) << 13);
  fiat_poly1305_uint128 x4 = ((fiat_poly1305_uint128)(arg1[2]) << 51);
  uint64_t x5 = ((arg1[1]) << 26);
  uint64_t x6 = (arg1[0]);
  fiat_poly1305_uint128 x7 = (x6 + (x5 + x4));
  uint64_t x8 = (uint64_t)(x7 >> 64);
  uint64_t x9 = (uint64_t)(x7 & UINT64_C(0xffffffffffffffff));
  uint64_t x10 = (x3 + x2);
  uint64_t x11 = (x8 + x10);
  uint64_t x12 = (x11 & UINT64_C(0xffffffffffffffff));
  uint64_t x13 = (x1 & UINT64_C(0xffffffffffffffff));
  uint64_t x14 = (x9 & UINT64_C(0xffffffffffffffff));
  uint64_t x15 = (x12 & UINT64_C(0xffffffffffffffff));
  out1[0] = x14;
  out1[1] = x15;
  out1[2] = x13;
}
```


jadephilipoom:
Looking at the code, you're doing one reduce step, which will not always be
enough to fully reduce -- you'll end up with a smaller and equivalent
number modulo s-c, but not necessarily the minimal one. Is that what you
want? If so, it should not be called canonicalize.

JasonGross:
@zx2c4 do you want only one reduce step, it do you want more?

@jadephilipoom is there some number that will always be enough, and do we have a proof of it?  Alternatively, maybe I would make it configurable? (I presume repeat_reduce is the thing to use?)
And if it's not going to canonicalize, should we go back to calling it convert_bases_and_reduce?

jadephilipoom:
Does the reduction in Arithmetic/Freeze.v work for what you're trying to
do? I believe that one produces minimal residue, but it's possible it
requires specific preconditions to do so -- you'll have to take a look at
the proof and see if it matches this use case. (What exactly is this use
case? It's not clear from the PR.)

You can determine the number of reduce steps you need from the modulus +
representation. We probably have code for it somewhere -- look for things
that call reduce or repeat_reduce.

JasonGross:
The precondition of `freeze` is that the value is less than twice the prime, and it works by subtracting the prime from the value, and then conditionally adding back the prime iff the subtraction resulted in a non-zero borrow.  @zx2c4 do you want this in the code?

@jadephilipoom I'm actually not sure that we're doing reduction in the right place; in particular, it seems like base-conversion is going to result in having a high component that overflows the weight, and reduce isn't going to catch this?  Are you sure that we don't want to change `chained_carries_no_reduce` to `chained_carries` for this code?

The reference code that @zx2c4 sent me was
```c
static void convert_to_base2_64(void *ctx)
{
       struct poly1305_arch_internal *state = ctx;
       u32 cy;

       if (!state->is_base2_26)
               return;

       cy = state->h[0] >> 26; state->h[0] &= 0x3ffffff; state->h[1] += cy;
       cy = state->h[1] >> 26; state->h[1] &= 0x3ffffff; state->h[2] += cy;
       cy = state->h[2] >> 26; state->h[2] &= 0x3ffffff; state->h[3] += cy;
       cy = state->h[3] >> 26; state->h[3] &= 0x3ffffff; state->h[4] += cy;
       state->hs[0] = ((u64)state->h[2] << 52) | ((u64)state->h[1] <<
26) | state->h[0];
       state->hs[1] = ((u64)state->h[4] << 40) | ((u64)state->h[3] <<
14) | (state->h[2] >> 12);
       state->hs[2] = state->h[4] >> 24;
#define ULT(a, b) ((a ^ ((a ^ b) | ((a - b) ^ b))) >> (sizeof(a) * 8 - 1))
       cy = (state->hs[2] >> 2) + (state->hs[2] & ~3ULL);
       state->hs[2] &= 3;
       state->hs[0] += cy;
       state->hs[1] += (cy = ULT(state->hs[0], cy));
       state->hs[2] += ULT(state->hs[1], cy);
#undef ULT
       state->is_base2_26 = 0;
}
```

jadephilipoom:
This code:
1. Does a [0;1;2;3] carry chain (ignoring the last digit at position 4 and
letting things accumulate there if the number is too large). This is
precisely what chained_carries_no_reduce does, as opposed to
chained_carries, so if we were replicating directly we'd use that.
2. Converts to base 2^64 using shifts and ors.
3. I'm not 100% clear on what ULT does here, but I assume some kind of
modular reduction -- is it actually pseudomersenne reduction, though? It's
also not totally clear to me whether it produces minimal residue or just
below 2^130.

To make a fiat-crypto template for this operation, we need to generalize it
to something that works for many moduli. To do that, we need to know what
the requirements are, and that's still not clear to me. Is this canonical?
What are the preconditions? Which kinds of moduli is it expected to work
for? Is it faster than what we already have?

Without knowing these things, my first guess would be:

(* requires 0 <= eval p < 2 * (s-c) *)
     Definition canonicalize (sn dn : nat) (p : list Z) : list Z :=
  let p' := chained_carries_no_reduce dw dn p' (seq 0 (pred sn)) in
      Positional.reduce dw dn s c (convert_bases sn dn p').

Then there would need to be a proof similar to freeze, saying the result is
actually canonical. But at this point, it seems to be doing freeze in a
different way, and I want a justification for why one would do this instead
of the freeze we already have. Not to say there aren't reasons, but if
we're going to have two versions of a function we should be able to explain
the difference between use cases.

PR #641, opened 2020-01-16
Improve expressivity of parsing

JasonGross:
Perhaps it is overkill, but Fiat-Crypto now includes a small parser
combinator library for parsing command-line options.  We can now parse
things like `2^(1.1 * 255) + 0xF + 0o1 + 0b1010.1010 + 2^(255/10)`.

The actual impetus for this change was to allow parsing of lists, to be
used to parse lists of bounds.


JasonGross:
Build failure is unrelated

PR #653, opened 2020-02-07
Remove .travis.yml

### LINKED ISSUE #652: Should we use GitHub Actions rather than Travis?
We have had many problems with Travis.  They are exacerbated by the fact that we need multiple stages to run within the time limit.  I've just become aware that [GitHub Actions](https://github.com/features/actions) can apparently do CI stuff and are unlimited for public repos; maybe we should use those?

Thoughts @andres-erbsen ?

JasonGross:
As I've now set up GitHub actions for the CI (in [.github/workflows](https://github.com/mit-plv/fiat-crypto/tree/master/.github/workflows) / [Actions](https://github.com/mit-plv/fiat-crypto/actions)), this closes #652 (though we still need to add GitHub actions to sp2019latest and disable travis on this repo).


JasonGross:
Andres and Jade have no option on this, so merging.

Issue #655, opened 2020-02-08
32-bit p521 uses uint128


JasonGross:
https://github.com/mit-plv/fiat-crypto/blob/b3c26501432fd1ec5dd75e0ba136b667106fd097/p521_32.c#L129-L143

The issue is that we claim to synthesize 17-limb 32-bit p521, but 1.1 * (521 / 17) > 32.  (We can't have a bounds multiplier greater than 1.044 in this case).  What should be done?

@andres-erbsen ?

JasonGross:
Maybe we should also have a check that the bounds are all within the bitwidth...

andres-erbsen:
I think we don't care about p521 on 32-bit right now, so let's drop it and figure out these issues when we need it?

PR #656, opened 2020-02-08
Ensure that input/output bounds fit in bitwidth

JasonGross:
This is one way of dealing with #655 (changing from 17 limbs to 18
limbs).


JasonGross:
Note, however, that we still end up using uint128 for adding the partial products.  Not sure what to do about this...

PR #660, opened 2020-02-14
Add Java backend to Fiat-Crypto

JasonGross:
We can generate Java code for all of the 32-bit files.  We cannot
generate Java code for the 64-bit files because Java does not seem to
support uint128 nor 64x64 -> 64x64 multiplication.  <strike>We take advantage of
the fact that most operations commute with modulo to use signed integers
to hold unsigned values, and we always use `>>>` because (at least
currently) right-shift is always unsigned in fiat-crypto.  Some of these
choices are perhaps a bit sketchy; it'd be nice to find a way to hook
fiat-crypto up to some sort of verified semantics for the various
backends.</strike>

cc @peterdettman @dghgit, who I believe were the people talking with me about adding Java to fiat-crypto at HACS


JasonGross:
Oh, hrm, we probably need to be more clever about casting between different sizes, because casting 2^32 - 1 to a long should give 2^32 - 1, not 2^64 - 1.

JasonGross:
Andres pointed out to me that my analysis of signed integers commuting with modulo was incorrect.  I'll need to revise bounds analysis to actually produce signed integers.

JasonGross:
I am going to merge this; we can iterate on it in the future.

PR #668, opened 2020-03-04
Relax bounds on literals during stringification

JasonGross:
This results in fewer useless casts on literals, and fixes the
`to_bytes` Go synthesis issue.

The underlying assumption required by this change is that integer
literals don't need to be casted to be recognized as the narrowest
bitwidth that they fit into of the bitwidths we allow.  (This is
trivially true in C, at least.)

cc @andres-erbsen and @davidben for changes to the C files


andres-erbsen:
I looked at the C files and didn't see anything obviously wrong. I think go is fine with large integers, and in C the printer already uses UINT64_C and such. In general, I think that the printer for literals should deal with getting them interpreted correctly (like here), instead separately of inserting casts around them :+1: .

PR #670, opened 2020-03-06
Inversion (WIP)

bshvass:
The following PR is an attempt at implementing a constant time inversion algorithm by D.J Bernstein (available at https://gcd.cr.yp.to/papers.html#safegcd).
At the moment it is just a subroutine, called divstep2 in the paper, which is implemented and proven functionally correct. This is the content of src/Arithmetic/Inv.v. It can be printed to C by the code in src/PushButtonSynthesis/TestInv.

Since it is only a subroutine, there is a C implementation of the remaining steps in test-inversion-c/test_bls.c (this is the inversion function; the main is a test).
An implementation using the divstep2 as generated by Fiat is also available in the RELIC toolkit (https://github.com/relic-toolkit/relic see commit bd87b62e0acb1d65b360d104059d29ebd676cd97)
This achieves quite a good performance (about twice as slow as hand optimized, non-constant time extended euclidean) and there are still many optimizations which can be done.

To be able to generate the full algorithm, however, we need bounded loops and/or function calls to be supported. 
Also one optimization (which is the implementation presented in the Bernstein paper) requires multiplication of 64-bit numbers and saturated multi limbs numbers (e.g. 128-bit, 192-bit etc. (NOT mod p)). Is this something that is already supported? I was able to find a sat_mul function, but not one which was reifiable (in src/Arithmetic/Saturated.v).


src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+  (a mod 2 ^ m <? 2 ^ (m - 1)) = negb ((- a) mod 2 ^ m <? 2 ^ (m - 1)).
+Proof.
+  assert (0 < 2 ^ m) by (apply Z.pow_pos_nonneg; lia).  
+  assert (a mod 2 ^ m + (- a mod 2 ^ m) = 2 * 2 ^ (m - 1)) by  
+      (rewrite !Zmod_eq, Z_div_nz_opp_full, Pow.Z.pow_mul_base, Z.sub_simpl_r by lia; ring).
+  destruct (a mod 2 ^ m <? 2 ^ (m - 1)) eqn:E1;
+    destruct (- a mod 2 ^ m <? 2 ^ (m - 1)) eqn:E2;
+    rewrite ?Z.ltb_lt, ?Z.ltb_ge in *; try reflexivity; lia. Qed.
+
+Lemma opp_mod2 a : a mod 2 = - a mod 2.
```
JasonGross: Belongs in `src/Util/ZUtil/Modulo.v`

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+Proof. rewrite !Zmod_odd, Z.odd_opp; reflexivity. Qed.
+
+Lemma select_push cond a b f (H : length a = length b) :
+  f (select cond a b) = Z.zselect cond (f a) (f b).
```
JasonGross: I think this belongs in `Section select` in `src/Arithmetic/Core.v`.  @jadephilipoom do you agree?

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+Proof. unfold select, Z.zselect.
+       destruct (cond =? 0); subst; simpl.
+       - rewrite (map_ext _ fst), ListUtil.map_fst_combine, <- H, firstn_all; reflexivity.
+       - rewrite (map_ext _ snd), ListUtil.map_snd_combine, H, firstn_all; reflexivity. Qed.
+
+(* **************************************************************************** *)
+(* Section for lemma relating to bitwise operations and the Z.testbits function *)
+(* **************************************************************************** *)
+
+Lemma land_pow2 a b :
```
JasonGross: This belongs in `src/Util/ZUtil/Land.v`, I believe

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+  apply Z.bits_inj_iff; red; intros; rewrite Z.land_spec.
+  destruct (Z.testbit a b) eqn:E.
+  - destruct (dec (b = n)); subst.
+    + now rewrite E, andb_true_l.
+    + now rewrite Z.pow2_bits_false, andb_false_r.
+  - rewrite Z.testbit_0_l; destruct (dec (b = n)); subst.
+    + now rewrite E, andb_false_l.
+    + now rewrite Z.pow2_bits_false, andb_false_r. Qed.
+
+Lemma testbit_small_neg a b
```
JasonGross: Belongs in `src/Util/ZUtil/Testbit.v`

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+      (Hb : 0 < b) :
+  Z.testbit a b = true.
+Proof.
+  destruct Ha; apply Z_le_lt_eq_dec in H; destruct H.
+  - apply Z.bits_iff_neg with (n:=b) in H0; [assumption|].
+    apply Log2.Z.log2_lt_pow2_alt; try lia.
+  - subst; replace (- 2 ^ b) with ((-1) * 2 ^ b) by ring.
+    rewrite Z.mul_pow2_bits, Z.sub_diag, Z.bit0_odd by lia; reflexivity. Qed.
+
+Lemma land_pow2_small a b
```
JasonGross: Belongs in `src/Util/ZUtil/Land.v`

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+      (Ha : 0 <= a < 2^b) :
+  a &' 2^b = 0.
+Proof. now rewrite land_pow2, Testbit.Z.bits_above_pow2. Qed.
+
+Lemma land_pow2_small_neg a b
```
JasonGross: Belongs in `src/Util/ZUtil/Land.v`

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+      (Ha : - 2^b <= a < 0)
+      (Hb : 0 < b) :
+  a &' 2^b = 2^b.
+Proof. now rewrite land_pow2, testbit_small_neg. Qed.
+
+Lemma lor_add a b (Hand : a &' b = 0) : a ||' b = a + b.
```
JasonGross: Belongs in `src/Util/ZUtil/Lor.v`

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+Proof. rewrite <- Z.lxor_lor, Z.add_nocarry_lxor by assumption; reflexivity. Qed.
+
+Lemma lor_small_neg a b
```
JasonGross: Belongs in `src/Util/ZUtil/Lor.v`

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+      (Ha : - 2^b <= a < 0)
+      (Hb : 0 < b) :
+  2^b ||' a = a.
+Proof.
+  apply Z.bits_inj_iff; red; intros; rewrite Z.lor_spec.
+  destruct (dec (b = n)); subst.
+  - now rewrite (testbit_small_neg a), orb_true_r.
+  - now rewrite Z.pow2_bits_false, orb_false_l. Qed.
+
+Lemma testbit_large a b
```
JasonGross: Belongs in `src/Util/ZUtil/Testbit.v`

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+      (Hb : 0 < b) :
+  Z.testbit a (b - 1) = true.
+Proof.
+  destruct (Z.testbit a (b - 1)) eqn:E; try reflexivity.
+  rewrite Z.testbit_false in * by lia.
+  apply Z.div_exact in E; [|lia].
+  rewrite Div.Z.div_between_1 in E; auto with zarith. 
+  rewrite Pow.Z.pow_mul_base, Z.sub_simpl_r; lia. Qed.
+
+Lemma land_div2 a b (Ha : 0 <= a < 2^(b + 1))  :
```
JasonGross: Belongs in `src/Util/ZUtil/Land.v`

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+      assert ((2 ^ b * (1 + c mod 2 ^ (n - b))) <= 2 ^ b * 2 ^ (n - b)) by (apply Zmult_le_compat_l; lia); lia. }
+    split; [|lia].
+    assert (0 <= 2 ^ b * (c mod 2 ^ (n - b))); try lia; apply Z.mul_nonneg_nonneg; try lia.
+    apply Z.mod_pos_bound; try lia; apply Z.pow_pos_nonneg; lia.
+
+    rewrite (Z.add_comm _ (2 * _)), <- Z.div2_odd, (Z.mul_comm (_ / _) _).
+    replace (2 ^ n) with (2 ^ b * 2 ^ (n - b)).
+    { rewrite <- Z.mul_assoc, <- Z.add_assoc, <- Z.mul_add_distr_l, <- Modulo.Z.div_mod'' by (apply Z.pow_nonzero; lia).
+      ring. }
+    rewrite <- Z.pow_add_r, Zplus_minus; lia. Qed.
```
JasonGross: I think you are looking for `Z.testbit_add_shiftl_high` in `src/Util/ZUtil/Shift.v`.  But please instead add a version of this combined lemma using `Z.shiftl` instead of `Z.add` to `src/Util/ZUtil/Shift.v`, and then the only thing you have to do here is convert `2^b * c` to `c << b` and then you can just apply that lemma.

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+  - now rewrite Pow.Z.base_pow_neg, Z.land_0_r.
+  - rewrite land_pow2, Z.div2_bits, Testbit.Z.bits_above_pow2; lia. Qed.
+
+Lemma testbit_split a b c n
+      (Ha : 0 <= a < 2^b)
+      (Hb : 0 < b)
+      (Hn : 0 <= n) :
+  Z.testbit (a + 2^b * c) n = if (n <? b)
+                              then Z.testbit a n
+                              else Z.testbit c (n - b).
```
JasonGross: Hmm, is there a reason to have this combined lemma rather than converting `2^b * c` to `c << b` at the use site?  (The latter is easier on file organization.)

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+      apply Div.Z.div_nonneg; lia. 
+      intros; specialize (Hz z); tauto.
+    + specialize (Hz a (or_introl eq_refl)); lia. Qed.  
+
+(* ************************************************************ *)
+(* Section for proving arithmetic shifting functionally correct *)
+(* ************************************************************ *)
+
+(* note that this is only true since coq integer division rounds towards -infinity *)
+Lemma arshift_small m a
```
JasonGross: Please move `arshift` to `ZUtil.Definitions.arithmetic_shiftr` and move this to a new file ``src/Util/ZUtil/ArithmeticShiftr.v`

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+    + apply Znot_le_gt in n. rewrite land_pow2_small_neg; try lia.
+      rewrite lor_small_neg; try lia. split.
+      apply Zdiv_le_lower_bound; lia.
+      apply Zdiv_lt_upper_bound; lia.  
+  - lia.
+  - subst; simpl in *.
+    destruct (Z_dec a 0) as [[]|]; subst; try lia; try reflexivity.
+    destruct (Z_dec a (-1)) as [[]|]; subst; try lia; reflexivity. Qed.
+
+Lemma arshift_large m a
```
JasonGross: idem

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+      (Hm : 0 < m) :
+  arshift m a = a / 2 + 2 ^ (m - 1).
+Proof.
+  assert (0 < 2 ^ (m - 1)) by (apply Z.pow_pos_nonneg; lia).
+  unfold arshift.
+  rewrite land_pow2, testbit_large, <- Z.div2_spec, Z.div2_div, lor_add; try lia.
+  rewrite Z.land_comm, land_div2; try lia. 
+  rewrite Z.sub_simpl_r; lia. Qed.
+  
+Lemma arshift_spec m a
```
JasonGross: idem

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+Lemma length_sat_add machine_wordsize n f g
+      (mw0 : 0 < machine_wordsize)
+      (Hf : length f = n)
+      (Hg : length g = n) :
+  length (sat_add machine_wordsize n f g) = n.
+Proof.
+  unfold sat_add, Rows.add.
+  rewrite Rows.length_flatten. reflexivity.
+  apply (uwprops machine_wordsize mw0).
+  intros. inversion H; subst. reflexivity. inversion H0. subst.
```
JasonGross: Please don't use autogenerated names.  Also consider using the `distr_length` tactic and/or `autorewrite with distr_length`, and add lemmas like this to the `distr_length` hint database.  (I might be naming it wrong; there might not be an underscore.)

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+(* ********************************** *)
+
+Lemma length_sat_one machine_wordsize n :
+  length (sat_one machine_wordsize n) = n.
+Proof. apply length_encode_no_reduce. Qed.
+
+Lemma length_sat_zero n :
+  length (sat_zero n) = n.
+Proof. apply repeat_length. Qed.
+(* Lemma length_sat_zero machine_wordsize n : *)
```
JasonGross: Why is this commented out?

src/Arithmetic/JumpDivstep.v
```diff
@@ -0,0 +1,46 @@
+Require Import Inv.
```
JasonGross:
```suggestion
Require Import Crypto.Arithmetic.Inv.
```

src/Arithmetic/JumpDivstep.v
```diff
@@ -0,0 +1,46 @@
...
+Require Import Coq.Bool.Bool.
+Require Import Coq.ZArith.ZArith.
+Require Import Coq.Lists.List.
+Require Import Coq.nsatz.Nsatz.
+Require Import Lia.
```
JasonGross:
```suggestion
Require Import Coq.micromega.Lia.
```

src/Arithmetic/JumpDivstep.v
```diff
@@ -0,0 +1,46 @@
...
+Require Import Crypto.Util.ZUtil.Notations.
+Require Import Crypto.Arithmetic.Partition.
+Require Import Crypto.Util.Decidable.
+
+Import Definitions.
+Import Positional.
+Import WordByWordMontgomery.
+Import Z.
+
+Open Scope Z.
```
JasonGross:
```suggestion
Local Open Scope Z_scope.
```

src/PushButtonSynthesis/TestInv.v
```diff
@@ -0,0 +1,169 @@
...
+          | None => 0
+          end.
+
+Definition decode := from_montgomerymod machine_wordsize n m m'.
+Definition enc0 :=
+  (* Eval compute in encodemod machine_wordsize n m m' 0. *)
+  [0; 0; 0; 0; 0; 0].
+Definition enc2 :=
+  (* Eval compute in encodemod machine_wordsize n m m' 2. *)
+  [3608227726454314319; 13347543502301691909; 6296135691958860625; 10026531341796875211; 7850492651313966083; 1291314412115845772].
```
JasonGross: Why write this out?

src/PushButtonSynthesis/TestInv.v
```diff
@@ -0,0 +1,169 @@
...
+Definition unary_word_out_with_carry :=
+  (Some r[0~>2^machine_wordsize-1], Some r[0~>1])%zrange.
+Definition unary_bit_out :=
+  (Some r[0~>1])%zrange.
+
+Definition saturated_bounds_list : list (option zrange)
+  := Eval compute in List.repeat (Some r[0 ~> 2^machine_wordsize-1]%zrange) n.
+Definition saturated_bounds : ZRange.type.option.interp (base.type.list (base.type.Z))
+  := Some saturated_bounds_list.
+Print saturated_bounds_list.
```
JasonGross: Is this leftover cruft?

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+Definition twosc_pos m a :=
+  dlet b := twosc_opp m a in
+        (b >> (m - 1)) &' 1.
+
+(* Saturated addition of multi-limb integers *)
+Definition sat_add machine_wordsize n f g :=
+  fst (Rows.add (uweight machine_wordsize) n f g).
+
+(* Saturated (logical) right shift *)
+Definition sat_rshift machine_wordsize n f :=
```
jadephilipoom: I'd call this something like `sat_shiftr1`, to make it clear you're only shifting by one bit.

test-inversion-c/bls12_381_q_64.c
```diff
@@ -0,0 +1,3514 @@
...
+  fiat_test_addcarryx_u64(&x20, &x21, x19, x11, x8);
+  uint64_t x22;
+  fiat_test_uint1 x23;
+  fiat_test_addcarryx_u64(&x22, &x23, x21, x9, x6);
+  uint64_t x24;
+  fiat_test_uint1 x25;
+  fiat_test_addcarryx_u64(&x24, &x25, x23, x7, x4);
+  uint64_t x26;
+  fiat_test_uint1 x27;
+  f
```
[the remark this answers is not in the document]
JasonGross: I don't see any literals... @andres-erbsen what are you expecting?

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+(* twos complement evaluation *)
+Notation twosc m a :=
+  (if ((a mod 2 ^ m) <? 2 ^ (m - 1)) then a mod 2 ^ m else a mod 2 ^ m - 2 ^ m).
+
+(* Custom evaluation function for multi-limbg twos complement *)
+Definition eval_twosc machine_wordsize n f :=
+  twosc (machine_wordsize * Z.of_nat n) (eval (uweight machine_wordsize) n f).
+
+(* arithmetic right shift *)
+Definition arshift (m a : Z) :=
```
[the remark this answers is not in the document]
JasonGross: If so, they should get better names, I think.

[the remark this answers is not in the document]
JasonGross: If so, they should get longer names, I think.  It was not clear to me that `twosc` was "twos complement" until just now (I thought it had something to do with the `s-c` encoding we use for primes...)

JasonGross:
The standard way that things here fail to reify is that we don't support `Q`, and some of the weight stuff uses `Q`, iirc.  We could just add support for `Q`, that might make things nicer rather than having to give the numerator and denominator separately.

JasonGross:
> I also can't think of a reason why the operations at https://github.com/mit-plv/fiat-crypto/blob/master/src/Arithmetic/Saturated.v#L798 shouldn't be reifiable -- but perhaps I am missing something here? What happened when you tried to reify it?

The issue is not in fact `Q` (though I made a branch that adds support for reifying `Q` [here](https://github.com/JasonGross/fiat-crypto/tree/support-Q), but idk if it's worth merging, since we don't actually use `Q` anymore.  Thoughts?)
In fact, all of these operations do work fine, as evidenced by adding the following succeeding lines to `CompilersTestCases.v`:
```coq
Require Import Arithmetic.Saturated.
Import Saturated.Rows.
Notation Reify v := (ltac:(let x := Reify v in exact x)) (only parsing).
Definition radd := Reify add.
Definition rsub := Reify sub.
Definition rconditional_add := Reify conditional_add.
Definition rconditional_sub := Reify conditional_sub.
Definition rsub_then_maybe_add := Reify sub_then_maybe_add.
Definition rmul := Reify mul.
Definition radjust_s := Reify adjust_s.
Definition rsat_reduce := Reify sat_reduce.
Definition rrepeat_sat_reduce := Reify repeat_sat_reduce.
Definition rmulmod := Reify mulmod.
Definition rdivmod := Reify divmod.
```

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+
+(* An implementation of the divsteps2 algorithm from "Fast constant-time gcd computation and modular inversion." 
+   by D. J. Bernstein et al. See the file TestInv.v in /PushButtonSynthesis for generation of C-code.
+   For a C-implementation using this generated code to implement modular inversion, see test-inversion-c/test_bls.c. *)
+
+Local Coercion Z.of_nat : nat >-> Z.
+
+Open Scope Z_scope.
+
+Infix "||'" := Z.lor (at level 50) : Z_scope .
```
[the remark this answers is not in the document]
bshvass: Good point. Note that the `|` in `Util/Notations.v` is the unicode character `∣` (0x2223), which confused me. I just changed this.

src/Arithmetic/Inv.v
bshvass: Thanks for point me to `Shift.v`! This lemma was already there (`Z.testbit_add_shiftl_full`).

src/Arithmetic/Inv.v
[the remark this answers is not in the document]
JasonGross: Oh, wow, I did not realize this.  I think we may want to just delete `Reserved Infix "∣" (at level 50).` because otherwise it may conflict with various Ltac notations as well as the notations for inductives and `match`, but changing `∣'` to `|'` should be fine, and I support this.

huitseeker:
Hey @bshvass, let me know if can help out with this or if you mind me just jumping in?

inversion-c/test-inversion.c
```diff
@@ -0,0 +1,42 @@
+#include "inversion.c"
+#include <string.h>
+#include <time.h>
+
+int main() {
+  uint64_t res[LIMBS], out[LIMBS], g[LIMBS], g2[LIMBS], g3[LIMBS];
+  uint8_t a[BYTES];
+  
+  srand(time(0));
```
JasonGross, line 9: If we're going to be running tests, please write the seed to stdout, and allow passing the seed as a command-line argument, so that if it fails, we can reproduce the failing behavior.

inversion-c/test-inversion.c
```diff
@@ -0,0 +1,42 @@
...
+    fiat_test_to_bytes(a,out);
+
+    if (a[0] != 1) {
+      printf("FAIL\n");
+      return 2;
+    }    
+    for (i = 1; i < BYTES; i++) {
+      if (a[i] != 0) {
+	printf("FAIL\n");
+      	return 1;
```
JasonGross, line 34: Please use spaces rather than tabs
```suggestion
        printf("FAIL\n");
        return 1;
```

src/Arithmetic/Core.v
```diff
@@ -1414,6 +1414,13 @@ Module Positional.
       length p = n -> length q = n ->
       length (select cond p q) = n.
     Proof using Type. clear dependent weight. distr_length; omega **. Qed.
+
+    Lemma select_push cond a b f (H : length a = length b) :
+      f (select cond a b) = Z.zselect cond (f a) (f b).
+    Proof. unfold select, Z.zselect.
```
JasonGross, line 1420:
```suggestion
    Proof using Type. unfold select, Z.zselect.
```
(Assuming this is the correct set of assumptions)

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1222 @@
...
+Require Import Crypto.Util.ZUtil.Div.
+Require Import Crypto.Util.ZUtil.Lor.
+Require Import Crypto.Util.ZUtil.Land.
+Require Import Crypto.Util.ZUtil.ArithmeticShiftr.
+
+Import Positional.
+Import ListNotations.
+Import Definitions.
+
+(* An implementation of the divsteps2 algorithm from "Fast constant-time gcd computation and modular inversion."
```
JasonGross, line 29: Nit for better coqdocing
```suggestion
(** An implementation of the divsteps2 algorithm from "Fast constant-time gcd computation and modular inversion." 
```

src/Arithmetic/Inv.v
```diff
@@ -0,0 +1,1466 @@
...
+(* note the corner case condition: when f is exactly 2 to the mw-1'th power, then -f = f and 
+   so checking that -f is negative does not work in that case. 
+   This is not really an issue, since it just requires that our integers are small (which they are)
+   Long term, we would like to add comparison operators to the supported C language *)
+Definition twosc_pos m a :=
+  dlet b := twosc_opp m a in
+        (b >> (m - 1)) &' 1.
+
+(* Saturated addition of multi-limb integers *)
+Definition sat_add machine_wordsize n f g :=
```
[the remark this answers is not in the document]
JasonGross: Alternatively,
```suggestion
Local Definition sat_add machine_wordsize n f g :=
```
will cause the unqualified name to not be available even after `Import`

src/CLI.v
```diff
@@ -531,6 +533,114 @@ Module ForExtraction.
     End __.
   End Parameterized.
 
+  Module WordByWordMontgomeryInversion. 
+    Local Instance api : PipelineAPI
+      := {
+          parse_args (args : list string)
```
JasonGross, line 539: There's a recent change I made that changed the format of `PipelineAPI` to make it a bit more structured, so you'll have to update this, sorry.  (Or I can suggest the appropriate change if you'd like, but I suspect it'll be pretty easy to copy from the word-by-word montgomery api).

src/ExtractionHaskell/test.v
```diff
@@ -0,0 +1,4 @@
+Require Import Crypto.StandaloneHaskellMain.
+
+(*Redirect "/tmp/test.hs"*) Recursive Extraction test.main.
+(* cat /tmp/test.hs.out | sed -f haskell.sed  > ../../test.hs *)
```
JasonGross, line 4: Was this leftover and you accidentally committed it?

src/PushButtonSynthesis/WordByWordMontgomeryDivstepReificationCache.v
```diff
@@ -0,0 +1,102 @@
...
+Import Associational Positional.
+
+Fixpoint sq_and_mul_mod (a : Z) (p : positive) (x : Z) :=
+  let res :=
+  match p with
+  | xO xs => (sq_and_mul_mod (((a mod x) * (a mod x)) mod x) xs x) mod x
+  | xI xs => ((a mod x) * (sq_and_mul_mod (((a mod x) * (a mod x)) mod x) xs x)) mod x
+  | xH => a mod x
+  end in res.
+Require Import Lists.List.
```
JasonGross, line 31: Please move this higher up in the file, above the first definition

src/PushButtonSynthesis/WordByWordMontgomeryDivstepReificationCache.v
```diff
@@ -0,0 +1,102 @@
...
+Compute seq 1 10. Import ListNotations.
+Compute fold_right (fun a b => a - b) 1 []. Compute Z.log2 3.
```
JasonGross, line 33: These should be removed before we merge

src/PushButtonSynthesis/WordByWordMontgomeryDivstepReificationCache.v
```diff
@@ -0,0 +1,102 @@
...
+
+Import Associational Positional.
+
+Fixpoint sq_and_mul_mod (a : Z) (p : positive) (x : Z) :=
+  let res :=
+  match p with
+  | xO xs => (sq_and_mul_mod (((a mod x) * (a mod x)) mod x) xs x) mod x
+  | xI xs => ((a mod x) * (sq_and_mul_mod (((a mod x) * (a mod x)) mod x) xs x)) mod x
+  | xH => a mod x
+  end in res.
```
JasonGross, line 30: Does this belong here, or in some arithmetic file?

bshvass:
Hey @huitseeker! You are more than welcome to jump in if you have suggestions or improvements.

src/PushButtonSynthesis/WordByWordMontgomeryDivstep.v
```diff
@@ -0,0 +1,286 @@
...
+    
+  Definition sdivstep (prefix : string)
+    : string * (Pipeline.ErrorT (list string * ToString.ident_infos))
+    := Eval cbv beta in
+        FromPipelineToString
+          prefix "divstep" divstep
+          (docstring_with_summary_from_lemma!
+             prefix
+             (fun fname : string => ["The function " ++ fname ++ " computes a divstep."]%string)
+             (forall a, valid a)).
```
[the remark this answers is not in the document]
bshvass: Thanks for the offer. I am not sure how to get `strip_bounds_info` and `stringify_rec0` (from `Primitives.v` to recognise and properly convert the correctness properties to strings, so i would appreciate help with that. I have added the correctness properties to `COperationSpecifications` in the latest commit.

src/COperationSpecifications.v
```diff
@@ -540,6 +543,37 @@ Module WordByWordMontgomery.
...
+                                                         then (49 * mbits + 80) / 17
+                                                         else (49 * mbits + 57)/ 17))
+      /\ valid precomp.
+
+    Definition divstep_correct
+               (divstep :
+                  Z -> list Z -> list Z -> list Z -> list Z ->
+                  Z * list Z * list Z * list Z * list Z) : Prop
+      := forall (d : Z) f g v r, let '(d1,f1,g1,v1,r1) := divstep d f g v r in
+        valid v -> valid r ->
```
JasonGross, line 564: The correctness condition stringification does not support let-binders shared between preconditions and post conditions
```suggestion
      := forall (d : Z) f g v r,
        valid v -> valid r ->
         let '(d1,f1,g1,v1,r1) := divstep d f g v r in
```

src/PushButtonSynthesis/WordByWordMontgomeryDivstep.v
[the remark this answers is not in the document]
JasonGross: If you rebase this PR on top of #810 (which I'll be merging into master as soon as the CI passes on it) and commit the suggestion I make in https://github.com/mit-plv/fiat-crypto/pull/670/files#r434769186 , then the stringification routine should work for all three definitions you add.

JasonGross:
Please rebase this on the tip of master, to get rid of the conflicts, if that's not too much toruble (I generally perfer merging PRs by rebasing them rather than with a merge commit). Btw, if you want to submit updates to the ZUtil files as a separate PR, I can merge those in now, so there's less rebuilding to be done when switching between this branch and master.

bshvass:
I believe all requests have been fixed. I don't know if you have time to review the PR, @JasonGross? Thanks.

JasonGross:
Thanks!  I'll make another pass through it, and hopefully this can be merged soon.

src/COperationSpecifications.v
```diff
@@ -552,6 +555,40 @@ Module WordByWordMontgomery.
...
+      (eval (from_montgomery precomp) = ((m - 1) / 2) ^ (if Decidable.dec (mbits < 46)
+                                                         then (49 * mbits + 80) / 17
+                                                         else (49 * mbits + 57)/ 17))
+      /\ valid precomp.
+
+    Definition divstep_correct
+               (divstep :
+                  Z -> list Z -> list Z -> list Z -> list Z ->
+                  Z * list Z * list Z * list Z * list Z) : Prop
+      := Eval cbv [divstep_spec_full] in
```
jadephilipoom, line 576: I'd prefer to remove the compute and simply repeat the definition of `divstep_spec_full`; this is what the other correctness definitions do, and makes the definition more readable.

src/COperationSpecifications.v
```diff
@@ -552,6 +555,40 @@ Module WordByWordMontgomery.
            saturated_bounds
            selectznz.
 
+    (* Bernstein-Yang inversion *)
+    Definition msat_correct
+               (msat : list Z) :=
+      twos_complement_eval msat = m /\
+      valid msat.
+
+    Definition precomp_correct
```
jadephilipoom, line 564: The name "precomp" seems a little vague; if it doesn't have some other name we can use, maybe `divstep_precomp`?

src/Util/Notations.v
```diff
@@ -73,6 +73,7 @@ Reserved Infix ">>" (at level 30, no associativity).
 Reserved Infix ">>>" (at level 30, no associativity).
 Reserved Infix "&'" (at level 50). (* N.B.  If we used '&', it would conflict with [{ a : T & P}] for [sigT] *)
 Reserved Infix "&''" (at level 50).
+Reserved Infix "|'" (at level 50).
```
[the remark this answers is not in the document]
bshvass: The one in line 78 is a different character (∣ vs |). There is also an earlier comment on this, where Jason suggested just removing the one in line 78.

src/Util/ZUtil/Pow.v
```diff
@@ -65,4 +67,63 @@ Module Z.
   Lemma pow_mul_base a b : 0 <= b -> a * a ^ b = a ^ (b + 1).
   Proof. intros; rewrite <-Z.pow_succ_r, <-Z.add_1_r by lia; reflexivity. Qed.
   Hint Rewrite pow_mul_base using zutil_arith : pull_Zpow.
+
+  (* faster modular exponentation for computing (not sure where to put this) *)
```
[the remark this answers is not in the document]
bshvass: There is a `+ 1` making it `(a - 1) * 1 + 1 = a` when the bit is in `1` and `(a - 1) * 0 + 1 = 1` if the bit is `0`. I can add this as a comment.

src/PushButtonSynthesis/InversionReificationCache.v
```diff
@@ -0,0 +1,70 @@
...
+Module Export WordByWordMontgomeryInversion.
+  Import WordByWordMontgomery.WordByWordMontgomery.
+
+  Definition msat bitwidth n m := Partition.partition (uweight bitwidth) n m. (* m in saturated representation *)
+  (* Precomputed value for the inversion algorithm in the montgomery domain (not sure where to put this) *)
+  Definition precompmod bitwidth n m m' :=
+  let bits := (Z.log2 m) + 1 in
+  let i := if bits <? 46 then (49 * bits + 80) / 17 else (49 * bits + 57) / 17 in
+  let k := (m + 1) / 2 in
+  let precomp := (Z.sq_and_multiply_mod k i m) in
```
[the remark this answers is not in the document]
bshvass: I tried using `Z.pow` followed by `Z.modulo`, but the computation hang (presumably because the exponentiation gets too large, the base is of size ~p/2 and the exponent of size ~3 * log p). Using this instead fixed the issue. Now that I think about it though, we could probably just use a simpler power function which reduced after every multiplication (instead of doing the multiplications first).

spitters:
@jadephilipoom 
> Not a blocker for merging, but would it be feasible to create a Coq version of the full inversion algorithm (with loop, as in the C code) and prove it correct?

I thought loops were not supported yet?

bshvass:
> @jadephilipoom
> 
> > Not a blocker for merging, but would it be feasible to create a Coq version of the full inversion algorithm (with loop, as in the C code) and prove it correct?
> 
> I thought loops were not supported yet?

You could still prove a Coq version correct, just not one that is reifiable in fiat.

bshvass:
@jadephilipoom 
> It seemed like this was closer to merging, so I looked over it again. Thanks for the hard work!
>
Thanks for the comments. 
> Not a blocker for merging, but would it be feasible to create a Coq version of the full inversion algorithm (with loop, as in the C code) and prove it correct?

The paper-proof is, unfortunately, not too easy to formalize, but I am currently working on it. I think it is feasible, but it will probably rely on some axioms which fiat does not. 

> 
> I have a naming nitpick also; this algorithm is just called "inversion" in file names (e.g. `Arithmetic/Inv.v`, `PushButtonSynthesis/InversionReificationCache.v`), but the fiat-crypto repo also contains inversion by Fermat's little theorem, and there's some active work that will likely add more FLT-inversion code. So I think it might be a good idea to be specific here. Maybe BY for Bernstein-Yang, as in `BYInv.v` and `BYInversionReificationCache.v`?

I agree and will fix this.

src/Util/ZUtil/Pow.v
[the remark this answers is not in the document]
jadephilipoom: Ah, I see, I eyeballed the parentheses wrong -- that's fine then.

src/PushButtonSynthesis/InversionReificationCache.v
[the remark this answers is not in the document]
jadephilipoom: Okay, I see why it was needed. But it would be nice to avoid adding new code just for the precomputed value (and avoid creating a new top-level synthesized operation just for computing a constant).

What about following the model we used for the modular inverse in Montgomery multiplication: that is, take in the precomputed value as an argument/context variable, and require as a precondition that it has the arithmetically correct value? Then the caller can compute it however they want. You can add the square-and-multiply definition to its own ZUtil file (maybe renaming it something like `modexp` to make it clear that this is modular exponentiation, and not tie the name to the implementation), like we've done with modular inverse (`modinv`), and use that to compute the argument in the PushButtonSynthesis file.

See:
https://github.com/mit-plv/fiat-crypto/blob/9f08fb67c944da8d30673c28f027726faaa1db78/src/PushButtonSynthesis/WordByWordMontgomery.v#L117
 https://github.com/mit-plv/fiat-crypto/blob/9f08fb67c944da8d30673c28f027726faaa1db78/src/PushButtonSynthesis/WordByWordMontgomery.v#L166

jadephilipoom:
> You could still prove a Coq version correct, just not one that is reifiable in fiat.

Yep, this is what I meant. The background is that although loops are *currently* not supported, there's some work in that direction and it would be cool to have the proofs in place when we're ready for them.

spitters:
@jadephilipoom yes, that proof would be interesting, but is rather long, and it should probably not stall this PR.

jadephilipoom:
> @jadephilipoom yes, that proof would be interesting, but is rather long, and it should probably not stall this PR

Agreed; as I said, not a blocker for merging this

src/PushButtonSynthesis/InversionReificationCache.v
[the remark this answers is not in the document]
bshvass: I see the idea and have fixed it now. I have however not added a check, since I'm not sure exactly what equation you could verify. You could check that `modexp a b m = a ^ b mod m`, but this would again be slow.

src/Util/ZUtil/OnesFrom.v
```diff
@@ -0,0 +1,21 @@
...
+
+Require Import Crypto.Util.ZUtil.Tactics.SolveTestbit.
+
+Local Open Scope Z_scope.
+
+Module Z.
+  Lemma ones_from_spec_full m k i
+        (Hk : 0 <= k) :
+    Z.testbit (Z.ones_from m k) i = if (i <? 0) then false else ((m - k) <=? i) && (i <? m).
+  Proof. unfold Z.ones_from; solve_testbit. Qed.
```
JasonGross, line 15: This isn't quite `full`.  Could you write a version that has `if 0 <=? k then ... else ...`?  (Note that `Z.ones x = -1` for `x < 0`, which you can prove via `Z.ones_equiv`.  And `Z.testbit (-1) i = true` for non-negative `i`.)

src/Util/ZUtil/Pow.v
```diff
@@ -1,6 +1,8 @@
 Require Import Coq.ZArith.ZArith.
 Require Import Coq.micromega.Lia.
 Require Import Crypto.Util.ZUtil.Hints.Core.
+Require Import Crypto.Util.ListUtil.
+Require Import Lists.List.
```
JasonGross, line 5:
```suggestion
Require Import Coq.Lists.List.
```

src/Util/ZUtil/ModExp.v
```diff
@@ -0,0 +1,27 @@
...
+    - simpl; rewrite  <- Zmult_mod_idemp_r, <- IHp; reflexivity. Qed.
+
+  Definition modexp a p m := modexp_nat a (Z.to_nat p) m.
+
+  Lemma modexp_correct a p m
+        (Hp : 0 <= p)
+        (Hm : 1 < m) :
+    modexp a p m = a ^ p mod m.
+  Proof.
+    unfold modexp; destruct p; simpl; rewrite modexp_nat_correct, ?Zpower_pos_nat by assumption; first [reflexivity|lia]. Qed.
```
JasonGross, line 26: Here is a version that is asymptotically more efficient for large exponents, and a bit simpler, based on the definition of `Z.pow`:
```suggestion
Local Open Scope Z_scope.

Module Z.
  Definition modexp_pos a p m := Pos.iter (fun z => (a * z) mod m) (1 mod m) p.
  Definition modexp a p m
    := match p with
       | 0 => 1 mod m
       | Z.neg _ => 0
       | Z.pos p => modexp_pos a p m
       end.

  Lemma modexp_pos_correct a p m
    : modexp_pos a p m = Z.pow_pos a p mod m.
  Proof.
    cbv [Z.pow_pos modexp_pos].
    erewrite Pos.iter_swap_gen with (f := fun a => a mod m).
    { reflexivity. }
    { cbv beta; intros; now rewrite Zmult_mod_idemp_r. }
  Qed.

  Lemma modexp_correct a p m
    : modexp a p m = a ^ p mod m.
  Proof.
    cbv [modexp Z.pow]; destruct p;
      now rewrite ?modexp_pos_correct, ?Zmod_0_l.
  Qed.
```

src/Util/ZUtil/Tactics/SolveRange.v
```diff
@@ -0,0 +1,69 @@
...
+
+(** Extension of lia to solve ranges involving bitwise operations and power *)
+
+Ltac solve_range :=
+    repeat match goal with
+    | [ |- context[?a ^ ?b] ] => match goal with
+                               | [ _ : 0 <= a ^ b |- _ ] => fail 1
+                               | [ |- _ ] =>
+                                 assert (0 <= a ^ b) by (apply Z.pow_nonneg; lia)
+                               end
```
JasonGross, line 19: We have a tactic that does this:
```coq
Require Import Crypto.Util.Tactics.UniquePose.
```
and then you can just do
```suggestion
    | [ |- context[?a ^ ?b] ] => unique assert (0 <= a ^ b) by (apply Z.pow_nonneg; lia)
```

src/Util/ZUtil/Tactics/SolveRange.v
```diff
@@ -0,0 +1,69 @@
...
+    | [ |- _ |' _ < 2 ^ _ ] => apply Z.lor_range
+    | [ |- Z.truncating_shiftl _ _ _ < 2 ^ _ ] => apply Z.truncating_shiftl_range
+    | [ |- 0 <= _ / _ ] => apply Z.div_nonneg
+    | [ |- 0 <= _ ^ _ ] => apply Z.pow_nonneg
+    | [ |- 0 <= _ |' _ ] => apply Z.lor_nonneg
+    | [ |- 0 <= _ &' _ ] => apply Z.land_nonneg
+    | [ |- 0 <= _ >> _ ] => apply Z.shiftr_nonneg
+    | [ |- 0 <= _ << _ ] => apply Z.shiftl_nonneg
+    | [ |- 0 <= Z.truncating_shiftl _ _ _ ] => apply Z.truncating_shiftl_nonneg
+    | [ |- 0 < _ ^ _ ] => apply Z.pow_pos_nonneg
```
JasonGross, line 47: Maybe add the `Z.truncating_shiftl` case to `Z.zero_bounds` and then you can
```suggestion
    | _ => progress Z.zero_bounds
```
if you `Require Import Crypto.Util.ZUtil.Tactics.ZeroBounds`

src/Util/ZUtil/Tactics/SolveTestbit.v
```diff
@@ -0,0 +1,131 @@
...
+
+Import Notations.
+
+(* Tactic for solving equality of testbits of operators. testbit_pos assumes all
+testbit indices will be positive, but can probably be faster in those cases *)
+
+Local Lemma mod_pow2_ones a m :
+  a mod 2 ^ m = if (Z_lt_dec m 0) then 0 else a &' Z.ones m.
+Proof. destruct (Z_lt_dec m 0). rewrite Z.pow_neg_r, Zmod_0_r; lia.
+       symmetry; apply Z.land_ones; lia. Qed.
```
JasonGross, line 21: Please move this to `Testbit.v` and make it non-local

src/Util/ZUtil/Tactics/SolveTestbit.v
```diff
@@ -0,0 +1,131 @@
...
+
+Local Lemma bits_1 m :
+  Z.testbit 1 m = if Z.eq_dec m 0 then true else false.
+Proof.
+  destruct (Z.eq_dec m 0); subst. reflexivity.
+  destruct (Z_lt_dec m 0). rewrite Z.testbit_neg_r by lia. reflexivity.
+  rewrite Z.bits_above_log2; simpl; try reflexivity; lia. Qed.
```
JasonGross, line 28: Please move this to `Testbit.v` and make it non-local

src/Util/ZUtil/Tactics/SolveTestbit.v
```diff
@@ -0,0 +1,131 @@
...
+            let E := fresh in
+            destruct (a =? b) eqn:E; [rewrite Z.eqb_eq in E; subst|rewrite Z.eqb_neq in E]
+          | [ H: 0 <= ?a < 2 ^ ?m |- context[Z.testbit ?a ?i] ] =>
+            assert (m <= i) by lia;
+            assert (2 ^ m <= 2 ^ i) by (apply Z.pow_le_mono; lia);
+            rewrite (Z.bits_above_log2 a i) by (try apply Log2.Z.log2_lt_pow2_alt; lia)
+          | [ |- Z.testbit _ _ = Z.testbit _ _ ] =>
+            apply f_equal2; lia
+          | [ |- _ <= _ < _] => auto with zarith
+          end; simpl; subst);
```
JasonGross, line 62: Please don't `simpl` here.  Instead use `cbn [<some explicit list of things you want to unfold>]`.  Tactics with `simpl` in them tend to not scale very well.

src/Util/ZUtil/Tactics/SolveTestbit.v
```diff
@@ -0,0 +1,131 @@
...
+          | [ |- context[?a =? ?b] ] =>
+            let E := fresh in
+            destruct (a =? b) eqn:E; [rewrite Z.eqb_eq in E; subst|rewrite Z.eqb_neq in E]
+          | [ H: 0 <= ?a < 2 ^ ?m |- context[Z.testbit ?a ?i] ] =>
+            assert (m <= i) by lia;
+            assert (2 ^ m <= 2 ^ i) by (apply Z.pow_le_mono; lia);
+            rewrite (Z.bits_above_log2 a i) by (try apply Log2.Z.log2_lt_pow2_alt; lia)
+          | [ |- Z.testbit _ _ = Z.testbit _ _ ] =>
+            apply f_equal2; lia
+          | [ |- _ <= _ < _] => auto with zarith
```
JasonGross, line 61:
```suggestion
          | [ |- _ <= _ < _] => solve [ auto with zarith ]
```
otherwise if you add something below this case it'll never trigger

src/Util/ZUtil/Tactics/SolveTestbit.v
```diff
@@ -0,0 +1,131 @@
...
+  destruct (Z_lt_dec m 0). rewrite Z.testbit_neg_r by lia. reflexivity.
+  rewrite Z.bits_above_log2; simpl; try reflexivity; lia. Qed.
+
+Hint Rewrite Z.bits_opp Z.shiftr_spec Z.shiftl_spec Z.lor_spec Z.land_spec Z.ones_spec_full Z.bits_0 bits_1 andb_false_r andb_true_r orb_false_r orb_true_r mod_pow2_ones : testbit_pos_rewrite.
+
+Hint Rewrite <- Z.ones_equiv : testbit_pos_rewrite.
+
+Ltac solve_pos_testbit :=
+  repeat (match goal with
+          | [ |- _ ] => progress autorewrite with testbit_pos_rewrite
```
JasonGross, line 36:
```suggestion
          | _ => progress autorewrite with testbit_pos_rewrite
```

src/Util/ZUtil/Tactics/SolveTestbit.v
```diff
@@ -0,0 +1,131 @@
...
+  apply Z.bits_opp; lia. Qed.
+
+Local Lemma pow2_bits_full m i :
+  Z.testbit (2 ^ m) i =
+  if (Z_lt_dec m 0) then false else if (Z.eq_dec i m) then true else false.
+Proof.
+  destruct (Z_lt_dec m 0); [now rewrite Z.pow_neg_r, Z.bits_0|].
+  destruct (Z.eq_dec i m); subst.
+  - apply Z.pow2_bits_true; lia.
+  - apply Z.pow2_bits_false; lia. Qed.
```
JasonGross, line 90: All of these belong as non-local lemmas in Testbit.v, I believe

src/Util/ZUtil/Tactics/SolveTestbit.v
```diff
@@ -0,0 +1,131 @@
...
+          | [ H: 0 <= ?a < 2 ^ ?m |- context[Z.testbit ?a ?i] ] =>
+            assert (m <= i) by lia;
+            assert (2 ^ m <= 2 ^ i) by (apply Z.pow_le_mono; lia);
+            rewrite (Z.bits_above_log2 a i) by (try apply Log2.Z.log2_lt_pow2_alt; lia)
+          | [ |- Z.testbit _ _ = Z.testbit _ _ ] =>
+            apply f_equal2; lia
+          | [ |- false = _ ] => rewrite Z.testbit_neg_r by lia
+          | [ |- _ = false ] => rewrite Z.testbit_neg_r by lia
+          | [ |- _ <= _ < _] => auto with zarith
+          end; simpl; subst);
```
JasonGross, line 127: Same comment about not using `simpl`, but using `cbn` with a whitelist instead

src/Util/ZUtil/Tactics/SolveTestbit.v
```diff
@@ -0,0 +1,131 @@
...
+            destruct (a =? b) eqn:E; [rewrite Z.eqb_eq in E; subst|rewrite Z.eqb_neq in E]
+          | [ H: 0 <= ?a < 2 ^ ?m |- context[Z.testbit ?a ?i] ] =>
+            assert (m <= i) by lia;
+            assert (2 ^ m <= 2 ^ i) by (apply Z.pow_le_mono; lia);
+            rewrite (Z.bits_above_log2 a i) by (try apply Log2.Z.log2_lt_pow2_alt; lia)
+          | [ |- Z.testbit _ _ = Z.testbit _ _ ] =>
+            apply f_equal2; lia
+          | [ |- false = _ ] => rewrite Z.testbit_neg_r by lia
+          | [ |- _ = false ] => rewrite Z.testbit_neg_r by lia
+          | [ |- _ <= _ < _] => auto with zarith
```
JasonGross, line 126:
```suggestion
          | [ |- _ <= _ < _] => solve [ auto with zarith ]
```

src/PushButtonSynthesis/WordByWordMontgomeryReificationCache.v
```diff
@@ -274,9 +274,10 @@ Local Opaque reified_%s_gen. (* needed for making [autorewrite] not take a very
          SuchThat (is_reification_of reified_one_gen onemod)
          As reified_one_gen_correct.
   Proof.
-    (* Time cache_reify (). *)
+    Time cache_reify ().
+    (* Note that the following speed up breaks compilation of binaries /benjamin *)
     (* we do something faster *)
-    Time cache_reify_faster_1arg ().
+    (* Time cache_reify_faster_1arg ().  *)
```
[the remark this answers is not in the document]
bshvass: I do need 1 in the montgomery domain, if that is what you mean? If you use cache_reify_faster_1arg, then compilation breaks (at least on my machine)

src/Util/ZUtil/ModExp.v
bshvass: Thanks for the suggestion!

bshvass:
Thanks for the review @JasonGross! I believe I have addressed all suggestions in my latest commit.

src/PushButtonSynthesis/WordByWordMontgomeryReificationCache.v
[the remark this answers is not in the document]
JasonGross: It seemed to me like you were changing from `cache_reify_faster_1arg` to `cache_reify` on a definition that already existed and worked fine with `cache_reify_faster_1arg`.  I'm aware that this tactic does indeed cause compilation failures in some cases, but I'm a bit confused by a compilation failure suddenly showing up in an unchanged definition, and am wondering what I'm missing.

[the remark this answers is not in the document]
bshvass: Ah, I believe it worked previously since `["one", sone]` wasn't in `known_functions`
https://github.com/mit-plv/fiat-crypto/blob/6ef489d015c527a7504b9e6aac961fbdd1f80260/src/PushButtonSynthesis/WordByWordMontgomery.v#L903-L914
That is, both having `sone` as a known function and using `cache_reify_faster_1arg` breaks compilation

inversion-c/p224_32.c
```diff
@@ -0,0 +1,63 @@
...
+  WORD h[LIMBS];
+  if (f[SAT_LIMBS - 1] >> (WORDSIZE - 1)) {
+    OPP(h, v);
+    for (int k = 0; k < LIMBS; k++) v[k] = h[k];
+  }
+
+  MUL(out, v, precomp);
+
+  return;
+}
```
JasonGross, line 63: Why not factor this out into an `inversion_template.c` file which you can then include into all of these C files so that you don't need to duplicate the code?  (Not necessary for merging this PR, though.)  And similarly for the test files.

src/Util/ZUtil/Definitions.v
```diff
@@ -119,4 +119,41 @@ Module Z.
...
+  Definition sign_extend old_m new_m a :=
+    dlet q := Z.zselect (sign_bit old_m a) 0 (ones_at old_m (new_m - old_m)) in
+          q |' a.
+
+  Definition arithmetic_shiftr m a k :=
+    dlet q := Z.zselect (sign_bit m a) 0 (ones_from m k) in
+          q |' (a >> k).
+
+  Definition twos_complement m a :=
+    (if ((a mod 2 ^ m) <? 2 ^ (m - 1)) then a mod 2 ^ m else a mod 2 ^ m - 2 ^ m).
```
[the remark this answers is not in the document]
JasonGross: Would you be willing to merge this comment, so that we have a record of this?

JasonGross (changes requested):
Sorry again for the delay (thesis writing is taking much of my time and energy), and thanks for bearing with me.  Assuming all changes were addressed, this looks good to me once the CI passes.  The current issues are:
1. `fiat_p256_one` is a name conflict with a BoringSSL symbol; I'd suggest just using the constant inline wherever you need it, or else renaming it (e.g., to `set_one` or `assign_one` or similar)
The error is
> `../crypto/fipsmodule/ec/p256.c:50:30: error: ‘fiat_p256_one’ redeclared as different kind of symbol
> static const fiat_p256_felem fiat_p256_one = {0x1, 0xffffffff00000000,
Alternatively, we can ask @davidben if a definition like
```c
static void fiat_p256_one(uint64_t out1[4]) {
  out1[0] = 0x1;
  out1[1] = UINT64_C(0xffffffff00000000);
  out1[2] = UINT64_C(0xffffffffffffffff);
  out1[3] = UINT32_C(0xfffffffe);
}
```
is enough for BoringSSL, or how to make the files play nicely together.  (My inclination, though, is to just rename it.)
2. It looks like you forgot to regenerate the Go, Java, JSON, and Rust files (just run `make` and it'll build them all)

bshvass:
I am encountering a problem with synthesis of bedrock2-code (all other languages seem to synthesize). I get the error 

`Error occured during translation to bedrock2. This is likely because a part of the input expression either had unsupported integer types (bedrock2 requires that all integers have the same size) or contained an unsupported operation.`

How exactly is this to be interpreted? Does it mean that you cannot have functions that take a word-sized integer and a multi-limb-sized integer and return a pair of such? Or simply that all integer-types must be the same.

I tried simplifying the function (`divstep`) to simply be the identity, but the issue persists. I have included the generated .tmp-file below.

```
check_args
/* Autogenerated: src/ExtractionOCaml/bedrock2_word_by_word_montgomery --lang bedrock2 --static --no-wide-int --widen-carry --widen-bytes --split-multiret --no-select p256 64 '2^256 - 2^224 + 2^192 + 2^96 - 1' mul square add sub opp from_montgomery to_montgomery nonzero selectznz to_bytes from_bytes one msat divstep divstep_precomp */
/* curve description: p256 */
/* requested operations: mul, square, add, sub, opp, from_montgomery, to_montgomery, nonzero, selectznz, to_bytes, from_bytes, one, msat, divstep, divstep_precomp */
/* m = 0xffffffff00000001000000000000000000000000ffffffffffffffffffffffff (from "2^256 - 2^224 + 2^192 + 2^96 - 1") */
/*                                                                    */
/* NOTE: In addition to the bounds specified above each function, all */
/*   functions synthesized for this Montgomery arithmetic require the */
/*   input to be strictly less than the prime modulus (m), and also   */
/*   require the input to be in the unique saturated representation.  */
/*   All functions also ensure that these two properties are true of  */
/*   return values.                                                   */
/*  */
/* Computed values: */
/* eval z = z[0] + (z[1] << 64) + (z[2] << 128) + (z[3] << 192) */
/* bytes_eval z = z[0] + (z[1] << 8) + (z[2] << 16) + (z[3] << 24) + (z[4] << 32) + (z[5] << 40) + (z[6] << 48) + (z[7] << 56) + (z[8] << 64) + (z[9] << 72) + (z[10] << 80) + (z[11] << 88) + (z[12] << 96) + (z[13] << 104) + (z[14] << 112) + (z[15] << 120) + (z[16] << 128) + (z[17] << 136) + (z[18] << 144) + (z[19] << 152) + (z[20] << 160) + (z[21] << 168) + (z[22] << 176) + (z[23] << 184) + (z[24] << 192) + (z[25] << 200) + (z[26] << 208) + (z[27] << 216) + (z[28] << 224) + (z[29] << 232) + (z[30] << 240) + (z[31] << 248) */

#include <stdint.h>
#include <memory.h>

// LITTLE-ENDIAN memory access is REQUIRED
// the following two functions are required to work around -fstrict-aliasing
static inline uintptr_t _br2_load(uintptr_t a, size_t sz) {
  uintptr_t r = 0;
  memcpy(&r, (void*)a, sz);
  return r;
}

static inline void _br2_store(uintptr_t a, uintptr_t v, size_t sz) {
  memcpy((void*)a, &v, sz);
}




In fiat_p256_divstep:
Stringification failed on the syntax tree:
(λ x1 x2 x3 x4 x5,
  expr_let x6 := x2[0] (* : uint64_t *) in
  expr_let x7 := x2[1] (* : uint64_t *) in
  expr_let x8 := x2[2] (* : uint64_t *) in
  expr_let x9 := x2[3] (* : uint64_t *) in
  expr_let x10 := x2[4] (* : uint64_t *) in
  expr_let x11 := x3[0] (* : uint64_t *) in
  expr_let x12 := x3[1] (* : uint64_t *) in
  expr_let x13 := x3[2] (* : uint64_t *) in
  expr_let x14 := x3[3] (* : uint64_t *) in
  expr_let x15 := x3[4] (* : uint64_t *) in
  expr_let x16 := x4[0] (* : uint64_t *) in
  expr_let x17 := x4[1] (* : uint64_t *) in
  expr_let x18 := x4[2] (* : uint64_t *) in
  expr_let x19 := x4[3] (* : uint64_t *) in
  expr_let x20 := x5[0] (* : uint64_t *) in
  expr_let x21 := x5[1] (* : uint64_t *) in
  expr_let x22 := x5[2] (* : uint64_t *) in
  expr_let x23 := x5[3] (* : uint64_t *) in
  ((uint64_t, ??, ??, ??)((uint64_t, ??, ??)((uint64_t, ??)((uint64_t)x1, (uint64_t)x6 :: (uint64_t)x7 :: (uint64_t)x8 :: (uint64_t)x9 :: (uint64_t)x10 :: []), (uint64_t)x11 :: (uint64_t)x12 :: (uint64_t)x13 :: (uint64_t)x14 :: (uint64_t)x15 :: []), (uint64_t)x16 :: (uint64_t)x17 :: (uint64_t)x18 :: (uint64_t)x19 :: []), (uint64_t)x20 :: (uint64_t)x21 :: (uint64_t)x22 :: (uint64_t)x23 :: [])
)
ERROR-CONTAINING OUTPUT:
/*
 * Input Bounds:
 *   in0: [0x0 ~> 0xffffffffffffffff]
 *   in1: [[0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff]]
 *   in2: [[0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff]]
 *   in3: [[0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff]]
 *   in4: [[0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff]]
 * Output Bounds:
 *   out0: [0x0 ~> 0xffffffffffffffff]
 *   out1: [[0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff]]
 *   out2: [[0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff]]
 *   out3: [[0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff]]
 *   out4: [[0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff]]
 */
uintptr_t internal_fiat_p256_divstep(uintptr_t out1, uintptr_t out2, uintptr_t out3, uintptr_t out4, uintptr_t in0, uintptr_t in1, uintptr_t in2, uintptr_t in3, uintptr_t in4) {
  uintptr_t x18, x0, x19, x1, x20, x2, x21, x3, x22, x4, x23, x5, x24, x6, x25, x7, x26, x8, x27, x9, x28, x10, x29, x11, x30, x12, x31, x13, x32, x14, x33, x15, x34, x16, x35, x17, ERROR, out0, x36, x37, x38, x39, x40;
  /*skip*/
  x0 = _br2_load((in1)+((uintptr_t)0ULL), sizeof(uintptr_t));
  x1 = _br2_load((in1)+((uintptr_t)8ULL), sizeof(uintptr_t));
  x2 = _br2_load((in1)+((uintptr_t)16ULL), sizeof(uintptr_t));
  x3 = _br2_load((in1)+((uintptr_t)24ULL), sizeof(uintptr_t));
  x4 = _br2_load((in1)+((uintptr_t)32ULL), sizeof(uintptr_t));
  /*skip*/
  x5 = _br2_load((in2)+((uintptr_t)0ULL), sizeof(uintptr_t));
  x6 = _br2_load((in2)+((uintptr_t)8ULL), sizeof(uintptr_t));
  x7 = _br2_load((in2)+((uintptr_t)16ULL), sizeof(uintptr_t));
  x8 = _br2_load((in2)+((uintptr_t)24ULL), sizeof(uintptr_t));
  x9 = _br2_load((in2)+((uintptr_t)32ULL), sizeof(uintptr_t));
  /*skip*/
  x10 = _br2_load((in3)+((uintptr_t)0ULL), sizeof(uintptr_t));
  x11 = _br2_load((in3)+((uintptr_t)8ULL), sizeof(uintptr_t));
  x12 = _br2_load((in3)+((uintptr_t)16ULL), sizeof(uintptr_t));
  x13 = _br2_load((in3)+((uintptr_t)24ULL), sizeof(uintptr_t));
  /*skip*/
  x14 = _br2_load((in4)+((uintptr_t)0ULL), sizeof(uintptr_t));
  x15 = _br2_load((in4)+((uintptr_t)8ULL), sizeof(uintptr_t));
  x16 = _br2_load((in4)+((uintptr_t)16ULL), sizeof(uintptr_t));
  x17 = _br2_load((in4)+((uintptr_t)24ULL), sizeof(uintptr_t));
  /*skip*/
  /*skip*/
  x18 = x0;
  x19 = x1;
  x20 = x2;
  x21 = x3;
  x22 = x4;
  x23 = x5;
  x24 = x6;
  x25 = x7;
  x26 = x8;
  x27 = x9;
  x28 = x10;
  x29 = x11;
  x30 = x12;
  x31 = x13;
  x32 = x14;
  x33 = x15;
  x34 = x16;
  x35 = x17;
  x36 = ERROR;
  x37 = ERROR;
  /*skip*/
  x38 = ERROR;
  /*skip*/
  x39 = ERROR;
  /*skip*/
  x40 = ERROR;
  /*skip*/
  out0 = x36;
  _br2_store((out1)+((uintptr_t)0ULL), x37, sizeof(uintptr_t));
  /*skip*/
  _br2_store((out2)+((uintptr_t)0ULL), x38, sizeof(uintptr_t));
  /*skip*/
  _br2_store((out3)+((uintptr_t)0ULL), x39, sizeof(uintptr_t));
  /*skip*/
  _br2_store((out4)+((uintptr_t)0ULL), x40, sizeof(uintptr_t));
  /*skip*/
  return out0;
}


Error occured during translation to bedrock2. This is likely because a part of the input expression either had unsupported integer types (bedrock2 requires that all integers have the same size) or contained an unsupported operation.
```

JasonGross:
This is a question for @jadephilipoom , but my current guess is that the bedrock2 output doesn't support functions that return multiple arrays/lists.  @jadephilipoom can you confirm?

I'd be fine with adjusting the Makefile to simply not generate divstep for the bedrock2 output, while we're working on getting it supported, so we can merge this sooner.  If you want help with doing this, let me know (I expect it's about a 5--10 minute project for me to adjust the Makefile.)

jadephilipoom:
> This is a question for @jadephilipoom , but my current guess is that the bedrock2 output doesn't support functions that return multiple arrays/lists. @jadephilipoom can you confirm?

No, it should support multiple arrays in the return values. Although it wouldn't shock me if there were errors, since none of the other operations test that functionality. If it's changed from `fun (x1 : Z) (x2 x3 x4 x5 : list Z) => (x1, x2, x3, x4, x5)` to `fun (x1 : Z) (x2 x3 x4 x5 : list Z) => x1`, does the issue persist?

bshvass:
@jadephilipoom Both `fun (x1 : Z) (x2 x3 x4 x5 : list Z) => x1` and `fun (x1 : Z) (x2 x3 x4 x5 : list Z) => x2` synthesize correctly, but `fun (x1 : Z) (x2 x3 x4 x5 : list Z) => (x1, x2)` and `fun (x1 : Z) (x2 x3 x4 x5 : list Z) => (x2, x3)` both fail. On the other hand `fun (x1 x2 : Z) (x3 .. : listZ) => (x1, x2)` also does not synthesize correctly, which doesn't really agree with the fact that addcarryx and mulx synthesizes properly?
@JasonGross It should be disabled in my latest commit.

JasonGross (approved):
Sorry, I just saw your updates now.  Thanks for doing all this working, and bearing with us through the long review process!

PR #672, opened 2020-03-06
Update the Rust printer to use the 128-bits type

huitseeker:
128-bits integers are available [since Rust 1.26](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1260-2018-05-10)


JasonGross:
Thanks!

Issue #627, opened 2020-01-06
Rust output uses 128-bit integers in places they are not needed


JasonGross:
This is causing issues in the Go output (which doesn't support 128-bit integer infix arithmetic), which I'm basing on Rust.

Consider the difference in `subborrowx_u51` between C and Rust:
https://github.com/mit-plv/fiat-crypto/blob/ec9a757aba9ebafc43f069d2eeeeb623eb342d4d/curve25519_64.c#L59
https://github.com/mit-plv/fiat-crypto/blob/ec9a757aba9ebafc43f069d2eeeeb623eb342d4d/fiat-rust/src/curve25519_64.rs#L61

I believe this is due to the difference in how binop casts are computed in C vs Rust:
https://github.com/mit-plv/fiat-crypto/blob/ec9a757aba9ebafc43f069d2eeeeb623eb342d4d/src/Stringification/C.v#L390-L413
https://github.com/mit-plv/fiat-crypto/blob/ec9a757aba9ebafc43f069d2eeeeb623eb342d4d/src/Stringification/Rust.v#L278-L295

I haven't tracked down what exactly is going wrong, but I'm creating this issue to track it.

JasonGross:
I've refactored things a bit to make the operations clearer.  Probably we can do the thing in Rust that I set up for Go, where we cast early if the operations commute with mod: https://github.com/mit-plv/fiat-crypto/blob/7b8225838744df7546f01fc34c18138698fafe51/src/Stringification/Go.v#L300-L316

huitseeker:
@JasonGross I do have a working solution, which instead of pre-casting to the desired type in case of commutation modulo, pre-casts to the lub of the desired type with the encountered types. 

Branch here: https://github.com/huitseeker/fiat-crypto/tree/fix_rust_bw
(passes various downstream tests)

I'll rebase on your recent bound relaxation, check and open a PR.

Beware: the fact Go casts to the desired type may indicate a bug in Go printing, I'm not suficiently familiar with the language (and its possible implicit casts) to tell.

JasonGross:
Is it possible to add the relevant rust tests to this repo somehow?

huitseeker:
@JasonGross not without incurring a more significant burden than you want to import (and maintain) here: the supply chain is pretty complicated, as they rely on the calibra fork of the curve5519 repo:
https://github.com/calibra/curve25519-dalek

[...]

PR #675, opened 2020-03-09
Fix superfluous rust casts in commutation cases (#627)

huitseeker:
This addresses #627 partially, in the sense that it fixes the issue on Rust, but may reveal an issue with the Go printer.

The Rust printer does not take into account that some mod bin ops commute with mod, which in Go allows to cast them early :
https://github.com/mit-plv/fiat-crypto/blob/7b8225838744df7546f01fc34c18138698fafe51/src/Stringification/Go.v#L300-L316

Hence, Rust inserts per-operand upper bound casts for correctness, which occasionally (substraction) lead to 128-bit use (see original issue). 

However, it is incorrect in Rust ([counterexample](https://github.com/mit-plv/fiat-crypto/issues/627#issuecomment-591014018)) to use a direct cast to `desired_type`, as it results in a type that's too small when an upcast is needed to fit the operands. The fix, applied in the present PR, is to cast to the lub of the operands with the desired type.

Since the current PR does not modify the Go printer, this may require a follow-up fix to address the issue in Go, if it presents itself there as well.

This branch was tested in curve25519 unit tests, test branch [here](https://github.com/huitseeker/curve25519-dalek/tree/rust_promotion_bw-2).


huitseeker:
Oops, not what I intended to open — this is too wide. Will fix and reopen

PR #676, opened 2020-03-09
"Fix" superfluous rust casts in commutative-under-mod cases

huitseeker:
This addresses #627 partially, in the sense that I think it fixes the issue on Rust, but may reveal an issue with the Go printer.

The Rust printer does not take into account that some mod bin ops commute with mod, which in Go allows to cast them early :
https://github.com/mit-plv/fiat-crypto/blob/7b8225838744df7546f01fc34c18138698fafe51/src/Stringification/Go.v#L300-L316

Hence, Rust inserts per-operand upper bound casts for correctness, which occasionally (substraction) lead to 128-bit use (see original issue). 

However, it does not work in Rust ([counterexample](https://github.com/mit-plv/fiat-crypto/issues/627#issuecomment-591014018)) to use an unbounded operand cast to `desired_type`, ~~as it can result in a type that's too small to fit the output of the binop. The present PR uses the lub of operands and `desired_type` as a return bound.~~
=> see discussion below: substraction implementaiton assumes wrapping semantics, which Rust does not allow.

```ocaml
       => let ct := ToString.int.union t1 t2 in
          let desired_type' := Some (ToString.int.union ct desired_type) in
          if bin_op_commutes_with_mod_pow2 idc
          then
            (desired_type', (Some desired_type, Some desired_type))
```

*Since the current PR does not modify the Go printer, this may require a follow-up fix to address the issue in Go, if it presents itself there as well*.

This branch was tested in curve25519 unit tests, test branch [here](https://github.com/huitseeker/curve25519-dalek/tree/rust_promotion_bw-2).

(2d iteration of #675 )


JasonGross:
I wonder if the issue with the Go code is that the operations only commute with unsigned modulo.  I'm going to go investigate this shortly.

src/Stringification/Rust.v
```diff
@@ -231,13 +231,31 @@ Module Rust.
...
        => match desired_type with
           | Some desired_type
             => let ct := ToString.int.union t1 t2 in
-               let desired_type' := Some (ToString.int.union ct desired_type) in
+              let desired_type' := Some (ToString.int.union ct desired_type) in
+              if bin_op_commutes_with_mod_pow2 idc
+              then
+                (* these operations commute with mod under modulo, so we just *)
+                (* pre-cast them to their upper bound with the desired type   *)
+                (desired_type', (Some desired_type, Some desired_type))
```
JasonGross: This casts the individual operands to the desired type, and then casts the result up to the lub.  I think this is neither right nor sound, (and part of that is my fault); I'm going to submit an alternative that should hopefully fix the issue.

JasonGross: Ah, no, my analysis is incorrect, nevermind

JasonGross:
Ah, I think I see what's going on here.  Well, almost.  The issue is that the suggestion of https://github.com/mit-plv/fiat-crypto/issues/627#issuecomment-584289241 assumes that signed overflow/underflow is well-defined to be wrapping behavior.  However, Rust sometimes requires some operations to not underflow/overflow.  Is this analysis correct?  Is it just signed subtraction?  Both Go and Java seem to be okay with ignoring overflow, so I don't think there's any issue with the Java/Go code.

src/Stringification/Rust.v
JasonGross: I'm a bit confused how the current design is better.  It seems like it's just putting the `i128` in different places.  Since in general the casts in fiat-crypto tend to guarantee lack of overflow, it seems that
```suggestion
                (Some desired_type, (Some desired_type, Some desired_type))
```
might be better?  Can you check if this works?

JasonGross:
> I wonder if the issue with the Go code is that the operations only commute with unsigned modulo. I'm going to go investigate this shortly.

This is not, in fact, the case

src/Stringification/Rust.v
huitseeker: :man_facepalming: indeed! Done.

huitseeker:
Indeed, Rust indeed [disallows wrapping semantics](https://github.com/rust-lang/rfcs/blob/master/text/0560-integer-overflow.md). It is possible to implement this behavior by using the [explicit types that correspond to it](https://doc.rust-lang.org/std/num/struct.Wrapping.html).

[Test output](https://gist.github.com/huitseeker/619291d51abc4174b5459381fc2a7f55) contains tests that directly relate to the backend arithmetic, which I've linked below:

failures:
    [constants::test::test_eight_torsion](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/constants.rs#L98-L104)
    [constants::test::test_four_torsion](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/constants.rs#L107-L113)
    [constants::test::test_sqrt_ad_minus_one](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/constants.rs#L169-L174)
    [constants::test::test_sqrt_constants_sign](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/constants.rs#L134-L140)
    [constants::test::test_sqrt_minus_one](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/constants.rs#L126-L131)
    [constants::test::test_two_torsion](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/constants.rs#L116-L122)
    [field::test::a_invert_vs_inverse_of_a_constant](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/field.rs#L335-L341)
    [field::test::a_mul_a_vs_a_squared_constant](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/field.rs#L314-L318)
    [field::test::a_p58_vs_ap58_constant](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/field.rs#L398-L402)
    [field::test::a_square2_vs_a_squared_constant](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/field.rs#L328-L332)
    [field::test::a_square_vs_a_squared_constant](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/field.rs#L321-L325)
    [field::test::batch_invert_a_matches_nonbatched](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/field.rs#L344-L356)
    [field::test::batch_invert_empty](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/field.rs#L457-L459)
    [field::test::conditional_negate](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/field.rs#L430-L440)
    [field::test::equality](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/field.rs#L405-L410)
    [field::test::from_bytes_highbit_is_ignored](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/field.rs#L421-L427)
    [field::test::sqrt_ratio_behavior](https://github.com/dalek-cryptography/curve25519-dalek/blob/805cdbfd3c0aa5c8a2ea11dcc0f002807fd617e5/src/field.rs#L359-L395)

fiat-rust/src/curve25519_64.rs
```diff
@@ -56,11 +54,11 @@ pub fn fiat_25519_addcarryx_u51(out1: &mut u64, out2: &mut fiat_25519_u1, arg1:
...
 #[inline]
 pub fn fiat_25519_subborrowx_u51(out1: &mut u64, out2: &mut fiat_25519_u1, arg1: fiat_25519_u1, arg2: u64, arg3: u64) -> () {
-  let x1: i64 = ((((((arg2 as i128) - (arg1 as i128)) as i64) as i128) - (arg3 as i128)) as i64);
+  let x1: i64 = (((arg2 as i64) - (arg1 as i64)) - (arg3 as i64));
   let x2: fiat_25519_i1 = ((x1 >> 51) as fiat_25519_i1);
-  let x3: u64 = (((x1 as i128) & (0x7ffffffffffff as i128)) as u64);
+  let x3: u64 = ((x1 as u64) & 0x7ffffffffffff);
   *out1 = x3;
-  *out2 = (((0x0 as fiat_25519_i2) - (x2 as fiat_25519_i2)) as fiat_25519_u1);
+  *out2 = (0x0 - (x2 as fiat_25519_u1));
```
JasonGross: The panic occurs at this line, presumably because `x2` is an `i8` with value `-1` which becomes a `u8` with value `2^8-1` which then underflows a `u8` when we try to subtract it from 0.

JasonGross:
One possible solution:  Should we just emit all of the types to be `Wrapping`?  Or we could keep the needless casts to int128

huitseeker:
> Did you test this version with the test-suite? It looks like it should still panic.

Oh boy, yes, quite a bit of mixups in my branches! Thanks for catching this!

fiat-rust/src/curve25519_64.rs
huitseeker: Yes! I am managing a half-dozen branches for demonstration purposes and messed up one of them.

huitseeker:
> One possible solution: Should we just emit all of the types to be `Wrapping`? Or we could keep the needless casts to int128

A large number of cryptographic libraries in the Rust ecosystem, including the dalek suite, have a [`no_std`](https://rust-embedded.github.io/book/intro/no-std.html) option for running without the standard library in embedded environments. 

If fiat-crypto was to generate code that imports `std::num::Wrapping` rather than primitive integers, it would severely limit the willingness of any downstream project to integrate this generated code.

With that said, it may form an alternate generation target, in the future.

huitseeker:
OK, took me a while to get my ducks in a row: I was actually tripped by cargo's cache in testing.

I resolved to hard-coding specific fiat-crypto commits in [my testing branch](https://github.com/huitseeker/curve25519-dalek/tree/rust_promotion_bw-2). Hence 0cd2bfba4265dfcda746095a0e09b914c52d2dc1 which pre-casts each operand to the lub works out:
```
huitseeker@pegasus➜huitseeker/tmp/curve25519-dalek(rust_promotion_bw-2)» grep huitseeker Cargo.toml                                                                                                                                                                   [14:33:43]
fiat-crypto = { git="https://github.com/huitseeker/fiat-crypto/", rev = "0cd2bfba4265dfcda746095a0e09b914c52d2dc1", optional = true}
huitseeker@pegasus➜huitseeker/tmp/curve25519-dalek(rust_promotion_bw-2)» cargo test -q --features="std fiat_u64_backend" --no-default-features                                                                                                                        [14:34:02]

running 102 tests
.................................................................................................... 100/102
..
test result: ok. 102 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

However, as you mentioned, it's just putting the i128 in different places.

JasonGross:
> => Shall we close this 

Yes, let's

> and #627 as won't fix (the 128-bit members are, well necessary),

Maybe.  The panic, as far as I can tell, isn't on the 128 bit numbers, but on the way we negate a signed number to turn it into an unsigned one.  There's actually an alternative fix that will take somewhat more work on my end, but is doable (eventually).  The real issue here is that by the time we get to cast-insertion, we've lost the information about what the numbers are actually bounded by.  So in Java and Go (but apparently not Rust), we can just take advantage of the fact that signed overflow wraps and that operations commute with modulo to cast things early.  However, the real solution to this, I think, will be to parameterize the bounds-analysis-and-cast-insertion-pass over a language-specific description of how to insert casts, and then the stringification phase can just transcribe the casts.  This will let us take advantage of early casting in languages like Go and Java (this might actually allow Java to generate native 64-bit code for curve25519 even though it has no 128-bit integers), while ensuring that we don't rely on signed wrapping behavior in Rust.  I'll make an issue for this.

> and open an issue for having an alternate transpilation target to Rust + Wrapping?

Sure

PR #683, opened 2020-03-12
Parameterize the main functions on supported langs

JasonGross:
This should allow extremely easy adaptation of the binaries to support
bedrock2.  Namely, we just need to inhabit `supported_languagesT := list
(string * ToString.OutputLanguageAPI)` with a list containing
`"bedrock2"` paired with a way of converting PHOAS ASTs to strings, and
then we can do something like
```coq
Require Import Coq.Lists.List.
Require Import Coq.Strings.String.
Require Import Crypto.CLI.
Require Import Crypto.StandaloneOCamlMain.
Require Import Crypto.Bedrock2.Stringification.
Local Open Scope string_scope.
Local Open Scope list_scope.

Local Instance bedrock2_supported_languages : ForExtraction.supported_languagesT
  := ForExtraction.default_supported_languages
       ++ [("bedrock2", OutputBedrock2API)].

Module UnsaturatedSolinas.
  Definition main : unit
    := main_gen ForExtraction.UnsaturatedSolinas.PipelineMain.
End UnsaturatedSolinas.

Module WordByWordMontgomery.
  Definition main : unit
    := main_gen ForExtraction.WordByWordMontgomery.PipelineMain.
End WordByWordMontgomery.

Module SaturatedSolinas.
  Definition main : unit
    := main_gen ForExtraction.SaturatedSolinas.PipelineMain.
End SaturatedSolinas.

Module BaseConversion.
  Definition main : unit
    := main_gen ForExtraction.BaseConversion.PipelineMain.
End BaseConversion.
```
and make copies of the various two-line `.v` files in
`src/ExtractionOCaml/` and add the relevant targets to the `Makefile`.

cc @jadephilipoom

### DIFF
--- a/src/CLI.v
+++ b/src/CLI.v
@@ -178,7 +178,9 @@ Module ForExtraction.
          => inr [res]
        end.
 
-  Definition supported_languages : list (string * ToString.OutputLanguageAPI)
+  Class supported_languagesT := supported_languages : list (string * ToString.OutputLanguageAPI).
+
+  Definition default_supported_languages : supported_languagesT
     := [("C", ToString.OutputCAPI)
        ; ("Rust", Rust.OutputRustAPI)
        ; ("Go", Go.OutputGoAPI)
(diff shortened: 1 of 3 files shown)


jadephilipoom:
Awesome, thanks Jason!

PR #684, opened 2020-03-12
[WIP] Initial support for bedrock2 output binaries

JasonGross:
@jadephilipoom Here is the code needed to make bedrock2 output CLI binaries.  I've left one axiom for you to fill, which describes how to convert PHOAS ASTs to lists of strings.  I'm happy to answer questions (and/or add documentation) about what the various arguments and datatypes are.

I think we should not merge this until the axiom is filled (though I'm open to other opinions); feel free to take over this branch / PR.

Also, open question: Should the binaries (and the language option) be called bedrock, or bedrock2?

Note that this is on top of #683


jadephilipoom:
Moved this to #686 , so I can take over it without needing to push to your repo.

PR #694, opened 2020-03-23
Bedrock2 computable valid

jadephilipoom:
Added a boolean definition of the inductive `valid_func` precondition for the bedrock2 backend, meaning you can fill in the precondition with `vm_compute`. Required a little bit of wrangling (e.g. making `valid_expr` work on `type.type` instead of `type.base`).


JasonGross:
I think you need to rebase this on master to catch the recent bump to the rewriter submodule (or else you need to bump the rewriter submodule yourself).

PR #703, opened 2020-03-28
Add a GH Actions Matcher for Coq, and disable the Go one

JasonGross:


JasonGross:
Hmm, I guess we should disable the warning matcher in versions earlier than master, to not get duplicate messages?  And maybe we should fix the existing warnings?

PR #712, opened 2020-04-05
Move all .c files to fiat-c/src

JasonGross:
Now the C files are more like the other generated files.  This adds more
uniformity.

Possibly we should move them all to a generated/ folder eventually?

cc @davidben for visibility for this change in location of .c files.  Requesting @andres-erbsen review about moving them / organization (does it seem good / reasonable to you?)


andres-erbsen:
LGTM

PR #686, opened 2020-03-13
[WIP] Bedrock2 CLI

jadephilipoom:
Rebased version of #684 because I don't want to push to Jason's fork.

PR #722, opened 2020-04-07
[WIP] Bedrock2 CLI

JasonGross:
I accidentally closed #686 and then force-pushed before reopening it, so am opening this here.


JasonGross:
@jadephilipoom 
> As a sidebar, I think it's kind of confusing that you get C output by default on a binary called `bedrock2_word_by_word_montgomery` -- can we change that at some point?

Sure, I've updated the generic CLI code to default to the first element of the supported languages list, and added a commit just now to this PR that moves the bedrock2 option to the head of the list, so it should now default to bedrock2

JasonGross:
Now on top of #723, so we should wait to merge this until that is merged.

jadephilipoom:
Okay, by my count there are still the following things left to do on this:

- [ ] add new rewrite rules that translate `Z.add_{with_}get_carry` and `Z.sub_{with_}get_borrow` splitting at something other than the bitwidth into an expression that is translatable to bedrock2
- [ ] figure out how to translate `Z.zselect` into bedrock2

These are unlikely to be super-quick tasks, so I'd like to get this into a passing state and then merge it so I can work on those in separate branches. I've been having trouble making progress because this branch keeps changing under my feet (due to COVID, I'm exclusively on my laptop right now, so rebuilding takes ages).

JasonGross:
Okay, I've made it so this is now only on top of #731.  I plan to merge #731 as soon as it passes the CI.
I've also made it so that this avoids synthesizing functions that rely on selectznz, and thus this should also pass the CI.  So once this passes the CI, feel free to merge.

JasonGross:
(Sorry for all of the churn in this branch.  I expect I'm now done touching it.)

JasonGross:
Running gcc on the generated files fails with messages like
```
fiat-bedrock2/src/curve25519_64.c:20:27: error: unknown type name ‘uintptr_t’; did you mean ‘__uint128_t’?
 void fiat_25519_carry_mul(uintptr_t in0, uintptr_t in1, uintptr_t out0) {
                           ^~~~~~~~~
                           __uint128_t
```

PR #723, opened 2020-04-07
Optimize adc that generates no carry into add

JasonGross:
This optimization changes things like `(uint64_t, uint0_t)(Z.add_with_get_carry(2^64, (uint64_t)x26, (uint64_t)x8, 0))` into `(uint64_t)x26 + (uint64_t)x8`.  This is valid because bounds analysis guarantees no carry.  This is needed for getting the bedrock2 word-by-word montgomery code to synthesize correctly.

cc @jadephilipoom because this breaks the fancy-machine stuff (I'm currently investigating)

Requesting review from @andres-erbsen and @davidben for changes in the C code

Note that there is an unfortunate amount of churn because the (unused but named) carry variables have disappeared, changing the names of variables on many lines of code.


JasonGross:
@jadephilipoom The issue is that (for, e.g., montgomery), the addc in
```
ADDC (-128) t2 t2 scratch
```
is guaranteed to be carryless.  Hence the new code optimizes it into `scratch + _ + _ >> 128`.  I see two possible solutions:
1. add a fancy rewrite rule to recognize versions of `Z.add_with_get_carry_full` that are just additions, when the additions don't overflow
2. add a flag to disable this new optimization, and disable it for fancy machine

What do you think we should do?

jadephilipoom:
> What do you think we should do?

The new rewrite rule would be the good and correct plan, but I also expect it to be a painful in proofs. I'd say make a flag and disable it for fancy, at least for now.

JasonGross:
I will just disable the rewrite rule when the fancy-options are passed.  The proofs probably won't be too hard, I expect, but the rewrite rules are kind-of a pain to write, given all the casts everywhere, so I'd rather avoid writing them if I can.

andres-erbsen:
sounds like a good idea and I trust that you can get something like this right. I will not look at the code unless you point me at something particular you'd like another pair of eyes on

davidben:
Sorry about the late response. I benchmarked this and it's faster! :-D

7b42002318b6b7663de620774c6a0e1fdf178abf (the commit this PR is based on)
```
Did 25324 ECDH P-256 operations in 5093774us (4971.6 ops/sec)
Did 70000 ECDSA P-256 signing operations in 5042517us (13882.0 ops/sec)
Did 29536 ECDSA P-256 verify operations in 5051229us (5847.3 ops/sec)
```

79965e9a15d887dbdd0990929992f44382b5bb89 (this PR)
```
Did 27768 ECDH P-256 operations in 5009024us (5543.6 ops/sec)
Did 76000 ECDSA P-256 signing operations in 5005671us (15182.8 ops/sec)
Did 33332 ECDSA P-256 verify operations in 5024180us (6634.3 ops/sec)
```

davidben:
I should footnote, those numbers are for x86_64. They also include the following patch:

```
static void fiat_p256_cmovznz_u64(uint64_t* out1, fiat_p256_uint1 arg1, uint64_t arg2, uint64_t arg3) {
  fiat_p256_uint1 x1 = (!(!arg1));
  uint64_t x2 = ((fiat_p256_int1)(0x0 - x1) & UINT64_C(0xffffffffffffffff));
  // Note this line has been patched from the synthesized code to add value
  // barriers.
  //
  // Clang recognizes this pattern as a select. While it usually transforms it
  // to a cmov, it sometimes further transforms it into a branch, which we do
  // not want.
  uint64_t x3 = ((value_barrier_u64(x2) & arg3) | (value_barrier_u64(~x2) & arg2));
  *out1 = x3;
}
```

`value_barrier_u64` is defined as below. It's a goofball hack to stop the compiler from turning cmovs back into branches, because apparently it's doing that now. :-(
```
static inline uint64_t value_barrier_u64(uint64_t a) {
#if !defined(OPENSSL_NO_ASM) && (defined(__GNUC__) || defined(__clang__))
  __asm__("" : "+r"(a) : /* no inputs */);
#endif
  return a;
}
```

JasonGross:
> Sorry about the late response. I benchmarked this and it's faster! :-D

Great, thanks!  I shall merge then.  (I'm also curious about 25519, if you have those numbers)

davidben:
> I'm also curious about 25519, if you have those numbers

I thought this PR didn't affect 25519, or were you just interested in general comparisons?

JasonGross:
> > I'm also curious about 25519, if you have those numbers
> 
> I thought this PR didn't affect 25519, or were you just interested in general comparisons?

Oh, oops, I was misremembering which files this PR impacted.

PR #724, opened 2020-04-08
Also run coqchk on the CI

### LINKED ISSUE #717: Add coqchk to the CI
It'll slow things down a bit, but will give us more trust.  Probably worth it.  Also probably relatively easy.

JasonGross:
Closes #717

Let's see how much this slows things down on the CI...

### DIFF
--- a/.github/workflows/coq-macos.yml
+++ b/.github/workflows/coq-macos.yml
@@ -115,3 +115,5 @@ jobs:
         name: build-outputs
         path: .
       if: always ()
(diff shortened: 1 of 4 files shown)


JasonGross:
The validate target adds 40--45 minutes to each build.  Do you think it's worth it @andres-erbsen @jadephilipoom ?

JasonGross:
TODO for me: make this run only on master and not PRs

PR #733, opened 2020-04-09
Generate all functions in all bedrock2 files

JasonGross:
<strike>This is a draft PR on top of #722, #723, #731, and #732, which we can use to track the state of being able to generate all functions in all bedrock2 files.  The one and only commit specific to this PR (at the time of me making it) eliminates the "filter out the non-building stuff" in the Makefile, and presumably will eventually be the commit that gets merged once we finish making bedrock2 able to synthesize all the functions.  (Well, we'll also need to regenerate the .c files in fiat-bedrock2.)</strike>


JasonGross:
@jadephilipoom This works now!  (At least locally on my machine)

JasonGross:
@jadephilipoom @andres-erbsen Note that we need to set `-Wno-error=tautological-compare`.  It's unfortunate that the rewriter can't yet deal with non-linear patterns to optimize away self-comparisons, though it's possible we could be more clever in the initial rewrite rules to avoid this case, for example by having special cases for when `Z.add_with_get_carry_full` and `Z.sub_with_get_borrow_full` have a `0` as one or more of their arguments.

PR #741, opened 2020-04-11
Print cabal config

JasonGross:
Also add code folding to the config printout


JasonGross:
Apparently cabal config does not live in ~/.cabal/config on the runners.

PR #744, opened 2020-04-12
Use dune to build Coq on Windows

JasonGross:


JasonGross:
We now use opam to get Coq on Windows

PR #749, opened 2020-04-16
New rewrite rules for add/sub with carries/borrows (for bedrock2)

jadephilipoom:
As discussed in #686 and #722, the bedrock2 translation was failing on `to_bytes` due to two problems, one of which was adds-with-carry and subs-with-borrow that split on a bit index other than the word size (e.g. `Z.add_with_get_carry (2^51) c x y`). These expressions weren't matching the `split_multiret` rewrite rules and were showing up as assignments to two variables, which bedrock2 doesn't support. This PR fixes that issue by adding new rules that match on non-word-size splits and split the return values.

Since we still lack a bedrock2-expressible transformation for `Z.zselect` (the second problem, which I'll start working on now), generating `to_bytes` still fails. But with many fewer errors than before!

### DIFF
--- a/src/Rewriter/Rules.v
+++ b/src/Rewriter/Rules.v
@@ -937,6 +937,20 @@ Section with_bitwidth.
                        = (dlet low := singlewidth (singlewidth x * singlewidth y) in
                               dlet high := singlewidth (Z.mul_high (cstZ rpow2_bitwidth ('(2^bitwidth))) (singlewidth x) (singlewidth y)) in
                               (singlewidth low, singlewidth high)))
+              ; (forall x y n,
+                    0 <= lgcarrymax <= bitwidth
+                    -> 0 < n < bitwidth
+                    -> 0 < bitwidth < n + lgcarrymax
+                    -> singlewidth_carry (Z.add_get_carry_full ('(2^n)) (singlewidth x) (singlewidth y))
+                       = (dlet sum_xy := singlewidth (singlewidth x + singlewidth y) in
+                              dlet carry_xy := carrywidth (Z.ltz (singlewidth sum_xy) (singlewidth x)) in
+                              dlet low := singlewidth (Z.land (singlewidth sum_xy) ('(2^n - 1))) in
+                              dlet high :=
+                            singlewidth
+                              (Z.add (singlewidth (Z.shiftr (singlewidth sum_xy) ('n)))
+                                     (singlewidth (Z.shiftl (carrywidth carry_xy) ('(bitwidth - n)))))
+                                in
+                              (singlewidth low, singlewidth high)))
              ]
            ; mymap
                do_again
--- a/src/Rewriter/Rules.v
+++ b/src/Rewriter/Rules.v
@@ -957,6 +957,27 @@ Section with_bitwidth.
                                                             (singlewidth ('(bitwidth - Z.log2 two_pow_n))))))
                             in
                               (singlewidth low, singlewidth high)))
+              ; (forall c x y two_pow_n r,
+                    0 <= lgcarrymax <= bitwidth
+                    -> two_pow_n = 2 ^ Z.log2 two_pow_n
+                    -> 0 < Z.log2 two_pow_n < bitwidth
+                    -> bitwidth + 1 < Z.log2 two_pow_n + lgcarrymax
+                    -> two_pow_n ∈ r
+                    -> singlewidth_carry
+                         (Z.add_with_get_carry_full (cstZ r ('(two_pow_n))) (carrywidth c) (singlewidth x) (singlewidth y))
+                       = (dlet sum_cx := singlewidth (carrywidth c + singlewidth x) in
+                              dlet carry_cx := carrywidth (Z.ltz (singlewidth sum_cx) (singlewidth x)) in
+                              dlet sum_cxy := singlewidth (singlewidth sum_cx + singlewidth y) in
+                              dlet carry_cxy := carrywidth (Z.ltz (singlewidth sum_cxy) (singlewidth y)) in
+                              dlet carry := singlewidth (Z.add (carrywidth carry_cx) (carrywidth carry_cxy)) in
+                              dlet low := singlewidth
+                                            (Z.land (singlewidth sum_cxy) (singlewidth ('(two_pow_n - 1)))) in
+                              dlet high := singlewidth
+                                             (Z.add (singlewidth (Z.shiftr (singlewidth sum_cxy)
+                                                                           (singlewidth ('(Z.log2 two_pow_n)))))
+                                                    (singlewidth (Z.shiftl (carrywidth carry)
+                                                                           (singlewidth ('(bitwidth - Z.log2 two_pow_n)))))) in
+                              (singlewidth low, singlewidth high)))
              ]
            ; mymap
                do_again
--- a/src/Rewriter/Rules.v
+++ b/src/Rewriter/Rules.v
@@ -937,22 +937,24 @@ Section with_bitwidth.
                        = (dlet low := singlewidth (singlewidth x * singlewidth y) in
                               dlet high := singlewidth (Z.mul_high (cstZ rpow2_bitwidth ('(2^bitwidth))) (singlewidth x) (singlewidth y)) in
                               (singlewidth low, singlewidth high)))
-              ; (forall x y n rpow2_n,
+              ; (forall x y two_pow_n r,
                     0 <= lgcarrymax <= bitwidth
-                    -> 0 < n < bitwidth
-                    -> bitwidth < n + lgcarrymax
-                    -> 2^n ∈ rpow2_n
+                    -> two_pow_n = 2 ^ Z.log2 two_pow_n
+                    -> 0 < Z.log2 two_pow_n < bitwidth
+                    -> bitwidth < Z.log2 two_pow_n + lgcarrymax
+                    -> two_pow_n ∈ r
                     -> singlewidth_carry
-                         (Z.add_get_carry_full (cstZ rpow2_n ('(2^n))) (singlewidth x) (singlewidth y))
+                         (Z.add_get_carry_full (cstZ r ('(two_pow_n))) (singlewidth x) (singlewidth y))
                        = (dlet sum_xy := singlewidth (singlewidth x + singlewidth y) in
                               dlet carry_xy := carrywidth (Z.ltz (singlewidth sum_xy) (singlewidth x)) in
                               dlet low := singlewidth
-                                            (Z.land (singlewidth sum_xy) (singlewidth ('(2^n - 1)))) in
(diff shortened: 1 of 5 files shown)


src/Rewriter/RulesProofs.v
```diff
@@ -676,19 +678,77 @@ Lemma multiret_split_rewrite_rules_proofs (bitwidth : Z) (lgcarrymax : Z)
...
+  all: try solve[
+  try lazymatch goal with
        | [ |- ?x mod ?y = _ mod ?y ]
          => apply f_equal2; [ | reflexivity ];
-              progress cbv [Z.ltz]; break_innermost_match; Z.ltb_to_lt; Z.div_mod_to_quot_rem
-       end.
-  all: repeat match goal with
+              cbv [Z.ltz]; break_innermost_match; Z.ltb_to_lt
+       end; Z.div_mod_to_quot_rem;
+   repeat match goal with
```
JasonGross, line 751: This is a lot of proof.  Would it be reasonable to factor it into helper lemmas or general-purpose tactics?

src/Rewriter/Rules.v
```diff
@@ -937,6 +937,90 @@ Section with_bitwidth.
...
                               dlet high := singlewidth (Z.mul_high (cstZ rpow2_bitwidth ('(2^bitwidth))) (singlewidth x) (singlewidth y)) in
                               (singlewidth low, singlewidth high)))
+              ; (forall x y two_pow_n r,
+                    0 <= lgcarrymax <= bitwidth
+                    -> two_pow_n = 2 ^ Z.log2 two_pow_n
+                    -> 0 < Z.log2 two_pow_n < bitwidth
+                    -> bitwidth < Z.log2 two_pow_n + lgcarrymax
+                    -> two_pow_n ∈ r
+                    -> singlewidth_carry
+                         (Z.add_get_carry_full (cstZ r ('(two_pow_n))) (singlewidth x) (singlewidth y))
```
JasonGross, line 947:
```suggestion
                         (Z.add_get_carry_full (cstZ r ('two_pow_n)) (singlewidth x) (singlewidth y))
```

src/Rewriter/Rules.v
```diff
@@ -937,6 +937,90 @@ Section with_bitwidth.
...
+                            in
+                              (singlewidth low, singlewidth high)))
+              ; (forall c x y two_pow_n r,
+                    0 <= lgcarrymax <= bitwidth
+                    -> two_pow_n = 2 ^ Z.log2 two_pow_n
+                    -> 0 < Z.log2 two_pow_n < bitwidth
+                    -> bitwidth + 1 < Z.log2 two_pow_n + lgcarrymax
+                    -> two_pow_n ∈ r
+                    -> singlewidth_carry
+                         (Z.add_with_get_carry_full (cstZ r ('(two_pow_n))) (carrywidth c) (singlewidth x) (singlewidth y))
```
JasonGross, line 967:
```suggestion
                         (Z.add_with_get_carry_full (cstZ r ('two_pow_n)) (carrywidth c) (singlewidth x) (singlewidth y))
```

src/Rewriter/Rules.v
```diff
@@ -937,6 +937,90 @@ Section with_bitwidth.
...
+                                                                           (singlewidth ('(bitwidth - Z.log2 two_pow_n)))))) in
+                              (singlewidth low, singlewidth high)))
+              ; (forall x y two_pow_n r,
+                    0 <= lgcarrymax <= bitwidth
+                    -> two_pow_n = 2 ^ Z.log2 two_pow_n
+                    -> 0 < Z.log2 two_pow_n < bitwidth
+                    -> bitwidth < Z.log2 two_pow_n + lgcarrymax
+                    -> two_pow_n ∈ r
+                    -> singlewidth_carry
+                         (Z.sub_get_borrow_full (cstZ r ('(two_pow_n))) (singlewidth x) (singlewidth y))
```
JasonGross, line 988:
```suggestion
                         (Z.sub_get_borrow_full (cstZ r ('two_pow_n)) (singlewidth x) (singlewidth y))
```

src/Rewriter/Rules.v
```diff
@@ -937,6 +937,90 @@ Section with_bitwidth.
...
+                            in
+                              (singlewidth low, carrywidth high)))
+              ; (forall c x y two_pow_n r,
+                    0 <= lgcarrymax <= bitwidth
+                    -> two_pow_n = 2 ^ Z.log2 two_pow_n
+                    -> 0 < Z.log2 two_pow_n < bitwidth
+                    -> bitwidth < Z.log2 two_pow_n + lgcarrymax
+                    -> two_pow_n ∈ r
+                    -> singlewidth_carry
+                         (Z.sub_with_get_borrow_full (cstZ r ('(two_pow_n))) (carrywidth c) (singlewidth x) (singlewidth y))
```
JasonGross, line 1008:
```suggestion
                         (Z.sub_with_get_borrow_full (cstZ r ('two_pow_n)) (carrywidth c) (singlewidth x) (singlewidth y))
```

src/Util/ZUtil/AddGetCarry.v
```diff
@@ -145,9 +157,50 @@ Module Z.
...
   Proof. fin_div_ltz. Qed.
 
+  Lemma sub_sub_div_ltz s x y c :
+    0 <= c < s
+    -> 0 <= x < s
+    -> 0 <= y < s
+    -> - ((x - y - c) / s) =
+       Z.ltz x ((x - y) mod s) + Z.ltz ((x - y) mod s) ((x - y - c) mod s).
+  Proof.
+    intros. rewrite (PullPush.Z.sub_mod_l (x-y)).
```
JasonGross, line 167: Please import the relevant file so you can just write `Z.sub_mod_l`.  Or consider doing `push_Zmod; rewrite <- !sub_div_ltz by auto with zarith; pull_Zmod`?

src/Util/ZUtil/AddGetCarry.v
```diff
@@ -145,9 +157,50 @@ Module Z.
...
+  Proof.
+    intros.
+    rewrite <-Z.div_add by auto with zarith.
+    match goal with
+      |- context [?x * 2^(?a-?b) * 2^?b] =>
+      replace (x * 2^(a-b) * 2^b) with (x * 2^a)
+        by (rewrite <-Z.mul_assoc, <-Z.pow_add_r by auto with zarith;
+            repeat (f_equal; try lia))
+    end.
+    fin_div_ltz.
```
JasonGross, line 204: I think it might be a bit nicer to do this proof in the following three steps, if they work and are sufficiently performant:
1. rewrite with the lemma that turns `2^(a-b)` into `2^a/2^b`
2. pose whatever positivity facts are needed about `2^a` and `2^b`
3. `Z.div_mod_to_quot_rem; nia`

But feel free to leave as-is

src/Util/ZUtil/AddGetCarry.v
```diff
@@ -145,9 +157,50 @@ Module Z.
...
+  (* TODO: should this lemma go elsewhere? Where? *)
+  Lemma ltz_mod_pow2_small x y z :
+    0 < z ->
+    (Z.ltz x y) mod (2 ^ z) = Z.ltz x y.
+  Proof.
+    cbv [Z.ltz]; break_match; intros; Z.ltb_to_lt;
+      apply Z.mod_small; auto with zarith.
+  Qed.
+
+  (* TODO: should this lemma go elsewhere? Where? *)
```
JasonGross, line 187: Probably in `ZUtil/Div.v`, unless that creates circular dependencies, in which case put it in `ZUtil/Modulo.v`.

jadephilipoom:
Thanks for the comments, Jason!

The CI passes everything else, but fails on Windows in the "clone Coq" stage with the message `The system cannot find the path specified.`. This looks to me like a CI problem and not a code problem, but it's working on master.

src/Rewriter/RulesProofs.v
jadephilipoom: I tried pretty hard already. This is about 1/4 the size of my original proof and I don't know how to make it smaller than that. Probably I could offload more to the match statement that happens after `Z.div_mod_to_quot_rem`, but it's hard to know what I'm looking at once the proof is in that form and I haven't been successful.

src/Util/ZUtil/AddGetCarry.v
jadephilipoom: I had to help it out with some extra rewrites that added dependencies to the file. Let me know which version you prefer.

jadephilipoom:
Hmm, Windows build now failing in the `deps` stage with `No rule to make target 'rewriter'`. I restarted it.

JasonGross:
The windows build is flaky, I'll look into it, but don't block this PR on having the windows build succeed.

src/Util/ZUtil/AddGetCarry.v
```diff
@@ -145,9 +158,30 @@ Module Z.
...
+    fin_div_ltz.
+  Qed.
+
   Lemma sub_get_borrow_full_ltz s x y :
     0 <= x < s
     -> 0 <= y < s
     -> snd (Z.sub_get_borrow_full s x y) = Z.ltz x ((x - y) mod s).
   Proof. rewrite sub_get_borrow_full_div; fin_div_ltz. Qed.
+
+  (* TODO: should this lemma go elsewhere? Where? *)
```
JasonGross, line 179: Maybe make a new ZUtil file called Ltz and put the lemmas that are just about Z.ltz in it?

src/Util/ZUtil/Modulo.v
```diff
@@ -6,6 +6,7 @@ Require Import Crypto.Util.ZUtil.Tactics.LtbToLt.
 Require Import Crypto.Util.ZUtil.Tactics.ReplaceNegWithPos.
 Require Import Crypto.Util.ZUtil.Tactics.PullPush.Modulo.
 Require Import Crypto.Util.ZUtil.Div.
+Require Import Crypto.Util.ZUtil.Divide.
```
JasonGross, line 9: What's the difference between Div and Divide, and why do we have both of them?

src/Util/ZUtil/AddGetCarry.v
jadephilipoom: Probably a good idea, I'll do that.

src/Util/ZUtil/Modulo.v
jadephilipoom: Divide is Z.divide, notation `|`, as in "4 divides 12 but 5 does not divide 12". Here I imported it for the lemma `Z.divide_pow_le : forall a b c, b <= c -> a^b | a^c`, which helps me prove `2^a mod 2^b = 0`.

src/Util/ZUtil/AddGetCarry.v
jadephilipoom: But maybe separate from this PR, since it would involve moving some other files too and seems off-topic. I'll merge this first and then move the lemmas.

PR #752, opened 2020-04-16
Add 32-bit bedrock2 builds

### LINKED ISSUE #751: bedrock2: 32-bit curve25519 code on CI?
like curve25519_64.c, but for a 32-bit machine.

JasonGross:
Closes #751

<strike>Except that it doesn't actually work.  I've committed the temporary files with the error messages, where the 32-bit builds fail in ways that the 64-bit builds succeed.</strike>


jadephilipoom:
I'm not sure where you're pulling in the bedrock2 translation parameters, but my `Defaults` file had a 64-bit default and won't work on 32-bit. I'm currently developing a 32-bit test case and in the process 32-bit default parameters.

@andres-erbsen why does BasicC32Semantics not have a `parameters_ok` instance?

jadephilipoom:
Working test case (with an axiom for `Semantics.parameters_ok` on the `bedrock2_32bit` branch. It breaks `Stringification.v` because that's where the parameters are pulled in, so I'm working on fixing that (without committing to a specific machine word size and splitting into `Stringification32` and `Stringification64`, which seems very silly).

jadephilipoom:
It would help me considerably if something in the `OutputLanguageAPI` got the word size as an argument. As it is, it looks like all I can do is try to pick the parameters based on the input bounds. Should I just do that, or is there a reasonably simple way to pass me the word size?

JasonGross:
It's pretty easy to have ToFunctionLines, header, footer, etc, get machine_wordsize as a parameter; let me go do that.

jadephilipoom:
> It's pretty easy to have ToFunctionLines, header, footer, etc, get machine_wordsize as a parameter; let me go do that.

I only need ToFunctionLines, the others are fine.

JasonGross:
The issue with carry_square should be fixed by #756.  Not sure why 32-bit and 64-bit are different here, though.

JasonGross:
@jadephilipoom This should be ready to go now, assuming the CI passes.

PR #757, opened 2020-04-17
Add support for 32-bit bedrock2 translations

jadephilipoom:
This PR:
- separates out the `Bedrock/Defaults.v` file into three files: one for global defaults, one for 32-bit-specific defaults, and one for 64-bit-specific defaults
- creates a 32-bit test in `Bedrock/Tests`
- modifies `Bedrock/Stringification.v` to select the translation parameters based on the word size

Put together, this gives us working 32-bit code from the CLI. I've tested it on all currently supported operations for curve25519 and p256. The changes should make #752 work!

### DIFF
--- a/src/Bedrock/Tests/Proofs/X25519_32.v
+++ b/src/Bedrock/Tests/Proofs/X25519_32.v
@@ -1,0 +1,322 @@
+Require Import Coq.ZArith.ZArith.
+Require Import Coq.Strings.String. (* should go before lists *)
+Require Import Coq.Lists.List.
+Require Import coqutil.Word.Interface.
+Require Import coqutil.Word.Properties.
+Require Import coqutil.Map.Interface.
+Require Import coqutil.Map.Properties.
+Require Import bedrock2.Array.
+Require Import bedrock2.BasicC32Semantics.
+Require Import bedrock2.Scalars.
+Require Import bedrock2.ProgramLogic.
+Require Import bedrock2.WeakestPrecondition.
+Require Import bedrock2.WeakestPreconditionProperties.
+Require Import bedrock2.Map.Separation.
+Require Import Crypto.BoundsPipeline.
+Require Import Crypto.Arithmetic.Core.
+Require Import Crypto.Arithmetic.ModOps.
+Require Import Crypto.Bedrock.Defaults.
+Require Import Crypto.Bedrock.Defaults32.
+Require Import Crypto.Bedrock.Types.
+Require Import Crypto.Bedrock.Tactics.
+Require Import Crypto.Bedrock.Proofs.Func.
+Require Import Crypto.Bedrock.Proofs.ValidComputable.Func.
+Require Import Crypto.Bedrock.Util.
+Require Import Crypto.Language.API.
+Require Import Crypto.Util.Tactics.BreakMatch.
+Require bedrock2.Map.SeparationLogic. (* if imported, list firstn/skipn get overwritten and it's annoying *)
+Local Open Scope Z_scope.
+
+Import Language.Compilers.
+Import Associational Positional.
+
+Require Import Crypto.Util.Notations.
+Import Types.Notations ListNotations.
+Local Open Scope Z_scope.
+Local Open Scope string_scope.
+
+Require Import Crypto.Bedrock.Tests.X25519_32.
+Import X25519_32.
+Local Coercion name_of_func (f : bedrock_func) := fst f.
+
+Axiom BasicC32Semantics_parameters_ok : Semantics.parameters_ok BasicC32Semantics.parameters.
+(* TODO: why does BasicC32Semantics not have a Semantics.parameters_ok instance? *) }
+Existing Instance BasicC32Semantics_parameters_ok.
+
+Section Proofs.
+  Context (n : nat := 10%nat)
+          (s : Z := 2^255)
+          (c : list (Z * Z) := [(1,19)])
+          (machine_wordsize : Z := 32).
+
+  (* requires some kind of proof about decimal stringification *)
+  Lemma decimal_varname_gen_unique :
+    forall i j : nat, varname_gen i = varname_gen j <-> i = j.
+  Admitted.
+
+  Instance p_ok : Types.ok.
+  Proof.
+    constructor.
+    { exact BasicC32Semantics_parameters_ok. }
+    { reflexivity. }
+    { exact decimal_varname_gen_unique. }
+  Admitted.
+
+  Local Notation M := (s - Associational.eval c)%Z.
+  Local Notation eval :=
+    (eval (weight limbwidth_num limbwidth_den) n).
+
+  Definition Bignum (addr : Semantics.word) (x : Z) :
+    Semantics.mem -> Prop :=
+    Lift1Prop.ex1
+      (fun xs =>
+         sep (emp (eval (map word.unsigned xs) = x))
+             (sep (emp (length xs = n))
+                  (array scalar (word.of_Z word_size_in_bytes)
+                         addr xs))).
+
+  Instance spec_of_mulmod_bedrock : spec_of mulmod_bedrock :=
+    fun functions =>
+      forall x y px py pout old_out t m
+             (Ra Rr : Semantics.mem -> Prop),
+        sep (sep (Bignum px x) (Bignum py y)) Ra m ->
+        sep (Bignum pout old_out) Rr m ->
+        WeakestPrecondition.call
+          (p:=@semantics default_parameters)
+          functions mulmod_bedrock t m
+          (px :: py :: pout :: nil)
+          (fun t' m' rets =>
+             t = t' /\
+             rets = []%list /\
+             sep (Bignum pout ((x * y) mod M)%Z) Rr m').
+
+  Lemma mulmod_valid_func :
+    valid_func (mulmod (fun H3 : API.type => unit)).
+  Proof.
+    apply valid_func_bool_iff.
+    vm_compute; reflexivity.
+  Qed.
+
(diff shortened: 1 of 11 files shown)


src/Bedrock/Tests/Proofs/X25519_32.v
```diff
@@ -0,0 +1,323 @@
...
+Section Proofs.
+  Context (n : nat := 10%nat)
+          (s : Z := 2^255)
+          (c : list (Z * Z) := [(1,19)])
+          (machine_wordsize : Z := 32).
+
+  (* requires some kind of proof about decimal stringification *)
+  Lemma decimal_varname_gen_unique :
+    forall i j : nat, varname_gen i = varname_gen j <-> i = j.
+  Admitted.
```
JasonGross, line 56: If you rebase on top of https://github.com/mit-plv/fiat-crypto/pull/758 (which I hope to merge soon, but which unfortunately might slightly conflict with this PR because it changes some of the string <-> number function names (because they've been upstreamed in Coq)), you'll get an easy proof that the composition `nat -> string -> nat` is the identity function.  You should then be able to prove this fact pretty easily

src/Bedrock/Tests/Proofs/X25519_32.v
```diff
@@ -0,0 +1,323 @@
...
+
+  Instance p_ok : Types.ok.
+  Proof.
+    constructor.
+    { exact BasicC32Semantics_parameters_ok. }
+    { reflexivity. }
+    { exact decimal_varname_gen_unique. }
+  Admitted.
```
JasonGross, line 64: What remains in this proof?  Why does it need to be admitted?

src/Bedrock/Tests/Proofs/X25519_32.v
```diff
@@ -0,0 +1,323 @@
...
+    valid_func (mulmod (fun H3 : API.type => unit)).
+  Proof.
+    apply valid_func_bool_iff.
+    vm_compute; reflexivity.
+  Qed.
+
+  (* TODO: ask Jason for help *)
+  Lemma mulmod_Wf :
+    Wf.Compilers.expr.Wf3 mulmod.
+  Admitted.
```
JasonGross, line 104: Two options:
1. The principled way: Pass `mulmod` through `GeneralizeVar` and then use [`Wf3_GeneralizeVar : forall {t} (e : Expr t), expr.Wf (ident:=ident) e -> expr.Wf3 (GeneralizeVar (e _))`](https://github.com/mit-plv/rewriter/blob/6f9c96c5f1b78813a0db8ade7da6a5799b6f0df6/src/Rewriter/Language/Wf.v#L1501) combined with [`BoundsPipeline_correct`](https://github.com/mit-plv/fiat-crypto/blob/17e3c5bc3953dfa1538df5f62a38bc6f396b1e3f/src/BoundsPipeline.v#L792-L820) which essentially gives you that if `Wf e` and `BoundsPipeline e = Success e2`, then `Wf e2`, and then use [`prove_Wf`](https://github.com/mit-plv/rewriter/blob/6f9c96c5f1b78813a0db8ade7da6a5799b6f0df6/src/Rewriter/Language/Wf.v#L1604)
2. The easy way (which is basically recomputing Wf on the output of the pipeline): ask me to add a tactic `prove_Wf3` to the rewriter (I expect it'll take me about 15 minutes) that knows how to prove `Wf3` on concrete syntax trees, and then use that.

src/Bedrock/Tests/Proofs/X25519_32.v
```diff
@@ -0,0 +1,323 @@
...
+      (expr.interp (@Compilers.ident_interp)
+                   (mulmod API.interp_type)
+                   x y) = n.
+  Proof.
+    vm_compute. reflexivity.
+  Qed.
+
+  (* TODO: here's where we need to use the FC pipeline to say things about
+         correctness *)
+  (* this can be changed to use type.app_curried if that's easier *)
```
JasonGross, line 117: That will probably make things easier.

src/Bedrock/Tests/Proofs/X25519_32.v
```diff
@@ -0,0 +1,323 @@
...
+  Lemma mulmod_correct x y :
+    eval
+      (map word.wrap
+           (expr.interp
+              (@Compilers.ident_interp) (mulmod API.interp_type)
+              (map word.unsigned x) (map word.unsigned y))) =
+    ((eval (map word.unsigned x) * eval (map word.unsigned y)) mod M)%Z.
+  Admitted.
```
JasonGross, line 125: The relevant lemma is [`BoundsPipeline_correct`](https://github.com/mit-plv/fiat-crypto/blob/17e3c5bc3953dfa1538df5f62a38bc6f396b1e3f/src/BoundsPipeline.v#L792-L820)

src/Bedrock/Tests/Proofs/X25519_32.v
```diff
@@ -0,0 +1,323 @@
...
+
+  (* TODO : move *)
+  Lemma substring_0_0 :
+    forall s, substring 0 0 s = "".
+  Proof.
+    clear. destruct s; reflexivity.
+  Qed.
```
JasonGross, line 132: Maybe this should be moved to `Util/Strings/String.v`?

src/Bedrock/Tests/Proofs/X25519_32.v
```diff
@@ -0,0 +1,323 @@
...
+
+  Lemma varname_gen_startswith v i :
+    v = varname_gen i ->
+    String.startswith "x" v = true.
+  Proof.
+    cbn [varname_gen default_parameters]. intro.
+    subst. cbn. rewrite substring_0_0.
+    reflexivity.
+  Qed.
```
JasonGross, line 141: Should lemmas like this be factored out between the 32-bit and the 64-bit code?

src/Bedrock/Tests/Proofs/X25519_32.v
jadephilipoom: The `prove_Wf3` tactics sounds great, if you're willing to put it together! I also need to do this proof in the 64-bit test case, so it would be cool to have a general-purpose tactic.

jadephilipoom:
I'll wait until you merge #758 and then rebase on top of that and change the string handling before merging.

The 32-bit test case is shamelessly copied from the 64-bit one, and the review comments have highlighted some sloppiness in both of them. I'll fix some of those issues before merging as well, in both files.

JasonGross:
I've merged #758

src/Bedrock/Tests/Proofs/X25519_32.v
JasonGross: https://github.com/mit-plv/fiat-crypto/commit/692462f904fb442e16bd73739046db30a358f10c and https://github.com/mit-plv/rewriter/commit/677d3f226e9cd26a8148490f44c780c7f8304f9f. It was easier than I estimated and only took about 10 minutes.  (I haven't tested it though; try it out and let me know how it works.)

src/Bedrock/Tests/Proofs/X25519_32.v
jadephilipoom: How do I get to use the `carry_mulmod` correctness proof? Right now `carry_mulmod` is under a call to `Reify`.

JasonGross: You have a proof that `mulmod_ = ErrorT.Success mulmod` called `mulmod_eq`.  The fact that your call to `Reify` is inlined into the definition of `mulmod_` is going to make things a bit painful, in large part because it's so big (because it unfolds a bunch of things during reification).  I would suggest factoring it out into a separate constant, a la the reification cache examples in PushButtonSynthesis.  Either way, you'll need to prove that the reified constant is `Wf` and also that the interpretation of the reified constant is pointwise-equal to `carry_mulmod` (this can be done by unfolding `interp` and using `reflexivity`). See https://github.com/mit-plv/fiat-crypto/blob/692462f904fb442e16bd73739046db30a358f10c/src/PushButtonSynthesis/ReificationCache.v#L25-L26 and https://github.com/mit-plv/fiat-crypto/blob/692462f904fb442e16bd73739046db30a358f10c/src/PushButtonSynthesis/UnsaturatedSolinasReificationCache.v#L54-L57 .  It's up to you whether you just factor out the call to `Reify` into a separate definition, but inline the correctness proofs, or factor it out entirely, or whether you depend on some (or even all) of the existing PushButtonSynthesis machinery, e.g., by using `carry_mul` at https://github.com/mit-plv/fiat-crypto/blob/692462f904fb442e16bd73739046db30a358f10c/src/PushButtonSynthesis/UnsaturatedSolinas.v#L284-L292 and the `carry_mul` correctness proof at https://github.com/mit-plv/fiat-crypto/blob/692462f904fb442e16bd73739046db30a358f10c/src/PushButtonSynthesis/UnsaturatedSolinas.v#L539-L542 which uses the correctness condition defined at https://github.com/mit-plv/fiat-crypto/blob/692462f904fb442e16bd73739046db30a358f10c/src/COperationSpecifications.v#L215-L221 .  How much you factor out probably depends on how much you want this example to be instructive vs just a test of the output machinery and correctness proof.

src/Bedrock/Tests/Proofs/X25519_32.v
jadephilipoom: Correct, this is done now.

src/Bedrock/Tests/Proofs/X25519_32.v
jadephilipoom: It was admitted by mistake; fixed now.

PR #758, opened 2020-04-18
Remove string cruft that's been upstreamed

JasonGross:
Since we now no longer support Coq 8.8, we can rely on definitions in
the standard library to print and parse strings.

The upshot is that now we have many fewer lines of code, and we have a
couple more proofs about the fact that `parse ∘ print` is the identity.


JasonGross:
Mac failure is probably spurious, and I've reported am issue in GitHub Actions about it.

PR #760, opened 2020-04-20
Instantiate correctness proofs for bedrock2 tests

jadephilipoom:
As discussed in #757, this change removes the remaining admits in the bedrock2 translation tests with the help of Jason's shiny new `prove_Wf3` tactic (which I can now confirm works beautifully and quickly!) and the existing `PushButtonSynthesis` infrastructure.

Now, when I `Print Assumptions mulmod_bedrock_correct` in the 64-bit test, the only dependencies are:
```
FunctionalExtensionality.functional_extensionality_dep : 
forall (A : Type) (B : A -> Type) (f g : forall x : A, B x),
(forall x : A, f x = g x) -> f = g
SortedListString.TODO_andres : False
 used in string_strict_order to prove: SortedList.parameters.strict_order
                                         String.ltb
SortedList.TODO_andres : False
 used in map_ok to prove: forall (m : map.rep)
                            (k k' : SortedList.parameters.key),
                          k <> k' ->
                          map.get (map.remove m k') k = map.get m k
 used in map_ok to prove: forall (m : map.rep)
                            (k : SortedList.parameters.key),
                          map.get (map.remove m k) k = None
 used in map_ok to prove: forall (m : map.rep)
                            (k : SortedList.parameters.key)
                            (v : SortedList.parameters.value)
                            (k' : SortedList.parameters.key),
                          k <> k' ->
                          map.get (map.put m k' v) k = map.get m k
 used in map_ok to prove: forall (m : map.rep)
                            (k : SortedList.parameters.key)
                            (v : SortedList.parameters.value),
                          map.get (map.put m k v) k = Some v
```

I believe the first comes from the rewriter, and the second two from coqutil. @andres-erbsen just out of curiosity, how close are you to completing those todos? Is it worth it for me to try to prove them?

The 32-bit `Print Assumptions` is the same, except it has an extra axiom for the `Semantics.parameters_ok` instance that is missing from `BasicC32Semantics` (see mit-plv/bedrock2#151).

Relevant prior discussion:
https://github.com/mit-plv/fiat-crypto/pull/757#discussion_r410706758
https://github.com/mit-plv/fiat-crypto/pull/757#discussion_r410717055

PR #763, opened 2020-04-21
Enable bedrock2 translation for Z.zselect

### LINKED ISSUE #750: Translate `Z.zselect` to bedrock2
Currently, synthesis of `to_bytes` and `selectznz` fails on bedrock2 because we don't have a good translation for `Z.zselect` (bedrock2 doesn't have a conditional-move or select instruction). I have an idea for how to translate it:
```
Z.zselect c x y ==
    let a := sub 0 (eq c 0) in
    let b := xor a 0xFFFFFFFF in
    or (and x a) (and y b)
```
This expression would be expressible in bedrock2. To implement it, we could do one of two things. One option is that I implement this transformation in the bedrock2 backend. This is possible but clunky; because it requires setting a variable (`a`) that is used twice, it would need to be implemented as a special case at the higher `cmd` level of abstraction, as opposed to the `expr` level where instruction-level translations usually live.

The other option is to add a new rewrite rule, and possibly a new CLI flag to indicate when we should use the new rule. This is probably nicer, except for the wrinkle that I don't think we support `xor` right now in the AST. Do we? If we need to add support for a new instruction, I'd actually vote for adding `Z.lnot` instead of `Z.xor` here; I'm using `xor` to implement `not` because bedrock2 doesn't have a `not` operation, and it would be straightforward to write that `not` gets translated to `xor` with 1s. That way if bedrock2 does later support `not`, the change would be really simple and at the appropriate level of abstraction.

So, succinctly, my questions are:
@andres-erbsen does this translation seem reasonable to you?
@JasonGross do we support `Z.lxor` or `Z.lnot`? How much work would be needed to support `Z.lnot` and a new rewrite rule to transform `Z.zselect`?

jadephilipoom:
Closes #750
On top of #760 

Enables translation of zselects by, as discussed in #750, adding a new rewriter pass that translates selects into a bedrock2-expressible form, adding a new CLI flag indicating whether the rewriter pass should be run, and then introducing support for a few new patterns in the bedrock2 backend.

After this is merged and `--no-select` flags are added to the bedrock2 build targets, #733 should work.

### DIFF
--- a/src/Bedrock/Proofs/ValidComputable/Expr.v
+++ b/src/Bedrock/Proofs/ValidComputable/Expr.v
@@ -153,17 +153,31 @@ Section Expr.
     | _ => false
     end.
 
+  Definition is_literalz {t}
+             (e : @API.expr (fun _ => unit) t) (z : Z) : bool :=
+    match e with
+    | expr.Ident _ (ident.Literal base.type.Z x) => Z.eqb x z
+    | _ => false
+    end.
+  Definition valid_expr_select_bool {t}
+             rc (e : @API.expr (fun _ => unit) t) : bool :=
+    match e with
+    | expr.Ident _ ident.Z_zselect => negb rc
+    | _ => false
+    end.
+
+  (* TODO: too many boolean arguments, change to a "mode" inductive *)
   (* add an extra flag to translate binops because otherwise the inductive
      hypothesis won't be strong enough to give information about *both*
      arguments to a binop *)
   Fixpoint valid_expr_bool' {t}
-           (binop_only shift_only require_casts : bool)
+           (binop_only shift_only select_only require_casts : bool)
            (e : @API.expr (fun _ => unit) t) {struct e} : bool :=
     if binop_only
     then match e with
          | expr.App type_Z (type.arrow type_Z type_Z) f x =>
            (valid_expr_binop_bool require_casts f)
-             && valid_expr_bool' false false true x
+             && valid_expr_bool' false false false true x
          | _ => false
          end
     else
@@ -171,10 +185,23 @@ Section Expr.
       then match e with
            | expr.App type_Z (type.arrow type_Z type_Z) f x =>
              (valid_expr_shift_bool require_casts f)
-               && valid_expr_bool' false false true x
+               && valid_expr_bool' false false false true x
            | _ => false
            end
       else
+        if select_only
+        then match e with
+             | expr.App type_Z (type.arrow type_Z type_Z) f x =>
+               (valid_expr_bool' false false true require_casts f)
+                 && is_literalz x 0
+             | expr.App type_Z
+                        (type.arrow type_Z
+                                    (type.arrow type_Z type_Z)) f x =>
+               (valid_expr_select_bool require_casts f)
+                 && valid_expr_bool' false false false true x
+             | _ => false
+             end
+        else
         match e with
         | expr.App type_nat _ f x =>
           (* only thing accepting nat is nth_default *)
@@ -182,25 +209,28 @@ Section Expr.
           then false
           else valid_expr_nth_default_bool (expr.App f x)
         | expr.App type_Z type_Z f x =>
-          if valid_expr_bool' true false require_casts f
+          if valid_expr_bool' true false false require_casts f
           then (* f is a binop applied to one valid argument;
                 check that x (second argument) is also valid *)
-            valid_expr_bool' false false true x
-          else if valid_expr_bool' false true require_casts f
+            valid_expr_bool' false false false true x
+          else if valid_expr_bool' false true false require_casts f
                then (* f is a shift applied to one valid
                        argument; check that x (shifting index) is a
                        valid shifter *)
                  valid_shifter x
-               else (* must be a cast *)
-                 (valid_expr_App1_bool require_casts f)
-                   && valid_expr_bool' false false false x
+               else if valid_expr_bool' false false true require_casts f
+                    then (* f is a select; make sure x = 2^w-1 *)
+                      is_literalz x (2^Semantics.width-1)
+                    else (* must be a cast *)
+                      (valid_expr_App1_bool require_casts f)
+                        && valid_expr_bool' false false false false x
         | expr.App type_ZZ type_Z f x =>
           (* fst or snd *)
           (valid_expr_App1_bool require_casts f)
-            && valid_expr_bool' false false true x
+            && valid_expr_bool' false false false true x
         | expr.App type_ZZ type_ZZ f x =>
           (valid_expr_App1_bool require_casts f)
-            && valid_expr_bool' false false false x
+            && valid_expr_bool' false false false false x
         | expr.Ident _ (ident.Literal base.type.Z z) =>
           is_bounded_by_bool z max_range || negb require_casts
         | expr.Ident _ (ident.Literal base.type.nat n) =>
@@ -210,7 +240,7 @@ Section Expr.
         | _ => false
         end.
 
-  Definition valid_expr_bool {t} := @valid_expr_bool' t false false.
+  Definition valid_expr_bool {t} := @valid_expr_bool' t false false false.
 
   Lemma valid_expr_App1_bool_type {t} rc (e : API.expr t) :
     valid_expr_App1_bool rc e = true ->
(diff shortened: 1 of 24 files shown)


jadephilipoom:
Local test:
```
jadep@toto ~/plv/fiat-crypto/src/ExtractionOCaml (git)-[bedrock2_zselect] % ./bedrock2_unsaturated_solinas --lang=bedrock2 --no-wide-int --widen-carry --widen-bytes --split-multiret --no-select --no-primitives '25519' '5' '2^255 - 19' '64' selectznz
/* Autogenerated: ./bedrock2_unsaturated_solinas --lang=bedrock2 --no-wide-int --widen-carry --widen-bytes --split-multiret --no-select --no-primitives 25519 5 '2^255 - 19' 64 selectznz */
/* curve description: 25519 */
/* requested operations: selectznz */
/* n = 5 (from "5") */
/* s-c = 2^255 - [(1, 19)] (from "2^255 - 19") */

/* Computed values: */
/* carry_chain = [0, 1, 2, 3, 4, 0, 1] */

#include <stdint.h>


/*
 * Input Bounds:
 *   in0: [0x0 ~> 0x1]
 *   in1: [[0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff]]
 *   in2: [[0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff]]
 * Output Bounds:
 *   out0: [[0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff], [0x0 ~> 0xffffffffffffffff]]
 */
void fiat_25519_selectznz(uintptr_t in0, uintptr_t in1, uintptr_t in2, uintptr_t out0) {
  uintptr_t x5, x10, x0, x11, x6, x13, x1, x14, x7, x16, x2, x17, x8, x19, x3, x20, x9, x22, x4, x23, x12, x15, x18, x21, x24, x25, x26, x27, x28, x29;
  /*skip*/
  x0 = *(uintptr_t*)((in1)+((uintptr_t)0ULL));
  x1 = *(uintptr_t*)((in1)+((uintptr_t)8ULL));
  x2 = *(uintptr_t*)((in1)+((uintptr_t)16ULL));
  x3 = *(uintptr_t*)((in1)+((uintptr_t)24ULL));
  x4 = *(uintptr_t*)((in1)+((uintptr_t)32ULL));
  /*skip*/
  x5 = *(uintptr_t*)((in2)+((uintptr_t)0ULL));
  x6 = *(uintptr_t*)((in2)+((uintptr_t)8ULL));
  x7 = *(uintptr_t*)((in2)+((uintptr_t)16ULL));
  x8 = *(uintptr_t*)((in2)+((uintptr_t)24ULL));
  x9 = *(uintptr_t*)((in2)+((uintptr_t)32ULL));
  /*skip*/
  /*skip*/
  x10 = ((uintptr_t)-1ULL)+((in0)==((uintptr_t)0ULL));
  x11 = (x10)^((uintptr_t)18446744073709551615ULL);
  x12 = ((x5)&(x10))|((x0)&(x11));
  x13 = ((uintptr_t)-1ULL)+((in0)==((uintptr_t)0ULL));
  x14 = (x13)^((uintptr_t)18446744073709551615ULL);
  x15 = ((x6)&(x13))|((x1)&(x14));
  x16 = ((uintptr_t)-1ULL)+((in0)==((uintptr_t)0ULL));
  x17 = (x16)^((uintptr_t)18446744073709551615ULL);
  x18 = ((x7)&(x16))|((x2)&(x17));
  x19 = ((uintptr_t)-1ULL)+((in0)==((uintptr_t)0ULL));
  x20 = (x19)^((uintptr_t)18446744073709551615ULL);
  x21 = ((x8)&(x19))|((x3)&(x20));
  x22 = ((uintptr_t)-1ULL)+((in0)==((uintptr_t)0ULL));
  x23 = (x22)^((uintptr_t)18446744073709551615ULL);
  x24 = ((x9)&(x22))|((x4)&(x23));
  x25 = x12;
  x26 = x15;
  x27 = x18;
  x28 = x21;
  x29 = x24;
  /*skip*/
  *(uintptr_t*)((out0)+((uintptr_t)0ULL)) = x25;
  *(uintptr_t*)((out0)+((uintptr_t)8ULL)) = x26;
  *(uintptr_t*)((out0)+((uintptr_t)16ULL)) = x27;
  *(uintptr_t*)((out0)+((uintptr_t)24ULL)) = x28;
  *(uintptr_t*)((out0)+((uintptr_t)32ULL)) = x29;
  /*skip*/
  return;
}
```

src/Bedrock/Proofs/ValidComputable/Expr.v
```diff
@@ -153,54 +153,84 @@ Section Expr.
...
     end.
 
+  Definition is_literalz {t}
+             (e : @API.expr (fun _ => unit) t) (z : Z) : bool :=
+    match e with
+    | expr.Ident _ (ident.Literal base.type.Z x) => Z.eqb x z
+    | _ => false
+    end.
+  Definition valid_expr_select_bool {t}
+             rc (e : @API.expr (fun _ => unit) t) : bool :=
```
JasonGross, line 163: I'm a bit confused by this one.  What's `rc`?  Why not just return `true` in the zselect case and then do `is_zselect && negb rc`?

src/Bedrock/Proofs/ValidComputable/Expr.v
```diff
@@ -153,54 +153,84 @@ Section Expr.
...
+    match e with
+    | expr.Ident _ ident.Z_zselect => negb rc
+    | _ => false
+    end.
+
+  (* TODO: too many boolean arguments, change to a "mode" inductive *)
```
JasonGross, line 169: Why not do this in this PR?

src/Rewriter/Rules.v
```diff
@@ -1056,4 +1056,33 @@ Section with_bitwidth.
...
+               these can be succinctly expressed as 0-c *)
+            (forall rc c,
+                singlewidth (Z.zselect (cstZ rc c)
+                                       (singlewidth ('0))
+                                       (singlewidth ('(2^bitwidth - 1))))
+                = singlewidth (Z.zselect (cstZ rc c)
+                                         (singlewidth ('0))
+                                         (singlewidth ('(2^bitwidth - 1)))))
+            ; (forall rc c x y,
+                  (0 <= bitwidth) ->
```
JasonGross, line 1074:
```suggestion
                  0 <= bitwidth ->
```

src/Bedrock/Proofs/ValidComputable/Expr.v
jadephilipoom: It's going to be a pretty verbose change and deserves to be handled separately, I think.

src/Bedrock/Proofs/ValidComputable/Expr.v
jadephilipoom: `rc` is "require casts", and although it has to be false here it's easier in the proofs if I leave it a variable (I've also done this with e.g. shifts).

JasonGross:
> After this is merged and `--no-select` flags are added to the bedrock2 build targets, #733 should work.

@jadephilipoom This fixes the unsaturated solinas builds, but `opp` still fails to synthesize for all wbw montgomery primes, and `from_montgomery` fails as well for some of them.  The logs with error messages are in #733, see, e.g., [`p224_64.c.tmp`](https://github.com/mit-plv/fiat-crypto/pull/733/files#diff-26b2aa2c1ff27ff58abe64f2dd4ac1ec) for errors in both of them.  It looks like `opp` is failing on lines like
```
  expr_let x2 := -x1[0] (* : uint64_t *) in
```
and it looks like `from_montgomery` is failing on lines like
```
 expr_let x85 := Z.ltz(0, (uint64_t)(-(Z.ltz((uint64_t)x77, (uint64_t)x83) + Z.ltz((uint64_t)x83, (uint64_t)x84)))) (* : uint64_t *) in
```
if I'm doing the translation correctly in my head (the error line is `  x87 = ((uintptr_t)0ULL)<(ERROR);`)

jadephilipoom:
That's because there's no translation for `opp` right now. Should be straightforwardly implementable as `0-x`, though. I'll fix it.

Issue #764, opened 2020-04-21
Clean up/automate bedrock2 tests


jadephilipoom:
The code in `Bedrock/Tests/` is maturing into less of a one-off sanity check and more of a suite of examples. For that purpose, it should be a lot less sloppy than it is right now. In particular, it should have:

- [ ] a once-and-for-all proof that composes the bedrock2 backend correctness proof with the bounds pipeline correctness proof to go straight from the specification to bedrock2
- [ ] automation allowing people to easily create and prove correct full new suites of field operations by just plugging in the compile-time parameters
- [ ] usage examples that compose operations (ideally up to curve-arithmetic level of abstraction)

PR #765, opened 2020-04-21
Bedrock2 poly1305 test with proofs

### LINKED ISSUE #761: feature request: bedrock2 32-bit poly1305

jadephilipoom:
Closes #761

Easy to add, just a copy-pasting of 32-bit curve25519. Right now, like the other tests, it translates multiplication and addition and has proofs only about multiplication. There's no reason that adding more would be complicated except that I'd like to make stuff nice and clean and automated before I do more copy-pasting (see #764).


andres-erbsen:
Makes sense. :+1:

PR #767, opened 2020-04-21
Generate all functions for solinas bedrock2 files

JasonGross:
Also generate more functions for montgomery bedrock2 files.  Generating
opp still fails for all primes, generating from_montgomery still fails
for some primes.


jadephilipoom:
Try rebasing on master now -- f3a74f9 should have fixed the opp translation.

JasonGross:
Indeed, it works!  I'll close this in favor of the other PR

PR #782, opened 2020-05-06
Add support for comments and stripping annotations

### LINKED ISSUE #779: Add support for comments in the output languages
> For proposal two, would that let me write
> `Definition f (x y : Z) := let __ := note "called with" (x, y) in x+x.` and have `// note: called with (1, 2) ` in the output? That'd be awesome.

_Originally posted by @andres-erbsen in https://github.com/mit-plv/fiat-crypto/issues/778#issuecomment-623182524_

----

> Hm, I'd have to add support for `string` as a type; I could do that. You'd also need `dlet` rather than `let`, and I'd need to do some munging to get it to pretty-print the AST, but, yes, I think I could make your `note` identifier work. Let me think a bit more about it, and try some things. I'll make a separate issue.

_Originally posted by @JasonGross in https://github.com/mit-plv/fiat-crypto/issues/778#issuecomment-623183992_

JasonGross:
After `Import Crypto.Language.PreExtra`, it's now possible to do `dlet _
:= ident.comment stuff in ...` and have `stuff` appear in the
synthesized code as a comment.  There's room for improvement with
regards to quoting and handling multi-line comments, but it works for
debugging purposes.

Note that bounds analysis does not add casts underneath comments, though
I believe let-lifting still happens, and the comments are still
definitely subject to rewriting.

If you want comments whose variable dependencies are eliminated during
the very last dead-code-elimination phase of the stringification
pipeline, use `ident.comment_no_keep`.  Both `ident.comment` and
`ident.comment_no_keep` have type `forall {A}, A -> unit`.

There are now also passes `StripAnnotations` and `StripAllAnnotations`;
the former takes an expression and input bounds, and strips cast
annotations which are guaranteed to be no-ops given the bounds; it's
proven to preserve `Wf` and `Interp`-relatedness.  The
`StripAllAnnotations` pass strips *all* cast annotations, even the ones
that might be doing necessary truncation or modulo computations, and
does not require input bounds.  It's proven to preserve `Wf` but has no
`Interp` proofs.

Closes #779

cc @andres-erbsen, what do you think?

### DIFF
--- a/src/AbstractInterpretation/AbstractInterpretation.v
+++ b/src/AbstractInterpretation/AbstractInterpretation.v
@@ -37,7 +37,7 @@ Module Compilers.
       Context (abstract_domain' : base_type -> Type)
               (annotate : forall (is_let_bound : bool) t, abstract_domain' t -> @expr var t -> UnderLets (@expr var t))
               (bottom' : forall A, abstract_domain' A)
-              (abstract_interp_ident : forall t, ident t -> type.interp abstract_domain' t).
+              (skip_annotations_under : forall t, ident t -> bool).
 
       Definition abstract_domain (t : type)
         := type.interp abstract_domain' t.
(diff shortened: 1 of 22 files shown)


andres-erbsen:
I looked at the test and it seems reasonable. Is there anything else you'd like me to review?

JasonGross:
Nope, just the test seems good

Issue #785, opened 2020-05-09
to_montgomery and feedback on rust code


mimoo:
Hello hello.

As @elichai pointed out, there is no `to_montgomery` function produced which limits how fiat-crypto can be used. I'm currently implementing ECDSA with P-256 in Rust and was hopping to use fiat-crypto for this.

Some other feedback about the generated rust:

* documentation before a function should be using `//!` so that they are displayed when generating documentation (via `cargo doc`)
* most functions are not efficient at the moment because the input and the output needs to be two different bytearrays. Indeed in rust you cannot do something like:
  ```
  let a = [0u32; 4];
  operation(&mut a, &a);
  ```
  so to use the fiat operations I have to create a new bytearray to store the result every time. This might get optimized out by the compiler but I'm not sure. I have no real short-term solution here, but an idea would be to provide two different APIs: one that provide the result in place (and for this to happen all functions need to be implemented as methods on a wrapper struct), and the current one.

JasonGross:
> As @elichai pointed out, there is no `to_montgomery`

Should be fixed by #786, see also #784.

> * documentation before a function should be using `//!` so that they are displayed when generating documentation (via `cargo doc`)

I'll make a PR for this, thanks.

>  This might get optimized out by the compiler but I'm not sure

Can you check?  It seems like the compiler should be able to optimize this out; the functions are all declared `#[inline]`, and don't assign to the output until the very end of the function, so it seems like if you define your own wrapper function for each of these, that just allocates the temporary arrays and does copying and then calls the fiat-crypto inline functions, the compiler should be able to optimize out the intermediate allocation.  (And, if not, this seems like it should be a request for the Rust compiler, not for fiat-crypto.)

JasonGross:
I'm going to consider this issue closed with the merging of #787.  I'll draft a new release now.  Feel free to open a new issue about wrapper methods if you want to, and think it should be something we handled rather than having the Rust compiler handling.

PR #787, opened 2020-05-09
Use Rust docstring comments

JasonGross:
Addresses part of #785

@mimoo , is this good?

Note that this conflicts with #786, and whichever is merged second will need to regenerate the rust files to include both changes.


fiat-rust/src/curve25519_64.rs
```diff
@@ -17,20 +17,18 @@ pub type fiat_25519_u2 = u8;
...
+ //!
+ //! Input Bounds:
+ //!   arg1: [0x0 ~> 0x1]
+ //!   arg2: [0x0 ~> 0x7ffffffffffff]
+ //!   arg3: [0x0 ~> 0x7ffffffffffff]
+ //! Output Bounds:
+ //!   out1: [0x0 ~> 0x7ffffffffffff]
+ //!   out2: [0x0 ~> 0x1]
 #[inline]
 pub fn fiat_25519_addcarryx_u51(out1: &mut u64, out2: &mut fiat_25519_u1, arg1: fiat_25519_u1, arg2: u64, arg3: u64) -> () {
```
JasonGross, line 33: @mimoo this errors with
```
error: expected outer doc comment
  --> src/curve25519_64.rs:20:2
   |
20 |  //! The function fiat_25519_addcarryx_u51 is an addition with carry.
   |  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: inner doc comments like this (starting with `//!` or `/*!`) can only appear before items

error: aborting due to previous error

error: could not compile `fiat-crypto`.
```
How is it supposed to look?

mimoo: I'm so sorry, I meant `///` not `//!` :D 

`//!` is supposed to be used at the start of the file to provide "module documentation"

mimoo: See this for an example: https://github.com/dalek-cryptography/curve25519-dalek/blob/master/src/scalar.rs

and the corresponding generated doc: https://doc.dalek.rs/curve25519_dalek/scalar/struct.Scalar.html

JasonGross: Thanks!

PR #789, opened 2020-05-11
Allow more array sizes in bedrock2 backend

### LINKED ISSUE #771: generated bedrock2 code says load_word when load 1 byte is indended
in `from_bytes` this causes incorrect values to be loaded (and probably undefined behavior at loading the last word). (from https://github.com/mit-plv/bedrock2/pull/155#issue-411559079)

jadephilipoom:
Fixes #771

The solution implemented here is pretty much exactly as described in https://github.com/mit-plv/fiat-crypto/issues/771#issuecomment-622944404 . Not too complicated really, just a lot of threading through new requirements.

I've tested the code by putting the generated bedrock2 for curve25519 64-bit into BoringSSL instead of the generated C and running their test suite. Before it was failing because `to_bytes` and `from_bytes` had the wrong function signatures, but it now works, so I think we can declare victory.


JasonGross:
I wonder if we can get our CI to run BoringSSL's test suite with the C and bedrock2 code...

PR #792, opened 2020-05-12
Display definitions for `eval` and `bytes_eval`

### LINKED ISSUE #791: include the definition of eval in a comment
One thought is that we could print the definition of eval specialized to the particular weight function in a comment at the top. This would have the benefit of being automatically generated, being guaranteed to be in sync with the code, and being precise

https://github.com/mit-plv/fiat-crypto/issues/785#issuecomment-626473878

JasonGross:
Closes #791

This should resolve the question of endianess of encoding.

@andres-erbsen (and/or @jadephilipoom) Please review the format of the additional comments in the C files, and let me know if this seems good to you.  (Some particular questions: Is putting these under "Computed values" good?  Is it okay to display large numbers in hex, or should I make sure that all shifts get displayed in decimal?)


andres-erbsen:
Please parenthesize like (a<<7)+42, and perhaps let's more generally avoid assuming anything about relative precedence of bitwise and arithmetic operators in printing. https://en.cppreference.com/w/c/language/operator_precedence has + bind tighter than << for example.

I think hex is fine.

JasonGross:
@andres-erbsen Should be fixed now.  (Technically we're printing `eval` and `bytes_eval` as Gallina rather than C, but I've updated it to pretty much always parenthesize things.  Unfortunately, this means that `a & b & c` and `a | b | c` will now always be parenthesized either to the left or to the right.  Fortunately, we'll no longer have somewhat ambiguous expressions like `a & b | c & d | f` without parentheses.)

JasonGross:
Merging based on ack on signal

PR #793, opened 2020-05-18
[coq] Adapt to coq/coq#11980 : Improvements on Print Assumptions

ejgallego:
Please merge [for Coq's master CI].

### DIFF
--- a/output-tests/Crypto.Fancy.Barrett256.prod_barrett_red256_correct.Assumptions.expected
+++ b/output-tests/Crypto.Fancy.Barrett256.prod_barrett_red256_correct.Assumptions.expected
@@ -1,9 +1,4 @@
 Axioms:
-FunctionalExtensionality.functional_extensionality_dep : forall (A : Type)
-                                                           (B : A -> Type)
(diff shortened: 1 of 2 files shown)


JasonGross:
This is not backwards compatible.  Perhaps instead there are some options we can pass to `diff` to get it to ignore whitespace including new lines?  Or else can we emit a `Set Printing Width` command to make `Print Assumptions` print each type on a single line?

ejgallego:
> This is not backwards compatible. Perhaps instead there are some options we can pass to `diff` to get it to ignore whitespace including new lines? Or else can we emit a `Set Printing Width` command to make `Print Assumptions` print each type on a single line?

I'm afraid we cannot provide user-level output compatibility upstream; I think the simplest approach is to only run output tests for Coq `master`

JasonGross:
Let's try #794 instead

ejgallego:
> Let's try #794 instead

Thanks @JasonGross , that seems a bit overkill tho; maybe you could use something like `patdiff` if you don't care about spacing?

PR #794, opened 2020-05-18
[coq] Backwards-compatible: Adapt to coq/coq#11980 : Improvements on Print Assumptions

JasonGross:
Backwards compatible version of #793

Closes #793


Zimmi48:
Looks like CI passed.

JasonGross:
I will merge once I'm at a computer and can strip out the intervening wip/debugging commits.

Issue #797, opened 2020-05-20
p448_32 fails to build on bedrock2 backend


JasonGross:
There are a lot of Z.combine_at_bitwidth operations left over.  Not sure why, I'll try to dig into this.

JasonGross:
Ah, I've figured it out, it's because there are operations like
```
expr_let x962 := Z.combine_at_bitwidth(32, (uint32_t)x779, (uint32_t)x780) + Z.combine_at_bitwidth(32, (uint32_t)x958, (uint32_t)x961) (* : [0x0 ~> 0x1055c28e3f0a3d758] *) in
```
where the sum fails to fit within bounds.

JasonGross:
See also https://github.com/mit-plv/fiat-crypto/pull/798

JasonGross:
From discussions with @andres-erbsen, the reason is that our loose bounds are 3 times our tight bounds, which are 1.1 times the limb-width.  (This is coded for in https://github.com/mit-plv/fiat-crypto/blob/99e2ae04406fe321f4f2a8ad95d12a8f053ad65e/src/UnsaturatedSolinasHeuristics.v#L57-L58 )

Andres said that the relevant paper for p448 is Goldilocks and Ridinghood by Michael Hamburg:
![](https://user-images.githubusercontent.com/396076/82395902-8e924800-9a1a-11ea-836c-c3a06c403225.png)
![](https://user-images.githubusercontent.com/396076/82395928-9d78fa80-9a1a-11ea-9078-0feb06081892.png)

If we adjust the loose bounds multiplier to 2, though, that doesn't work, because subtraction adds balance.  In curve25519 x64, this is
https://github.com/mit-plv/fiat-crypto/blob/99e2ae04406fe321f4f2a8ad95d12a8f053ad65e/fiat-c/src/curve25519_64.c#L292-L296
In p448 x32, this is
```c
  uint32_t x1 = ((UINT32_C(0x1ffffffe) + (arg1[0])) - (arg2[0]));
  uint32_t x2 = ((UINTv32_C(0x1ffffffe) + (arg1[1])) - (arg2[1]));
  uint32_t x3 = ((UINT32_C(0x1ffffffe) + (arg1[2])) - (arg2[2]));
  uint32_t x4 = ((UINT32_C(0x1ffffffe) + (arg1[3])) - (arg2[3]));
  uint32_t x5 = ((UINT32_C(0x1ffffffe) + (arg1[4])) - (arg2[4]));
[...]

PR #799, opened 2020-05-20
Use tighter loose bounds

JasonGross:
We now compute the loose bounds to be as tight as possible given the
tight bounds: they are the tightest bounds that will let us add two
tightly-bounded field elements, and will also let us subtract two
tightly-bounded field elements (with balance), without needing to carry.

This is probably required for having 32-bit p448 work, but even with
this, p448 is still not working; see
https://github.com/mit-plv/fiat-crypto/issues/797.  Seems worth merging
anyway.

@andres-erbsen @davidben please review either the changes in the bounds in the comments, or the changes in the definition of `loose_upperbounds` in `src/UnsaturatedSolinasHeuristics.v`, or both.

@jadephilipoom Please review my changes to your proof of `loose_bounds_within_base_access_sizes` in `src/Bedrock/Tests/Proofs/X25519_64.v` (the upper bounds are no longer `repeat r n`, so I had to use a lemma about relaxing them, and then relaxed them to the repetition of the max element).


hannesm:
I came here while reading #797 which I'm interested in. I understand this does not bring fiat-crypto to p448-32, but already goes some way. I was wondering (since this has multiple positive reviews) what is missing to get this PR merged (apart from a rebase)?

JasonGross:
My impression from https://github.com/mit-plv/fiat-crypto/issues/797#issuecomment-631711634 was that determining headroom is more complicated than just "how many additions / subtractions do you want to be able to do in a row without carrying?" and hence the one-line change here (changing `headroom_add_count` from 2 to 1) does not actually make progress conceptually towards handling p448-32.  I did in fact merge the earlier underlying changes, i.e., factoring the computation in a way that lets us just declare how many times we want to be able to add in a row and bounds for that.

I think the thing standing in the way of merging a PR like this is having an understanding of what the right interface to present here is.  Should we compute the loose bounds by asking people to specify how many additions they'd like to perform in a row (and encode it somehow in the types so that users can in fact perform that many additions in a row)?  It's not clear to me what the landscape looks like around choosing bounds.

If it's useful to you, I'm happy to make `headroom_add_count` an argument to the binary, but again I think this is the wrong strategy in the absence of a vision about what should be tunable parameters of the implementation, and what should be computed automagically and how.  Perhaps @bitwiseshiftleft or @kevaundray has thoughts on what a global plan for this should look like.

hannesm:
Thanks for your explanation @JasonGross. Indeed that sounds like this PR may not be relevant for the p448-32 goal. I'll read up on #797 to figure out the missing bits and pieces when I find enough time and headroom.

PR #803, opened 2020-05-21
Fix limb count heuristics

### LINKED ISSUE #801: get_possible_limbs thinks there are no valid limbs for 2^448 - 2^224 - 1
https://github.com/mit-plv/fiat-crypto/blob/ccf644f710b931dad8339aa63c1ce6e95d19a623/src/UnsaturatedSolinasHeuristics.v#L96-L111

```coq
Compute @get_possible_limbs (2^448) [(2^224,1); (1,1)] 64
```
gives `[]`, and similarly for 32 bits.

jadephilipoom:
Fixes #801 

Tests:
```
Compute @get_possible_limbs (2^448) [(2^224,1); (1,1)] 64. (* [8%nat; 10%nat; 13%nat] *)
Compute @get_possible_limbs (2^448) [(2^224,1); (1,1)] 32. (* [18%nat; 20%nat; 21%nat; 22%nat; 23%nat; 24%nat; 25%nat; 26%nat; 27%nat] *)
Compute @get_possible_limbs (2^480) [(2^240,1); (1,1)] 64. (* [8%nat; 10%nat; 11%nat; 12%nat; 13%nat; 14%nat; 15%nat] *)
Compute @get_possible_limbs (2^480) [(2^240,1); (1,1)] 32. (* [18%nat; 20%nat; 22%nat; 23%nat; 24%nat; 25%nat; 26%nat; 27%nat; 28%nat; 29%nat] *)
Compute @get_possible_limbs (2^192) [(2^64, 1); (1, 1)] 64. (* [4%nat; 5%nat] *)
Compute @get_possible_limbs (2^192) [(2^64, 1); (1, 1)] 64. (* [9%nat; 10%nat; 11%nat] *)
```

### DIFF
--- a/src/UnsaturatedSolinasHeuristics.v
+++ b/src/UnsaturatedSolinasHeuristics.v
@@ -97,14 +97,11 @@ Section ___.
           (c : list (Z * Z))
           (machine_wordsize : Z).
   (** given a parsed prime, pick out all plausible numbers of (unsaturated) limbs *)
-  (** we want to leave enough bits unused to do a full solinas
-      reduction without carrying; the number of bits necessary is the
-      sum of the bits in the negative coefficients of p (other than
-      the most significant digit), i.e., in the positive coefficients
-      of c *)
+  (** an unsaturated implementation will necessarily need at least as many limbs
+      as a saturated one, so search starting there *)
   Let num_bits_p := Z.log2_up s.
-  Let unused_bits := sum (map (fun t => Z.log2_up (fst t)) c).
-  Let min_limbs := Z.to_nat (Qceiling (num_bits_p / (machine_wordsize - unused_bits))).
+  Let nlimbs_saturated := Z.to_nat (Qceiling (num_bits_p / machine_wordsize)).
+  Let min_limbs := nlimbs_saturated.
   (* don't search past 2x as many limbs as saturated representation; that's just wasteful *)
   Let result := filter (fun n => overflow_free n s c machine_wordsize) (seq min_limbs min_limbs).
   Definition get_possible_limbs : list nat


jadephilipoom:
One note here: with the loose bounds multiplier set to 3, some nice possibilities (like 16-limb p448 on 32-bit) are excluded due to overflow. I double checked and it does indeed seem the bounds are too loose in that case (but not with the bounds multiplier set to 2).

JasonGross:
> One note here: with the loose bounds multiplier set to 3, some nice possibilities (like 16-limb p448 on 32-bit) are excluded due to overflow. I double checked and it does indeed seem the bounds are too loose in that case (but not with the bounds multiplier set to 2).

This seems potentially like an argument in favor of #799 (cc @andres-erbsen )

JasonGross:
Should we add test-cases for this?

jadephilipoom:
Yes. I vote for just including a solinas multi-tap prime on our generated
code tests (the smallest one I quoted there, 2^192 - 2^64 - 1, would be a
nice compact one.)

JasonGross:
> Yes. I vote for just including a solinas multi-tap prime on our generated code tests

Oh, heh, I was thinking of doing tests for all of the primes.  The first 40 (out of the 80 in the primes list) take about 30 seconds in `native_compute`, and about 240 seconds in `vm_compute` (@maximedenes might be interested in this speedup).  Unfortunately doing all 80 is prohibitively expensive, but I think doing the first 40 in the native compiler is reasonable.

jadephilipoom:
Does the first 40 include many multi-tap ones, though? iirc they were mostly towards the end of the list.

JasonGross:
Oh, indeed, the multitap primes are the final 19.  But we can do all of these primes in 2.027 seconds in the native compiler.  So maybe we have as a test the first 30 and the last 19?

jadephilipoom:
Sure, although keep in mind the primes with huge coefficients on their multi-taps are designed for Montgomery and might *correctly* not have any options for limb counts.

JasonGross:
Hm, unfortunately, with this PR, the computation is much, much slower.  So I guess we can't get all of them...

JasonGross:
> (the smallest one I quoted there, 2^192 - 2^64 - 1, would be a nice compact one.)

Huh, this one isn't in primes.txt.  Should we add it? https://github.com/mit-plv/fiat-crypto/blob/9ac37a1cc1cfe8d8cc741ed2fad07156f748aeca/primes.txt#L66-L85

JasonGross:
Btw, `2^205 - 45*2^198 - 1` is not listed as a montgomery-specific prime, but there are no options generated for it.  This is the only one of the ones labeled "two or more taps" like this.

JasonGross:
(All of the "two taps, golden ratio" primes get possibilities generated for them, though it takes 50s to generate all the possibilities)

PR #806, opened 2020-05-25
Reorganize generation of files

JasonGross:
There's a bit more Makefile magic now, but it should hopefully be easier
to add new primes and new languages to the Makefile, now that there is
less copy-pasting of boilerplate.

We also add targets `all-except-generated` and `lite-generated-files`
which should help with https://github.com/coq/coq/pull/12405

@andres-erbsen Please comment on if this is too much makefile magic, and if I should add more documentation or do things in some other way.  (The main thing I'm against is having to copy dozens of lines of code for each new prime and/or language that we want to add to the build.)


JasonGross:
We might get rid of this later if we decide it's too much magic

PR #807, opened 2020-05-27
Better computation of loose bounds

JasonGross:
We now compute the loose bounds by recording the number of additions
and/or balanced subtractions we want to be able to do in a row before we
need to multiply and carry.  Note that balanced subtraction always takes
a bit more overhead than addition.

This is a more conservative variant of #799 that doesn't actually change
the bounds, and makes the genuine change of #799 easier to see, namely,
reducing `headspace_add_count` from `2` to `1`.


JasonGross:
@andres-erbsen @jadephilipoom Could you review the changes I made here to the definition of `loose_upperbounds` in `src/UnsaturatedSolinasHeuristics.v`?

PR #808, opened 2020-06-01
Bedrock2 interface improvements

jadephilipoom:
Long-overdue merge of the new bedrock2 interface that I demoed last week. This is probably too big of a change to ask anyone to review, and I've gotten informal code-walkthrough reviews along the way, so I'll consider myself free to merge this once the CI passes. If anyone wants to take a look, though, it would be welcome.


src/Bedrock/Arrays/ByteBounds.v
```diff
@@ -0,0 +1,141 @@
...
+Local Open Scope Z_scope.
+Import ListNotations.
+Import Partition.
+
+Section ByteBounds.
+  Context {p : Types.parameters} {ok : Types.ok}.
+  Context (n : nat).
+
+  Definition byte_range : ZRange.zrange :=
+    {|ZRange.lower:=0; ZRange.upper:=2^8-1|}.
```
JasonGross, line 26: If you `Require Import Crypto.Util.ZRange`, you can use the `r[0 ~> 2^8-1]%zrange` notation, which might be nicer.

src/Bedrock/Arrays/ByteBounds.v
```diff
@@ -0,0 +1,141 @@
...
+  Definition byte_bounds : list (option ZRange.zrange) :=
+    repeat (Some byte_range) n.
+
+  Lemma byte_bounds_range_iff x :
```
JasonGross, line 30: I think this proof should be split into the following lemmas, some of which we probably already have:
- a lemma in FoldBool that says `fold_andb_map f x y = true <-> Forall2 f x y`
- a lemma, maybe in ListUtil, that says `Forall2 P x y <-> (length x = length y /\ Forall (fun '(a, b) => P a b) (combine x y))`
- a lemma that we probably already have in ListUtil saying `Forall P x <-> forall n v, nth_error x n = Some v -> P x`
- a lemma that we almost certainly already have in ListUtil saying `nth_error (combine x y) n = match nth_error x n, nth_error y n with ...`
- a lemma that we maybe already have saying `nth_error (repeat v n) m = if (m <=? n)%nat then Some v else None`

Then most of this lemma can be taken care of by just using `setoid_rewrite` with these lemmas, and I expect the final but of reasoning to be pretty short.  It might be a bit more work proving all of these lemmas, but I think most of them already exist, and adding the ones that don't should make proofs like this easier in the future.

src/Bedrock/Arrays/MaxBounds.v
```diff
@@ -0,0 +1,120 @@
...
+
+  Existing Instance semantics_ok.
+
+  (* TODO: make Expr.v import this *)
+  Definition max_range : ZRange.zrange :=
+    {|ZRange.lower:=0; ZRange.upper:=2^Semantics.width-1|}.
+  Definition max_bounds : list (option ZRange.zrange) :=
+    repeat (Some max_range) n.
+
+  Lemma max_bounds_range_iff x :
```
JasonGross, line 31: You should be able to prove this in an almost identical way.  Might be worth factoring the composition of the lemmas I named above, even, so that you can just apply/rewrite with a common lemma.

src/Bedrock/Arrays/MaxBounds.v
```diff
@@ -0,0 +1,120 @@
...
+               | |- _ /\ _ => split
+               | |- S _ = S _ => f_equal
+               | _ => progress Z.ltb_to_lt
+               | _ => congruence
+               | _ => lia
+               end.
+  Qed.
+
+
+  Lemma map_word_wrap_bounded' r x m :
```
JasonGross, line 68: Might be worth generalizing this so you're not proving the same thing twice?

src/Bedrock/Interfaces/Tactics.v
```diff
@@ -0,0 +1,329 @@
...
+         end.
+
+Ltac next_argument :=
+  (exists 1%nat); sepsimpl; cbn [firstn skipn];
+  [ solve [eauto using firstn_length_le] | ].
+
+Ltac ssubst :=
+  repeat match goal with
+         | H : literal (word.unsigned _) (eq _) |- _ =>
+           inversion H as [H']; clear H;
```
JasonGross, line 72: You probably want to let H' := fresh before this, otherwise it'll fail to use this case if H' exists

src/Bedrock/Interfaces/Tactics.v
```diff
@@ -0,0 +1,330 @@
...
+                        | exists ptr ]
+         end.
+Ltac exists_all_placeholders out_array_ptrs :=
+  lazymatch out_array_ptrs with
+  | cons ?p ?ptrs' =>
+    let bits := get_value_of_pointer p in
+    let zs :=
+        lazymatch type of bits with
+        | list word.rep => constr:(map word.unsigned bits)
+        | list Byte.byte => constr:(map byte.unsigned bits)
```
JasonGross, line 274: Please remove the warning by doing
```suggestion
        | list Init.Byte.byte => constr:(map byte.unsigned bits)
```

src/Bedrock/Interfaces/Tactics.v
```diff
@@ -0,0 +1,330 @@
...
+        | ?t => fail "unexpected placeholder type " t
+        end in
+    let words :=
+        lazymatch type of bits with
+        | list word.rep => constr:(bits)
+        | list Byte.byte =>
```
JasonGross, line 280: Please remove the warning by doing
```suggestion
        | list Init.Byte.byte =>
```

src/Bedrock/Interfaces/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,759 @@
...
+    Context {ok : Types.ok}.
+    Existing Instance semantics_ok.
+
+    Lemma Bignum_of_bytes n addr bs :
+      length bs = (n * Z.to_nat word_size_in_bytes)%nat ->
+      Lift1Prop.iff1
+        (array ptsto (word.of_Z 1) addr bs)
+        (Bignum addr (map word.of_Z
+                          (eval_bytes (width:=Semantics.width) bs))).
+    Admitted. (* TODO *)
```
JasonGross, line 110: Are you planning to remove this Admitted before merging?

src/Bedrock/Interfaces/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,759 @@
...
+
+    Lemma Bignum_to_bytes n addr x :
+      list_Z_bounded_by (max_bounds n) (map word.unsigned x) ->
+      Lift1Prop.iff1
+        (Bignum addr x)
+        (array ptsto (word.of_Z 1) addr (encode_bytes x)).
+    Admitted. (* TODO *)
```
JasonGross, line 117: Are you planning to remove this `Admitted` before merging?

src/Bedrock/Interfaces/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,759 @@
...
+       (Qnum (inject_Z (Z.log2_up M) / inject_Z (Z.of_nat n)))
+       (QDen (inject_Z (Z.log2_up M) / inject_Z (Z.of_nat n)))).
+  Local Notation eval := (eval weight n).
+
+  (* for to/from bytes *)
+  Local Definition limbwidth :=
+    (Z.log2_up (s - Associational.eval c) / Z.of_nat n)%Q.
+  Local Definition n_bytes :=
+    (Freeze.bytes_n (Qnum limbwidth) (Qden limbwidth) n).
+  Let prime_bytes_upperbound_list :=
```
JasonGross, line 155: Should the PushButtonSynthesis files export more of these things, perhaps?

src/Bedrock/Interfaces/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,759 @@
...
+    | ?b => fail "unable to select access size for bound " b
+    end.
+
+  Ltac select_length bounds :=
+    lazymatch bounds with
+    | Some saturated_bounds => constr:(n)
+    | Some loose_bounds => constr:(n)
+    | Some tight_bounds => constr:(n)
+    | Some prime_bytes_bounds => constr:(n_bytes)
+    | ?b => fail "unable to select array length for bound " b
```
JasonGross, line 176:
```suggestion
    | ?b => fail "unable to select array length for bound" b
```
The space is inserted automatically, no need to double it

src/Bedrock/Interfaces/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,759 @@
...
+    map (fun v : Z => Some {| ZRange.lower := 0; ZRange.upper := v |})
+        prime_bytes_upperbound_list.
+
+  Ltac select_access_size bounds :=
+    lazymatch bounds with
+    | Some saturated_bounds => constr:(access_size.word)
+    | Some loose_bounds => constr:(access_size.word)
+    | Some tight_bounds => constr:(access_size.word)
+    | Some prime_bytes_bounds => constr:(access_size.one)
+    | ?b => fail "unable to select access size for bound " b
```
JasonGross, line 167: This inserts two spaces.  Instead do:
```suggestion
    | ?b => fail "unable to select access size for bound" b
```

src/Bedrock/Interfaces/UnsaturatedSolinas.v
jadephilipoom: No, this will be a later merge. It's not actually a dependency for the interface, and doesn't show up in Print Assumptions for any of the examples. It's just something to make it easier for a caller to convert unassigned memory to Bignum-sized chunks.

src/Bedrock/Interfaces/UnsaturatedSolinas.v
jadephilipoom: Same as previous comment.

src/Bedrock/Interfaces/UnsaturatedSolinas.v
jadephilipoom: That would be convenient, yeah -- particularly `n_bytes` and `prime_bytes_bounds`. I just compute `limbwidth` and `prime_bytes_upperbound_list` on the way to those.

JasonGross: I've just pushed 14f9ea3a7042ddd3d6a00498772adf3e50b850da exposing these

src/Bedrock/Interfaces/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,759 @@
...
+
+  Ltac prove_operation_correctness :=
+    intros;
+    let Hcorrect :=
+        lazymatch goal with H : _ = ErrorT.Success _ |- _ => H end in
+    apply_correctness_in Hcorrect; [ | eassumption .. ];
+    fold weight in *; specialize_to_args Hcorrect;
+    postcondition_from_correctness.
+
+  Ltac make_operation inbounds outbounds out :=
```
JasonGross, line 190: With 14f9ea3a7042ddd3d6a00498772adf3e50b850da, you should be able to do
```suggestion
  Ltac make_operation out :=
    let inbounds := (eval cbv beta in (BoundsPipeline.Pipeline.arg_bounds_of_pipeline out)) in
    let outbounds := (eval cbv beta in (BoundsPipeline.Pipeline.out_bounds_of_pipeline out)) in
```

src/Bedrock/Interfaces/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,759 @@
...
+    Context (inname_gen_unique : unique inname_gen)
+            (outname_gen_unique : unique outname_gen).
+
+    Lemma relax_to_max_bounds x :
+      list_Z_bounded_by loose_bounds x ->
+      list_Z_bounded_by (max_bounds n) x.
+    Proof. apply relax_list_Z_bounded_by; auto. Qed.
+
+    (* TODO : this proof is going to be way more annoying than it needed to be,
+    since there doesn't appear to be any encode_no_reduce_partitions *)
```
JasonGross, line 547: Is there a plan to add `encode_no_reduce_partitions`?

src/Bedrock/Names/VarnameGenerator.v
```diff
@@ -0,0 +1,101 @@
...
+(* TODO: move this? *)
+Lemma append_eq_r_iff s1 s2 s3 :
+  s1 ++ s2 = s1 ++ s3 <-> s2 = s3.
+Proof.
+  induction s1; cbn [append]; split;
+    try inversion 1; intros; auto; [ ].
+  apply IHs1. auto.
+Qed.
+
+(* TODO: move this? *)
```
JasonGross, line 17: Yes, this belongs in the string util file

src/Bedrock/Synthesis/ReifiedOperation.v
```diff
@@ -0,0 +1,59 @@
...
+  | |- Func.valid_func _ => eapply valid_func_bool_iff
+  | _ => idtac
+  end;
+  [ match goal with
+    | |- _ = make_bedrock_func _ _ _ =>
+      vm_compute; reflexivity end | .. ];
+  match goal with
+  | |- expr.Wf3 _ => abstract (prove_Wf3 ())
+  | |- valid_func_bool ?x = true =>
+    abstract (vm_compute; reflexivity)
```
JasonGross, line 56: You can get a 2x improvement from branch (not sure if it matters) by writing
```suggestion
    abstract vm_cast_no_check (eq_refl true)
```

src/Bedrock/Synthesis/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,444 @@
...
+           spec_of_sub := sub_spec;
+           spec_of_opp := opp_spec;
+           spec_of_selectznz := selectznz_spec;
+           spec_of_to_bytes := to_bytes_spec;
+           spec_of_from_bytes := from_bytes_spec |}.
+
+Ltac handle_easy_preconditions :=
+  lazymatch goal with
+  | |- ZRange.type.option.is_tighter_than _ _ = true =>
+    abstract (vm_compute; reflexivity)
```
JasonGross, line 330:
```suggestion
    abstract vm_cast_no_check (eq_refl true)
```

src/Bedrock/Util.v
```diff
@@ -658,10 +740,230 @@ Section Maps.
     let H := fresh in
     intro H; specialize (H k); destruct H.
     { tauto. }
-    { auto. } 
+    { auto. }
   Qed.
 End Maps.
 
+(* proofs about list_Z_bounded_by *)
```
JasonGross, line 747: These belong in some other file, I think.  Maybe `src/COperationSpecifications.v`, or maybe a new file in `src/Util`?

src/Bedrock/Interfaces/UnsaturatedSolinas.v
jadephilipoom: There's a plan to replace all instances of `encode_no_reduce` with `partition` -- the former is an early version before we did saturated arithmetic and needed something more robust. I have a local branch that does this, but have been waiting on merging this PR to apply it.

src/Bedrock/Util.v
jadephilipoom: Is there a reason `COperationSpecifications.v` is using its own homegrown `is_tighter_than` instead of `ZRange.type.option.is_tighter_than`?

jadephilipoom: Hmm, I prefer it here to either of those locations. It depends on AbstractInterpretation (for the bounds-relaxation proof), so I don't think it should go in `src/Util`, but it also seems very out of place in `COperationSpecifications`. By the way, why is `COperationSpecifications` using its own version of `is_tighter_than` in the first place, instead of `ZRange.type.option.is_tighter_than`?

JasonGross: I didn't want COperationSpecifications to depend on any of the PHOAS / compilers stuff; I wanted it to have relatively minimal dependencies

jadephilipoom:
The Windows build failure looks very much unrelated (it's failing to build dependencies), so I'm going to go ahead and merge.

PR #813, opened 2020-06-17
encode_no_reduce -> partition

### LINKED ISSUE #468: NewPipeline Arithmetic: partition vs encode
@jadephilipoom  Why do both `partition` and `encode` exist? Is there actually a need for both variants, or can we get rid of one of them? Oh, and there is also `recursive_partition`...

jadephilipoom:
Very belatedly fixes #468 

When we did the unsaturated version of the core arithmetic library, we created `encode` (which converts a number to limbs by putting the number all in the first digit and then carrying) and also `encode_no_reduce`, which does the same thing except omitting the reduction step in carrying so the number doesn't get reduced mod M. `encode_no_reduce` was only used in a few places, e.g. for bounds, and is easily superseded by the more robust and easy to reason about `partition` definition that we wrote for saturated arithmetic. This PR replaces it with `partition` and uses that replacement to prove a couple of admits I had been waiting on the change for.

Note: this change means `chained_carries_no_reduce` is no longer used in `Core.v` and should probably be moved to `BaseConversion.v`. I'm hoping I can also remove all the enormous proofs about `nth_default` and carries from `Core.v`, which are major culprits to adding time and complexity to that file, but that will be a different PR.

### DIFF
--- a/src/Bedrock/Util.v
+++ b/src/Bedrock/Util.v
@@ -154,6 +154,13 @@ Section Lists.
       constructor; auto.
   Qed.
 
+  Lemma Forall_repeat (R : A -> Prop) n x :
+    R x -> Forall R (repeat x n).
+  Proof.
+    induction n; intros; cbn [repeat];
+      constructor; auto.
+  Qed.
+
   Section Partition.
     Fixpoint partition_equal_size' {T}
              (n : nat) (xs acc : list T) (i : nat)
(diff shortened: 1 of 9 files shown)


src/Bedrock/Util.v
```diff
@@ -812,6 +819,92 @@ Section ListZBoundedBy.
...
+    (match b0 with
+     | Some r =>
+       ZRange.is_bounded_by_bool x0 r = true
+     | None => True
+     end /\ list_Z_bounded_by bs xs).
+  Proof.
+    cbv [list_Z_bounded_by]. cbn [FoldBool.fold_andb_map].
+    cbv [ZRange.is_bounded_by_bool].
+    split; destruct b0; intros;
+      repeat match goal with
```
JasonGross, line 834: I think you might be able to replace this entire part of the proof with `destruct b0; rewrite !Bool.andb_true_iff; firstorder.` (maybe even intuition rather than firstorder)?

src/Bedrock/Util.v
```diff
@@ -812,6 +819,92 @@ Section ListZBoundedBy.
...
+    (match b0 with
+     | Some r =>
+       ZRange.is_bounded_by_bool x0 r = true
+     | None => True
+     end /\ list_Z_bounded_by bs xs).
+  Proof.
+    cbv [list_Z_bounded_by].
+    rewrite FoldBool.fold_andb_map_snoc.
+    cbv [ZRange.is_bounded_by_bool].
+    split; destruct b0; intros;
```
JasonGross, line 856: Same as above

PR #814, opened 2020-06-18
Fix w.r.t. coq/coq#12532.

ppedrot:
Should be backwards compatible in theory.

### DIFF
--- a/src/AbstractInterpretation/Proofs.v
+++ b/src/AbstractInterpretation/Proofs.v
@@ -628,6 +628,8 @@ Module Compilers.
                           | match goal with
                             | [ |- bool ] => constructor (* ??? *)
                             end ].
+      Unshelve.
+      all: match goal with [ |- bool ] => constructor (* ??? *) end.
       Qed.
 
       Lemma interp_eval_with_bound'


JasonGross:
I'm thoroughly confused by this PR.  Why is there a change from coq/coq#12532?

JasonGross:
Also, I think I'd prefer wrapping the entire `repeat` in another `repeat unshelve`, but I can go do this

ppedrot:
> Why is there a change from coq/coq#12532?

I don't know for sure, I believe there is a slight difference in shelving. The previous code was using a very weird sequence likely to lose evar data, while the new one is using a code similar to auto and typeclasses eauto.

JasonGross:
I've tracked down where the initial `???` bit came from, and removed it with afd4506607e2eeedb6201809034718717e37de94 .

I'm still very confused by this PR, though.  Is `eauto` now solving more goals because it's leaving more things shelved?

JasonGross:
Should be fixed by d5cfd788354c09cd93ba78922af1700934d8643d (hopefully I haven't broken anything in pushing it)

Issue #708, opened 2020-03-30
Add C tests to the CI


JasonGross:
Maybe we can clone BoringSSL and compile it's test-suite?  Or probably there's a better way?  Maybe?  cc @andres-erbsen @davidben

andres-erbsen:
We used to have C tests before the new pipeline. Perhaps we can just resurrect them?

JasonGross:
Yes, that would be ideal

PR #819, opened 2020-06-24
Test C and bedrock2 code with BoringSSL

JasonGross:
Significant work on #708

@davidben Please review for this being the correct way to patch, build, and run tests on BoringSSL.

### DIFF
--- a/.github/workflows/c.yml
+++ b/.github/workflows/c.yml
@@ -36,3 +36,4 @@ jobs:
       run: EXTRA_CFLAGS="" etc/ci/test-fiat-c-boringssl.sh fiat-c/src
     - name: BoringSSL bedrock2 test
       run: EXTRA_CFLAGS="$(make bedrock2-extra-cflags SKIP_INCLUDE=1 2>/dev/null)" etc/ci/test-fiat-c-boringssl.sh fiat-bedrock2/src
+      continue-on-error: true
--- a/.github/workflows/c.yml
+++ b/.github/workflows/c.yml
@@ -19,9 +19,20 @@ jobs:
       run: |
         sudo sed -i 's/azure\.//' /etc/apt/sources.list
         sudo apt-get -o Acquire::Retries=30 update -q
-        sudo apt-get -o Acquire::Retries=30 install g++-7 libssl-dev -y --allow-unauthenticated
+        sudo apt-get -o Acquire::Retries=30 install -y --allow-unauthenticated \
+             g++-7 libssl-dev \
+             ninja-build libunwind-dev cmake
         sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 60 --slave /usr/bin/g++ g++ /usr/bin/g++-7
+    - name: Set up Go 1.13
+      uses: actions/setup-go@v2-beta
+      with:
+        go-version: 1.13
+      id: go
     - name: make only-test-c-files CC=gcc
       run: make only-test-c-files CC=gcc EXTERNAL_DEPENDENCIES=1
     - name: make only-test-bedrock2-files CC=gcc
       run: make only-test-bedrock2-files CC=gcc EXTERNAL_DEPENDENCIES=1
+    - name: BoringSSL C test
+      run: EXTRA_CFLAGS="" etc/ci/test-fiat-c-boringssl.sh fiat-c/src
+    - name: BoringSSL bedrock2 test
+      run: EXTRA_CFLAGS="$(make bedrock2-extra-cflags SKIP_INCLUDE=1 2>/dev/null)" etc/ci/test-fiat-c-boringssl.sh fiat-bedrock2/src
--- a/etc/ci/test-fiat-c-boringssl.sh
+++ b/etc/ci/test-fiat-c-boringssl.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+# USAGE: $0 SUBCOMPONENT (e.g., fiat-c/src)
+
+################################################################################
+# Tests for BoringSSL
+################################################################################
+echo "::group::Cloning BoringSSL"
+({
+    set -ex
+    rm -rf boringssl
+    git clone https://boringssl.googlesource.com/boringssl || exit $?
+}) || exit $?
+echo "::endgroup::"
+
+SUBCOMPONENT="$1"
+if [ -z "$SUBCOMPONENT" ]; then
+    echo "ERROR: Missing argument"
+    echo "USAGE: $0 SUBCOMPONENT"
+    echo "Example SUBCOMPONENTS include fiat-c/src, fiat-bedrock2/src"
+    exit 1
+fi
+SUBCOMPONENT_PATH="$(cd "$SUBCOMPONENT" && pwd)"
+
+pushd boringssl >/dev/null
+
+echo "::group::Patching BoringSSL"
+({
+    set -ex
+    ( cd third_party/fiat && for i in *.h; do cp "${SUBCOMPONENT_PATH}/${i/.h/.c}" "$i" || exit $?; done ) || exit $?
+    ( cd third_party/fiat && git --no-pager diff )
+}) || exit $?
+echo "::endgroup::"
+
+echo "::group::Building BoringSSL"
+({
+    set -ex
+    mkdir build
+    cd build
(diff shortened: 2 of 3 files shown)


.github/workflows/c.yml
```diff
@@ -19,9 +19,15 @@ jobs:
       run: |
         sudo sed -i 's/azure\.//' /etc/apt/sources.list
         sudo apt-get -o Acquire::Retries=30 update -q
-        sudo apt-get -o Acquire::Retries=30 install g++-7 libssl-dev -y --allow-unauthenticated
+        sudo apt-get -o Acquire::Retries=30 install -y --allow-unauthenticated \
+             g++-7 libssl-dev \
+             ninja-build libunwind-dev cmake
```
davidben, line 24: You'll also need a copy of Go. Or does the CI have that by default?

.github/workflows/c.yml
```diff
@@ -19,9 +19,15 @@ jobs:
...
         sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 60 --slave /usr/bin/g++ g++ /usr/bin/g++-7
     - name: make only-test-c-files CC=gcc
       run: make only-test-c-files CC=gcc EXTERNAL_DEPENDENCIES=1
     - name: make only-test-bedrock2-files CC=gcc
       run: make only-test-bedrock2-files CC=gcc EXTERNAL_DEPENDENCIES=1
+    - name: BoringSSL C test
+      run: etc/ci/test-fiat-c-boringssl.sh fiat-c/src
+    - name: BoringSSL bedrock2 test
+      run: etc/ci/test-fiat-c-boringssl.sh fiat-bedrock2/src
```
davidben, line 33: I'm not familiar with GitHub workflows, so a sanity-check: if this starts failing but not the others, it won't be super confusing/disruptive?

I can't promise it won't accidentally break due to a BoringSSL issue or because we've moved some files around. If that happens, I'll certainly work with you to fix it, but I don't want that to halt all work on your end.

etc/ci/test-fiat-c-boringssl.sh
```diff
@@ -0,0 +1,54 @@
...
+
+pushd boringssl >/dev/null
+
+echo "::group::Patching BoringSSL"
+({
+    set -ex
+    ( cd third_party/fiat && for i in *.h; do cp "${SUBCOMPONENT_PATH}/${i/.h/.c}" "$i" || exit $?; done ) || exit $?
+    ( cd third_party/fiat && git --no-pager diff )
+
+    sed 's/"-Werror"/"-Werror", "-Wno-error=unused-function"/g' -i util/BUILD.toplevel
```
davidben, line 32: If you're using the CMake build, you can ignore `util/BUILD.toplevel`. We have.... too many build systems for reasons. :-(

etc/ci/test-fiat-c-boringssl.sh
```diff
@@ -0,0 +1,54 @@
...
+    sed 's/-Werror/-Werror -Wno-error=unused-function/g' -i CMakeLists.txt
```
davidben, line 33: BTW, if you're outputting a file that's meant to be #included somewhere (given the symbols are `static` anyway), there may be something to be said for the files being named `foo.h` and maybe to mark them with `__attribute__((unused))`. We use this definition:

```
#if defined(__GNUC__) || defined(__clang__)
#define OPENSSL_UNUSED __attribute__((unused))
#else
#define OPENSSL_UNUSED
#endif
```

(Looking at both `__GNUC__` and `__clang__` is a little goofy. Normally Clang aims to be GCC-compatible, so it defines `__GNUC__` and `__attribute__` is indeed a GNU C extension. But `clang-cl` is Clang in MSVC-compatible mode. That defines `__clang__`, `_MSC_VER`, but *not* `__GNUC__`. That usually does the right thing, but it still implements a lot of the GCC extensions and warning flags, so sometimes we need to explicitly include it.)

etc/ci/test-fiat-c-boringssl.sh
```diff
@@ -0,0 +1,54 @@
...
+    cd build || exit $?
+    cmake -GNinja .. || exit $?
+    ninja || exit $?
+}) || exit $?
+echo "::endgroup::"
+
+echo "::group::Testing BoringSSL"
+({
+    set -ex
+    ninja run_tests || exit $?
```
davidben: Looks like this step is failing. Guessing you need to either `cd build` or use `ninja -C build`.

etc/ci/test-fiat-c-boringssl.sh
```diff
@@ -0,0 +1,54 @@
...
+}) || exit $?
+echo "::endgroup::"
+
+echo "::group::Building BoringSSL"
+({
+    set -ex
+    mkdir build || exit $?
+    cd build || exit $?
+    cmake -GNinja .. || exit $?
+    ninja || exit $?
```
davidben, line 43: Optional: if your CI has a new enough `cmake`, you can write `cmake -GNinja -B build` and `ninja -C build` without either `mkdir` or `cd`.

etc/ci/test-fiat-c-boringssl.sh
davidben: (Oops, I think we collided midair.)

.github/workflows/c.yml
JasonGross: We can always just comment out the tests.  Or move this test into a separate file.  I've currently added this to the test that tests that the C code compiles, so we don't lose out on much if this starts failing (in particular it's completely separate from all of the Coq builds)

etc/ci/test-fiat-c-boringssl.sh
JasonGross: Ah, neat, thanks!  Updated.

etc/ci/test-fiat-c-boringssl.sh
JasonGross: Moved to #820.  This is indeed a good idea, I think, but beyond the scope of this PR.  In any case, it turns out we need this for bedrock2, for `-Wno-error=unused-but-set-variable -Wno-error=tautological-compare`

etc/ci/test-fiat-c-boringssl.sh
JasonGross: Updated, thanks

.github/workflows/c.yml
JasonGross: Updated, thanks

etc/ci/test-fiat-c-boringssl.sh
```diff
@@ -0,0 +1,52 @@
...
+pushd boringssl >/dev/null
+
+echo "::group::Patching BoringSSL"
+({
+    set -ex
+    ( cd third_party/fiat && for i in *.h; do cp "${SUBCOMPONENT_PATH}/${i/.h/.c}" "$i" || exit $?; done ) || exit $?
+    ( cd third_party/fiat && git --no-pager diff )
+
+    # As per https://github.com/mit-plv/fiat-crypto/pull/819#discussion_r445026866, we don't need to modify util/BUILD.toplevel
+    sed "s/-Werror/-Werror -Wno-error=unused-function ${EXTRA_CFLAGS}/g" -i CMakeLists.txt
```
davidben, line 33: Sorry, didn't realize this earlier. Does it work if you pass `-DCMAKE_C_FLAGS="-Wno-error=unused-function ${EXTRA_CFLAGS}"` to `cmake`? (You may also need to pass the same to `-DCMAKE_CXX_FLAGS`.) That might be more robust than a `sed` line.

etc/ci/test-fiat-c-boringssl.sh
JasonGross: Hm, my local machine has not a new-enough cmake, apparently, so I'm going to go back to the old way; I'm getting
```
CMake Error: The source directory "/home/jgross/Documents/repos/fiat-crypto/boringssl/build" does not exist.
```
or, if I make it and then pass `-B build`, I get
```
CMake Error: The source directory "/home/jgross/Documents/repos/fiat-crypto/boringssl/build" does not appear to contain CMakeLists.txt.
```
Locally I have cmake version 3.10.2

etc/ci/test-fiat-c-boringssl.sh
JasonGross: Yep, that seems to work, updated, thanks!

JasonGross:
C tests pass (bedrock2 tests fail as per #821, so I've marked them as allowed to fail), so I'm going to merge this.

cry42 (commented):
Change

PR #826, opened 2020-06-29
Move C delcarations up to the top of C code

### LINKED ISSUE #824: Mixed declarations and code
Howdy Fiat Folks,

```ISO C90 forbids mixed declarations and code```

This is a warning generated because of stuff like

```C
    uint64_t x8;
    fiat_secp384r1_mulx_u64(&x7, &x8, x6, (arg2[5]));
    uint64_t x9;
```

You probably think this is a dumb compiler warning and I should just turn it off. And I even agree with you. But it's in the coding guidelines for many projects. For example, [OpenSSL](https://www.openssl.org/policies/codingstyle.html) which is currently one of many things holding up [this PR](https://github.com/openssl/openssl/pull/12201).

Is there any way to prevent that with Fiat? I'd rather not resort to shamelessly hacking up your output C with some BS post-processing script.

In other news, our project [ECCKiila](https://gitlab.com/nisec/ecckiila) is now public, powered by Fiat! Thanks for all your impactful work.

JasonGross:
This is needed for ISO C90 compliance.

Fixes #824 , cc @bbbrumley

@andres-erbsen @davidben Does this seem like a reasonable thing to do to the C code?  (One alternative is to provide a `--no-mix-declarations-and-code` option or similar, and have users that care about ISO C90 compliance generate their own code and pass this flag.  But it seems harmless to me to do it always.)


bbbrumley:
Thanks so much for the quick reaction here, @JasonGross !

I tested your branch -- it looks swell to me, and squelches the warnings :)

JasonGross:
I'm going to merge this.  @davidben , if BoringSSL doesn't like these changes, I can backtrack on them or add an option or something.

davidben:
Just fine for BoringSSL. I think, in 2020, it's about time we declared our variables closer to their use (much more readable), but doing it the old way still compiles just fine.

(I believe you're allowed to move your decls down starting C99. Strictly speaking, MSVC doesn't implement C99, but it is perfectly fine with mixed decls.)

Issue #833, opened 2020-07-06
Need rewrite rules for new base conversion


jadephilipoom:
As discussed in a couple of recent fiat-crypto meetings and #813 (also related to #809), I've been trying to clean up `Core.v` by removing the enormous `nth_default` proofs about carry functions. As it stands, the only one being used is `nth_default_chained_carries_no_reduce`, which is used in base conversion to prove that `convert_bases` is equivalent to evaluating the input in the source base system and then evenly partitioning it in the destination base system (`convert_bases_partitions`). Therefore, removing the need for `nth_default` in that one proof will remove some of the largest and nastiest proofs from the core arithmetic library.

To that end, in the `columns-base-conversion` branch, I've written a definition of `convert_bases` that uses the saturated arithmetic library (`Columns.from_associational` and `Columns.flatten`) to convert between bases. It seems to work in my preliminary tests and the proofs pass. However, because the `Columns` definitions use `Z.add_get_carry_full`, the output ends up having add-get-carry statements that split at invalid bit indices, e.g.:
```
(uint40, uint8) (Z.add_get_carry (2^40) ((uint47) x) ((uint47) y))
```
Instead, we want an add and then a bitwise div/mod, e.g.:
```
dlet xy := (uint48 (Z.add (uint47 x) (uint47 y)) in
(uint40 (Z.shiftr (uint48 xy) (literal 8)), uint8 (Z.land (uint48 xy) (2^8-1))
```
I think existing rewrite rules will take care of translating a Z div/mod into a shiftr/land, but I don't know how to write a rule that will do the div/mod translation in the with-casts rules. Especially problematic is figuring out what the range of `x + y` should be, because it doesn't appear in the initial expression and I think (correct me if I'm wrong) the rewriter won't allow me to write a range in the new expression as an expression (e.g. `ZRange.four_corners rx ry`).

I see two main solutions:
1. Write a rewrite rule that recognizes such add-get-carry statements and changes them into div/mod
2. Change Columns definitions to use something other than `Z.add_get_carry_full` when some boolean is set, or parameterize Columns over your desired implementation of div/mod (which we can then plug in as `Z.add_get_carry_full` for saturated arithmetic and div/mod for base conversion)

[...]

PR #837, opened 2020-07-08
Assume that cast truncates more pervasively

JasonGross:
This should fix the issue with `None` in the output bounds in #833

Reviews requested for the conceptual change (that we are now ~always working under the assumption that `cast` truncates its input, rather than being agnostic about how `cast` behaves when things are out of range).

On top of #836.  Closes #836.

### DIFF
--- a/src/BoundsPipeline.v
+++ b/src/BoundsPipeline.v
@@ -485,7 +485,7 @@ Module Pipeline.
              out_bounds
   : ErrorT (Expr t)
     := (*let E := expr.Uncurry E in*)
-      let assume_cast_truncates := false in
+      let assume_cast_truncates := true in
       let opts := opts_of_method in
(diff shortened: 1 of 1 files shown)


jadephilipoom:
I'm not sure I quite follow why it helps to make this assumption more pervasive, and the change seems costly to me. Are we actually allowing integer overflow in intermediate expressions somewhere?

JasonGross:
The bounds analysis pass thinks that your new rewrite rule overflows, because the bounds rule for `Z.land` is not tight enough to match the bounds assigned by your rewrite rule.  We could instead/additionally tighten up the bounds inference on `Z.land`

jadephilipoom:
In that case, I advocate for just tightening the `Z.land` bounds instead, unless there's some additional reason to broaden the assumption.

PR #838, opened 2020-07-08
Change definition of `convert_bases`

### LINKED ISSUE #809: Core.v cleanup
The file holding the main logic for our arithmetic templates, `Arithmetic/Core.v`, is a bit overwhelming. It's over 1400 lines and contains mixed code and proofs (including some proofs that _really_ seem like they could be improved with another close look). I think it could be improved, and that it should be given it's a pretty important file. I'm willing to do it myself.

Some ideas:
- separate out code and proofs so it's easy to skim the templates
- separate associational and positional, and make Core.v a skeleton file that exports both
- more comments
- fix the huge slow proofs

What do you think @JasonGross @andres-erbsen ?

jadephilipoom:
Fixes #809 
See discussion in #813 and #833 

This PR reduces `Arithmetic/Core.v` by about 25% in size and (according to preliminary and probably noisy local tests) 57% in build time by removing `chained_carries_no_reduce` and some extremely large and slow proofs about `nth_default` after carrying.

To remove dependencies on these proofs, I changed the definition of `convert_bases` to use `Columns.flatten` instead of associational carries. For reasons discussed in #833, the best way to do this was to parameterize `Columns.flatten` over the implementation of add-get-carry. To make Fancy work with the new implementation, I needed to add a rewrite rule that removes superfluous masks.


jadephilipoom:
View diff without generated code over at #850

PR #839, opened 2020-07-08
Add a theory of bitlists and bitwise

JasonGross:
This is important for proving a tighter bound on the bitwise zrange
operations.


JasonGross:
If I can get the tritwise code working (which I'd still like to), then, yes.  Otherwise, it'll join the theory limbo.

Issue #706, opened 2020-03-30
Add Go tests to the CI


JasonGross:
It would be nice to be able to verify that our Go code works.  I'm not sure where to find Go tests, though.  cc @andres-erbsen

andres-erbsen:
I think the most obvious test would be to replace field arithmetic in https://github.com/golang/crypto/blob/master/curve25519/curve25519_generic.go with ours and then run https://github.com/golang/crypto/blob/master/curve25519/curve25519_test.go#L14-L32

I remember it being easier, though, but perhaps I just misremember

mdempsky:
It looks like if you implement the [curve25519 API](https://godoc.org/golang.org/x/crypto/curve25519), it should be easy enough to just run those same tests against it. So I think writing those functions would be the first step.

I'm assuming it would be easy enough to glue the current fiat_25519_* functions together, but could fiat-go generate the glue code itself? E.g., a function like:

    func fiat_25519_scalar_mult(out, scalar, base *[32]byte)

guaranteeing that "out = scalar * base"?

Aside: It looks like the Go backend might need some work still. The "from_byte" and "to_byte" functions are using `[32]uint64` instead of `[32]byte` (or `[32]uint8`).

https://github.com/mit-plv/fiat-crypto/blob/a26d5d416274a05111248165ae7a210194910337/fiat-go/src/curve25519_64.go#L616

https://github.com/mit-plv/fiat-crypto/blob/a26d5d416274a05111248165ae7a210194910337/fiat-go/src/curve25519_64.go#L764

mdempsky:
Go's generic scalarMult code is here: https://github.com/golang/crypto/blob/948cd5f35899cbf089c620b3caeac9b60fa08704/curve25519/curve25519_generic.go

I tried hacking the from_bytes and to_bytes code to operate on bytes instead of uint64, and gluing everything together with the generic scalarMult and feInvert functions along with this bit of adapter glue:

```
type fieldElement = [5]uint64

var (
	feAdd       = fiat_25519_add
	feSub       = fiat_25519_sub
	feMul       = fiat_25519_carry_mul
	feMul121666 = fiat_25519_carry_scmul_121666
	feSquare    = fiat_25519_carry_square

	feFromBytes = fiat_25519_from_bytes
	feToBytes   = fiat_25519_to_bytes
)

func feCopy(dst, src *fieldElement) { *dst = *src }
func feOne(out *fieldElement)       { *out = fieldElement{1} }

[...]

PR #843, opened 2020-07-12
Don't --widen-bytes in Go

JasonGross:
As per
https://github.com/mit-plv/fiat-crypto/issues/706#issuecomment-657147516

I'm not sure why this doesn't result in too-tight integer ranges for the
outputs of `bits.Add{32,64}`, but apparently it doesn't.  I should
probably look into this sometime, if only to understand it...


andres-erbsen:
I don't know why this works either.

JasonGross:
Ah, I should have updated the PR comment.  The reason it worked is that for the methods other than conversion to/from bytes, we don't extend the list with uint8, and so widening carry widens all the way to bitwidth.  On the other hand, this causes https://github.com/mit-plv/fiat-crypto/issues/846, which I'm not sure how to solve, and would be happy for your insight on (maybe to be discussed during the next fiat-crypto meeting).

This passes the CI, so I'm going to merge it.

PR #848, opened 2020-07-13
Use `s` rather than `weight n` in `n_bytes`

JasonGross:
As per
https://github.com/mit-plv/fiat-crypto/issues/706#issuecomment-657787914,
this allows us to deal with unsaturated WBW Montgomery implementations
without allocating more bytes than we should.

Note for the future: we should possibly move the definition of `bytes_n`
out of `Freeze.v`, it seems like a slightly weird location for it.

Requesting reviews for the change to the definition of `bytes_n` in `Freeze.v` (and/or the corresponding changes to list lengths in the generated code), not for any of the proof changes that resulted from it.


mdempsky:
fiat-go/src/p224_64.go looks good to me. (Didn't look at the rest of the commit.)

JasonGross:
Need to rebase this and regenerate the Go files and look into the CI failure.  Them I'll merge once CI passes.

PR #850, opened 2020-07-14
Change definition of `convert_bases` (minus generated code)

jadephilipoom:
This is #838 minus the generated code, just to make the diff more readable. Should not be merged.


src/Util/ZRange/Operations.v
```diff
@@ -175,9 +175,30 @@ Module ZRange.
...
+                 else 0 in
+      let high := if ((0 <=? lower x)%Z && (0 <=? lower y)%Z)%bool
+                  then (* both arguments must be positive *)
+                    if ((0 <=? upper x)%Z && (0 <=? upper y)%Z)%bool
+                    then Z.min (upper x) (upper y)
+                    else (* only get here if upper < lower *)
+                      Z.max (Z.max (lower x) (upper x))
+                            (Z.max (lower y) (upper y))
+                  else if ((upper x <? 0) && (upper y <? 0))%bool
+                       then 0 (* both arguments must be negative *)
```
JasonGross, line 197: This should be -1, not 0.  If both arguments are negative, the bitwise and is also negative.  I believe, in fact, that the bitwise and is always upperbounded by the max of its arguments, regardless of whether they are positive or negative.

src/Util/ZUtil/LandLorShiftBounds.v
```diff
@@ -114,6 +114,108 @@ Module Z.
...
+      auto using land_le_neg_l_lower, land_le_neg_l_upper.
+    Qed.
+    Lemma land_neg_r_range :
+      forall x y, 0 <= x -> y < 0 -> y <= Z.land x y <= x.
+    Proof.
+      intros; rewrite Z.land_comm. auto using land_neg_l_range.
+    Qed.
+
+    (* TODO : move *)
+    Lemma N2Z_inj_lor a b :
```
JasonGross, line 141: At the very least, this belongs in `Module N2Z`, not `Module Z`.

jadephilipoom:
Comments incorporated and code merged at #838

Issue #554, opened 2019-06-12
Unsaturated solinas fails on on p224


JasonGross:
Synthesis fails on
```
./src/ExtractionOCaml/unsaturated_solinas 'p224' '4' '2^224 - 2^96 + 1' '64'
```
With a couple of issues:
- bounds analysis cannot prove that the lower bound on carry is non-negative.
- `to_bytes` immediately calls sbb in a way that gives a borrow bit which is maybe-a-borrow-but-maybe-a-carry `Z.sub_with_get_borrow(2^56, 0, x1[0], 1)`
- the lower bounds on opp are negative:
```c
In fiat_p224_opp:
Computed bounds (Some [Some [-0x119999999999997 ~> 0x2], Some [0xe6646666666667 ~> 0x1fffe0000000000], Some [0xe6666666666665 ~> 0x1fffffffffffffe], Some [0xe6666666666665 ~> 0x1fffffffffffffe]]) are not tight enough (expected bounds not looser than (Some [Some [0x0 ~> 0x34ccccccccccccb], Some [0x0 ~> 0x34ccccccccccccb], Some [0x0 ~> 0x34ccccccccccccb], Some [0x0 ~> 0x34ccccccccccccb]])).
The bounds [-0x119999999999997 ~> 0x2] are looser than the expected bounds [0x0 ~> 0x34ccccccccccccb]
When doing bounds analysis on the syntax tree:
/*
 * Input Bounds:
 *   arg1: [[0x0 ~> 0x119999999999999], [0x0 ~> 0x119999999999999], [0x0 ~> 0x119999999999999], [0x0 ~> 0x119999999999999]]
 * Output Bounds:
 *   out1: None
 */
void f(int128 out1[4], const uint64_t arg1[4]) {
  int64_t x1 = (0x2 - (int64_t)(arg1[0]));
  uint64_t x2 = (UINT64_C(0x1fffe0000000000) - (arg1[1]));
  uint64_t x3 = (UINT64_C(0x1fffffffffffffe) - (arg1[2]));
  uint64_t x4 = (UINT64_C(0x1fffffffffffffe) - (arg1[3]));
  out1[0] = x1;
  out1[1] = x2;
  out1[2] = x3;
  out1[3] = x4;
}
```

Is this an issue with alternating TAPs?  An issue with us not supporting karatsuba or goldilocks?  An issue with computations of carry chains (we're using `[0, 3, 1, 0, 2, 3, 1, 0]`)?  Something else?

[...]

PR #853, opened 2020-07-22
Parameterize sub over balance, not coef

JasonGross:
This is the infrastructure work towards dealing with issues like #554
(at least the part involving balance in sub and opp).

Still to come is the more general way of computing balance.


src/UnsaturatedSolinasHeuristics.v
```diff
@@ -91,11 +104,15 @@ else:
   (** Allow enough space to do one subtraction w/o carrying *)
   Definition headspace_sub_count : nat := 1.
 
+  Definition balance : list Z
+    := let weight := weight (Qnum limbwidth) (Qden limbwidth) in
+       scmul weight n coef (encode weight n s c (s - Associational.eval c)).
```
JasonGross, line 109: @jadephilipoom If you change only this definition, the time to compile and get new files should be only about 5 minutes or so.  Alternatively, you can just fill in the parameters and do something like
```coq
Require Import Crypto.Util.Strings.Show.
Require Import Coq.Strings.String.
Open Scope string_scope.
Existing Instance PowersOfTwo.show_Z.
(* Or: *)
(* Existing Instance Hex.show_Z. *)
Compute show false (@balance 4 (2^224) [(2^96,1);(1,-1)]).
```
at the bottom of this file (or just close the section right after defining `balance`), and see if the balance is what you expect right after defining.

PR #854, opened 2020-07-23
Update bedrock2

jadephilipoom:
Updates bedrock2 submodule to the latest version, which includes some upstreaming of lemmas from `Bedrock/Util.v`, so the code is modified to use the upstreamed versions. I also made the CI skip bedrock2 for 8.10, since bedrock2 does not support it and the latest updates are broken on 8.10.


.github/workflows/coq.yml
```diff
@@ -15,11 +15,11 @@ jobs:
       matrix:
         env:
         - { COQ_VERSION: "8.11.1", COQ_PACKAGE: "coq-8.11.1 libcoq-8.11.1-ocaml-dev", SKIP_BEDROCK2: "" , SKIP_VALIDATE: "" , PPA: "ppa:jgross-h/many-coq-versions-ocaml-4-05" }
-        - { COQ_VERSION: "8.10.2", COQ_PACKAGE: "coq-8.10.2 libcoq-8.10.2-ocaml-dev", SKIP_BEDROCK2: "" , SKIP_VALIDATE: "" , PPA: "ppa:jgross-h/many-coq-versions-ocaml-4-05" }
+        - { COQ_VERSION: "8.10.2", COQ_PACKAGE: "coq-8.10.2 libcoq-8.10.2-ocaml-dev", SKIP_BEDROCK2: "1" , SKIP_VALIDATE: "" , PPA: "ppa:jgross-h/many-coq-versions-ocaml-4-05" }
```
JasonGross, line 18: Minor nit: Please remove the space from after `"1"` so everything still lines up space-wise

PR #855, opened 2020-07-25
Tighter bounds

jadephilipoom:
So you know how we compute tight bounds for unsaturated solinas by encoding `s-1` and then multiplying by 1.1? Turns out it also works just to add 1 to each limb. This makes both tight and loose bounds smaller and more readable, makes `to_bytes` work for p224 (see #554), and doesn't break any of our currently compiled examples (including the ladderstep example).

I left the multiplicative factor in, in case sometime we need to multiply the bound, but set the default value to 1.


JasonGross:
Neat!  Is this the right direction though?  And do we want to parameterize over the additive factor?

jadephilipoom:
I mean, smaller tight bounds should be better as long as our multiplication/carry routines can satisfy them. If it works on our examples and we retain the ability to loosen the bounds if we later want, I think it's an improvement. Or do we have some strong reason to think multiplying by 1.1 is the right magic number?

JasonGross:
> Or do we have some strong reason to think multiplying by 1.1 is the right magic number?

I'm not aware of anything, so I'd say good to go.  @andres-erbsen ?

Issue #748, opened 2020-04-14
How can/should we get good constant-time C code for cmov?


JasonGross:
I should footnote, those numbers are for x86_64. They also include the following patch:

```
static void fiat_p256_cmovznz_u64(uint64_t* out1, fiat_p256_uint1 arg1, uint64_t arg2, uint64_t arg3) {
  fiat_p256_uint1 x1 = (!(!arg1));
  uint64_t x2 = ((fiat_p256_int1)(0x0 - x1) & UINT64_C(0xffffffffffffffff));
  // Note this line has been patched from the synthesized code to add value
  // barriers.
  //
  // Clang recognizes this pattern as a select. While it usually transforms it
  // to a cmov, it sometimes further transforms it into a branch, which we do
  // not want.
  uint64_t x3 = ((value_barrier_u64(x2) & arg3) | (value_barrier_u64(~x2) & arg2));
  *out1 = x3;
}
```

`value_barrier_u64` is defined as below. It's a goofball hack to stop the compiler from turning cmovs back into branches, because apparently it's doing that now. We really need language-level support for this stuff. :-(
```
static inline uint64_t value_barrier_u64(uint64_t a) {
#if !defined(OPENSSL_NO_ASM) && (defined(__GNUC__) || defined(__clang__))
  __asm__("" : "+r"(a) : /* no inputs */);
#endif
  return a;
}
```

_Originally posted by @davidben in https://github.com/mit-plv/fiat-crypto/pull/723#issuecomment-613639650_

davidben:
Long-term, I think we need language support for this sort of thing. In the meantime, yeah, the compilers seem to have gotten smart enough, so we have a derpy function we can sprinkle in as needed. It's terrible and there is no principled answer for where to use it, but... so it goes. :-(

jadephilipoom:
One more addition to a long list of reasons to eventually try to bypass the C compiler by producing something closer to machine code.

davidben:
That too. :-) Although projects often need to support random generic targets, so anything close to machine code tends to be in addition to rather than instead of C. I'm hopeful the long-term answer is compiler support.

davidben:
(Then again, `value_barrier` doesn't work for no-asm builds anyway...)

JasonGross:
@chjj said in #840 that
[...]

PR #856, opened 2020-07-28
Generate value_barrier wrappers for cmovznz code

JasonGross:
This is significant progress towards #748

Note that rather than checking for whether or not `OPENSSL_NO_ASM` is
defined, we instead check for the definition of a curve-specific
`*_NO_ASM` variable, where the prefix is the uppercased prefix of the
curve description.  This lets us not be tied to OpenSSL.  (I presume
that BoringSSL can do something like
```c
#ifdef OPENSSL_NO_ASM
#define FIAT_25519_NO_ASM
#define FIAT_P256_NO_ASM
[...]
#endif
```
cc @davidben)


JasonGross:
I'm going to merge this, @davidben let me know if you want me to change anything in the future

davidben:
Nah, this should work for us.

PR #857, opened 2020-08-03
Generalized sub/opp balance for unsaturated solinas

jadephilipoom:
See discussion in #554 

This PR generalizes the formula for computing sub/opp balance so that it works on primes with very low limbs (e.g. p224, whose lowest limb in base 2^56 is 1). The key idea is to borrow from higher limbs so that each limb of the balance is higher than the `tight_upperbounds` -- a strategy borrowed from the unsaturated-solinas p224 implementation in OpenSSL. We still can't synthesize unsaturated solinas code for p224 because we would need balance for reduce, and that comes with further complications (see #554).

Also included in the PR is a tweak to make `to_bytes` work for p224 (and other primes with low limbs) and unsaturated solinas -- using the same function we use to distribute sub/opp balance nicely across limbs, we distribute the encoding of the prime so that it doesn't cause problems in `to_bytes`.

### DIFF
--- a/src/PushButtonSynthesis/UnsaturatedSolinas.v
+++ b/src/PushButtonSynthesis/UnsaturatedSolinas.v
@@ -117,8 +117,9 @@ Section __.
   Local Notation balance := (balance n s c).
 
   Definition m : Z := s - Associational.eval c.
-  Definition m_enc : list Z
-    := encode (weight (Qnum limbwidth) (Qden limbwidth)) n s c m.
+  Definition m_enc : list Z :=
+    let M := encode (weight (Qnum limbwidth) (Qden limbwidth)) n s c m in
+    distribute_balance n s c M.
 
   (* We include [0], so that even after bounds relaxation, we can
        notice where the constant 0s are, and remove them. *)
@@ -156,6 +157,13 @@ Section __.
   Lemma length_saturated_bounds_list : List.length saturated_bounds_list = n.
   Proof using Type. cbv [saturated_bounds_list]; now autorewrite with distr_length. Qed.
   Hint Rewrite length_saturated_bounds_list : distr_length.
+  Lemma length_m_enc : List.length m_enc = n.
+  Proof using Type.
+    cbv [m_enc distribute_balance distribute_balance_step].
+    apply fold_right_invariant; intros;
+      break_innermost_match; now autorewrite with distr_length.
+  Qed.
+  Hint Rewrite length_m_enc : distr_length.
 
   (** Note: If you change the name or type signature of this
         function, you will need to update the code in CLI.v *)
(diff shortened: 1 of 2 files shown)


src/PushButtonSynthesis/UnsaturatedSolinas.v
```diff
@@ -153,6 +170,13 @@ Section __.
   Lemma length_saturated_bounds_list : List.length saturated_bounds_list = n.
   Proof using Type. cbv [saturated_bounds_list]; now autorewrite with distr_length. Qed.
   Hint Rewrite length_saturated_bounds_list : distr_length.
+  Lemma length_m_enc : List.length m_enc = n.
+  Proof using Type.
+    cbv [m_enc distribute_balance distribute_balance_step].
```
JasonGross, line 175: I think instead we should have a `length_distribute_balance` lemma.

src/UnsaturatedSolinasHeuristics.v
```diff
@@ -104,9 +105,53 @@ else:
...
-    := let weight := weight (Qnum limbwidth) (Qden limbwidth) in
-       scmul weight n coef (encode weight n s c (s - Associational.eval c)).
+      := let weight := weight (Qnum limbwidth) (Qden limbwidth) in
+         let B := encode weight n s c (s - Associational.eval c) in
+         let B := scmul weight n coef B in
+         distribute_balance tight_upperbounds B.
+
+  Lemma balance_length : length balance = n.
+  Proof.
+    cbv [balance scmul distribute_balance distribute_balance_step].
```
JasonGross, line 149: Again, I think it's better to have a `length_distribute_balance`.

src/UnsaturatedSolinasHeuristics.v
```diff
@@ -139,18 +184,56 @@ else:
       Z.div_mod_to_quot_rem; nia. }
   Qed.
 
-  Lemma length_balance : List.length balance = n.
-  Proof using Type. cbv [balance]; now autorewrite with distr_length. Qed.
+  (* TODO: shouldn't this be in Positional? *)
```
JasonGross, line 187: Yes.  Why isn't it?

jadephilipoom:
> I think I also want to try the algorithm I proposed (which I think is a general replacement for `distribute_balance` that may be simpler to understand and reason about), and see how that fares, before we merge this, if you're okay with that.

Sure, that's fine. I'll leave this until you've tested your version.

src/UnsaturatedSolinasHeuristics.v
JasonGross: In fact, it seems to already be in positional...

JasonGross:
Closed by #860

PR #860, opened 2020-08-03
Generalized sub/opp balance for Unsaturated Solinas (alternate version)

JasonGross:
Alternative to #857.

Rather than redistributing an existing encoding, this version simply
encodes a number using `partition` that, when added to the relevant
bounds on a limb-by-limb basis without carrying, gives something which
evaluates to the desired result.

The limb-by-limb addition ensures the bounds we want, and the encoding
scheme ensures the correct value.  Furthermore, we don't need to know
ahead-of-time how many copies of the prime we need to add. This is
pleasingly simple to me.

Additionally, as in #857, the tweak to make `to_bytes` work for p224
(and other primes with low limbs) and unsaturated solinas.  The tweak
here is a bit uglier than in #857, because balancing here doesn't
guarantee the same value under `eval`, only something congruent to that
value.  So we need to ensure that we're encoding at least one copy of
the prime.

Note that this is on top of #858 and #859, which are mostly about utility lemmas and reorganization, and also that this PR adds another test case (computing balance for a bunch of primes), so the relevant diff is more like +148,-16.  This is a bit larger than #857 primarily because we're also proving that the limbs of the balanced number are in fact not less than the given minimum values, which is there primarily as a sanity check.


src/Arithmetic/Balance.v
```diff
@@ -0,0 +1,75 @@
...
+    Local Notation eval := (eval weight) (only parsing).
+
+    (** Sometimes we want to ensure that each limb of our encoded
+        number [x] is larger than the corresponding limb of a given
+        bound [minvalues].  We can ensure this by encoding [x -
+        minvalues] in a way that ensures each limb is non-negative,
+        and then adding back [minvalues].  We extend [minvalues] with
+        zeros for convenience, so we don't need a proof that [length
+        minvalues = n] for the proofs about [encode_balanced]. *)
+    Definition encode_balanced (minvalues : list Z) (x : Z) : list Z
```
jadephilipoom, line 33: `encode_distributed`? I think calling it "balanced" for being distributed across limbs is confusing.

src/Arithmetic/Core.v
```diff
@@ -816,6 +816,28 @@ Module Positional.
   Proof using Type. clear -Hf Hf; cbv [add]; distr_length.               Qed.
   Hint Rewrite @length_add : distr_length.
 
+  Definition add_no_carry (f g:list Z) : list Z
+    := map2 Z.add f g.
```
jadephilipoom, line 820: `Positional.add` already doesn't carry. I think it has exactly this behavior.

JasonGross: Oh, cool, I'll use that instead

JasonGross:
> It seems like this method didn't make the proofs any easier

Compare:
This version:
```coq
   Lemma eval_encode_distributed minvalues x
      : 0 < s - Associational.eval c <= weight n
        -> eval n (encode_distributed minvalues x) mod (s - Associational.eval c)
           = x mod (s - Associational.eval c).
    Proof using wprops.
      intros; cbv [encode_distributed]; rewrite eval_add, eval_partition by (auto; distr_length).
      assert (forall x, x mod (s - Associational.eval c) < s - Associational.eval c) by auto with zarith.
      rewrite (Z.mod_small (_ mod (s - _)) (weight _)) by now eauto using Z.lt_le_trans with zarith.
      pull_Zmod; now autorewrite with zsimplify_fast.
    Qed.
    Hint Rewrite eval_encode_distributed : push_eval.

  Lemma eval_balance : eval (weight (Qnum limbwidth) (Qden limbwidth)) n balance mod (s - Associational.eval c) = 0.
  Proof using Hs_c_bounded wprops.
    clear -Hs_c_bounded wprops.
    cbv [balance]; rewrite eval_encode_distributed by auto.
    now autorewrite with zsimplify_fast.
  Qed.
```

vs the version of #857 
```coq
  Lemma eval_distribute_balance_step minvalues i x :
    eval (weight (Qnum limbwidth) (QDen limbwidth)) n
         (distribute_balance_step minvalues i x)
    = eval (weight (Qnum limbwidth) (QDen limbwidth)) n x.
  Proof using Hs_nz Hs_c_nz Hs_n Hn_nz.
    clear -Hs_nz Hs_c_nz Hs_n Hn_nz wprops.
    cbv [distribute_balance_step].
    break_innermost_match; [ | lia ].
    rewrite eval_from_associational by auto with zarith.
    autorewrite with push_eval. cbn [fst snd].
    rewrite eval_to_associational by auto.
    ring_simplify.
    do 2 match goal with
         | |- context [?a * ?b * (?c / ?a)] =>
           replace (a * b * (c / a)) with (a * (c / a) * b) by lia;
             rewrite (Modulo.Z.mul_div_eq_full c a) by auto with zarith
         end.
    rewrite weight_multiples by auto. lia.
  Qed.

  Lemma eval_distribute_balance minvalues x :
    eval (weight (Qnum limbwidth) (QDen limbwidth))
         n (distribute_balance minvalues x)
    = eval (weight (Qnum limbwidth) (QDen limbwidth)) n x.
  Proof using Hs_nz Hs_c_nz Hs_n Hn_nz.
    clear -p Hs_nz Hs_c_nz Hs_n Hn_nz wprops.
    cbv [distribute_balance].
    apply fold_right_invariant; [ reflexivity | ].
    intros; rewrite eval_distribute_balance_step; auto.
  Qed.

  Lemma eval_balance : eval (weight (Qnum limbwidth) (Qden limbwidth)) n balance mod (s - Associational.eval c) = 0.
  Proof using Hs_nz Hs_c_nz Hs_n Hn_nz.
    clear -p Hs_nz Hs_c_nz Hs_n Hn_nz wprops.
    cbv [balance]. rewrite eval_distribute_balance.
    repeat first [ rewrite Z_mod_same_full
                 | rewrite eval_scmul by auto
                 | rewrite eval_encode by auto with zarith
                 | reflexivity
                 | progress (pull_Zmod; autorewrite with zsimplify_fast; push_Zmod) ].
  Qed.
```

JasonGross:
Here is my argument that the two algorithms should give exactly the same answer in all cases where their evaluation is
equal and the input does not exceed the maximum of the weight and
lower-bound in any limb except possibly the highest one.  (This last
condition morally says that we have not already rebalanced for a larger
minimum value.)  Here is my argument:

The specification of the algorithm of #857 (henceforth algorithm (1))
is that it computes the minimum value that each limb can have subject to
the following constraints:
1. the overall evaluation must equal that of the input
2. each limb must be greater than or equal to the specified lower bound
3. each limb remains unchanged iff, after borrowing what's necessary for
   the lower limbs from the current limb, the current limb is already
   above the minimum value.

The specification of the algorithm of #860 (henceforth algorithm (2))
is that it computes limbwise the number:
```
partition (eval (input - minvalues) mod p) + minvalues
```

The intuition is that both algorithms in fact compute the minimum
possible borrow for each limb.

The argument:
Note that if both algorithms give results that evaluate to the same
value, then both algorithms can be expressed as borrowing some amount
from each non-lowest-limb to add into the next-lower-limb, by the
properties of modular arithmetic (in particular that `x = x / d + x mod
d` for every `x` and any positive `d`).

Then the two algorithms compute different limbs iff they borrow
different amounts from some limb to add to the next-lower-limb.
Suppose, for contradiction, that there is a such a limb.  Say that the
lowest such limb has index `i+1`, and that (1) borrows `b1` from limb
`i+1` for limb `i`, while (2) borrows `b2` from limb `i+1` to add to
limb `i`.

Let us compute the output limbs of (2).

By the specification of `partition`, limb `i` of the output of (2) is
equal to
```
minvalues[i] + (eval(input - minvalues) mod p) mod (weight(i+1)) / weight i
```
The assumption that the evaluations of (1) and (2) are equal is the
assumption that `eval(input - minvalues)` is already reduced modulo `p`.
Hence we have that limb `i` of the output of (2) is equal to
```
minvalues[i] + eval(input - minvalues) mod (weight(i+1)) / weight i
```

Since `(x mod weight(i+1)) / weight i` is less than `weight(i+1) /
weight i`, we have that the amount we add to `minvalues[i]` is at most 1
more than we would need to borrow to get exactly `minvalues[i]`.
Furthermore, if we can get exactly `minvalues[i]` by borrowing some
amount, then we borrow exactly that amount, and not more, because the
remainder will be smaller than `weight i` in that case.

Hence if we borrowed any less than `b2`, then we'd end up with something
less than `minvalues[i]`.

By the constraints on (1), `b1 ≤ b2`.

Hence `b1 = b2`, in violation of our assumption that they differ in this
limb.

Qed

JasonGross:
I've updated this PR with both algorithms.  Contrary to the consensus at the group meeting, I've left the code computing the balance with the algorithm originally proposed in this PR, for two reasons:
1. There are some primes that the algorithm of #857 computes the incorrect balance for (the balance is lower than the bound given).  One such example is 64-bit 2^224 - 2^96 + 1 with 5 limbs; the tight upper bounds are `[2^45, 2^45, 2^45, 2^45, 2^44]`, and the computed balance is `[2^45+2, 2^45-1, 2^46-2^7-1, 2^46-2, 2^45-2]` while the desired balance is `[2^45+2, 2^46-1, 2^46-2^7-2, 2^46-2, 2^45-2]` (perhaps there is an off-by-one-error somewhere?)
2. (much less important) the algorithm of this PR is about 2x faster than the one of #857, taking only 35.45s to compute balances for all primes in Tests.v vs the 75.38s of #857

@jadephilipoom @andres-erbsen can you re-review?

jadephilipoom:
> Compare:

Yes, if you include the proof of `distribute_balance` and exclude the additional lemmas you needed to prove this version, then it has fewer lines. I'll point out that the proofs were simple enough that I put almost no effort into making them smaller. But the number of lines was never the issue -- I remain unconvinced that this is *conceptually* simpler. The OpenSSL code for p224 is doing a procedure that makes sense and can be explained in a couple of lines. In my view, the original version of the algorithm from #857 straightforwardly encodes this intuition. This version *might* always do the same thing in *well-behaved* cases, but it's very non-obvious and I don't see a strong reason to go with the less obvious encoding and include the arithmetic library (via `add` and `partition`) in the chain of reasoning, even if it makes the proofs fewer lines.

Regarding the proof of equivalence: can you easily produce a Coq proof for this? Given we have proof tools available and few test cases, I'd prefer that.

The issue producing different bounds for 5-limb p224 was an error I made in pulling the reduce-balance changes out of the PR for sub/opp; the indices list needs to be reversed, otherwise it distributes starting from the second-highest limb and going down to 0. I pushed the fix to #857; it's only one line, so should be easy to add here too.

This isn't the hill I'm going to die on. I remain unconvinced, but I am unwilling to devote more time and energy to this issue, and if you insist then I won't block using this version.

JasonGross:
> I pushed the fix to #857; it's only one line, so should be easy to add here too.

I've incorporated the fix.  I also think it's not worth devoting more energy here; sorry for pushing so hard, and thanks for explaining.  I've re-reverted the commit so that we're using your version, now that it works.  Shall we merge this now once the CI passes?

jadephilipoom:
Sounds good to me, thanks!

PR #865, opened 2020-08-12
Add rupicola as a submodule

jadephilipoom:
Creates a submodule dependency for rupicola so we can start using those tools for higher-level functions. Since rupicola also depends on bedrock2 and coqutil, those dependencies are now through rupicola (this ensures that fiat-crypto and rupicola are using the same version of bedrock2). That meant updating the bedrock2 version, so unfortunately the changes are a bit mixed up here. Almost all of the line-count change is due to a change in how bedrock2 prints to C. No code from rupicola is used yet; this change just lays the groundwork.

The Windows test is currently failing, because of some problem finding the dependencies that was unclear to me and looked like possibly a submodule initialization issue (it seems to be showing huge git diffs that suggest maybe the submodule wasn't cloned at all). Following our usual convention of not blocking on the Windows build, I would like to merge despite that failure.


JasonGross:
Nice work!  I think we should not let *random* Windows build failures block the merge, but the errors here are things that would genuinely make developing on Windows more painful.  Let's at least try to fix them before merging this.  The current issue is that checking out Rubicola on Windows will by default change the line endings of all of the files and git will then mark them as changed.  Try copying our [.gitattributes](https://github.com/mit-plv/fiat-crypto/blob/master/.gitattributes) file to Rubicola

Makefile
```diff
@@ -311,12 +312,15 @@ REWRITER_FOLDER := rewriter
 REWRITER_SRC := $(REWRITER_FOLDER)/src
 COQPRIME_FOLDER := coqprime
 COQPRIME_SRC := $(COQPRIME_FOLDER)/src
-BEDROCK2_FOLDER := bedrock2/bedrock2
+BEDROCK2_FOLDER := rupicola/bedrock2/bedrock2
```
JasonGross, line 315: Might be a bit cleaner if you move the `RUBICOLA` variables up so you can do
```suggestion
BEDROCK2_FOLDER := $(RUBICOLA_FOLDER)/bedrock2/bedrock2
```

PR #869, opened 2020-08-14
Copy montgomery-ladder code from Rupicola

jadephilipoom:
For now, just copies over the crypto-specific parts of the montgomery ladder example from rupicola. I also added in a file for the specification of groups, added a file that instantiates the `FieldRepresentation` class with the representation fiat-crypto uses for field elements (just to make sure it worked properly), and did a round of renaming to be more specific (`bignum` -> `felem`).

The rupicola proofs all work but have **not** yet been hooked up to the field implementations; that is, the bedrock2 backend produces some code for finite field operations, and proves that each operation obeys a certain spec (the ones in `COperationSpecifications`). The montgomery ladder proof says that, assuming you have bedrock2 implementations of field operations which obey the *different* specs in `Bedrock/Specs/Field.v`, then you've got a bedrock2 implementation that matches the Gallina code for Montgomery ladder.

The two field specifications are very similar but need to be explicitly proven to match. The format of `COperationSpecifications` specs makes this a little nasty (because it locks me into a function that operates on `list Z`, which bedrock2 implementations don't produce), so the matching might be done by having the bedrock2 backend use the specs in `Bedrock/Specs/Field.v` throughout, and then prove during synthesis (on the abstraction level of `Bedrock/Field/Synthesis/Generic`) that this does in fact match with the `COperationSpecifications` specs.

Coming up:
- fixing the spec mismatch, as described above
- proving that the ladderstep and montladder Gallina implementations match the Gallina implementations from `Curves/Montgomery/XZ.v` so we can use the proofs from `XZProofs` to say that montgomery ladder is a group scalarmult
- X25519! (still needs scalar implementations and inverse, but can be proven assuming those for now)


JasonGross:
Looking forward to getting this all hooked up!  Is there anything particular you want me to review?

jadephilipoom:
> Looking forward to getting this all hooked up! Is there anything particular you want me to review?

Most of the code has already gone through review when it was being added to rupicola, and it's a lot of code, so I don't need anything fine-grained. Let's say:
- naming sanity check (do names of files, directories, and major definitions make sense?)
- code-hygeine check for newly created, non-copied files (`Bedrock/Field/Interface/Representation.v` and `Bedrock/Specs/Group.v`, both short)
- are specs in `src/Bedrock/Specs/{Field,ScalarField,Group}.v` readable and logical? (There's some extra code in `ScalarField.v` in a section called `Compile` that will be moved out of specs; feel free to ignore that for now.)

From Jason I especially want a check on the code hygiene and naming; from Andres I especially want a check on the specs, because they're going to need to interface with the proofs in `src/Curves` that he worked on forever ago. Each of you can feel free to ignore the points I asked for from the other if the review feels like too much.

jadephilipoom:
Andres also see specs on #870, which uses F instead of Z for field specs

src/Bedrock/Field/Interface/Representation.v
```diff
@@ -0,0 +1,32 @@
...
+Section Representation.
+  Context {p : Types.parameters}.
+  Context (n : nat) (weight : nat -> Z)
+          (loose_bounds tight_bounds : list (option zrange)).
+
+  Definition eval_words : list word -> Z :=
+    fun ws =>
+      Positional.eval weight n (map word.unsigned ws).
+
+  Instance frep : FieldRepresentation :=
```
JasonGross, line 21: Note that `Instance`s don't survive sections unless you prefix them with `Global`.  I'd suggest either making it a `Global Instance` or a `Definition`

src/Bedrock/Specs/Group.v
```diff
@@ -0,0 +1,52 @@
...
+Class GroupParameters :=
+  { G : Type;
+    eq : G -> G -> Prop;
+    add : G -> G -> G;
+    zero : G;
+    opp : G -> G;
+    scalarmult : Z -> G -> G;
+
+    (* function names *)
+    scmul : string;
```
JasonGross, line 17: Why does only this one get a name?  And why is the name part of the spec?

src/Bedrock/Specs/Group.v
```diff
@@ -0,0 +1,52 @@
...
+    GElem : word -> gelem -> Semantics.mem -> Prop;
+  }.
+
+Section Specs.
+  Context {semantics : Semantics.parameters}
+          {scalar_representaton : ScalarRepresentation}.
+  Context {group_parameters : GroupParameters}
+          {group_representaton : GroupRepresentation (G:=G)}.
+
+  Instance spec_of_scmul : spec_of scmul :=
```
JasonGross, line 37: Again, `Instance` disappears after `Section` ending, I'd suggest either `Definition` or `Global Instance`

src/Bedrock/Specs/Group.v
jadephilipoom: It's the function name for the bedrock2 implementation, which might be something like `X25519_scalarmult` or something else that depends on the particular group being implemented. `scmul` is the only one that gets a name because as of now it's the only one we're making a bedrock2 implementation for. I changed the comment to make this a little clearer; let me know if you think it needs more.

src/Bedrock/Field/Interface/Representation.v
jadephilipoom: It looks like I can't use the `{ _ := _ }` notation with a `Definition`, so I've changed it and then changed it back because I think this way is much more readable.

JasonGross: You can use `{| _ := _ |}` with `Definition`. The only reason you'd need `{ _ := _ }` is if you need fancier bidirectional type inference of the parameters to the record type.  If you do need that, then I'd suggest using an explicit `Local Instance`, maybe with a comment.  (Alternatively, you can use `Arguments Build_Foo param1 ... paramn & arg1 ... argm` to use the new bidi type inference feature with normal terms)

jadephilipoom:
> feel free to just replace one implementation with the other, at least as long as we keep the boringssl version around and perhaps otherwise as well.

Actually, in my local branch I tried proving them equivalent and it's super easy. I already did it. So I think we might as well keep both.

> There are also a couple of convention issues that I would push on if this was completely new code, but given that rupicola already has the opposite conventions and this code is still better kept in fiat-crypto than rupicola, I don't think these should be considered blocking. 

Yes, there's supposed to be some refactoring happening in rupicola soon and I was originally waiting on merging this stuff so that the refactor wouldn't have to happen on this code twice -- but given I'm going to have a real job next week, I thought it was higher priority to get this in while I still have free time.

Issue #736, opened 2020-04-09
Remove axioms


JasonGross:
On Coq master, [`coqchk` successfully runs on fiat-crypto and the parts of bedrock and the rewriter that fiat-crypto uses](https://github.com/mit-plv/fiat-crypto/runs/574704310?check_suite_focus=true) (I'm working on adding it to the CI in https://github.com/mit-plv/fiat-crypto/pull/724).  However, it reports a number of axioms:
```
CONTEXT SUMMARY
===============

* Theory: Set is predicative
  
* Axioms:
    Crypto.Rewriter.PerfTesting.StandaloneOCamlMain.printf_float
    Crypto.Algebra.Field.isomorphism_to_subfield_field
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.Rquot2
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.Rquot1
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.R1_def
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.R0_def
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.Rmult_def
    Coq.ssr.ssrunder.Under_rel.Over_rel
    Coq.Reals.Rdefinitions.RinvImpl.Rinv_def
    Crypto.Bedrock.Tests.Proofs.X25519_64.mulmod_Wf
    Crypto.Curves.Edwards.AffineProofs.E.Kchar_ge_2
    Crypto.StandaloneOCamlMain.printf_char
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.Rrepr
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.Rplus
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.Rmult
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.Rabst
    Crypto.Bedrock.Tests.Proofs.X25519_64.mulmod_correct
    Crypto.StandaloneOCamlMain.string_init
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.Ropp
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.Rlt
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.R1
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.R0
    Coq.Reals.Rdefinitions.RbaseSymbolsImpl.R
[...]

PR #874, opened 2020-08-17
Remove the axiom in Curves/Edwards/AffineProofs

JasonGross:
This will make progress towards #736

@andres-erbsen Please comment on my naming, and whether `of_Z_absorbs_homomorphism` seems okay, or if there's a different name you'd suggest

### DIFF
--- a/src/Algebra/Ring.v
+++ b/src/Algebra/Ring.v
@@ -427,6 +427,42 @@ Section of_Z.
   Qed.
 End of_Z.
 
+Section of_Z_absorbs_homomorphism.
+  Context {R Req Rzero Rone Ropp Radd Rsub Rmul}
+          {Rring : @ring R Req Rzero Rone Ropp Radd Rsub Rmul}.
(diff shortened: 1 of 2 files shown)


andres-erbsen:
I can't think of a great name here. `two_homomorphisms_of_Z` is an alternative that comes to mind.

JasonGross:
Hm, I think I'll leave it as is, if you're okay with that.  (It's not clear to me how `two_homomorphisms_of_Z` describes what's going on here.)

andres-erbsen:
how does the current name describe what's going on? I think it's fine as an arbitrary name, though

JasonGross:
It says that when you compose `of_Z` with ring homomorphisms, the homomorphisms are "absorbed" into the `of_Z`.

JasonGross:
Hmmm, maybe a better factoring of this proof would be proving that the composition of two homomorphisms is a homomorphism (do we have this already?), and proving that any two ring homomorphisms out of `Z` are pointwise-equal?  (i.e., proving that `Z` is the initial ring)  Anyway, that's a task for the future

PR #876, opened 2020-08-18
Added info on dependencies package

dderjoel:
Apparently, the package `ocamlfind` has been renamed to `ocaml-findlib`
Added convenience cmd lines to install the required dependencies.

 (Checked package names in brew / pacman and apt package-lists)

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,10 +18,21 @@ Building
 
 This repository requires [Coq](https://coq.inria.fr/) [8.9](https://github.com/coq/coq/releases/tag/V8.9.0) or later.
 Note that if you install Coq from Ubuntu aptitude packages, you need `libcoq-ocaml-dev` in addition to `coq`.
-Note that in some cases (such as installing Coq via homebrew on Mac), you may also need to install `ocamlfind`.
+Note that in some cases (such as installing Coq via homebrew on Mac), you may also need to install `ocaml-findlib` and `ocaml-num`.
 If you want to build the bedrock2 code, you need [Coq 8.10](https://github.com/coq/coq/releases/tag/V8.10.0) or later (otherwise you can pass `SKIP_BEDROCK2=1` to `make`).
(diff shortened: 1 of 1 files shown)


JasonGross:
@dderjoel Do you recall why `ocaml-num` was added here?  It seems to me that if you're installing Coq with a package manager, you shouldn't also need `ocaml-num`.  Furthermore, newer versions of Coq no longer use `num` and instead use `zarith`, so I'm wondering if I should update `ocaml-num` to "`ocaml-num` or `ocaml-zarith`" or if I should remove it entirely.

dderjoel:
Checking on OS X (2020 MacBook Air, M1, Darwin macm1 20.3.0 Darwin Kernel Version 20.3.0: Thu Jan 21 00:06:51 PST 2021; root:xnu-7195.81.3~1/RELEASE_ARM64_T8101 arm64), I needed to install `coq`, `ocaml-findlib` and `coreutils` (coreutils was needed to get realpath) `brew install coq ocaml-findlib coreutils`; then `make -j4 standalone-ocaml` was able to create the standalone bins to src/ExtractionOcaml/{unsaturated_solinas, word_by_word_montgomery, ...}.

On Arch (Linux intel8gen 5.11.14-arch1-1 #1 SMP PREEMPT Wed, 14 Apr 2021 12:06:34 +0000 x86_64 GNU/Linux) `pacman -S coq ocaml-findlib` was enough

On Ubuntu 21.04 (Linux ryzen7 5.11.0-14-generic #15-Ubuntu SMP Thu Apr 8 21:39:23 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux)  `apt install coq ocaml-findlib libcoq-ocaml-dev` (as in the Readme) was enough to build the standalone-ocaml's

Long story short: I think its safe to remove `ocaml-num` from the deps.

JasonGross:
Great, thanks!  I'll remove num and add coreutils on Mac

PR #878, opened 2020-08-20
Rupicola libonly

jadephilipoom:
See mit-plv/rupicola#7

The rupicola submodule build was slow because of some large examples, so I've added a `lib` target that builds the core library only (similar to bedrock2 `noex`). It's much faster now!

This is a pretty local change to the bedrock2 stuff, so I'll assume I'm free to merge once CI passes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -408,7 +408,7 @@ install-bedrock2:
 	$(MAKE) --no-print-directory -C $(BEDROCK2_FOLDER) install
 
 rupicola: bedrock2
-	$(MAKE) --no-print-directory -C $(RUPICOLA_FOLDER)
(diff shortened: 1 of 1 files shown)


JasonGross:
`readlink` on Mac OS does not support `-f`.  So instead of
```
$(shell git ls-files "$(LIBDIR)/*.v" | xargs readlink -f)
```
you probably want either
```
$(abspath $(shell git ls-files "$(LIBDIR)/*.v"))
```
or
```
$(realpath $(shell git ls-files "$(LIBDIR)/*.v"))
```
depending on whether you just want absolute paths, or if you also want to resolve symlinks to their underlying paths.

PR #879, opened 2020-09-01
Wrap bedrock2 C functions for BoringSSL

JasonGross:
Per discussion on the fiat-crypto call today, we now wrap the bedrock2 C functions so that BoringSSL can consume them.

Specific Questions:
1. Should there be a comment about this wrapper bit being unverified?
2. Should we make the wrapped functions `static`, or not?  Should we prefix them with `internal_`, or something else?


src/Bedrock/Standalone/Stringification.v
```diff
@@ -102,6 +105,37 @@ Definition bedrock_func_to_lines (f : func)
...
+                 (("*" ++ r0 ++ " = " ++ name)%string, outnames' ++ innames)
+            end%list in
+     ((precall ++ "(")
+        ++ (String.concat
+              ", "
+              (List.map
+                 (** TODO: Is there a better way to do this? *)
+                 (fun n => "(uintptr_t)" ++ n)
+                 all_args))
+        ++ ")")%string.
```
JasonGross, line 137: @jadephilipoom @andres-erbsen The way arguments are handled in bedrock2 is relatively finicky (I think, for example, that it reorders all list-typed returns to the beginning, followed by the arguments, followed by all but the last non-list-typed returns, and then the final non-list-type return is returned from the function; it's also not entirely clear to me how to get the C declaration types).  Do you have any better / more complete ideas for this bit of code?

fiat-bedrock2/src/curve25519_64.c
```diff
@@ -930,14 +971,19 @@ void fiat_25519_from_bytes(uintptr_t out0, uintptr_t in0) {
   return;
 }
 
+void fiat_25519_from_bytes(uint64_t out1[5], const uint64_t arg1[32]) {
```
JasonGross, line 974: Somehow we'll need to figure out how to know that this is supposed to be `uint8_t` ... :-(
```suggestion
void fiat_25519_from_bytes(uint64_t out1[5], const uint8_t arg1[32]) {
```

src/Bedrock/Standalone/Stringification.v
jadephilipoom: Not quite: it puts the list-typed returns at the beginning and then follows with the arguments, and then all non-list-type returns are returned from the function.

As for better ideas, what exactly are you trying to accomplish? This code is pretty opaque.

JasonGross: C functions can't return multiple values, so I think the bedrock2 C printer only returns the final noon-list value and passes the rest in as pointers.

I'm trying to write a C wrapper which obeys the calling conventions of C stringification code, inserts casts and passes them in to the bedrock2 functions.

andres-erbsen:
1. I think a comment along the lines of "the following code is not covered by coq proofs" would be helpful.
2. Wrapped functions should be static. `internal_` prefix sounds fine, or anything else really; I've also seen `_impl` suffix.

src/Bedrock/Standalone/Stringification.v
andres-erbsen: There are two translations going on here: fiat-crypto->bedrock2 translation changes returned list to destination-passing-style arrays at the beginning of the arguments, and bedrock2->C translation returns all values after the first by pointers at the end of the arguments. I think the latter is never triggered in fiat-crypto code? It's indeed awkward that the conventions are the opposite, and if one is to be changed, I think we should try to match BoringSSL.

JasonGross:
I've updated it to include comments.

I'll have to think about how to fix the issue with smaller integer types (maybe I can re-derive them from bounds?)

andres-erbsen:
perhaps `NOTE` instead of `WARNING`?

JasonGross:
> perhaps `NOTE` instead of `WARNING`?

Done

> I think the code @jadephilipoom wrote derives integer sizes from bounds, and I think it is only done for lists.

I've inserted a kludge of re-relaxing the bounds to integer sizes except this time explicitly allowing 8-bit integers.  Hopefully this will work with BoringSSL now.

JasonGross:
Success!  The bedrock2 code builds with BoringSSL and all tests pass!

jadephilipoom (approved):
Looks good to me now that the tests pass.

Also, just a note: although it's okay to occasionally request my review on PRs going forward (especially ones like this one, related to the bedrock2 backend), I may not respond very quickly or at all because work on this project is now happening only on my evenings and weekends.

PR #880, opened 2020-09-07
Connect bedrock2 montladder with algebraic scalarmult

jadephilipoom:
In terms of the "next steps" discussion on #869, this PR is the second bulletpoint :
 > - proving that the ladderstep and montladder Gallina implementations match the Gallina implementations from `Curves/Montgomery/XZ.v` so we can use the proofs from `XZProofs` to say that montgomery ladder is a group scalarmult

The punchline here is `scmul_func_correct`, which goes through `XZProofs` to prove that a bedrock2 implementation of group scalar multiplication corresponds to the specification in `Bedrock/Specs/Group.v`, where the group is instantiated with a Montgomery curve. `Print Assumptions` tells me it's closed under the global context, and `Check`ing it shows the following:
```coq
M.scmul_func_correct
     : forall char_ge_3 : Ring.char_ge 3,
       Ring.char_ge 5 ->
       Ring.char_ge 12 ->
       Ring.char_ge 28 ->
       forall (a b : F M_pos) (scmul : string) (b_nonzero : b <> 0%F),
       ((1 + 1 + 1 + 1) * a24)%F = (a - (1 + 1))%F ->
       (forall r : F M_pos, (r * r)%F <> (a * a - (1 + 1 + 1 + 1))%F) ->
       forall scalar_field_parameters : ScalarFieldParameters,
       ScalarFieldParameters_ok ->
       forall scalar_field_representation : ScalarRepresentation,
       program_logic_goal_for (M.scmul_func scmul)
         (forall
            functions : list (string * (list string * list string * cmd)),
          M.spec_of_from_bytes functions ->
          spec_of_montladder functions ->
          M.spec_of_to_bytes functions ->
          M.spec_of_scmul char_ge_3 a b scmul b_nonzero
            (M.scmul_func scmul :: functions))
where
?semantics : [ |- Semantics.parameters]
?semantics_ok : [ |- Semantics.parameters_ok ?semantics]
?field_parameters : [ |- FieldParameters]
?field_parameters_ok : [ |- FieldParameters_ok]
?field_representation : [ |- FieldRepresentation]
?field_representation_ok : [ |- FieldRepresentation_ok]
```

In order to make sure the field representation typeclasses are possible to instantiate, I've kept up a file `Field/Interface/Representation.v` that instantiates `FieldRepresentation` and `FieldRepresentation_ok`, but it is not currently hooked up to anything. The other typeclass instances (e.g. `FieldParameters`, `ScalarFieldParameters_ok`) are pretty simple and I'm not worried about them.

The top-level scalar multiplication function is implemented by:
1. stack-allocating the intermediate values needed by the Montgomery ladder
2. calling `from_bytes` on the input
3. calling the Montgomery ladder function
4. calling `to_bytes` to serialize the result
5. deallocating intermediates

Getting everything to play nicely together (rupicola-generated specs, bedrock2 stack allocation, algebraic proofs, etc) required a fair amount of side work. Other changes you'll see in this PR include:
- a proven correspondence between `Bignum` and byte arrays (this could be generalized into a correspondence between `array scalar` and `array ptsto` and upstreamed to bedrock2)
- `Placeholder` is now `Memory.anybytes` instead of just an alias for `FElem`, in order to fit with stack allocation and simplify specifications (they no longer need to take several arguments for all the placeholder values because there's an existential in `Memory.anybytes`; now specs only need the pointers)
- field compilation proofs are now better automated because I didn't want to change them all one by one
- scalar fields now have a parameter that declares how many bits they have, and this is used as the loop bound for Montgomery ladder instead of an extra standalone parameter

Next steps now are to:
- make the code generated in `Bedrock/Field` proven against the specs in `Bedrock/Specs/Field` instead of `COperationSpecifications`
- create an X25519 end-to-end example with assumed implementations for missing pieces like field-element inverse

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -108,6 +108,7 @@ src/Bedrock/Field/Translation/Proofs/ValidComputable/Func.v
 src/Bedrock/Group/Loops.v
 src/Bedrock/Group/Point.v
 src/Bedrock/Group/ScalarMult/LadderStep.v
+src/Bedrock/Group/ScalarMult/MontgomeryEquivalence.v
 src/Bedrock/Group/ScalarMult/MontgomeryLadder.v
 src/Bedrock/Group/ScalarMult/ScalarMult.v
 src/Bedrock/Specs/Field.v
--- a/_CoqProject
+++ b/_CoqProject
@@ -105,9 +105,11 @@ src/Bedrock/Field/Translation/Proofs/VarnameSet.v
 src/Bedrock/Field/Translation/Proofs/ValidComputable/Cmd.v
 src/Bedrock/Field/Translation/Proofs/ValidComputable/Expr.v
 src/Bedrock/Field/Translation/Proofs/ValidComputable/Func.v
+src/Bedrock/Group/Loops.v
 src/Bedrock/Group/Point.v
 src/Bedrock/Group/ScalarMult/LadderStep.v
 src/Bedrock/Group/ScalarMult/MontgomeryLadder.v
+src/Bedrock/Group/ScalarMult/ScalarMult.v
 src/Bedrock/Specs/Field.v
 src/Bedrock/Specs/Group.v
 src/Bedrock/Specs/ScalarField.v
--- a/src/Bedrock/Field/Common/Util.v
+++ b/src/Bedrock/Field/Common/Util.v
@@ -667,11 +667,11 @@ Section Scalars.
 
   Lemma scalar_to_bytes a x :
     Lift1Prop.iff1
+      (scalar a x)
       (array ptsto (word.of_Z 1) a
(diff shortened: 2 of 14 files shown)


andres-erbsen:
LGTM; Very exciting. I read through the non-tactic code and confirmed baseline reasonableness, and that it indeed connects the pieces I was expecting. The next steps can make sense as separate PRs so I think no need to keep this off master.

Don't bother generalizing the bedrock2 subarray lemma unles you need the general version, I am currently experimenting with alternative symbolic representations of consecutive structures in memory and may very well land on something slightly different anyway. And IIUC better handling of stackalloc is scheduled after a rupicola refactor which we think would break it anyway?

PR #882, opened 2020-10-05
Adapt w.r.t. coq/coq#13139.

ppedrot:
Do not merge yet.


ppedrot:
Actually, this should be backwards compatible, so it can be merged as-is.

PR #883, opened 2020-10-11
Clean up field specifications

jadephilipoom:
This is part of my work-in-progress on connecting the bedrock2 backend to the new field specifications and scalarmult implementation. Mainly, this change:
- moves the scalar field compilation lemmas to a new `ScalarField/` directory, so that the structure matches `Field/` and `Group/` and doesn't clutter the spec with non-trusted code
- changes the argument order for field operations so that the output pointer comes first instead of last, to match BoringSSL and the existing bedrock2 backend implementations
- introduces a separate separation-logic frame for each input instead of one for all inputs, because it should be okay to call `add(out, x, x)` (see #828)

### DIFF
--- a/src/Bedrock/Group/ScalarMult/LadderStep.v
+++ b/src/Bedrock/Group/ScalarMult/LadderStep.v
@@ -243,24 +243,24 @@ Print ladderstep_body.
 *)
 (* ladderstep_body =
  * fun field_parameters : FieldParameters =>
- *  (cmd.call [] add [expr.var "X2"; expr.var "Z2"; expr.var "A"];;
- *  cmd.call [] square [expr.var "A"; expr.var "AA"];;
- *  cmd.call [] sub [expr.var "X2"; expr.var "Z2"; expr.var "B"];;
- *  cmd.call [] square [expr.var "B"; expr.var "BB"];;
- *  cmd.call [] sub [expr.var "AA"; expr.var "BB"; expr.var "E"];;
- *  cmd.call [] add [expr.var "X3"; expr.var "Z3"; expr.var "C"];;
- *  cmd.call [] sub [expr.var "X3"; expr.var "Z3"; expr.var "D"];;
- *  cmd.call [] mul [expr.var "D"; expr.var "A"; expr.var "DA"];;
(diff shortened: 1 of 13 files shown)


src/Bedrock/Group/ScalarMult/LadderStep.v
```diff
@@ -243,24 +243,24 @@ Print ladderstep_body.
 *)
 (* ladderstep_body =
  * fun field_parameters : FieldParameters =>
- *  (cmd.call [] add [expr.var "X2"; expr.var "Z2"; expr.var "A"];;
```
JasonGross, line 246: Do you want to put this in one of the checked output files, rather than manually maintaining the comment?

jadephilipoom: Yes, that's a good idea. I'll add it.

andres-erbsen:
:+1: 

> introduces a separate separation-logic frame for each input instead of one for all inputs, because it should be okay to call add(out, x, x) (see #828)

is this change actually included? Maybe I missed it.

src/Bedrock/Specs/Field.v
```diff
@@ -77,55 +74,41 @@ Class FieldRepresentation_ok
...
+Notation binop_spec name op xbounds ybounds outbounds :=
+  (forall! (x y : felem) (px py pout : word),
+      (fun Rr mem =>
+         bounded_by xbounds x
+         /\ bounded_by ybounds y
+         /\ (exists Ra, (FElem px x * Ra)%sep mem)
+         /\ (exists Ra, (FElem py y * Ra)%sep mem)
+         /\ (Placeholder pout * Rr)%sep mem)
+        ===> name @ [pout; px; py] ===>
+        (fun _ =>
```
jadephilipoom, line 100: > is this change actually included? Maybe I missed it.

@andres-erbsen Here's the code that changes from one input frame to two for all two-input field operations. I made a single notation each for the spec of binary operations and unary operations to reduce boilerplate, so this change applies to all binary ops (for unary ops, of course, no change was required).

JasonGross (approved):
I have not read the code changes super-carefully, but seems good to me

PR #887, opened 2020-11-05
Make translate_cmd handle pairs

### LINKED ISSUE #886: Issues with the bedrock2 backend on functions taking a list as input and returning a tuple
> @jadephilipoom Both `fun (x1 : Z) (x2 x3 x4 x5 : list Z) => x1` and `fun (x1 : Z) (x2 x3 x4 x5 : list Z) => x2` synthesize correctly, but `fun (x1 : Z) (x2 x3 x4 x5 : list Z) => (x1, x2)` and `fun (x1 : Z) (x2 x3 x4 x5 : list Z) => (x2, x3)` both fail. On the other hand `fun (x1 x2 : Z) (x3 .. : listZ) => (x1, x2)` also does not synthesize correctly, which doesn't really agree with the fact that addcarryx and mulx synthesizes properly?

_Originally posted by @bshvass in https://github.com/mit-plv/fiat-crypto/pull/670#issuecomment-709117953_

------

My guess is that the commonality is that all of these functions take a list as input and return a tuple.  AFAIK, we don't currently have any other functions which both take at least one list as input and also return a tuple of things.

JasonGross:
This allows things like `fun (x y : Z) => (x, y)` and `fun (x y : Z) (z
: list Z) => (x, y, z)` to synthesize correctly.  This fixes #886 and is partial work towards making bedrock2 output handle divstep.

### DIFF
--- a/src/Bedrock/Field/Translation/Cmd.v
+++ b/src/Bedrock/Field/Translation/Cmd.v
@@ -62,33 +62,53 @@ Section Cmd.
     : nat (* number of variable names used *)
       * ltype t (* variables in which return values are stored *)
       * Syntax.cmd.cmd (* actual program *) :=
-    match e in expr.expr t0
-          return (nat * ltype t0 * Syntax.cmd.cmd) with
-    | expr.LetIn (type.base t1) (type.base t2) x f =>
-      let trx := assign nextn (translate_expr true x) in
-      let trf := translate_cmd (f (snd (fst trx))) (nextn + fst (fst trx)) in
-      ((fst (fst trx) + fst (fst trf))%nat,
-       snd (fst trf),
-       Syntax.cmd.seq (snd trx) (snd trf))
-    | expr.App
-        type_listZ type_listZ
-        (expr.App type_Z _ (expr.Ident _ (ident.cons _)) x) l =>
-      let trx := assign nextn (translate_expr true x) in
-      let trl := translate_cmd l (S nextn) in
-      ((fst (fst trx) + fst (fst trl))%nat,
-       snd (fst trx) :: snd (fst trl),
-       Syntax.cmd.seq (snd trx) (snd trl))
-    | expr.Ident type_listZ (ident.nil _) =>
-      (0%nat, [], Syntax.cmd.skip)
-    | expr.App _ _ f x =>
-      let v := translate_expr true (expr.App f x) in
-      assign nextn v
-    | expr.Ident _ i =>
-      let v := translate_expr true (expr.Ident i) in
-      assign nextn v
-    | expr.Var _ v =>
-      let v := translate_expr true (expr.Var v) in
-      assign nextn v
-    | _ => (0%nat, dummy_ltype _, Syntax.cmd.skip)
+    let default _
+        := match e in expr.expr t0
+                 return (nat * ltype t0 * Syntax.cmd.cmd) with
+           | expr.LetIn (type.base t1) (type.base t2) x f =>
+             let trx := assign nextn (translate_expr true x) in
+             let trf := translate_cmd (f (snd (fst trx))) (nextn + fst (fst trx)) in
+             ((fst (fst trx) + fst (fst trf))%nat,
+              snd (fst trf),
+              Syntax.cmd.seq (snd trx) (snd trf))
+           | expr.App
+               type_listZ type_listZ
+               (expr.App type_Z _ (expr.Ident _ (ident.cons _)) x) l =>
+             let trx := assign nextn (translate_expr true x) in
+             let trl := translate_cmd l (S nextn) in
+             ((fst (fst trx) + fst (fst trl))%nat,
+              snd (fst trx) :: snd (fst trl),
+              Syntax.cmd.seq (snd trx) (snd trl))
+           | expr.Ident type_listZ (ident.nil _) =>
+             (0%nat, [], Syntax.cmd.skip)
+           | expr.App _ _ f x =>
+             let v := translate_expr true (expr.App f x) in
+             assign nextn v
+           | expr.Ident _ i =>
+             let v := translate_expr true (expr.Ident i) in
+             assign nextn v
+           | expr.Var _ v =>
+             let v := translate_expr true (expr.Var v) in
+             assign nextn v
+           | _ => (0%nat, dummy_ltype _, Syntax.cmd.skip)
+           end in
+    match invert_expr.invert_pair_cps e (@translate_cmd) (@translate_cmd) with
+    | Some (a, b)
+      => match t
+               return let A := match t with type.base (A * B) => type.base A | _ => _ end in
+                      let B := match t with type.base (A * B) => type.base B | _ => _ end in
+                      (_ -> _ * ltype A * _) -> (_ -> _ * ltype B * _) -> (unit -> _ * ltype t * _) -> _ * ltype t * _
+         with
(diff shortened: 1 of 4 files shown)


src/Bedrock/Field/Translation/Cmd.v
```diff
@@ -62,33 +62,53 @@ Section Cmd.
...
+             let v := translate_expr true (expr.App f x) in
+             assign nextn v
+           | expr.Ident _ i =>
+             let v := translate_expr true (expr.Ident i) in
+             assign nextn v
+           | expr.Var _ v =>
+             let v := translate_expr true (expr.Var v) in
+             assign nextn v
+           | _ => (0%nat, dummy_ltype _, Syntax.cmd.skip)
+           end in
```
JasonGross, line 94: Note that this block is unchanged, just reindented

src/Bedrock/Field/Translation/Expr.v
```diff
@@ -256,10 +276,10 @@ Section Expr.
            {t} (e : @API.expr ltype t) : rtype t :=
     if cast_exempt e
     then translate_cast_exempt require_cast e
-    else if require_cast
+    else if (require_cast && negb (is_pair e))%bool
          then
            match e in expr.expr t0 return rtype t0 with
-           | expr.App _ _ f x =>
+           | expr.App s d f x =>
```
JasonGross:
```suggestion
           | expr.App _ _ f x =>
```

src/Bedrock/Field/Translation/Cmd.v
```diff
@@ -62,33 +62,53 @@ Section Cmd.
...
+    let default _
+        := match e in expr.expr t0
+                 return (nat * ltype t0 * Syntax.cmd.cmd) with
+           | expr.LetIn (type.base t1) (type.base t2) x f =>
+             let trx := assign nextn (translate_expr true x) in
+             let trf := translate_cmd (f (snd (fst trx))) (nextn + fst (fst trx)) in
+             ((fst (fst trx) + fst (fst trf))%nat,
+              snd (fst trf),
+              Syntax.cmd.seq (snd trx) (snd trf))
+           | expr.App
```
jadephilipoom, line 74: I _think_ you could add a case like this to recognize (left-associative only) pairs and translate them:
```suggestion
           | expr.App
               (type.base t2) (type.base (base.type.prod t1 t2)).
               (expr.App (type.base t1) _ (expr.Ident _ (ident.pair _ _)) x) p =>
             let trx := assign nextn (translate_expr true x) in
             let trp := translate_cmd y (S nextn) in
             ((fst (fst trx) + fst (fst trp))%nat,
              (snd (fst trx), snd (fst trp)),
              Syntax.cmd.seq (snd trx) (snd trp))
           | expr.App
```
Might need to put two cases and specify that `x` is either a `list Z` or a `Z`. But still, I think this version is clearer and matches with existing cases better.

src/Bedrock/Field/Translation/Cmd.v
```diff
@@ -62,33 +62,53 @@ Section Cmd.
...
+           => fun translate_cmd_a translate_cmd_b _
+              => let tra := translate_cmd_a nextn in
+                 let trb := translate_cmd_b (fst (fst tra) + nextn)%nat in (** TODO(Jade) Please check if this is right *)
+                 ((fst (fst tra) + fst (fst trb))%nat,
+                  (snd (fst tra), snd (fst trb)),
+                  Syntax.cmd.seq (snd tra) (snd trb))
+         | _ => fun _ _ default => default tt
+         end a b default
+    | None
+      => default tt
```
jadephilipoom, line 112: This is way, way more complicated than I would expect it to be, and I'm concerned that it will make the (quite central!) `translate_cmd` definition much harder to read, understand, and modify.

src/Bedrock/Field/Translation/Cmd.v
JasonGross: Indeed, this will only work with concrete type variables for `t1`.  I could do that, but I'm very wary, because this will enormously blow up the term size (nested pattern matches are desugared to account for all of their cases) and make proving anything about this function significantly slower.  It's not scalable, and I think a better approach is to provide dedicated functions for each decomposition.

jadephilipoom: It doesn't need to be scalable. We don't need to support a huge set of patterns here. But what we do need to do is have some ability to understand this code in the future. The other cases do matches like this, so it's not going to be massively slowing things down. Translation to bedrock2 is reasonably fast; I don't think we have a need to heavily optimize for speed at this point, especially not if it comes with heavy tradeoffs.

JasonGross: The cost is not in reduction, the cost is in how long it takes to typecheck `translate_cmd` and prove it correct.  I expect it to add about a full second to the typechecking time of `translate_cmd` and somewhere between 30s and a couple minutes to the time it takes to run the proof script for `translate_cmd`-correctness.  I'll try it and get back to you with a performance diff

JasonGross: Also, why is it `(S nextn)` rather than `fst (fst trx) + nextn`?  What if the first component is a list?

JasonGross: Actually, no, this code only works if you know how many elements the tuple has ahead of time, and it still requires 2<sup>*k*</sup> cases where *k* is the number of elements of the tuple.  I have an open issue for supporting this sort of non-linearity (currently, it's forbidden to include the same variable name in two locations in a pattern) at https://github.com/coq/coq/issues/6320.  (Currently supporting code like this requires depending on Equations.)

JasonGross: I've pushed a change which hopefully makes this significantly more readable, by making more liberal use of option types and factoring out the tricky `match` `return` into a separate function which has a very obvious semantics.  What do you think?

jadephilipoom: > The cost is not in reduction, the cost is in how long it takes to typecheck `translate_cmd` and prove it correct. I expect it to add about a full second to the typechecking time of `translate_cmd` and somewhere between 30s and a couple minutes to the time it takes to run the proof script for `translate_cmd`-correctness. I'll try it and get back to you with a performance diff

I think this is where we disagree -- I think that a second to typechecking and a minute or so for proofs is a price I'm willing to pay for readability in this particular case. I wouldn't feel that way about just any function, but this one (along with `translate_func` and `translate_expr`) is one of the first that a person new to this code would read in order to see how the bedrock2 translation works. Especially with my having a different day job now, I want to make sure it's not too hard for someone else to get up to speed.

However, I'm very happy with the latest version! It's very readable to me and I'm pleased that it's fast too. Thanks!

> Also, why is it `(S nextn)` rather than `fst (fst trx) + nextn`? What if the first component is a list?

Because I wrote that on the fly without testing it -- you're correct.

jadephilipoom (approved):
Thanks for making the change more readable! Looks good to me now.

JasonGross:
Thanks, I'll merge this once I get the CI to pass on old versions of Coq.

@jadephilipoom I haven't extended the notion of valid expressions to handle pairs of lists; should I be adding that case somewhere?

JasonGross:
(And is there supposed to be a test somewhere (in the sanity checking of bedrock2 output, perhaps?) that the expression is valid?)

PR #895, opened 2020-11-10
Fix bedrock2 wrapper functions

### LINKED ISSUE #891: Unverified bedrock2 wrapper functions are incorrect on divstep
https://github.com/mit-plv/fiat-crypto/blob/4be7fee8f4d551d67e2f0f5de71e07593834a64f/src/Bedrock/Standalone/Stringification.v#L108-L137
divstep contains mixed integers and lists in the output.  Can @jadephilipoom or @andres-erbsen help me write a better version of this?  Roughly, I need to somehow perform the same algorithm as `make_innames` + `make_outnames` + `translate_func` + `c_func` to print the function prototype, except instead of starting with a function `nat -> string` and getting out `ltype`s, I already have `string`s where each `ltype` should be (and they are not parameterized by `nat`, but instead are already in `type.for_each_lhs_of_arrow` / type-indexed structures.  How can I get this to work?

JasonGross:
This is kind-of hacky.  Fixes #891 

It might be nice to factor things better in some way...


jadephilipoom (approved):
Can we either put the `*VarData.v` files in `Bedrock/Standalone` or make a `Bedrock/Field/Stringification` directory and put them there along with the current `Standalone/Stringification.v`? The `Bedrock/Field/Translation` directory is pretty complex already, but I think it would be good to try to maintain that it only contains things directly relevant to *translating to bedrock2*, which seems separate from *printing bedrock2*.

Also, it seems like we might at some point want to be more systematic about the structure in `Flatten` -- it would be nice if the new `FlattenVarData` could pull from some generalized file that `Flatten` also pulls from, so that they don't have to be manually kept in sync. But probably not the top priority at the moment.

Other than that, looks good to me!

JasonGross:
Belatedly, thanks for the review!  I've moved the files as requested, and will merge once the CI passes

PR #897, opened 2020-11-20
 Adapt to fixing dropped implicit arguments in Context

herbelin:
Implicit arguments in `Context` were discarded. Coq PR [#11390](https://github.com/coq/coq/pull/11390) preserves them. The present PR adapts fiat-crypto to the new semantics.

There were several places in fiat-crypto using implicit arguments in `Context`. Depending on whether it looked intended or not that they were there, or depending on how easy it was to adapt to a change, I sometimes removed the implicit arguments and sometimes preserve them.

As a result, the PR is backward-incompatible and has to merged synchronously with coq/coq#11390. If you'd prefer a backward-compatible change, the implicit arguments should instead be removed in the few places where it has an impact.

### DIFF
--- a/src/Arithmetic/Core.v
+++ b/src/Arithmetic/Core.v
@@ -440,7 +440,7 @@ Module B.
     Qed. Hint Rewrite eval_negate_snd : push_basesystem_eval.
 
     Section Carries.
-      Context {modulo_cps div_cps:forall {R},Z->Z->(Z->R)->R}.
+      Context {modulo_cps div_cps:forall R,Z->Z->(Z->R)->R}.
       Let modulo x y := modulo_cps _ x y id.
       Let div x y := div_cps _ x y id.
       Context {modulo_cps_id : forall R x y f, modulo_cps R x y f = f (modulo x y)}
@@ -730,7 +730,7 @@ Module B.
       .
 
       Section Carries.
-        Context {modulo_cps div_cps:forall {R},Z->Z->(Z->R)->R}.
+        Context {modulo_cps div_cps:forall R,Z->Z->(Z->R)->R}.
         Let modulo x y := modulo_cps _ x y id.
         Let div x y := div_cps _ x y id.
         Context {modulo_cps_id : forall R x y f, modulo_cps R x y f = f (modulo x y)}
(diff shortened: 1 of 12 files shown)


JasonGross:
Having implicit arguments is nice, bit we have a policy of retaining backwards compatibility.  Could you submit a backwards-compatible PR?  Thanks

herbelin:
I see two ways to retain compatibilities
- just remove the implicit arguments used in `Context`
- keep them when they were relevant, but add a `Local Arguments` to compensate that they were lost

For instance, this would turn
```
       Context (eta : forall {A B}, A * B -> A * B).
```
from file `src/Compilers/Eta.v` into either
```
       Context (eta : forall A B, A * B -> A * B).
```
or
```
       Context (eta : forall {A B}, A * B -> A * B).
       Local Arguments eta {A B}.
```
If you don't have a preference, I'll go for the second option (possibly adding a remark that 8.14 makes it unnecessary).

JasonGross:
I have a slight preference for the second option

herbelin:
> I have a slight preference for the second option

OK, I pushed a new version, hopefully backwards-compatible. Tell me if ok.

JasonGross:
Thanks!  LGTM assuming the CI passes

herbelin:
The failure on `Hint Resolve (ex_intro _ nil)` in `src/Experiments/NewPipeline/AbstractInterpretationWf.v` (not tested on Coq CI) is due to coq/coq#13139. Shall I fix it at the same time?

JasonGross:
Yes, please, and thanks!

herbelin:
While at it, I also "fixed" an `Extraction Language Ocaml` -> `Extraction Language OCaml` but this breaks compatibility and I don't know what to do.

JasonGross:
Maybe it is time to archive the NewPipeline folder of sp2019latest, given that the state it's in in sp2019latest isn't really interesting.  I'll think on it, and probably just delete it.

JasonGross:
If you rebase on the tip of sp2019latest, the extraction command should be gone

JasonGross:
Sorry, I missed some changes. sp2019latest now builds again, so if you rebase once more, it should hopefully work.

PR #900, opened 2020-12-13
Remove an invalid "at" occurrence of "rewrite"

herbelin:
PR coq/coq#13568 detected that an invalid occurrence was tacitly accepted by a call to `rewrite`. This fiat-crypto PR removes it so that fiat-crypto-legacy is successful on Coq CI.

This is a backward-compatible change that can be merged as soon as now.


JasonGross:
Thanks!

PR #903, opened 2021-01-20
First stab at parsing assembly for validating assembly output

JasonGross:
The format used here is (note that juxtaposition is used to denote whitespace-insensitive sequencing, "(foo)?" means "optionally foo", *italicized* things are nonterminals, and parentheses are used for grouping): `[` comma-separated list of *Instruction* `]`
- An *Instruction* is `{` *instruction key-value pair* `,` *dst key-value pair* `,` *src key-value pair* (`,`  *name key-value pair*)? `,`?  `}`
- An *instruction key-value pair* is `"instruction"` `:` `"` *OpCode* `"`
- A *dst key-value pair* is `"dst"` `:` `"` `[` comma-separated list of *arguments* `]`
- A *src key-value pair* is `"src"` `:` `"` `[` comma-separated list of *arguments* `]`
- A *name key-value pair* is `"name"` `:` `"` `[` comma-separated list of *names* `]`
- An *argument* is `"` *argument-string* `"`
- An *argument-string* is either a *register* or a *memory expression* or a *constant* or a *flag*
- A *register* is the name of any constructor of the inductive `REG`
- A *memory expression* is (`byte `)? `[` *register* ((`+` or `-`) *exact integer expression*)? `]`
- A *constant* is `CONST ` *exact integer expression*
- A *flag* is the name of any constructor of the inductive `FLAG`
- A *name* is any non-empty sequence of upper- or lower-case letters followed by zero or more decimal digits
- An *exact integer expression* is any valid mathematical expression that evaluates exactly to an integer and consists of binary literals (starting with `0b`), octal literals (starting with `0o`), decimal literals, hexadecimal literals (starting with `0x`), using decimal points for fractional parts and using `e`/`E` for scientific notation (excepting hexadecimal literals) or `p`/`P` for the binary version of scientific notation (in both cases, the exponent must be a literal of the same base as the part before the `e`/`E`/`p`/`P` and does not require a prefix), and uses only the operations `+`, `-`, `*`, `/`, and `^` and parentheses for grouping.  (Note that this is probably overly powerful, but I just reused the number parser that is used for parsing the expression for the prime.)  Note further that we require that every subterm of the expression evaluate to a rational number (so `(2^(1/2))^2` is not considered an exact integer expression).

cc @andres-erbsen , do you think the data-structures here are a good place to start symbolic evaluation in Coq?  Should we be parsing the assembly directly instead?  Should we make the argument lists match more closely with assembly (e.g., not mention flags in dst, not get the same register as dst and src in things like inc?  maybe just have one list of arguments?)


src/Assembly/Model.v
```diff
@@ -0,0 +1,606 @@
...
+      match f with OF => v | _ => ofv end).
+Definition set_flag (st : flag_state) (f : FLAG) (v : bool) : flag_state
+  := set_flag_internal st f (Some v).
+Definition havoc_flag (st : flag_state) (f : FLAG) : flag_state
+  := set_flag_internal st f None.
+Definition havoc_flags (st : flag_state) : flag_state
+  := (None, None, None, None, None, None).
+
+Definition reg_state := list N.
+Definition index_and_shift_and_bitcount_of_reg (r : REG) : nat * N * N
```
JasonGross, line 374: Fix ordering; also define index -> widest register and convenience function for displaying register state

JasonGross:
Merging on Andres' offline ack

Issue #820, opened 2020-06-24
Using .h rather than .c, and using `__attribute__((unused))`?


JasonGross:
> Optional/follow-up/whatever: if you're outputting a file that's meant to be #included somewhere (given the symbols are `static` anyway), there may be something to be said for the files being named `foo.h` and maybe to mark all symbols with `__attribute__((unused))`. We use this definition:
>
> ```
#if defined(__GNUC__) || defined(__clang__)
#define OPENSSL_UNUSED __attribute__((unused))
#else
#define OPENSSL_UNUSED
#endif
```
>
> (Looking at both `__GNUC__` and `__clang__` is a little goofy. Normally Clang aims to be GCC-compatible, so it defines `__GNUC__` and `__attribute__` is indeed a GNU C extension. But `clang-cl` is Clang in MSVC-compatible mode. That defines `__clang__`, `_MSC_VER`, but *not* `__GNUC__`. That usually does the right thing, but it still implements a lot of the GCC extensions and warning flags, so sometimes we need to explicitly include it.)
>
> That should avoid the need for these.

_Originally posted by @davidben in https://github.com/mit-plv/fiat-crypto/timeline_

**Comment by chjj** [comment]
Just a thought: if fiat is now going to be in the business of generating compiler-specific code to avoid warnings, it would also be nice to put `__extension__` in front of the `__int128` typedefs to get `gcc -std=c89 -pedantic` to stop complaining about them (note that clang does _not_ complain about this).

The logic would be similar, e.g.

``` c
#ifdef __GNUC__
#  define FIAT_EXTENSION __extension__
#else 
#  define FIAT_EXTENSION
#endif

FIAT_EXTENSION typedef unsigned __int128 fiat_uint128_t;
```

As far as I can tell, `__extension__` has been a GCC keyword almost as far as back as the commit history goes (since the early '90s at least).

`__attribute__((unused))` seems to have been supported from GCC 2.7 and up (also very old).

PR #906, opened 2021-02-03
C Stringification: prepend __extension__ before the typedef with __int128

hannesm:
This allows the generated C code to compile with gcc and -Wpedantic.

Suggested in #820 by @chjj

### DIFF
--- a/src/Stringification/C.v
+++ b/src/Stringification/C.v
@@ -104,8 +104,14 @@ Module Compilers.
                             "typedef signed char " ++ prefix ++ "int1;"]%string
                     else [])
                 ++ (if IntSet.mem uint128 bitwidths_used || IntSet.mem int128 bitwidths_used
-                    then ["typedef signed __int128 " ++ prefix ++ "int128;";
-                            "typedef unsigned __int128 " ++ prefix ++ "uint128;"]%string
+                    then ["#ifdef GNUC";
+                          "#  define FIAT_EXTENSION __extension__";
+                          "#else";
+                          "#  define FIAT_EXTENSION";
+                          "#endif";
+                          "";
+                          "FIAT_EXTENSION typedef signed __int128 " ++ prefix ++ "int128;";
+                          "FIAT_EXTENSION typedef unsigned __int128 " ++ prefix ++ "uint128;"]%string
                     else [])
                 ++ [""
                     ; "#if (-1 & 3) != 3"
--- a/src/Stringification/C.v
+++ b/src/Stringification/C.v
@@ -104,14 +104,15 @@ Module Compilers.
                             "typedef signed char " ++ prefix ++ "int1;"]%string
                     else [])
                 ++ (if IntSet.mem uint128 bitwidths_used || IntSet.mem int128 bitwidths_used
-                    then ["#ifdef __GNUC__";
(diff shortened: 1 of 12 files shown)


JasonGross:
Thanks!  Please also regenerate the C files (run `make c-files` or just `make` and let the build complete).  Let's also add `-Wpedantic` to the cflags we test with by adding it at https://github.com/mit-plv/fiat-crypto/blob/7b250931d9ec98a45f365b505f037f8a810a60f7/Makefile#L498

The current CI error is due to a packaging issue with my ppa which I hope will be fixed within an hour or so (but may take up to a couple of days).

hannesm:
@JasonGross sure, done in 602069d  (feel free to squash if that's preferred)

JasonGross:
Thanks!  This doesn't seem to work, though:
```
fiat-c/src/curve25519_64.c:24:31: error: ISO C does not support ‘__int128’ types [-Werror=pedantic]
 FIAT_EXTENSION typedef signed __int128 fiat_25519_int128;
```

JasonGross:
Maybe you meant `__GNUC__` rather than `GNUC`?

hannesm:
indeed, sorry about that. I did not notice since I'm using on my development machine clang and there no such warning is issued, even with `-Wpedantic`. I fixed it in f3fc8ce (using `__GNUC__`).

chjj:
It might be wise to have a different prefix for the extension macro in each file: `fiat_p224_extension`, `fiat_p256_extension` etc. Otherwise GCC will complain about macro redefinition if two files are included.

src/Stringification/C.v
```diff
@@ -104,8 +104,14 @@ Module Compilers.
...
-                    then ["typedef signed __int128 " ++ prefix ++ "int128;";
-                            "typedef unsigned __int128 " ++ prefix ++ "uint128;"]%string
+                    then ["#ifdef __GNUC__";
+                          "#  define FIAT_EXTENSION __extension__";
+                          "#else";
+                          "#  define FIAT_EXTENSION";
+                          "#endif";
+                          "";
+                          "FIAT_EXTENSION typedef signed __int128 " ++ prefix ++ "int128;";
+                          "FIAT_EXTENSION typedef unsigned __int128 " ++ prefix ++ "uint128;"]%string
```
JasonGross, line 114: As per @chjj 's suggestion:
```suggestion
                    then let FIAT_EXTENSION := (String.to_upper prefix ++ "FIAT_EXTENSION")%string in
                         ["#ifdef __GNUC__";
                          "#  define " ++ FIAT_EXTENSION ++ " __extension__";
                          "#else";
                          "#  define " ++ FIAT_EXTENSION;
                          "#endif";
                          "";
                          FIAT_EXTENSION ++ " typedef signed __int128 " ++ prefix ++ "int128;";
                          FIAT_EXTENSION ++ " typedef unsigned __int128 " ++ prefix ++ "uint128;"]%string
```

hannesm:
sure, as you like. (running `make c-files` again and will push an update). I'm not entirely sure which C compilers fiat-crypto targets -- in both clang and gcc we can actually emit `__extension__` unconditionally (they also both define `__GNUC__`)...

JasonGross:
Thanks!

PR #907, opened 2021-02-24
Adjust Go names

JasonGross:
As per https://github.com/golang/go/issues/40171#issuecomment-780139520


mdempsky:
I agree with the comments from https://github.com/golang/go/issues/40171#issuecomment-785314910:

Package names (i.e., the identifier after the `package` keyword at the top of the .go files) are usually just all lower-case without underscores. So a name like `package fiatcurve25519` (or even just `package curve25519`) would be more common than `FiatCurve25519`.

And also like mentioned, since imported identifiers are (almost) always used qualified, we tend to prefer avoiding repetition, so the caller can just write `fiatcurve25519.Add` instead of `fiatcurve25519.Curve25519Add`.

How feasible is restructuring file output? I think ideally, the directory layout would look like:

````
fiat-go/
  go.mod
  32/
    curve25519/
      curve25519.go  // package curve25519; func Add(...); ...
    p256/
      p256.go // package p256; func Add(...); ...
    ...
  64/
    curve25519/
      curve25519.go
    p256/
      p256.go
```

That would allow users to write something like:

```
import "github.com/mit-plv/fiat-crypto/fiat-go/32/curve25519"
...
    curve25519.Add(...)
```

which would be very idiomatic.

JasonGross:
> How feasible is restructuring file output?

Pretty easy, it's all controlled by the `Makefile`.  I'll aim to do this restructuring and renaming within the next couple of days.

mdempsky:
Great, SGTM!

As for creating the `go.mod` file, you should be able to just run `go mod init github.com/mit-plv/fiat-crypto/fiat-go` in the fiat-go directory. Or just manually create it as:

```
module github.com/mit-plv/fiat-crypto/fiat-go

go 1.15
```

And then as long as `go build ./...` then works from within the fiat-go directory, you should be good and Go users can start directly using your generated code.

JasonGross:
Nearly done refactoring here.  Now I just need to regenerate the files and push the change that makes the package names match the file names.  One remaining question: What should be done for `p448_solinas`?  Should the package name be `p448solinas`?  Should the file and directory names also become `p448solinas/p448solinas.go`?  (Currently the package is `p448` inside `p448_solinas.go`)

JasonGross:
Also, another question: `go build ./...` succeeds with no output after a second or two.  Is this how it's supposed to work?  (Note that I'm using go 1.13, not 1.15, idk if that matters.)

JasonGross:
> Now I just need to regenerate the files and push the change that makes the package names match the file names.

This change has been made and pushed here.

mdempsky:
> Now I just need to regenerate the files and push the change that makes the package names match the file names.

Sorry, I should have mentioned this up front: the file names don't really matter. Go compiles all .go files within a directory as a single compilation unit. It's only the directory path and the `package` line within the files that really matter.

That said, for single file packages, naming the file the same as the package is pretty common.

> What should be done for p448_solinas? 

`p448solinas` would be more idiomatic for package name.

> Also, another question: go build ./... succeeds with no output after a second or two. Is this how it's supposed to work?

Yeah, `go` tools tend to be quiet by default, and only print output if any errors are encountered. You can try running `go build -a -v ./...` if you want to confirm they're actually being built.

mdempsky:
The diffs look good.

Only suggestion I have is to get rid of the "src" subdirectory, and just putting the "32" and "64" directories directly within "fiat-go". But if having the "src" directory makes things easier for you, it's not a big deal either. It'll just appear in users' import directives.

JasonGross:
> `p448solinas` would be more idiomatic for package name.

Done



> Only suggestion I have is to get rid of the "src" subdirectory, and just putting the "32" and "64" directories directly within "fiat-go".

Done

-------

Thanks for your feedback!  I've updated the PR and will merge once the CI passes, unless there are more objections or suggestions.

mdempsky:
Everything looks good to me.

PR #908, opened 2021-02-25
Add support for interaction with files

JasonGross:
We now support the `--hints-file` and `--output-asm` arguments (both
unused) and the `--output` / `-o` argument.  The first two are initial
support for interacting with assembly hint files, though we don't
currently do anything with the arguments other than opening and reading
the files, and parsing the assembly.

cc @andres-erbsen, the relevant entry point is the function `Synthesize` at the bottom of `Primitives.v`.


JasonGross:
Going to merge this, having gone through it with @andres-erbsen in meeting with no objections

PR #914, opened 2021-03-25
Bump coqprime and rupicola

JasonGross:
This should hopefully eliminate uses of omega, making fiat-crypto
compatible with https://github.com/coq/coq/pull/13741


JasonGross:
@jadephilipoom @andres-erbsen @samuelgruetter The build fails with
```
File "./src/Bedrock/Group/ScalarMult/LadderStep.v", line 207, characters 4-9:
Error: The reference setup was not found in the current environment.
```
This is
https://github.com/mit-plv/fiat-crypto/blob/bf8a76e1ba4c1e795e74ed2968fc4a92bc9b21ee/src/Bedrock/Group/ScalarMult/LadderStep.v#L194-L208

How should this be fixed?

andres-erbsen:
Maybe @cpitclaudel knows about some Rupicola change that could have caused this?

cpitclaudel:
> Maybe @cpitclaudel knows about some Rupicola change that could have caused this?

Yes: Rupicola shouldn't be bumped, ideally.  Do you need to bump it?  The definitions that teh fiat-crypto stuff depend on changed a few months ago, and the fiat-crypto stuff needs to be ported when bumping rupicola.

If you need to bump your bedrock2 dependency transitively, make a branch in the rupicola repo and use that in your submodule.

JasonGross:
Yeah, we need to bump our dependencies to a version that doesn't use `omega` and instead use on `lia`, for compatibility with https://github.com/coq/coq/pull/13741.  I'll try making a branch

samuelgruetter:
bedrock2 stopped using omega a while ago (I think before https://github.com/mit-plv/bedrock2/tree/eeeb8eb9b52ff4bf460a8dcd8cc3a3b6ed325dbf, namely here: https://github.com/mit-plv/coqutil/commit/da5b2d3e5de32164fa9c6535f2146ceca9fe4bb3), so maybe not bumping bedrock2 would just work?

JasonGross:
https://github.com/mit-plv/bedrock2/commit/eeeb8eb9b52ff4bf460a8dcd8cc3a3b6ed325dbf uses https://github.com/mit-plv/coqutil/commit/c8006ceca816076b117c31d7feaefb5bbb850754 which is too old.  I'll try to be more fine-grained

JasonGross:
I think we need https://github.com/mit-plv/bedrock2/commit/f8325200d330d5c7db958edb16e4904b663f9c83 or newer

JasonGross:
Now fails with
```coq
File "./src/Bedrock/Group/ScalarMult/ScalarMult.v", line 120, characters 30-41:
Error: The reference Syntax.func was not found in the current environment.
```
This seems like a change in bedrock2?

andres-erbsen:
This sounds like bedrock2 is older than the caller expects? https://github.com/mit-plv/bedrock2/blame/master/bedrock2/src/bedrock2/Syntax.v#L39-L43

JasonGross:
Argh, is the bedrock2 commit history non-linear?      https://github.com/mit-plv/bedrock2/commit/f8325200d330d5c7db958edb16e4904b663f9c83 
was committed later than https://github.com/mit-plv/bedrock2/commit/eeeb8eb9b52ff4bf460a8dcd8cc3a3b6ed325dbf but it seems that neither is an ancestor of the other.  What's the oldest commit that includes both of these?

andres-erbsen:
https://github.com/mit-plv/bedrock2/commit/578e719f58f54988348856b23279031cdc2a3107
history

JasonGross:
Thanks @andres-erbsen !

PR #917, opened 2021-03-28
Bump bedrock2 via rupicola again

JasonGross:
There were still some uses of `omega`.  This should hopefully actually
eliminate all uses of omega, making fiat-crypto compatible with
https://github.com/coq/coq/pull/13741

### DIFF
--- a/.github/workflows/coq.yml
+++ b/.github/workflows/coq.yml
@@ -15,12 +15,12 @@ jobs:
       matrix:
         env:
         - { COQ_VERSION: "8.12.0", COQ_PACKAGE: "coq-8.12.0 libcoq-8.12.0-ocaml-dev", SKIP_BEDROCK2: "" , SKIP_VALIDATE: "" , PPA: "ppa:jgross-h/many-coq-versions-ocaml-4-05" }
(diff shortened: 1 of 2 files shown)


JasonGross:
This fails with
```
File "./src/Bedrock/Field/Translation/Parameters/Defaults.v", line 45, characters 2-265:
Error: Error: Non exhaustive pattern-matching: no clause found for pattern
expr.inlinetable _ _ _
```
@andres-erbsen @samuelgruetter (@jadephilipoom ) I believe this is the oldest version of bedrock2 compatible with coq/coq#13741.  Is this easily fixible?

samuelgruetter:
Fixing this specific error might be as easy as adding a line like `| expr.inlinetable _ _ index => error_free_expr index` (untested), but I cannot estimate how many further incompatibility errors would show up after this one...

JasonGross:
Hrm, it seems we need an even newer version to include https://github.com/tchajed/coq-record-update/commit/ad3f28f69b388063a1da245c3b7a8bd56a2330d9

JasonGross:
I guess we need at least https://github.com/mit-plv/bedrock2/commit/5fec31330c1c7f5ae84ed821cf3bd84140ec6486, hopefully that's not too new

PR #920, opened 2021-03-30
Remove src/Bedrock/Field/Synthesis/Generic/Copy.v

JasonGross:
The file does not build.  Until recently, this file was not tracked in _CoqProject, and was not
building.  It's still in git history, has not been touched in a month,
and was never tracked on the CI.


DIJamner (approved):
I didn't realize it wasn't in the build pipeline when I committed it, it used to at least work on my machine. It's fine to remove for now, once I get to it, I'll fix it up and re-add it.

PR #928, opened 2021-04-09
Pipe through an always-false equivalence checker

JasonGross:
This should be enough rope to check equivalence of PHOASTs with assembly
expressions.  It's plausible we want more control over the calling
conventions from the command line, for example, specifying:
- [x] whether to do inputs first or outputs first
- [x] whether arguments are RTL or LTR
- [ ] what to do when we run out of registers / how arguments are stored on
  the stack
- [ ] support for a number of common calling conventions

We may also want to emit better headers for the assembly files, rather
than replicating the header of the input file.

@andres-erbsen Please review `Equivalence.v`.  Are the relevant bits understandable?  The relevant bits to extend are:
- `Inductive EquivalenceCheckingError` for the various ways the equivalence checker can fail (and then the `show_lines` instance to print it that will be an incomplete `match` once there are more constructors)
- `Local Notation var t` (what data do you want to store in PHOAS var nodes?) (and the corresponding `reg_data_to_var`)
- the second `refine` in `Definition check_equivalence` (which is currently just `if false then Success else Failure`)


JasonGross:
TODO: add `--error-on-unused-hints` to make the pipeline error out if some global labels exist but are not matched to synthesized functions

src/Assembly/Equivalence.v
```diff
@@ -0,0 +1,278 @@
...
+Class assembly_output_first_opt := assembly_output_first : bool.
+Typeclasses Opaque assembly_output_first_opt.
+(** Should we assign registers to the arguments in left-to-right or right-to-left order? *)
+Class assembly_argument_registers_left_to_right_opt := assembly_argument_registers_left_to_right : bool.
+Typeclasses Opaque assembly_argument_registers_left_to_right_opt.
+(** Stack size (in bytes) *)
+Class assembly_stack_size_opt := assembly_stack_size' : option N.
+Typeclasses Opaque assembly_stack_size_opt.
+(** TODO(for Andres from Jason): What should this be? *)
+Definition default_assembly_stack_size := 40%N.
```
andres-erbsen, line 35: The literal argument to the first `sub rsp, LITERAL` in the code :).

src/Assembly/Equivalence.v
```diff
@@ -0,0 +1,278 @@
...
+       => fun 'tt => []
+     end%list.
+
+Definition reg_list_of_reg_data {t} : reg_data t -> list (REG * nat)
+  := match t with
+     | type.base t => @reg_list_of_base_reg_data t
+     | type.arrow _ _ => fun absurd : Empty_set => match absurd with end
+     end.
+
+Definition prod_sequence_reg_available' {E A B C}
```
andres-erbsen, line 117: I got lost about here

src/Assembly/Equivalence.v
```diff
@@ -0,0 +1,278 @@
...
+    Proof.
+      refine (
+          let reg_available := assembly_calling_registers (* registers available for calling conventions *) in
+          (* outputs come first, assuming assembly_output_first *)
+          let res := prod_sequence_reg_available
+                       [] assembly_output_first
+                       (fun reg_available => make_output_reg_data reg_available out_bounds)
+                       (fun reg_available => make_input_reg_data reg_available arg_bounds)
+                       reg_available in
+          (decl <- match res with
```
andres-erbsen, line 247: from here on generally makes sense

src/Assembly/Equivalence.v
```diff
@@ -0,0 +1,278 @@
...
+           (* we now construct the inputs to the equivalence checker:
+              - the list of registers and array sizes for outputs
+              - the list of registers and array sizes for inputs
+              - the stack size
+              - the expression specialized to some appropriate variable type
+              - the assembly expression is available in the context as asm *)
+           let output_register_list : list (REG * nat) := reg_list_of_base_reg_data output_registers in
+           let input_register_list : list (REG * nat) := reg_list_of_input_reg_data input_registers in
+           let stack_size : nat := N.to_nat assembly_stack_size in
+           let ast : API.expr (type.base (type.final_codomain t))
```
andres-erbsen, line 262: well ok I don't get this line either but probably that's fine

andres-erbsen (commented):
Overall looks good

src/Assembly/Equivalence.v
JasonGross: So this should not be a command-line argument at all, and should instead be parsed out of the assembly code?  Or something else?  (Is it only needed for the proofs and not for equivalence checking?)

andres-erbsen: I think this default is very good, but I'd prefer keeping the argument unless it is a lot of code to support it.

JasonGross: > this default

40, or "The literal argument to the first sub rsp, LITERAL"?

> I'd prefer keeping the argument unless it is a lot of code to support it.

Not very much code, and sounds good.

src/Assembly/Equivalence.v
JasonGross: The purpose of the next block of code is to use the bounds data from bounds analysis, together with a list of registers, to populate the dependently typed data structure which holds registers and array lengths for each argument/output (and I expect we'll want to use some variant of this for `var` nodes, though it's plausible that instead we want to somehow flatten things earlier?).

The `prod_sequence_reg_available` definition in particular is extruded for the following pattern: we sometimes (at product types, arrow types, and when combining inputs and outputs) have two functions of the shape "populate some data structure with available registers, drawing from a list of available registers, and either error with a count of how many more registers we'd need or else succeed and also tell me what registers are left over for the next call" (morally we're in a state monad + an error monad or two).  Furthermore, while the combined datastructure will always have the same shape, depending on arguments we may want to draw registers first for the first function (if we're going left to right) or first for the second function (if we're going right to left) (and there's a similar choice about outputs on the left vs outputs on the right).  This function captures this pattern, and the unprimed version captures the "which one first" bit.

If you think this should all be designed in some other way, let me know (or if you want to chat about how we should design the interface with PHOAS, I'm happy to chat about that).

src/Assembly/Equivalence.v
JasonGross: I think this line is somewhat important to get.  We start with a PHOAS `Expr`, which is parameterized over var types and headed by some abstraction nodes for the arguments to the function.  This line peels off all of the abstraction nodes and applies the underlying abstractions to the var data that we generated from assigning to each input variable a register and an array length.  So after this line (in the current state), the AST type no longer has any arrows, and the uses of the input variables are `Var` nodes holding their associated registers and array lengths.

JasonGross: In particular:
- `type.base` expresses that there are no more arrows
- `type.final_codomain t` expresses that it still has the same return type underneath all the arrows that were removed
- `invert_expr.smart_App_curried` is peeling off the `Abs` nodes all at once (probably it should be named `uncurried`)
-  `expr _` is specializing the AST to whatever var type we're using
-  `type.map_for_each_lhs_of_arrow (@reg_data_to_var) input_registers` says that we're using `reg_data_to_var` (to be customized above) to transform the assignment of registers to input variables (`input_registers`) into the var type we want, and doing this for all of the input arguments at once (`type.map_for_each_lhs_of_arrow`)

src/Assembly/Equivalence.v
JasonGross: Okay, I've added 20 loc to infer the stack size from the assembly code, and to fallback to 40 if there is no `sub rsp, LITERAL`.  (And kept the command line argument which can supersede that.)

JasonGross:
Rebased, now merging

PR #929, opened 2021-04-09
Bump bedrock2 via rupicola

JasonGross:
Let's see if we'll have any issues to solve here in order to integrate
https://github.com/mit-plv/bbv/pull/34, which will be required for
https://github.com/coq/coq/pull/14086.

### DIFF
--- a/.github/workflows/coq.yml
+++ b/.github/workflows/coq.yml
@@ -15,13 +15,13 @@ jobs:
       matrix:
         env:
         - { COQ_VERSION: "8.13.1", COQ_PACKAGE: "coq-8.13.1 libcoq-8.13.1-ocaml-dev", SKIP_BEDROCK2: "" , SKIP_VALIDATE: "" , PPA: "ppa:jgross-h/many-coq-versions-ocaml-4-05" }
-        - { COQ_VERSION: "8.12.2", COQ_PACKAGE: "coq-8.12.2 libcoq-8.12.2-ocaml-dev", SKIP_BEDROCK2: "" , SKIP_VALIDATE: "" , PPA: "ppa:jgross-h/many-coq-versions-ocaml-4-05" }
+        - { COQ_VERSION: "8.12.2", COQ_PACKAGE: "coq-8.12.2 libcoq-8.12.2-ocaml-dev", SKIP_BEDROCK2: "1", SKIP_VALIDATE: "" , PPA: "ppa:jgross-h/many-coq-versions-ocaml-4-05" }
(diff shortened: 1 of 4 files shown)


JasonGross:
Hm, this fails with
```
File "./src/Bedrock/Field/Translation/Proofs/LoadStoreList.v", line 300, characters 6-45:
Error:
Found no subterm matching "Z.land ?M7079 (Z.ones ?M7080)" in the current goal.
```

Any thoughts on what might have gone wrong when updating to the latest bedrock2?  @andres-erbsen @samuelgruetter @jadephilipoom ?

mrhaandi:
FYI
I got the same issue compiling `rupicola/src/Rupicola/Examples/CapitalizeThird/Properties.v` in line `423` (`rewrite Z.land_ones, Z.mod_small; [ reflexivity |  | lia ].`). This one was solvable by `unfold truncate_Z.` first.

JasonGross:
@mrhaandi Thanks!, that indeed solves the issue there.

I've fixed the errors (I hope I've gotten all of them).  There were a couple of changes due to the fact that `Scalars.scalar_of_bytes` seems to now require `32 <= Semantics.width` where it previously didn't.  @andres-erbsen @samuelgruetter was this intentional?

@jadephilipoom Can you take a look at the +8,-5 change of the commit I made and see if the added hypotheses seem reasonable to you?

mrhaandi:
@JasonGross I'm not sure that bedrock2 is entirely up to date.
When I look at your PR and navigate to `rupicola/bedrock2/deps/coqutil/src/coqutil/Z/div_mod_to_equations.v` I arrive at version `afaf...` which says `Lemma mod_0_r_ext x y : y = 0 -> x mod y = 0.`
Now the master branch of `bedrock2` has the correct `ltac:(...)` formulation of `x mod 0`, which is different from the above.
So this PR seems not to be a sufficient bump for https://github.com/coq/coq/pull/14086.

jadephilipoom:
> Can you take a look at the +8,-5 change of the commit I made and see if
the added hypotheses seem reasonable to you?

You shouldn't need these axioms. Semantics.parameters_ok (provided by
Types.ok) guarantees the width is either 32 or 64:
https://github.com/mit-plv/bedrock2/blob/1e44aff4ee5d8b2f054f5db9e04ade4bda562e06/bedrock2/src/bedrock2/Semantics.v#L59

JasonGross:
@jadephilipoom Thanks!  I've updated the proofs.

@mrhaandi I've bumped bedrock2 again to the latest version.

I'll merge once this passes the CI, unless anyone wants to object to my proof changes before then.

PR #934, opened 2021-04-15
run Dalek-crypto backend tests more directly

huitseeker:
- this uses the Dalek upstream repository to run backend tests, rather than the Novi fork used previously.
- this tests the 32-bit backend as well as the 64 bit backend.

/cc @isislovecruft for awareness of extent of backend tests running here.


JasonGross (approved):
Thanks!

PR #941, opened 2021-04-25
Make inversion run in constant time.

dfaranha:
Hi,

I've worked together with Benjamin Hvass (@bshvass) and Bas Spitters (@spitters) in the Fiat-Crypto implementation of divstep-based field inversion. Our C templates had a minor timing issue at the end of algorithm that was later fixed, so I'm pushing a manual fix here as well.

Thank you!


andres-erbsen:
LGTM, thanks!

The Windwos CI error is `          - ocaml-variants.4.10.0+mingw64c: curl error code 403`; I'm overruling it.

Issue #921, opened 2021-03-31
Using Fiat-generated Bedrock2 functions.


RasmusHoldsbjergCSAU:
Lately, I have been trying to use Bedrock2 functions for implementing elliptic curves.

We have worked out a simple example illustrating the functionality of the Bedrock2 framework that is needed for this, and the approach that we intend to follow.

We first synthesize Bedrock2 functions for a 383 prime, m. This is done here:
https://github.com/RasmusHoldsbjergCSAU/fiat-crypto/blob/master/src/Bedrock/Field/Synthesis/Examples/bls12prime.v

The example consists of a simple Bedrock2 function that does the following:
- Allocate 48 bytes of memory on the stack. This is the amount of memory needed to store a representation of one number mod m.
- Store the list [12; 0; 0; 0; 0; 0] into this memory.
- call the "to_montgomery" funct with the list as input

The function itself can be found here:
https://github.com/RasmusHoldsbjergCSAU/fiat-crypto/blob/master/src/Bedrock/Field/Synthesis/Examples/FunctionCallExample.v#L168

We prove that the function meets its specification here:
https://github.com/RasmusHoldsbjergCSAU/fiat-crypto/blob/master/src/Bedrock/Field/Synthesis/Examples/FunctionCallExample.v#L624

Note that a number of tactics are used in this proof.
These tactics mainly serve as glue between different means of addressing memory; namely "anybytes" "Scalar" and the Fiat-specific "Bignum".

Stackallocation uses "anybytes"
Store uses "Scalar"
and the specifications of Fiat-generated functions are phrased in terms of "Bignum".

I was hoping that someone (@andres-erbsen,  @JasonGross ?) could comment on the viability of using this approach. Does it comply with the intended use of Bedrock2/Fiat-crypto or is there an easier/better way of doing this?

A slightly simpler example can be found here:
https://github.com/RasmusHoldsbjergCSAU/fiat-crypto/blob/master/src/Bedrock/Field/Synthesis/Examples/quadratic_extension_bls12.v
This is an implementation of addition and multiplication in the quadratic field extension of F_m.
I say that it is simpler because we do not need to store any constants, as all the data needed is passed to the function as arguments.
Note that code is shared between the files "FunctionCallExample.v" and "quadratic_extension.bls12.v"

A number of technical lemmas are used as well. Ideally, these would be refactored into apriopriate repos, as is done here on my own fork of Fiat-crypto:
[...]

PR #944, opened 2021-04-27
Partial version of new bedrock2 synthesis engine

jadephilipoom:
tl; dr you can now generate (unsaturated solinas only for now) field operations in bedrock2 with proofs that they match the specifications that group operations use (i.e. the specs in `Bedrock/Specs/Field.v`)

The current bedrock2 synthesis backend produces proofs based on `COperationSpecifications`, while the group operations under `Bedrock/Group` expect field operations that match the specs under `Bedrock/Specs`. There are slight differences between the two, and this causes pain -- see https://github.com/mit-plv/fiat-crypto/issues/921 . To fix the issue, I've been working in small bits of free time over several months to re-implement the synthesis pipeline so it produces proofs that match the spec (and also incorporate some lessons learned from the previous attempt). You can find the new pipeline under `Bedrock/Field/Generic/New`, and see an example of generating field operations in the `Tests` section at the end of `Bedrock/Field/Generic/New/UnsaturatedSolinas.v`. My hope is that this new pipeline will eliminate many the aforementioned pain points and allow end-to-end verification of group operations. I think this is likely of interest to @RasmusHoldsbjergCSAU ?

This version is not as feature-complete as the previous one; right now, it has only unsaturated solinas, not word-by-word Montgomery (although many proofs should be reusable). It also doesn't have examples of usage except for the test case at the end of `Bedrock/Field/Synthesis/New/UnsaturatedSolinas.v`. For that reason, and to avoid messing up the ongoing work of others, I have not removed the old pipeline here. I'd ultimately like to replace the old pipeline with the new version. However, because this is not my day job, progress will continue to be a bit glacial...unless anyone wants to volunteer to collaborate!


src/Bedrock/Field/Common/Util.v
```diff
@@ -519,6 +519,37 @@ Section ListZBoundedBy.
              end.
   Qed.
 
+  Lemma list_Z_tighter_than_cons b1 b2 b1s b2s :
+    list_Z_tighter_than (b1 :: b1s) (b2 :: b2s) <->
+    (match b1 with
+     | Some r1 =>
+       match b2 with
```
JasonGross, line 526: Minor: I think this would be slightly easier to read of you did `match b1, b2 with`

src/Bedrock/Field/Common/Util.v
```diff
@@ -582,6 +613,73 @@ Section ListZBoundedBy.
...
+               Z.ltb_to_lt
+           | H : Forall _ (_ :: _) |- _ => inversion H; clear H; subst
+           | H : exists _, _ |- _ => destruct H; subst
+           | H : _ /\ _ |- _ => destruct H; subst
+           end.
+    progress cbn [ZRange.lower ZRange.upper].
+    split; auto with lia.
+  Qed.
+
+  Lemma partition_bounded_by_partition_ones x lg2s nb :
```
JasonGross, line 647: I'm wondering if we should put the bulk of these lemmas somewhere else, outside the bedrock2 directory? (The main issue is that I don't want the basic ZRange stuff depending on arithmetic, and I don't think it makes sense to make the spec stuff depend on fancy ZRange stuff, but I think it's fine, for example, to add a file for reasoning about ZRange lists to the ZRange directory, and then maybe to add a file about boundedness reasoning to the arithmetic directory.). What do you think?

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,599 @@
...
+          {field_parameters : FieldParameters}
+          (varname_gen_is_default :
+             varname_gen = default_varname_gen).
+
+  Context (n : nat) (s : Z) (c : list (Z * Z))
+          (M_eq : M = m s c)
+          (check_args_ok :
+             check_args n s c Semantics.width (ErrorT.Success tt)
+             = ErrorT.Success tt)
+          (* TODO: is this proven elsewhere/can it be proven in general? *)
```
JasonGross, line 83: Where are you pulling tight_bounds and max_bounds from?  I think I probably prove that the tight bounds are tighter than the loose bounds, so you can probably use transitivity for one of these proofs.

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,599 @@
...
+          (scmula24_func_eq : scmula24_func = b2_func scmula24_op)
+          (from_bytes_func_eq : from_bytes_func = b2_func from_bytes_op)
+          (to_bytes_func_eq : to_bytes_func = b2_func to_bytes_op).
+  Existing Instance semantics_ok.
+
+  Local Notation weight :=
+    (ModOps.weight (QArith_base.Qnum (limbwidth n s c))
+                   (Z.pos (QArith_base.Qden (limbwidth n s c))))
+      (only parsing).
+  (* Note: annoyingly, prime_bytes_bounds is an option type, unlike loose_bounds
```
JasonGross, line 107: Why not just change the definition?  I'm all for great consistency.

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,599 @@
...
+         (prime_bytes_upperbound_list s)).
+  Local Instance field_representation : FieldRepresentation
+    := field_representation
+         n (n_bytes s) weight
+         (UnsaturatedSolinasHeuristics.loose_bounds n s c)
+         (UnsaturatedSolinasHeuristics.tight_bounds n s c)
+         (prime_bytes_bounds_value).
+
+  Local Ltac specialize_correctness_hyp Hcorrect :=
+    cbv [feval feval_bytes bounded_by bytes_in_bounds
```
JasonGross, line 120: Somewhat recently developed rule of thumb for proof automation performance: if you issue `cbv` on a list of things, also issue `Strategy` so that `Qed` also knows to unfold those things early/late.  Not important to follow unless you're having trouble with Qed performance or you want to implement this uniformly.

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,599 @@
...
+  Lemma length_byte_bounds :
+    length prime_bytes_bounds_value = encoded_felem_size_in_bytes.
+  Proof.
+    autorewrite with distr_length.
+    apply length_prime_bytes_upperbound_list.
+  Qed.
+
+  Lemma modulus_fits_in_bytes :
+    (0 < m s c <= ModOps.weight 8 1 (n_bytes s))%Z.
+  Proof.
```
JasonGross, line 165: Minor: I'd suggest adding `Proof using` annotations to these proofs (ideally to all of them).  This both speeds up the quick build (maybe we should start testing and timing the quick build?) and also makes the proof more robust to changes in tactics like `lia`.  If you `Set Suggest Proof Using` (I do this in my `~/.coqrc` file), you get suggestions of annotations on Qed of unannotated proofs.

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,599 @@
...
+    (* Extract information from check_args *)
+    clear - check_args_ok. cbv [check_args] in check_args_ok.
```
JasonGross, line 167: Instead of cbv, I think you want to apply use_curve_good:
https://github.com/mit-plv/fiat-crypto/blob/50a022aa8ca6de2693aa7c84884cda844ca65884/src/PushButtonSynthesis/UnsaturatedSolinas.v#L245-L265

JasonGross (commented):
Nice!  Here are some comments, I'll continue reviewing later.

src/Bedrock/Field/Common/Util.v
jadephilipoom: I very much agree! The whole point of this `Util` file was to be a place where I put lemmas that could also make sense in other parts of the repo or in upstream projects. I moved some of them over time, but many of the ones here could be moved as well. It's just a matter of getting the time to do it and knowing exactly where to move things to. Maybe this could be a follow-up PR, though (by me or you or someone else)?

JasonGross: Sure, doing this in a follow-up PR seems fine

JasonGross (commented):
Here are some more comments.  Also, I just had an idea that may allow us to eliminate Wf3 and just use Wf everywhere instead, I'm going to go look into it

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
jadephilipoom: They're from `UnsaturatedSolinasHeuristics`, and yes, I could use transitivity for the `tight_bounds` one. Will do.

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
jadephilipoom: `Qed` performance hasn't been an issue at the moment, but thanks for the tip!

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,599 @@
...
+    apply F.eq_of_Z_iff; rewrite ?F.to_Z_of_Z;
+    cbv [M] in M_eq; rewrite ?M_eq; pull_Zmod.
+
+  Lemma mul_func_correct :
+    valid_func (res mul_op _) ->
+    expr.Wf3 (res mul_op) ->
+    forall functions,
+      spec_of_mul (mul_func :: functions).
+  Proof.
+    intros. cbv [spec_of_mul]. rewrite mul_func_eq.
```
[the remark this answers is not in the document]
jadephilipoom: I did consider this, but I think the previous pipeline suffered from too many tactic parameters that were hard to look up and figure out how to use. I did factor out most of the logic that was genuinely common to all proofs; although similar in structure, the proofs are different in small ways, and I think they're easier to debug with the cases separated a bit.

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,599 @@
...
+  (* solve all the remaining goals *)
+  lazymatch goal with
+  | |- _ = @ErrorT.Success ?ErrT unit tt =>
+    abstract (vm_cast_no_check (@eq_refl _ (@ErrorT.Success ErrT unit tt)))
+  | |- list_Z_tighter_than _ _ =>
+    abstract (vm_cast_no_check (@eq_refl bool true))
+  | |- valid_func _ =>
+    eapply valid_func_bool_iff;
+    abstract vm_cast_no_check (eq_refl true)
+  | |- expr.Wf3 _ => abstract (prove_Wf3 ()) (* this abstract is slow, but if it is removed the slowness moves to Qed instead *)
```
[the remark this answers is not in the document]
JasonGross: Now that I've removed the need for Wf3, this should no longer be needed.

src/Bedrock/Field/Synthesis/Generic/Signature.v
```diff
@@ -0,0 +1,705 @@
...
+                 | progress autounfold with equivalence ].
+  Ltac lists_autounfold :=
+    repeat first [ progress types_autounfold
+                 | progress autounfold with list_lengths
+                 | progress autounfold with lists_reserved ].
+
+  Ltac translate_func_precondition_hammer :=
+    lazymatch goal with
+    | |- valid_func _ => assumption
+    | |- Wf.Compilers.expr.Wf3 _ => assumption
```
JasonGross, line 219: Note (possibly to self): this should be changed when we rebase on master.  Which lemmas does it connect the Wf conditions on?

src/Bedrock/Field/Synthesis/Generic/Signature.v
```diff
@@ -0,0 +1,705 @@
...
+    | |- @eq (list word.rep) _ _ => reflexivity
```
JasonGross, line 220: In case you're not aware and like notations better, here's another way to write this line (feel free to take it or leave it as is your preference)
```suggestion
    | |- _ = _ :> list word.rep => reflexivity
```

src/Bedrock/Field/Synthesis/Generic/Signature.v
```diff
@@ -0,0 +1,705 @@
...
+              rewrite <-H end.
+            auto. } } }
+    Qed.
+  End ListUnop.
+
+  Section FromBytes.
+    Context {res : API.Expr (type_listZ -> type_listZ)}
+            (res_valid :
+               valid_func (res (fun _ : API.type => unit)))
+            (res_Wf3 : Wf.Compilers.expr.Wf3 res).
```
JasonGross, line 502: I think this sticks around and becomes Wf, though I'm not sure.  At this point, we no longer have that the expr came from the bounds pipeline, right?

src/Bedrock/Field/Synthesis/Generic/New/ComputedOp.v
```diff
@@ -0,0 +1,29 @@
...
+
+Record computed_op
+      {p : Types.parameters} {t}
+      {op : Pipeline.ErrorT (API.Expr t)}
+      {name : String.string}
+      {insizes outsizes inlengths}
+  :=
+  { res : API.Expr t;
+    b2_func : func;
+    res_eq : op = ErrorT.Success res;
```
JasonGross, line 16: What's the reason for removing the Wf field from this record?  (I'm trying to understand how best to adapt the PR to the changes I made where rather than reproving Wf3, we instead use the fact that everything that comes out of the synthesis pipeline is Wf)

src/Bedrock/Field/Synthesis/Generic/Signature.v
```diff
@@ -0,0 +1,705 @@
...
+              rewrite <-H end.
+            auto. } } }
+    Qed.
+  End FromBytes.
+
+  Section ToBytes.
+    Context {res : API.Expr (type_listZ -> type_listZ)}
+            (res_valid :
+               valid_func (res (fun _ : API.type => unit)))
+            (res_Wf3 : Wf.Compilers.expr.Wf3 res).
```
JasonGross, line 591: This too becomes Wf I think.  Btw, I've made `API.Wf` so we no longer need `Compilers` here.  In fact, if the are any uses of `Compilers.*` in bedrock2 code, that's probably a sign that I haven't made enough API aliases.  Are you aware of any other uses of `Compilers.*`?

src/Bedrock/Field/Synthesis/Generic/Signature.v
```diff
@@ -0,0 +1,705 @@
...
+    Context (byte_bounds_tighter_than_max :
```
JasonGross, line 592: This seems a bit misnamed, since you're saying that one encoding of byte bounds is tighter than another.  (Why do we have two non-judgmentally-equal encodings btw?  I spent some effort trying to make all the bounds encodings line up in judgmentally nice ways across their uses, so if there's a different syntactic form that works better for you, I'd be interested in knowing it and seeing if we can line them all up syntactically again.)

src/Bedrock/Field/Synthesis/Generic/Signature.v
```diff
@@ -0,0 +1,705 @@
...
+                       (API.interp (res _) (map word.unsigned x)))
+                = feval x)
+            (res_bounds : forall x,
+                bounded_by tight_bounds x ->
+                list_Z_bounded_by
+                  byte_bounds
+                  (API.interp (res _)
+                              (map word.unsigned x))).
+
+    Ltac equivalence_side_conditions_hook ::=
```
JasonGross, line 610: Does this assignment persist past the section?  If it's not supposed to, I'd suggest prefixing it with `Local` to make the intent clear, assuming Coq accepts that.

src/Bedrock/Field/Synthesis/Generic/Signature.v
```diff
@@ -0,0 +1,705 @@
...
+    - focus on unsat solinas first, make mul/add defs
+    - take mul/add defs through full synthesis/translation to get x25519 impls
+    - compose with montladder as demo to test
+    === merge ===
+    - generate other bin/unops
+    === merge ===
+    - make signature proofs for other signatures (select, from_bytes)
+    - generate all operations
+    === merge ===
+    - figure out what to do for copy and small_literal; general template that spits out the correct bedrock2?
```
JasonGross, line 697: Are these field operations?  We should be able to generate these through the bounds pipeline.  Or is the question how to translate the synthesized expr to bedrock2?

JasonGross (commented):
Nice work!  I've at least skimmed through all of it.

> The current bedrock2 synthesis backend produces proofs based on `COperationSpecifications`, while the group operations under `Bedrock/Group` expect field operations that match the specs under `Bedrock/Specs`. There are slight differences between the two, and this causes pain -- see https://github.com/mit-plv/fiat-crypto/issues/921 .

I'd like to understand the slight differences here; I'm wondering if it makes sense to try to minimize them by changing the specifications rather than by writing glue proofs.  I wasn't able to extract insight from #921; could you say what the relevant differences are?
Btw, the design constraints on `COperationSpecifications` are:
1. The form of the bounds and interpretation conditions should be uniformly definable without knowing the type of the operation being specified; this is so that we can line it up with the correctness condition on the bounds pipeline
2. It should be feasible to automatically transcribe the statements into documentation strings describing the generated code
3. The specifications should be complete in the sense of fully pinning down the behavior we want from the interpreted function
4. The specifications should be usable by consumers of the synthesis pipeline (currently only bedrock2)

> and also incorporate some lessons learned from the previous attempt

I'd be curious to know what these lessons are, if you could articulate them.

I'd also be curious to have a high-level overview of the code layout.  Additionally, I imagine a lot of the code falls into one of the following categories, and I'd be interested in knowing which code falls into which category, what other categories there are, and what the distribution between categories is.
- operation-independent glue lemmas tying PHOAS/PushButtonSynthesis to bedrock2/rupicola
- operation-independent glue tactics that are run again for each operation (presumably because the lemmas were too hard to state in general form?)
- operation-specific glue proofs
- interface specifications

jadephilipoom:
I was also hoping to address some of Jason's comments before merging.
Should be able to do it tonight after work.

JasonGross:
@jadephilipoom Do you mind if I force-push a rebased version directly to your branch and then push a commit that updates to the Wf3-less version there as well?

src/Bedrock/Field/Synthesis/Generic/Signature.v
```diff
@@ -0,0 +1,705 @@
...
+    Context (xbounds ybounds outbounds : bounds)
+            (op : F M_pos -> F M_pos -> F M_pos)
+            (outbounds_tighter_than_max :
+               list_Z_tighter_than outbounds (MaxBounds.max_bounds n))
+            (outbounds_length : length outbounds = n)
+            (res_eq : forall x y,
+                bounded_by xbounds x ->
+                bounded_by ybounds y ->
+                feval (map word.of_Z
+                           (API.interp (res _)
```
JasonGross, line 331: Why not `API.Interp res` instead?  (`API.Interp x` is the same as `API.interp (x _)`)

jadephilipoom:
> I've pushed to [`mit-plv:field-spec-connect2-rebased`](https://github.com/mit-plv/fiat-crypto/tree/field-spec-connect2-rebased), let me know if you want me to push directly to your branch.

Go ahead

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
jadephilipoom: I think I tried and it broke too many things and I gave up. `prime_bytes_bounds` is used in a lot of places.

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
jadephilipoom: It won't work for this one, I think. That's not `weight (Qnum limbwidth) (QDen limbwidth) n`, it's `weight 8 1 n_bytes` -- the proof that the modulus is smaller than the number of bits in the _byte_ representation. I believe it's not covered by `use_curve_good`.

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,599 @@
...
+  Derive fe25519_to_bytes
+         SuchThat (forall functions,
+                      spec_of_to_bytes
+                        (field_representation:=field_representation n s c)
+                        (fe25519_to_bytes :: functions))
+         As fe25519_to_bytes_correct.
+  Proof. Time derive_bedrock2_func to_bytes_op. Qed.
+End Tests.
+
+Print fe25519_mul.
```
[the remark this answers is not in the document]
jadephilipoom: Huh, thought I commented them out. Will fix.

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
```diff
@@ -0,0 +1,599 @@
...
+      erewrite map_byte_wrap_bounded
+        by eauto with bounds
+    end.
+  Ltac FtoZ :=
+    apply F.eq_of_Z_iff; rewrite ?F.to_Z_of_Z;
+    cbv [M] in M_eq; rewrite ?M_eq; pull_Zmod.
+
+  Lemma mul_func_correct :
+    valid_func (res mul_op _) ->
+    expr.Wf3 (res mul_op) ->
```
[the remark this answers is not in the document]
jadephilipoom: Yes, I do have that information.

src/Bedrock/Field/Synthesis/Generic/Signature.v
```diff
@@ -0,0 +1,705 @@
...
+Require Import Crypto.Bedrock.Field.Common.Tactics.
+Require Import Crypto.Bedrock.Field.Common.Types.
+Require Import Crypto.Bedrock.Field.Translation.Func.
+Require Import Crypto.Bedrock.Field.Translation.Proofs.Func.
+Require Import Crypto.Bedrock.Field.Interface.Representation.
+Require Import Crypto.Bedrock.Specs.Field.
+Require Import Crypto.COperationSpecifications.
+Require Import Crypto.Language.API.
+Require Import Crypto.Spec.ModularArithmetic.
+(*Import ZRange.Compilers.*)
```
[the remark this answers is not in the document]
jadephilipoom: Nope, will remove.

src/Bedrock/Field/Synthesis/Generic/Signature.v
jadephilipoom: `translate_func_correct`; just checked and on master it's now a `Wf` precondition (sorry, you probably already figured this out)

src/Bedrock/Field/Synthesis/Generic/Signature.v
jadephilipoom: I believe that is correct.

src/Bedrock/Field/Synthesis/Generic/Signature.v
jadephilipoom: Mostly it's `Wf3`, but grep also turned up a couple uses of `AbstractInterpretation.Compilers` and several uses of `*.Compilers.*` in `Stringification` (not sure if that matters).

src/Bedrock/Field/Synthesis/Generic/Signature.v
jadephilipoom: Whoops, these were notes-to-self and are quite outdated, I'll remove these before merging. Good thing I didn't say anything too silly there.

I think copy and small_literal are small and obvious enough to just write in bedrock2 directly. There's some pseudocode here: https://github.com/mit-plv/fiat-crypto/blob/bdbf2a55484ce47cdb1a0d5068d50f7e18c6517b/src/Bedrock/Specs/Field.v#L18

I think Dustin might have been working on this at some point? I recommended it as a possible learn-bedrock2 exercise some time ago.

src/Bedrock/Field/Synthesis/Generic/Signature.v
jadephilipoom: Probably either no good reason, or a very stupid quirk. I'll investigate.

src/Bedrock/Field/Synthesis/Generic/Signature.v
jadephilipoom: They're slightly different; `byte_bounds` in this context is the bounds for a serialized field element, while `ByteBounds.byte_bounds` is the maximum bounds for bytes (i.e. `0~>2^8-1` repeated n times). See:

https://github.com/mit-plv/fiat-crypto/blob/bdbf2a55484ce47cdb1a0d5068d50f7e18c6517b/src/Bedrock/Field/Common/Arrays/ByteBounds.v#L26-L29

The naming is not ideal, I agree. I could change `byte_bounds` to `encoded_felem_bounds`, or `ByteBounds.byte_bounds` to `max_byte_bounds`?

jadephilipoom:
> > The current bedrock2 synthesis backend produces proofs based on `COperationSpecifications`, while the group operations under `Bedrock/Group` expect field operations that match the specs under `Bedrock/Specs`. There are slight differences between the two, and this causes pain -- see #921 .
> 
> I'd like to understand the slight differences here; I'm wondering if it makes sense to try to minimize them by changing the specifications rather than by writing glue proofs. I wasn't able to extract insight from #921; could you say what the relevant differences are?

A few things. First, the `COperationSpecifications` conditions are fundamentally incompatible with bedrock2's postcondition-style semantics, because they are defined over a pure-Gallina implementation of the function, for example:

https://github.com/mit-plv/fiat-crypto/blob/bdbf2a55484ce47cdb1a0d5068d50f7e18c6517b/src/COperationSpecifications.v#L242-L247

In bedrock2, it's not possible to get hold of an interpretation of the implementation that is `list Z -> list Z`; you can only make statements about the state of your memory, trace, and local environment after the function is run.

Second, the bedrock2 functions need to know the exact memory layout, and they need pointers for output, so they have some extra pre- and post-conditions. Here's the bedrock2 version of the `from_bytes` spec:
```coq
  Definition spec_of_from_bytes : spec_of from_bytes :=
    forall! (bs : list byte) (px pout : word),
      (fun Rr mem =>
         (exists Ra, (FElemBytes px bs * Ra)%sep mem)
         /\ bytes_in_bounds bs
         /\ (Placeholder pout * Rr)%sep mem)
        ===> from_bytes @ [pout; px] ===>
        (fun _ =>
           liftexists X,
           (emp (feval X = feval_bytes bs /\ bounded_by tight_bounds X)
            * FElem pout X)%sep).
```

I'm not willing to go and change `COperationSpecifications` -- I tried it maybe  a year or so ago and wasn't successful. Possibly didn't know what I was doing. But my design constraint for the bedrock2 specs is that I need them to be as simple as possible and easy to reason about for the group code; I'm concerned that generating them from `COperationSpecifications` will result in an autogenerated mess that hinders higher-level proofs. I'd rather absorb the glue-code pain at this level of abstraction than pass it on to the next.

> > and also incorporate some lessons learned from the previous attempt
> 
> I'd be curious to know what these lessons are, if you could articulate them.

- Lots of code was repeated in the old version for functions with the same signature; this version of the pipeline has general lemmas for, for instance, binary and unary operations on field elements, which can be reused.
- The old version tried to make it easy to write one definition and get the full suite of field arithmetic for a particular curve. This version abandons that approach, because the one-shot tactic was slow and sometimes you only really care about a few operations and don't want to synthesize them all.
- I've steered away from one-size-fits-all one-line tactics this time, because they're a pain to debug, and instead tried to automate shared _pieces_ of proofs that are conceptually well-defined. In general, I've tried to steer towards a "less magic" approach.

> I'd also be curious to have a high-level overview of the code layout. Additionally, I imagine a lot of the code falls into one of the following categories, and I'd be interested in knowing which code falls into which category, what other categories there are, and what the distribution between categories is.

Maybe I should add a README or something. Do you mean the layout of this new pipeline or the bedrock2 backend in general? For this pipeline:
- Interface specifications for coordinate-field operations, group operations, and scalar field operations are all in `Bedrock/Specs`. These are parameterized over typeclasses that define, for instance, the exact names of functions and the memory layout of field/group elements. Instances of those typeclasses (still generic over the exact curve, but containing information like "field elements are arrays") are under `Bedrock/{Field, ScalarField}/Interface/Representation.v` (the group-level one is hackily shoved into `Bedrock/Group/ScalarMult/ScalarMult.v`, I was almost organized there)
- `Bedrock/Field/Synthesis/Generic/Signature.v` has what you call "operation-independent glue lemmas", I think -- it has lemmas for each function _signature_ e.g. `felem -> felem -> felem` (binop) or `list byte -> felem` (from_bytes) that link them to the specs in `Bedrock/Specs/Field.v`
- `Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v` has the proofs that link each unsaturated-solinas operation to its spec from `Bedrock/Specs/Field` using the `Signature` lemmas -- parameterized over the modulus, number of limbs, etc, but now we're talking about `mul` and `add`, not `binop`
- `Bedrock/Field/Synthesis/Generic/New/ComputedOp.v` has a record that bundles a computed `API.Expr`, a call to the bounds pipeline, and a bedrock2 function, along with proofs that say the computed `API.Expr` came out of the call to the bounds pipeline and that the bedrock2 function came from translating the computed `API.Expr`. This makes those pieces of information easier to handle in the proofs than, for instance, posing the call to the bounds pipeline and computing it and then later needing to get the information that the computed output is related to the original call.
- The `make_computed_op` tactic constructs the bundle described above, and needs to be run for each fully-specialized operation. The `derive_bedrock2_func` tactic solves the preconditions left over from the `UnsaturatedSolinas` level of abstraction, mostly preconditions on the algorithm parameters, and also needs to be run for each fully-specialized operation.
- The new pipeline, once complete, should replace everything currently under `Bedrock/Field/Synthesis/Specialized` and most of what's under `Bedrock/Field/Synthesis/Generic` (except for `Signature.v` and `New/`, because in my great wisdom I decided the best place for a new pipeline was smack in the middle of the old pipeline).

Does this answer the question?

jadephilipoom:
> > I've pushed to [`mit-plv:field-spec-connect2-rebased`](https://github.com/mit-plv/fiat-crypto/tree/field-spec-connect2-rebased), let me know if you want me to push directly to your branch.
> 
> Go ahead

Actually never mind, I'll just do it

andres-erbsen:
https://github.com/mit-plv/fiat-crypto/blob/bdbf2a55484ce47cdb1a0d5068d50f7e18c6517b/src/COperationSpecifications.v#L242-L247
This looks like it really should have been quantified over the output of the function, not the function itself. But let's not hold up this PR on that; I'll probably poke at the specifications when I start putting things together anyway.

jadephilipoom:
I didn't account for rebuild time; I won't be able to finish the code review comments tonight. If it's urgent feel free to merge as-is -- either way I'll get to it as soon as I can, in the next few days.

andres-erbsen:
No rush, I will be looking this late next week the earliest.

JasonGross:
Btw, the Mac build is expected to be broken and I'm working on it, this is https://github.com/coq/coq/issues/14228

RasmusHoldsbjergCSAU:
I have had a look at the new pipeline, it looks interesting.

The discussion in #921 emerged as a consequence of me trying to use Bedrock2-functions for implementing elliptic curve group operations. My initial approach left me with a function that suffered from 2 issues:

The first one was that pointers to intermediate variables had to be passed to the function as arguments. This was solved by using stack allocation to allocate memory for intermediate variables. Some glue needed to be written for stack allocation to work with Bignum.

The other one was more tricky. The preconditions of the functions required that arguments were separated in memory.
This was not ideal, as we would like to prove that the function is correct both in the case were inputs are separated, and in the case were the same pointer is passed as multiple arguments (this is particularly relevant in e.g. functions implementing extended field arithmetic, as these will often be used with overlapping inputs/outputs).

A Bedrock2 function implementing multiplication in a quadratic field extension can be found here:
https://github.com/RasmusHoldsbjergCSAU/fiat-crypto/blob/master/src/Bedrock/Field/Synthesis/Examples/quadratic_extension_bls12.v#L1175
Its specification is just below it in the same file. The proof is still a horrible mess at this point, so I would not recommend that you try and read it, but if you have an idea of how using the Bedrock2-functions from the new pipeline might make things easier, I would be happy to know your thoughts : )

On another note, at some point in the future I would like to have Bedrock2-functions generated from these:
https://github.com/RasmusHoldsbjergCSAU/QuadraticFieldExtensions/blob/master/Implementations/SOS/SOSMul.v
https://github.com/RasmusHoldsbjergCSAU/QuadraticFieldExtensions/blob/master/Implementations/SOS/SOSReduction.v

An alternative word-by-word Montgomery multiplication procedure, that separates the multiplication and reduction steps; how would one go about integrating these into the pipeline?

jadephilipoom:
> My initial approach left me with a function that suffered from 2 issues:

I believe both issues are now fixed:

> The first one was that pointers to intermediate variables had to be passed to the function as arguments. This was solved by using stack allocation to allocate memory for intermediate variables. Some glue needed to be written for stack allocation to work with Bignum.

I'm surprised by that; when did you try it? This example from the current master branch shows a bedrock2 function with stack allocation, plus proof:

https://github.com/mit-plv/fiat-crypto/blob/43475a54bd011b8910984c403443b5fd27824f70/src/Bedrock/Group/ScalarMult/ScalarMult.v#L120

You should be able to use the same strategy for any bedrock2 function. A few of the tactics there should be upstreamed to bedrock2 or at least moved somewhere more general.

> The other one was more tricky. The preconditions of the functions required that arguments were separated in memory.

The new version does not require input to be separated in memory:

https://github.com/jadephilipoom/fiat-crypto/blob/16ad19823074762af4528f34744261be311551cf/src/Bedrock/Specs/Field.v#L100-101

> A Bedrock2 function implementing multiplication in a quadratic field extension can be found here:
> https://github.com/RasmusHoldsbjergCSAU/fiat-crypto/blob/master/src/Bedrock/Field/Synthesis/Examples/quadratic_extension_bls12.v#L1175
> Its specification is just below it in the same file. The proof is still a horrible mess at this point, so I would not recommend that you try and read it, but if you have an idea of how using the Bedrock2-functions from the new pipeline might make things easier, I would be happy to know your thoughts : )

You should now be able to define the algorithm in generic form (i.e. not specialized to a specific curve) by requiring `FieldParameters` and `FieldRepresentation` instances (plus their `_ok` instances) as context variables. The `scalarmult` example I linked above uses this style. Then you can call field operations and also use `felem_copy` instead of manually copying field elements. (Currently `felem_copy` is not generated in the pipeline, but you can write one for your curve and prove it matches the spec, which is likely easier than proving it inside a bigger proof).

> On another note, at some point in the future I would like to have Bedrock2-functions generated from these:
> https://github.com/RasmusHoldsbjergCSAU/QuadraticFieldExtensions/blob/master/Implementations/SOS/SOSMul.v
> https://github.com/RasmusHoldsbjergCSAU/QuadraticFieldExtensions/blob/master/Implementations/SOS/SOSReduction.v
> 
> An alternative word-by-word Montgomery multiplication procedure, that separates the multiplication and reduction steps; how would one go about integrating these into the pipeline?

1. Go to `src/Bedrock/Specs/Field.v` and add fields to `FieldParameters` for the names of these functions (the new pipeline doesn't require you to generate all operations, so it would be fine for someone to choose to generate either mul-and-reduce or mul and reduce separately). This would be as simple as `mul_noreduce : string`, `reduce : string`.
2. Add specifications for the functions. See, for instance, the spec of `from_bytes` here: 
https://github.com/mit-plv/fiat-crypto/blob/43475a54bd011b8910984c403443b5fd27824f70/src/Bedrock/Specs/Field.v#L126-L135
Note that you probably can't use `binop_spec` and `unop_spec` for this one, because there's not an exactly analogous field operation. It's also not revealed at this level of abstraction that field elements are arrays, so to produce a double-size result from `mul` the easiest way is probably to have it return two field elements (one for the low half, one for the high half) and have `reduce` take in two field elements and return one.
3. Since these functions wouldn't use `binop_spec` and `unop_spec`, you'd need new lemmas for them in `src/Bedrock/Field/Synthesis/Generic/Signature.v`. You can largely follow the pattern from the other lemmas there, I imagine.
4. Finally, you'd need to create your own version of `src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v` that covers the operations you want. Again, you should be able to largely follow the pattern from the other lemmas.
5. The `Tests` section at the bottom of `UnsaturatedSolinas.v` generates several field operations for a specific curve; you should be able to do exactly the same thing for your new operations.

RasmusHoldsbjergCSAU:
Thanks for the response; using FieldParameters, FieldRepresentation and felem_copy will make things more streamlined : )

With respect to preconditions requiring inputs to be separated, the functions/specs generated by the old pipeline also did not require inputs to be separated. I meant to say that the functions I have written myself, and which make use of calls to functions generated by the pipeline, had this issue, and that it takes more work proving the general case. This is not really an issue with the pipeline or the generated function, but with the Bedrock2 framework, and is what is being discussed here #921. 

As far as I can tell the ladderstep implementation does require separated inputs:
https://github.com/jadephilipoom/fiat-crypto/blob/16ad19823074762af4528f34744261be311551cf/src/Bedrock/Group/ScalarMult/LadderStep.v#L79

With respect to integrating new functions into the pipeline; it looks doable; I will have a closer look when I have more time : )

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
jadephilipoom: There's no transitivity lemma for `list_Z_bounded_by` as far as I can tell. In the interests of merging the PR sooner to unblock other things, I'm leaving a TODO there instead.

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
JasonGross: `use_curve_good` is supposed to cover everything from `check_args`.  If there's some check that happens that's not present, we should just add that check to `use_curve_good`.

src/Bedrock/Field/Synthesis/Generic/Signature.v
jadephilipoom: I did investigate, and although they might be equivalent changing it is not straightforward enough for me to want to spend the time on it.

JasonGross: Ah, okay, there must be something subtle going on here with Ltac syntax matching or `rewrite`s or unification.  `API.Interp x` literally just unfolds to `API.interp (x _)`

jadephilipoom:
I rebased on the latest master (95ed951a989103bca3ec205aa48675f43fdd0207) and did my best to address the review comments. I also moved all of the files to do with the new pipeline under `Bedrock/Field/Synthesis/New` so they're not mixed in with the older version. Let me know if there's anything else I should address.

src/Bedrock/Field/Synthesis/Generic/Signature.v
jadephilipoom: Not even necessarily subtle; since it's just cosmetic, I didn't spend that much time trying to fix it. There's a mismatch between `expr.Interp` and `expr.interp` inside of several proofs when I just make the text-replace change and try to step through. Probably not too hard to fix given a little time.

src/Bedrock/Field/Synthesis/Generic/New/UnsaturatedSolinas.v
jadephilipoom: The property isn't in `check_args` directly, it's just inferrable from what's there.

JasonGross: Is there reason to directly unfold `check_args` and do the boolean reasoning in this function rather than to do the inference from the result of `use_curve_good`, which will have already done a bunch of the boolean processing?

JasonGross:
> Let me know if there's anything else I should address.

I'm happy to consider this good enough to merge if the CI passes, and fixup anything else later (I've far exceeded my budget of how much time I should be spending on fiat-crypto lately).

PR #945, opened 2021-04-27
Add another field inversion example

jedisct1:
Fiat-crypto cannot generate the full field inversion code, so having reference code or examples for the supported backends may be useful.

Add a Zig example for it, and hook it to CI by the way.


JasonGross:
The Makefile needs to be updated with the renamed files so the CI doesn't break:
https://github.com/mit-plv/fiat-crypto/blob/50a022aa8ca6de2693aa7c84884cda844ca65884/Makefile#L215

JasonGross (approved):

Issue #939, opened 2021-04-21
Z_static_cast should record whether it's a cast up or cast down


JasonGross:
> Just one thing: there is a single instruction for changing the size of a value (`Z_static_cast`), no matter if it extends or reduces the size.

Ah, I should be able to add an extra argument to `Z_static_cast` that says whether we're increasing the size or decreasing it.  I'll open an issue to remind myself to do it.
 
_Originally posted by @JasonGross in https://github.com/mit-plv/fiat-crypto/issues/936#issuecomment-823691564_

JasonGross:
In fact it would be nice to distinguish three different cases: (lossless) widen, lossless narrow, and truncating narrow.

jedisct1:
Hi Jason,

Do you think this is doable?

The generated Zig code currently works around this, but that workaround has a significant impact on compilation time.

JasonGross:
It's doable, but a bit non-trivial.  I don't have the time to work on it right now, but if you wanted to take a stab at it, the thing to do would be to add
```coq
Variant cast_kind := annotation | truncating.
```
Then update
https://github.com/mit-plv/fiat-crypto/blob/94919dcd99a03ba433204278f9e7145a3e781b24/src/Language/PreExtra.v#L34-L43
to take in such `cast_kind` arguments, add `cast_kind` to
https://github.com/mit-plv/fiat-crypto/blob/94919dcd99a03ba433204278f9e7145a3e781b24/src/Language/IdentifierParameters.v#L45-L50
Then, for each rewrite rule involving a cast, you need to either make it parametric over the kind of cast, or else go from any kind of cast to a truncating cast (for the few rules that do truncate).  There's no help from the typechecker here, so this part needs to be done carefully.  The code is at https://github.com/mit-plv/fiat-crypto/blob/94919dcd99a03ba433204278f9e7145a3e781b24/src/Rewriter/Rules.v#L301-L1179
The corresponding tactics and proofs in https://github.com/mit-plv/fiat-crypto/blob/master/src/Rewriter/RulesProofs.v need to be updated, as well as the lemmas in https://github.com/mit-plv/fiat-crypto/blob/master/src/CastLemmas.v.
Then you need to update https://github.com/mit-plv/fiat-crypto/blob/94919dcd99a03ba433204278f9e7145a3e781b24/src/AbstractInterpretation/AbstractInterpretation.v#L392-L401 to add casts in annotation mode.
[...]

PR #959, opened 2021-05-01
Zig Fiat-Crypto-generated code is now continuously tested

jedisct1:
The first bits of Fiat-generated code have landed into the Zig standard library.

The code is now continuously tested on all supported platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ Go | :white_check_mark: @mdempsky | :heavy_check_mark: | |
 Java | :x: Unmaintained | :heavy_check_mark: | :x: Known Buggy [#707](https://github.com/mit-plv/fiat-crypto/issues/707#issuecomment-763098543) | 
 JSON | Experimental | :white_check_mark: (only syntax) | | 
 Rust | :white_check_mark: | :heavy_check_mark: | :heavy_check_mark: ([Dalek Cryptography test-suite](https://github.com/dalek-cryptography/curve25519-dalek)) | 
-Zig | :white_check_mark: @jedisct1 | :heavy_check_mark: | | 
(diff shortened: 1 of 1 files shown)


JasonGross:
That's great to hear!

I'd like to reserve the heavy checkmark for tests that will be applied to pull requests before they are merged, i.e., backends where we expect that introducing, e.g., a transcription error changing `+` to `*` will result in our CI failing.

Is there a way to add a test to our CI that runs our code against the Zig crypto test-suite?

jedisct1:
Oh, absolutely, I can add that!

jedisct1:
Meanwhile, the check mark has been replaced by the white one.

JasonGross:
Great, thanks!  Could I trouble you too provide links to Zig's CI (if it's public) and to the part of the standard library that includes fiat-crypto (I couldn't find it with a brief Google search)

jedisct1:
Zig uses multiple CI systems, but this is the [main one](https://dev.azure.com/ziglang/zig/_build?definitionId=1&_a=summary&repositoryFilter=1&branchFilter=5%2C5%2C5%2C5%2C5%2C5%2C5%2C5), used to test Linux, macOS and Windows builds.

The first bits using Fiat-crypto are [here](https://github.com/ziglang/zig/tree/master/lib/std/crypto/pcurves) (generated file [here](https://github.com/ziglang/zig/blob/master/lib/std/crypto/pcurves/p256/p256_64.zig)) and other curves will follow quickly.

jedisct1:
The generated code is now directly tested in fiat-crypto's CI:

https://github.com/mit-plv/fiat-crypto/blob/master/fiat-zig/src/main.zig

in addition to the inversion code, that is also part of what is run by CI.

These tests are run in debug mode, which checks for arithmetic overflow. And will also check for unexpected truncation with help of #939

Issue #960, opened 2021-05-01
Curve25519 scalar field reduction


jadephilipoom:
More of a discussion than an issue. I was talking with @FiloSottile at HACS about the possibility of generating code for curve25519's scalar field using fiat-crypto. My first answer was "I think it uses Barrett reduction, and we don't have the mainstream pipeline set up for it", based on the `ref` implementation [0]. But further discussion about the `ref10` implementation sounded like pseudomersenne/unsaturated solinas reduction to me, so I said I would take a closer look. Here's what I found:

It *does* look like unsaturated solinas reduction, but a variant I've never seen before. In particular, the usual algorithm for unsaturated solinas is to take a modulus in the form `2^k-c`, split the result of multiplication (usually a no-op because of a cleverly chosen representation) at the kth bit so you have `(a + 2^kb) mod (2^k-c)`, and, since this is the same as `(a + cb) mod (2^k-c)`, just multiply the high half by `c` and add it to `a`, then repeat the process. The ref10 implementation [1] does essentially this procedure, except that `c` is positive (the modulus is `l = 2^252 + 27742317777372353535851937790883648493`). These lines [2], repeated over and over again, subtract 27742317777372353535851937790883648493 from the first half of the number. Essentially they're doing `(a + 2^kb) mod (2^k+c) = (a - cb) mod (2^k+c)` -- note the change in sign. I think this is using signed integers; otherwise I can't see how a line like [3] wouldn't underflow (The `crypto_int64` type is defined at [4], and I do believe this means signed).

Overall, I don't think we can support this exact algorithm right now without additional engineering, but it's still interesting to know about.

[0] https://github.com/floodyberry/supercop/blob/a351f2c29235512a042d4b6989d241a67e86ad23/crypto_sign/ed25519/ref/sc25519.c#L41
[1] https://github.com/floodyberry/supercop/blob/a351f2c29235512a042d4b6989d241a67e86ad23/crypto_sign/ed25519/ref10/sc_reduce.c#L35
[2] https://github.com/floodyberry/supercop/blob/a351f2c29235512a042d4b6989d241a67e86ad23/crypto_sign/ed25519/ref10/sc_reduce.c#L140-L146
[3] https://github.com/floodyberry/supercop/blob/a351f2c29235512a042d4b6989d241a67e86ad23/crypto_sign/ed25519/ref10/sc_reduce.c#L84
[...]

PR #961, opened 2021-05-01
Use a more flexible encoding for tight upperbounds

JasonGross:
We encode `2^Z.log2_up(s-c)-1` rather than `s-1` for cases when `c` is
negative.  We can't actually handle `c < 0` due to the preconditions for
`freeze`, but this brings us a step closer.

See #960


JasonGross:
The next step after this PR, btw, is to adjust `check_args` so that it knows which functions have been requested and will only check conditions for the functions that have been requested.  I'll aim to do this later today.

PR #963, opened 2021-05-01
Comsetic changes to the example Zig output files

jedisct1:
Multiple updates to the Zig backend
    
* Cosmetic:
      - Use // instead of /// for the header comments
      - Remove function prefixes - `const fiat = @import("...")` is a  more idiomatic way to create a namespace, and makes it easier to use multiple implementations of the same function.
      - Remove unneeded types for constant assignments.
    
* Security:
      - Disable checked arithmetic except in debug mode in order to avoid side channels.
    
* Tests:
      - Run the generated code against test vectors to check for regressions and inconsistencies between targets.


JasonGross (approved):
Will merge once CI passes

PR #966, opened 2021-05-03
 Add a CI test of our opam package

### LINKED ISSUE #964: Use setup-ocaml GH Action
We should stop building our own version of cygwin/OCaml/coq on Windows and Mac, and just use https://github.com/avsm/setup-ocaml to get opam and use opam to install Coq

JasonGross:
Closes #964


JasonGross:
Todo: fix return code of -h

PR #968, opened 2021-05-03
Add `ShowLevel`, rip out `with_parens` from `Show`

JasonGross:
Now `show` and `show_lines` no longer take a `with_parens` argument, and
instead that functionality is generalized into `ShowLevel` / `show_lvl`.
This is the first step towards having more flexibility in parenthesizing
printed expressions, in a more principled way.  It's sort-of a
test-drive of a more uniform way of level-based printing.


JasonGross:
Rebased, previously passed CI so merging now

PR #969, opened 2021-05-04
Switch to latest version of Haskell and Cabal

JasonGross:
Mainly to silence warnings/annotations.

Also switch from an unmaintained action to a maintained one.


JasonGross:
This will not happen for a while, due to https://github.com/coq/coq/issues/14256

Issue #949, opened 2021-04-28
Go code cosmetic fixes


FiloSottile:
- [ ] use `uint64` instead of `uint1` and drop addcarryxU64 and subborrowxU64 wrappers

mdempsky:
I think these would be nice improvements if they're easy and don't add complexity to fiat-crypto's output routines (which aren't verified), but I don't think any of these affect the assembly ultimately generated by the Go compilers?

armfazh:
@JasonGross  definitely types for differentiating between Montgomery domain. +1

mdempsky:
FWIW, the syntax rules for Go documentation are described at: https://golang.org/pkg/go/doc/#ToHTML

I think the main noteworthy things for fiat-go code are that blank lines separate paragraphs, and runs of indented lines get formatted with \<pre\>.

JasonGross:
> formatted with <pre>.



> * run `gofmt` on the output (which will only fix some of the whitespace

This one is a bit tricky logistically, I'd rather not make `gofmt` a requirement for synthesizing the go files via the Makefile.  I can change the output to match the spacing conventions of `gofmt`, though?

mdempsky:
> Which indented lines become `<pre>` blocks?

It looks like lines 3 and 4, 6 and 7, and 9 and 10 become pre blocks:

<details>
<summary>testing with `go doc`</summary>

```
mdempsky@mdempsky:/tmp/test$ ls
test.go
mdempsky@mdempsky:/tmp/test$ cat test.go
/*
line1
continued

line2
continued
  line3
  line4
continued

line5
continued

  line6
  line7
line8
continued

  line9
  line10
*/
package p
mdempsky@mdempsky:/tmp/test$ go doc .
package p // import "."

line1 continued

line2 continued

    line3
    line4

continued

line5 continued

    line6
    line7

line8 continued

    line9
    line10

```

</details>

> This one is a bit tricky logistically, I'd rather not make gofmt a requirement for synthesizing the go files via the Makefile.

Out of curiosity, what's the logistic difficulty here? I thought Go is pretty generally available nowadays.

> I can change the output to match the spacing conventions of gofmt, though?

gofmt can be somewhat finnicky about things; e.g., it'll format `1 * 2` and `3 + 4`, but then `1*2 + 3`.

However, to the extent you're able to better match the output, that sounds good and desirable.

JasonGross:
[...]

PR #970, opened 2021-05-04
Use operator precedence to print C code

JasonGross:
I'm attempting to make the printing code more uniform and factor out common patterns a bit more, in part to decrease the likelihood of transcription errors, but I'm afraid I might be making this code harder to read and understand.  I'd love feedback on this @andres-erbsen @jedisct1 (and anyone else interested in looking at the backend code).

This is also in part to allow implementing @FiloSottile's request in #949 to not have so many useless parentheses.


jedisct1:
This is certainly making the output look nicer.

However I'm not convinced that this is necessary. Extra parenthesis don't make the compiled code run any slower, and this is automatically generated code, that will not be manually modified.

The Coq code is far more complex to understand, which may hurt addition of new languages.

PR #982, opened 2021-05-18
Better JSON casts

JasonGross:
We now always insert casts at the outputs of operations indicating what
type fits the operation.  We do not pre-cast the input types ever.


JasonGross:
This actually makes things worse.

PR #983, opened 2021-05-18
Add named typedefs

JasonGross:
This PR adds named typedefs for field elements.

@mdempsky @FiloSottile does the Go code look good?  (This makes some progress towards https://github.com/mit-plv/fiat-crypto/issues/949)

@davidben Does this seem like a positive change to the C code?

@andres-erbsen Do you have thoughts?

@jedisct1 Does Zig have typedefs?  Do you want to make use of typedefs in the Zig code?

Questions:
- [ ] Which divstep things belong in the montgomery domain?
- [ ] What should we do about the various other kinds of bounds
  (saturated, bytes, etc)?
- [x] Should we synthesize relax? (Current answer: Yes)
- [ ] What should we do about selectznz? (should it take tight fe, loose
  fe, saturated fe; should we have a variant of saturated_relax that
  goes from loose to saturated?)


jedisct1:
This is a really useful and welcome change. 

The distinction between tight and lose field elements clear, without relying on comments. 

And yes, Zig could use this, too.

JasonGross:
What's the syntax for typedefs in Zig?

JasonGross:
(and in particular array typedefs, if those are treated specially)

jedisct1:
> What's the syntax for typedefs in Zig?

const type1 = type2;

This makes “type1” an alias for an existing “type2”.

JasonGross:
> Ex:
> 
> `const TightElement = [5]u64;`

Should this be `pub const TightElement = [5]u64;`, or are type definitions not subject to `pub`?

jedisct1:
Right, they should be pub.

JasonGross:
@jedisct1 I've pushed the Zig updates.  Does this look good to you?  (Do you want something other than snake_case for types?)

fiat-go/32/p224/p224.go
```diff
@@ -42,6 +42,16 @@ import "math/bits"
...
 
+// The type MontgomeryDomainFieldElement is a field element in the Montgomery domain.
+//
+// Bounds: [[0x0 ~> 0xffffffff], [0x0 ~> 0xffffffff], [0x0 ~> 0xffffffff], [0x0 ~> 0xffffffff], [0x0 ~> 0xffffffff], [0x0 ~> 0xffffffff], [0x0 ~> 0xffffffff]]
+type MontgomeryDomainFieldElement[7]uint32
+
+// The type NonMontgomeryDomainFieldElement is a field element NOT in the Montgomery domain.
+//
+// Bounds: [[0x0 ~> 0xffffffff], [0x0 ~> 0xffffffff], [0x0 ~> 0xffffffff], [0x0 ~> 0xffffffff], [0x0 ~> 0xffffffff], [0x0 ~> 0xffffffff], [0x0 ~> 0xffffffff]]
+type NonMontgomeryDomainFieldElement[7]uint32
```
FiloSottile, line 53:
```suggestion
type NonMontgomeryDomainFieldElement [7]uint32
```

Missing space.

fiat-go/32/poly1305/poly1305.go
```diff
@@ -30,6 +30,16 @@ package poly1305
 type uint1 uint8
 type int1 int8
 
+// The type LooseFieldElement is a field element with loose bounds.
+//
+// Bounds: [[0x0 ~> 0xc000000], [0x0 ~> 0xc000000], [0x0 ~> 0xc000000], [0x0 ~> 0xc000000], [0x0 ~> 0xc000000]]
```
FiloSottile, line 35:
```suggestion
// Bounds:
//
//   [[0x0 ~> 0xc000000], [0x0 ~> 0xc000000], [0x0 ~> 0xc000000], [0x0 ~> 0xc000000], [0x0 ~> 0xc000000]]
```

Making it a code block would improve readability.

fiat-go/32/poly1305/poly1305.go
```diff
@@ -30,6 +30,16 @@ package poly1305
 type uint1 uint8
 type int1 int8
 
+// The type LooseFieldElement is a field element with loose bounds.
```
FiloSottile, line 33:
```suggestion
// LooseFieldElement is a field element with loose bounds.
```

Go style omits prefixes like "The function" and "The type".

FiloSottile (commented):
Awesome! Looks good, just style/gofmt nits.

I assume these get prefixed and capitalized like everything else, so I can get them to come out as `p521TightFieldElement`?

fiat-go/32/p224/p224.go
JasonGross: Oops, thanks

JasonGross:
@FiloSottile 
> I assume these get prefixed and capitalized like everything else, so I can get them to come out as `p521TightFieldElement`?

The type names will be prefixed iff the function names are prefixed, so if you pass something other than `''` for the curve description, it will show up before the type.

I'm a bit confused by the capitalization request though.  Are you saying that typedefs follow a different convention than functions for public/private, or that these typedefs should not be exported?  Or perhaps I'm confused about what it means for a typedef to be public; I was assuming that it meant that the name gets exported, but perhaps in Go the name is always exported and it's just a matter of if the definition gets exported?  (You can kludge this by passing `--public-type-case camelCase` instead of `--public-type-case UpperCamelCase`, but I suspect this is semantically wrong, and you instead want something like `--typedefs-private` (or `--typedefs-static`, to match with the rest of the option naming convention which is inherited from C) to make the typedefs be emitted as private rather than public?)

fiat-go/32/poly1305/poly1305.go
JasonGross: Done

fiat-go/32/p224/p224.go
JasonGross: Done, thanks

fiat-go/32/poly1305/poly1305.go
JasonGross: Done

jedisct1:
> @jedisct1 I've pushed the Zig updates. Does this look good to you? (Do you want something other than snake_case for types?)

Output looks good! Types should be PascalCase, though.

FiloSottile:
> I'm a bit confused by the capitalization request though. Are you saying that typedefs follow a different convention than functions for public/private, or that these typedefs should not be exported? Or perhaps I'm confused about what it means for a typedef to be public; I was assuming that it meant that the name gets exported, but perhaps in Go the name is always exported and it's just a matter of if the definition gets exported? (You can kludge this by passing `--public-type-case camelCase` instead of `--public-type-case UpperCamelCase`, but I suspect this is semantically wrong, and you instead want something like `--typedefs-private` (or `--typedefs-static`, to match with the rest of the option naming convention which is inherited from C) to make the typedefs be emitted as private rather than public?)

Nono, types follow the same conventions as all other names.

I just wanted to make sure that I could use the curve description to namespace them and make them private, like I do in https://github.com/golang/go/blob/master/src/crypto/elliptic/internal/fiat/p521_fiat64.go. The idea is that I keep fiat code private and then expose a higher-level API wrapper. This also lets me put multiple fields in the same package.

Sounds like it will work without issues!

JasonGross:
> I just wanted to make sure that I could use the curve description to namespace them and make them private, like I do in https://github.com/golang/go/blob/master/src/crypto/elliptic/internal/fiat/p521_fiat64.go. The idea is that I keep fiat code private and then expose a higher-level API wrapper. This also lets me put multiple fields in the same package.
> 
> Sounds like it will work without issues!

@FiloSottile  Yes.  It should also work if instead of passing `--internal-static --public-function-case camelCase --public-type-case camelCase` you pass `--static --public-function-case UpperCamelCase --public-type-case UpperCamelCase` (`--static` says to make everything private), and there's a sense in which this is more semantically correct (also if this doesn't work that means there's a bug in my code, which would be worth reporting).

JasonGross:
> Types should be PascalCase, though.

@jedisct1 Is this for both public and private types?

JasonGross:
@FiloSottile do you have opinions / requests on these questions?

> * Which divstep things belong in the montgomery domain?
> * What should we do about the various other kinds of bounds
>   (saturated, bytes, etc)?
> * What should we do about selectznz? (should it take tight fe, loose
>   fe, saturated fe; should we have a variant of saturated_relax that
>   goes from loose to saturated?)

jedisct1:
> > Types should be PascalCase, though.
> 
> @jedisct1 Is this for both public and private types?

Yes, for all types

JasonGross:
I'm going to merge this so it doesn't bitrot; feel free to open issues about improvements you want

PR #991, opened 2021-05-25
Bump rupicola

andres-erbsen:
@jadephilipoom / @JasonGross  What is `src/Bedrock/Field/Synthesis/New/Signature.v` for and what do you think it would make sense for us to do about its breakage right now if we are trying to get a rupicola-based in-coq derivation of x25519 through? In particular, it would be convenient to have the ongoing work on master but also breaking all bedrock2-based code generation tests seems undesirable.


JasonGross:
This is part of Jade's new version of the pipeline, I think?  I've only made minor edits to work with changes I've made to, e.g., Wf3, so I'm not knowledgeable about this part.

andres-erbsen:
Ok, it looks like the changes that break `src/Bedrock/Field/Synthesis/New/Signature.v` are in https://github.com/mit-plv/fiat-crypto/pull/991/commits/9c6dec90d5aeede7d2db41c2658f33e3b1c2fe77#diff-812fa7d36366f4b964e6045e3e8d102df655e9168bfd6c027b7b3473e3f41b23R103 and that `src/Bedrock/Field/Synthesis/New/UnsaturatedSolinas.v` (which generates bedrock2 code for x25519) does depend on it, so we will need to fix it up.

JasonGross:
> I'm not sure where the information about the input lists should come from.

I expect that `bounded_by` (`H` and `H0`) will give you that the lengths of `x` and `y` are the same as the lengths of the corresponding bounds lists (`bin_xbounds` and `bin_ybounds`).  idk where that's defined, but you should have that tight bounds and loose bounds (and in fact all of your section-wide bounds) should have length `n`.

jadephilipoom:
> I expect that `bounded_by` (`H` and `H0`) will give you that the lengths of `x` and `y` are the same as the lengths of the corresponding bounds lists (`bin_xbounds` and `bin_ybounds`). 

Yep, that should be it.

jadephilipoom:
If you need to add a section variable saying that `bin_xbounds` and `bin_ybounds` have length `n`, that's perfectly fine -- it will be easy to prove at the next (`UnsaturatedSolinas.v`) level since all of those will be `tight_bounds` or `loose_bounds`.

andres-erbsen:
https://github.com/jadephilipoom/fiat-crypto/commit/9c6dec90d5aeede7d2db41c2658f33e3b1c2fe77#diff-df8e71cd8ec77f0d6a56cf6acde9baaf3f6e956fe366e7d5673c9e60b874723aL139-R58 swapped argument order (outputs-first to outputs-last), is that really intended? I see no reason, and I figure we thought about https://github.com/mit-plv/fiat-crypto/pull/832 when we did it, so I am leaning towards reverting that

andres-erbsen:
@DIJamner the last commit I pushed changed the specs, sorry. The specs for binary operations should now be final, though -- I think I fixed up the last proof about them. The same for unary operations will hopefully follow soon.

jadephilipoom:
> [jadephilipoom@9c6dec9#diff-df8e71cd8ec77f0d6a56cf6acde9baaf3f6e956fe366e7d5673c9e60b874723aL139-R58](https://github.com/jadephilipoom/fiat-crypto/commit/9c6dec90d5aeede7d2db41c2658f33e3b1c2fe77#diff-df8e71cd8ec77f0d6a56cf6acde9baaf3f6e956fe366e7d5673c9e60b874723aL139-R58) swapped argument order (outputs-first to outputs-last), is that really intended? I see no reason, and I figure we thought about #832 when we did it, so I am leaning towards reverting that

Feel free to revert.

Issue #707, opened 2020-03-30
Add Java tests to the CI


JasonGross:
We should test the Java code.  I'm not sure where to find Java tests, though. cc @peterdettman @dghgit,

peterdettman:
Hi Jason, happy to lend a hand here, but it may be a week or two.

jvz:
Here's a fairly simple combination of operations which give different results in Java and C: https://gist.github.com/jvz/0bd0ff04b47524da8b8aedc16e79a4eb

I've seen similar discrepancies in a few other operations similarly.

spitters:
This commit by @eorloff might solve this issue:
https://github.com/eorloff/fiat-crypto

PR #995, opened 2021-06-02
Buggy java code

eorloff:
We have been trying to use fiat-crypto to generate java code for BLS12-381. We initially tried using parameters like the ones used in the github actions for generating the java sources, but the generated casts did not work as expected. As far as I can tell the major problem is when we need to convert from int to long as the conversion, as the sign is preserved. 

I have in this merge request removed the usage of the `Box` class and instead created one element arrays. This allows always using primitives and thus avoided some problems with boxing and unboxing of the numbers. 

The static cast has been changed to always assume that we can use `Integer.toUnsignedLong` to ensure that the sign is not preserved when casting from int to long.

For the code to be functional we needed to remove the flag `only-signed`, thus allowing the generated code to handle unsigned values in the signed types. This should be safe as the operations used on the values is the same for 2-complement signed values and for unsigned values. The change was needed to avoid the usage of `long` to represent 32 bit unsigned values. The generated casts (or no cast at all) when calling a method with an int that expects a long since an implicit cast will happen that preserves the sign.

The command we ended up using to generate functional java code:
`word_by_word_montgomery  --lang Java --cmovznz-by-mul --widen-carry --widen-bytes --internal-static BLS12 32 4002409555221667393417789825735904156556882819939007885332058136124031650490837864442687629129015664037894272559787  mul square add sub opp from_montgomery to_montgomery nonzero selectznz to_bytes from_bytes one msat divstep divstep_precomp`


JasonGross:
This is awesome, thanks for tracking this down and figuring out how to fix it!  Please also remove the `--only-signed` flag from the Makefile and regenerate the Java files with `make java-files`.

Could you elaborate on what the issue is with boxing?

Optional bonus items:
- [ ] add a test to the test-suite a la the suggestion in https://github.com/mit-plv/fiat-crypto/issues/707#issuecomment-763098543
- [ ] assuming this in fact fixes the discrepancy, update the README to no longer mark the Java code as known buggy
- [ ] if you'd like, update the README table of backends to mark yourself as a maintainer of the Java backend

eorloff:
The primary problem I had with boxing was the inability to cast from Long to int. I have now changed the static cast to handle this case and therefore reintroduced the usage of Box. My only other concern is how the performance is when doing that much boxing and unboxing, but since I do not have any data to support the thought. 

I have merged master, and will generate the java-files tomorrow at the latest.

PR #1002, opened 2021-07-14
make accept-Crypto.Bedrock.Group.ScalarMult.MontgomeryLadder.montladd…

JasonGross:
…er_body

Idk why this file was previously empty


JasonGross:
Ah, it was empty because the output test was removed.

JasonGross:
Maybe we should remove the test from the makefile?

Issue #1008, opened 2021-08-08
add a flag to enable subst01 in selectznz


JasonGross:
> As a start, looking at the Selectznz in curve25519.go, shortening the lifetimes of the variables will help.

I'll aim to add a flag that allows this shortly.  If it's helpful, I can also have the cmovznz calls inlined, though I expect that'll take me a little bit more work

_Originally posted by @JasonGross in https://github.com/golang/go/issues/40171#issuecomment-894750084_

PR #1010, opened 2021-08-16
bump rupicola/bedrock2 and adapt to expr.inlinetable feature

xisanth:
#1008

### DIFF
--- a/src/Bedrock/Field/Translation/Parameters/Defaults.v
+++ b/src/Bedrock/Field/Translation/Parameters/Defaults.v
@@ -48,6 +48,7 @@ Section Defs.
     | expr.var x => negb (String.eqb x ERROR)
     | expr.load _ a => error_free_expr a
     | expr.op _ x y => (error_free_expr x && error_free_expr y)%bool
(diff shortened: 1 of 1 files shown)


JasonGross:
@samuelgruetter the commit is yours; do you know this person?  What's the state of this branch?

samuelgruetter:
This commit is part of the `compiler-integration` branch on which I worked in Dec 2020. The goal was to see if the bedrock2 compiler can compile fiat-crypto-generated bedrock2 code, and to invoke that risc-v code from ed25519 signature code compiled by gcc, and to run it inside qemu. It worked for some of the crypto functions, but not for all of them (ie for some I still had to keep using gcc instead of the bedrock2 compiler). I then concluded that instead of more debugging, we should "just" prove that it works, and that we'd find the bugs while trying to prove it, but that hasn't happened yet...

I think the reason we did not merge this is that at the time, the examples in rupicola depended on an old version of bedrock2, so the rupicola build failed when trying to bump the bedrock2 submodule of rupicola.

JasonGross:
Closing as this is outdated.  Perhaps this is just spam

PR #1017, opened 2021-09-21
Bump minimum Coq dependency to 8.11, use coqutil

JasonGross:
Now that Debian bullseye has reached stable, with it's Coq 8.12.0
package, we can bump our minimum version of Coq to 8.11, since the
latest Ubuntu LTS, 20.04 (focal), has 8.11.0.  This allows us to use
coqutil, which is now maintaining compatibility with Coq back to 8.11.

This is blocked on ~https://github.com/mit-plv/rupicola/pull/14~ https://github.com/mit-plv/rupicola/pull/15 (we should update the submodule to point to a commit on master once that's merged).
Before this is merged, we should also tag a final release compatible with Coq 8.9 and 8.10.


andres-erbsen:
port WIP at https://github.com/mit-plv/fiat-crypto/tree/bump-rupicola-noparamrecords

andres-erbsen:
I update the dependencies and got everything to build on 8.13, the work I did is at https://github.com/mit-plv/fiat-crypto/commits/bump-rupicola-noparamrecords-standalone and a version with this PR on top is at https://github.com/mit-plv/fiat-crypto/commits/bump-rupicola-noparamrecords

I am guessing the CI builds will fail on some other version, but maybe you can take over at this point anyway?

JasonGross:
> I am guessing the CI builds will fail on some other version, but maybe you can take over at this point anyway?

CI seems to pass.  I've made #1017, and will take over from here (tagging a last release before dropping support for older Coqs and merging).  Thanks!

PR #1023, opened 2021-10-01
 Add `--asm-ignore-unique-name-mismatch`

JasonGross:
Don't unconditionally ignore mismatch between the name of the assembly function and the name of the synthesized function.

@andres-erbsen Does making this a command-line flag seem reasonable to you (under the presumption that sometimes we might want to be warned if we're using the wrong assembly file, and also that we should be in strict mode by default to catch such mistakes earlier)?  Does the name seem fine?

### DIFF
--- a/src/BoundsPipeline.v
+++ b/src/BoundsPipeline.v
@@ -227,6 +227,9 @@ Typeclasses Opaque assembly_hints_lines_opt.
 (** Error if there are un-requested assembly functions *)
 Class error_on_unused_assembly_functions_opt := error_on_unused_assembly_functions : bool.
 Typeclasses Opaque error_on_unused_assembly_functions_opt.
+(** Ignore function-name mismatch errors when there's only one assembly function and only one actual function requested *)
+Class ignore_unique_asm_names_opt := ignore_unique_asm_names : bool.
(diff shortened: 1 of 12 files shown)


andres-erbsen:
afaik current usage of the checker reliably passes one function whose name does not match the fiat-crypto-generated name.  @dderjoel, do you think this should this be allowed, a warning, or an error?

JasonGross:
TODO for myself from call with Andres: flip the sense of the flag; if the check succeeds, it must have just been a name mismatch.  If the equivalence check fails, it would be nice to get this error/warning message, but it's not super-important.

dderjoel:
> TODO for myself from call with Andres: flip the sense of the flag; if the check succeeds, it must have just been a name mismatch. If the equivalence check fails, it would be nice to get this error/warning message, but it's not super-important.

I like this. Would not break anything now. Would this description would then be the default - if so, what would a flag change then ?

JasonGross:
I've flipped the sense and updated the help text to "By default, when there's only one assembly function in a --hints-file and only one function requested to be synthesized, the name of the assembly function is ignored.  Passing this flag disables this behavior, raising an error if the names mismatch regardless of whether there are multiple functions or just one."  @dderjoel, does this seem clear enough?

PR #1024, opened 2021-10-03
Bump minimum Coq dependency to 8.11, use coqutil (take 2)

JasonGross:
Version of #1017 that will hopefully actually work.

Now that Debian bullseye has reached stable, with it's Coq 8.12.0
package, we can bump our minimum version of Coq to 8.11, since the
latest Ubuntu LTS, 20.04 (focal), has 8.11.0. This allows us to use
coqutil, which is now maintaining compatibility with Coq back to 8.11.

Before this is merged, we should also tag a final release compatible with Coq 8.9 and 8.10.


JasonGross:
Note to self: I'll also have to update the dependencies in opam and maybe also the Coq CI now that we depend on coqutil everywhere.

JasonGross:
I'll merge this once https://github.com/mit-plv/fiat-crypto/actions/runs/1304396335 runs

PR #1025, opened 2021-10-03
[legacy] Adapt to coq/coq#13837 ("apply with" does not rename arguments)

SkySkimmer:
Should be backwards compatible.


JasonGross:
I'm confused how this is backwards compatible, but the CI says it passes, so I'll merge.  Did the old code accept both renamed and nonrenamed arguments?

JasonGross:
Thanks!

JasonGross:
Ah, thanks for the explanation

PR #1030, opened 2021-10-09
Zig: use @import("builtin") instead of the deprecated reexport

### LINKED ISSUE #1029: Zig code fails due to recent change in zig
```
error: get this from @import("builtin") instead of std.builtin
pub const mode = @compileError("get this from @import(\"builtin\") instead of std.builtin");
                 ^
./src/curve25519_32.zig:16:25: note: referenced here
const mode = std.builtin.mode; // Checked arithmetic is disabled in non-debug modes to avoid side channels
                        ^
./src/curve25519_32.zig:826:23: note: referenced here
    @setRuntimeSafety(mode == .Debug);
                      ^
The following command exited with error code 1:
```
cc @jedisct1

jedisct1:
Fixes #1029


JasonGross:
Thanks!

Issue #1022, opened 2021-09-30
removing FieldRepresentation in favor of Representation.frep


andres-erbsen:
I don't see us ever using a different implementation of `FieldRepresentation` than `frep` in `Representation.v` and thus I feel this interface serves no conceptual function. Can we remove it, or is it essential for preventing unfolding or something else? @jadephilipoom @DIJamner 

`FieldRepresentation_ok` says `FElem` has the same memory footprint as a `Placeholder`, which is defined right above as `Memory.anybytes p felem_size_in_bytes`, which means `felem_size_in_bytes` bytes. That count is further required to be 0 mod `width/8`; the corresponding quotient is exactly the `n` in `Representation.v`. So we're definitely dealing with an array of `n` words -- and the `felem` type is in fact instantiated with a list of words.

There is slightly more flexibility with other fields, perhaps most seriously `eval` could itself be a parameter instead of parametrizing over a weight function, but I still don't see a use case for different parameters, so I'd be inclined to squash it.

DIJamner:
I haven't looked under the FElem interface much, but looking at `frep`, it seems pretty fine to work with its implementation directly in any of the code that I've touched. Even if there is some code that wants to differentiate `FElem` and `Bignum`, which I haven't seen, we could probably reduce it to a local alias rather than a fully opaque class.

As far as ordering, I would like to prioritize the parameter changes so that I'm not working on two Rupicola branches longer than necessary. On the other hand, it probably makes sense to make this change, if we're going to, before trying to move any additional code to a unified predicate like we were considering since it will have some effect on that.

jadephilipoom:
Hmm, I think it's a useful layer of abstraction. `frep` already exposes things like `Bignum` and `list_Z_bounded_by` for bounds that the group layer should really not be unfolding; the FieldRepresentation interface basically enforces that rule. Sure, if we trust that everyone who ever writes anything at the group level will know better than to ever unfold `frep`, it wouldn't be an issue, but that seems more fragile to me.

[...]

PR #1034, opened 2021-10-20
reveal FElem:=Bignum in FieldRepresentation

andres-erbsen:
@DIJamner I think this should be enough to cmov `FElem`s  
@jadephilipoom gave general agreement over chat, feel free to nit here still  
#1022 remains open because the admit mentioned there still needs to be resolved


DIJamner:
The cmove I have in rupicola is implemented for a proposition that with the right instantiation is equal (`=`) to Bignum, so it will be easy to transport to `FElem` under this change.

PR #1046, opened 2021-11-12
Pin Haskell to 9.0.1

JasonGross:
As per https://github.com/coq/coq/issues/15180, 9.2.1 cannot compile our code :-(


JasonGross:
Let's see if #1048 fixes this, since I think that's a slightly better option

JasonGross:
Actually, I'll just make #1048 be on top of this one

PR #1047, opened 2021-11-12
Adapt w.r.t. coq/coq#15178.

ppedrot:
Should be backwards compatible, but let's check the result on the other PR first.


ppedrot:
This is indeed enough to make coq/coq#15178 go through, and the local CI shows this is backwards compatible, so we don't lose anything by merging this now.

PR #1048, opened 2021-11-14
Compatibility with Haskell 9.2.1

JasonGross:
Alternative to and closes #1046, per the response to
https://github.com/coq/coq/issues/15180


JasonGross:
Failing due to a Haskell regression on large files, reported upstream [here](https://gitlab.haskell.org/ghc/ghc/-/issues/20683).

PR #1051, opened 2021-11-16
Improve symbolic equivalence unification errors

JasonGross:
Error messages now contain explanations such as:
```
Unable to unify: [inr [191, 197, 199, 201]] == [inr [34, 40, 43, 49]]
index 0: [191, 197, 199, 201] != [34, 40, 43, 49]
index 0: 191 != 34
(add 64, [3, 8, 10, 190]) != (add 64, [3, 8, 10, 33])
index 3: 190 != 33
(and 64, [30, 31]) != (and 1, [30])
(and 64, [(selectznz, [28, 8, 29]), (const 1, [])]) != (and 1, [(selectznz, [28, 8, 29])])
Operation mismatch: and 64 != and 1
```

@andres-erbsen @dderjoel what do you think?  Any suggestions/requests?

### DIFF
--- a/src/Assembly/Equivalence.v
+++ b/src/Assembly/Equivalence.v
@@ -93,6 +93,108 @@ Inductive EquivalenceCheckingError :=
 | Code_after_ret
 .
 
+Fixpoint explain_array_unification_error
+         (singular plural : string)
+         (explain_idx_unification_error : idx -> idx -> list string)
+         (asm_array PHOAS_array : list idx) (start_idx : nat)
+  : list string
+  := match asm_array, PHOAS_array with
+     | [], []
+       => ["Internal Error: Unifiable " ++ plural]
+     | [], PHOAS_array
+       => ["The synthesized " ++ singular ++ " contains " ++ show (List.length PHOAS_array) ++ " more values than the assembly " ++ singular ++ "."]
+     | asm_array, []
+       => ["The assembly " ++ singular ++ " contains " ++ show (List.length asm_array) ++ " more values than the synthesized " ++ singular ++ "."]
+     | asm_value :: asm_array, PHOAS_value :: PHOAS_array
+       => if Decidable.dec (asm_value = PHOAS_value)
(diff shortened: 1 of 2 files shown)


dderjoel:
okay, that is helpful.

`64` and `1`, do they resolve to `(*64*) (old 64 64, []);` ([here](https://github.com/mit-plv/fiat-crypto/runs/4218951643?check_suite_focus=true#step:9:683))  and `(*1*) (old 64 1, []);` ([here](https://github.com/mit-plv/fiat-crypto/runs/4218951643?check_suite_focus=true#step:9:620)) ?
what does that `old` tell me? 
I still can't really see, where the mismatch is from the amd64-code and the c-code or coq representation starting in line [503](https://github.com/mit-plv/fiat-crypto/runs/4218951643?check_suite_focus=true#step:9:503).

JasonGross:
> would mean, that 64 bits from the left `and` is not the same as the 1 bits from the right hand `and`
> 
> If the left one is amd64, and the right one is coq, than it should be correct, right? Because `const 1` (expr. 31) is 1 bit wide, and `and`ing that with expression 31 gives the 1 bit wide expression, that we see on the right hand side (the selectznz).

Right, I'm working on fixing this



> `64` and `1`, do they resolve to `(*64*) (old 64 64, []);` ([here](https://github.com/mit-plv/fiat-crypto/runs/4218951643?check_suite_focus=true#step:9:683)) and `(*1*) (old 64 1, []);` ([here](https://github.com/mit-plv/fiat-crypto/runs/4218951643?check_suite_focus=true#step:9:620)) ?

No, these are just bitwidths

> I'm having trouble to understand which numbers in those lines are constants

Inline arguments to operators are constants.  So, `x` in `const x`, `x` in `add x`, etc

JasonGross:
After chatting with Andres: Not a total solution to debugging, but better than the current state, so I'll merge it and we can incrementally improve it.

PR #1049, opened 2021-11-15
added testcase p224 sub

dderjoel:

PR #1052, opened 2021-11-17
Add support for folding constants into and

JasonGross:
Supercedes and closes #1049

I got nerdsniped and wrote the rewrite rule myself.

The first commit of this PR is a reorganization of the associative-commutative machinery for operations, making more operations associative and commutative.  I tried to minimize the number of proofs in which we destruct the operation.  I decided on the following:

For an operation with identity, the spec is that if `identity o = Some id`, then:
- `interp_op G o xs = interp_op G o (List.filter (fun v => negb (v =? id)) xs)`
- `interp_op G o [] = Some id`

For an associative operation, the spec is:
- `interp_op G o xs = fold_right (fun v acc => acc <- acc; interp_op G o [v; acc])%option (interp_op G o []) xs`
- `interp_op G o [] = identity o`
- `interp_op G o ys = Some vys -> interp_op G o zs = Some vzs ->((xy <- interp_op G o [x; vys]; interp_op G o [xy; vzs]) = (yz <- interp_op G o [vys; vzs]; interp_op G o [x; yz]))%option` which is only used to prove `interp_op G o (List.concat xs) = (vxs <-- List.map (interp_op G o) xs; interp_op G o vxs)%option`, which is generally treated as the actual spec for this bullet point 

The third commit is @dderjoel 's test from #1049

### DIFF
--- a/fiat-amd64/8917_ratio11077_seed3401074160038646_sub_p224.asm
+++ b/fiat-amd64/8917_ratio11077_seed3401074160038646_sub_p224.asm
@@ -1,0 +1,56 @@
+SECTION .text
+	GLOBAL sub_p224
+sub_p224:
+sub rsp, 0x20
+mov rax, [ rsi + 0x0 ];x1, copying arg1[0] here, cause arg1[0] is needed in a reg. It has those deps: "x1--x2, size: 1"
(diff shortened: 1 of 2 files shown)


JasonGross:
Ack'd by Andres over zoom

PR #1061, opened 2021-11-30
Fix JSON bitwidths by avoiding rounding

### LINKED ISSUE #1058: Bitwidth differences in fiat-json output
I have been experimenting with code-generation based on the fiat-json output format. As an aside, I'd be interested to know if you consider that a good or bad idea. For now I'm trying to generate output that's mostly equivalent to fiat-go but from the fiat-json files.

One issue is the datatypes reported in the fiat-json output. For example, in the `to_bytes` function for 64-bit curve25519:

https://github.com/mit-plv/fiat-crypto/blob/edba07de3c424cb4d584044d3231afcaca67edd5/fiat-json/src/curve25519_64.json#L3891-L3903

https://github.com/mit-plv/fiat-crypto/blob/edba07de3c424cb4d584044d3231afcaca67edd5/fiat-go/64/curve25519/curve25519.go#L596-L598

A few other languages I looked at (C, Rust, Zig) all generate similar code.

So `u32` in the JSON seems like obviously the wrong type. Presumably `u64` or `u51` would be the most helpful here. The consumer of the fiat-json output somehow needs to know this is a 51-bit subtraction.

JasonGross:
Fixes #1058


JasonGross:
@dderjoel @mmcloughlin does this look good?

fiat-json/src/README.md
```diff
@@ -64,6 +65,12 @@ It describes a line of code or an expression:
 - whose arguments are the in the arguments field, which may be either
   subexpressions or variable names
 
+The suffixed `mulx_*`/`addcarryx_*`/`subborrowx_*` are used when the
+index of the carry location is not a natural machine bitwidth (natural
+bitwidths are powers of 2).  When the carry location is a natural
+machine bitwidth, the unsuffixed versions are used, and the size shows
+up in the datatype field.
```
mmcloughlin, line 72: Why did you decide to encode this information in a string vs an additional field? I'm not necessarily saying this is bad I'm just wondering about your thought process.

```json
      {
        "datatype": "(auto)",
        "name": ["x1", "x2"],
        "operation": "subborrowx",
        "operation_size": 51,
        "arguments": ["0x0", "arg1[0]", "0x7ffffffffffed"]
      },
```

Maybe if you wanted this to be potentially more flexible in future you'd nest it in a generic "parameters" field.

```json
      {
        "datatype": "(auto)",
        "name": ["x1", "x2"],
        "operation": "subborrowx",
        "parameters": {"size": 51},
        "arguments": ["0x0", "arg1[0]", "0x7ffffffffffed"]
      },
```

fiat-json/src/p224_32.json
```diff
@@ -10328,7 +10328,7 @@
         "arguments": [
           "x43",
           {
-            "datatype": "u32",
+            "datatype": "(auto)",
```
mmcloughlin, line 10331: This change has added `(auto)` to a lot of seemingly unrelated places? Is this happening anywhere the data size is not a power of two (or a standard bitwidth)?

That's fine if the consumer has the information to deduce what the size is? But does it? Or would it have to redo analysis that `fiat-crypto` has already done.

fiat-json/src/README.md
JasonGross: I was trying to minimize the change to the spec, but this is a good point, I guess I can add a `"parameters"` field.

fiat-json/src/p224_32.json
JasonGross: Good catch, this was my mistake in not realizing that I was passing `sz` recursively to `arith_to_string` and it was shadowing a name.  I'll fix this; thanks for catching it!

fiat-json/src/README.md
mmcloughlin: Completely your call. Seemed like a reasonable suggestion but I don't have all the information like you and @dderjoel do. Can we think of any other fields that might belong under `parameters`, or is this an over-generalization?

JasonGross: Updated

fiat-json/src/curve25519_64.json
```diff
@@ -3889,64 +3889,64 @@
...
+        "operation": "subborrowx_u51",
         "arguments": [
           "0x0",
           "arg1[0]",
           "0x7ffffffffffed"
         ]
       },
       {
-        "datatype": "u32",
+        "datatype": "(auto)",
```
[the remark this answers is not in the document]
JasonGross: Yes, `u64` is not really an accurate specification of the size of the two variables, and there's not really a reason for us to do the rounding/calculation, especially given that there's nothing keeping what we'd print here in sync with any changes in the codebase.

fiat-json/src/p224_32.json
JasonGross: Fixed

JasonGross:
I've pushed an updated version which adds the `parameters` field to everything

fiat-json/src/README.md
dderjoel: seems sensible to me so far.

dderjoel:
> I've pushed an updated version which adds the parameters field to everything. I decided to add it everywhere even when it's empty for uniformity, but I'm not sure if that's the best decision.

I like this. When is the parameter.size undefined? Does this only happen, if the datatype !== "(auto)"? And what does that mean then?  For `addcarryx`, `name[0]` is of type `u1` (implicitly through the operation) and `name[1]` has type `u<parameter.size>`? 
I do not particularly like the 'implicit' part here.

How about (optionally also putting the `width`s of the inputs in there.
```json
{ ...
"parameters": {
  "outWidths": [ 1, 51],
  "inWidths": [ 1, 51, 64],
  ...
}
}
```

mmcloughlin:
There seem to be three separate things at play that are all interrelated and often the same:

1. Minimum number of bits needed to store the values of the inputs and outputs
2. Datatype used for the inputs and outputs. This is often (or maybe always) the minimum bits rounded up to the next power of two, but it's at least conceivable the datatype chosen could be programming-language dependent.
3. Operation/computation to be performed, specifically in this case where the carry bit is.

Currently:

1. Minimum number of bits is provided implicitly in JSON for arguments and return values since we get lower and upper bounds. It's not available for all expressions. @dderjoel if I understand you, you're suggesting adding extra `*width` parameters for this? This could be useful, though I'm not sure I personally have a specific need for it yet, and I don't think it's required to fully specify to the consumer what computation is being specified?
2. Datatype is already covered by the `datatype` field. There's a question of whether this needs to be extended to optionally be a list of types, so we can specify that `addcarryx` and `subborrowx` return `["u64", "u1"]`. I think this would be more correct, though I'm fine for now just hard-coding that the second return type is `u1` in those cases.
3. We're proposing here that _operation parameters_ are added to the `parameters` field, specifically `size` parameter for `addcarryx` and `subborrowx`. Maybe `size` is too vague of a name and we should use `carry_position`, not sure?

Having written that I'm not sure how helpful it was 😂  Just thought it might help to get a bit more precise about what we're talking about.

Thanks for the work @JasonGross @dderjoel

mmcloughlin:
Screen Shot 2021-11-30 at 8 28 57 PM

Jaysus 😮 

> I decided to add it everywhere even when it's empty for uniformity, but I'm not sure if that's the best decision.

I'm fine with this field being missing when empty, which I think is the majority of cases. I'd go with that if I were you, but it doesn't matter too much either way.

dderjoel:
> I'm fine with this field being missing when empty, which I think is the majority of cases. I'd go with that if I were you, but it doesn't matter too much either way.

yeah, also doesn't matter to me, which one of  'parameters' or `parametes.size` is optional.


> There seem to be three separate things at play that are all interrelated and often the same:
> 
> 1. Minimum number of bits needed to store the values of the inputs and outputs
> 2. Datatype used for the inputs and outputs. This is often (or maybe always) the minimum bits rounded up to the next power of two, but it's at least conceivable the datatype chosen could be programming-language dependent.
> 3. Operation/computation to be performed, specifically in this case where the carry bit is.
> 
> Currently:
> 
> 1. Minimum number of bits is provided implicitly in JSON for arguments and return values since we get lower and upper bounds. It's not available for all expressions. @dderjoel if I understand you, you're suggesting adding extra `*width` parameters for this? 

Yes, but on parameter level for each operation (such as addcarryx) instead of only at method level (such as carry_mul)

> This could be useful, though I'm not sure I personally have a specific need for it yet, and I don't think it's required to fully specify to the consumer what computation is being specified?

Hmm, I guess if the consumer knows, one value `u26` instead of the generic round up to <bitwitdh of system>bits (say `u64`), It could potentially vectorize more values (casting them to `u32`). However, I don't know if that could be the case anyway for curve25519 specifically, or if (for any other prime), if fiat would then actually also choose `u32` instead.

> 2. Datatype is already covered by the `datatype` field. There's a question of whether this needs to be extended to optionally be a list of types, so we can specify that `addcarryx` and `subborrowx` return `["u64", "u1"]`. I think this would be more correct, though I'm fine for now just hard-coding that the second return type is `u1` in those cases.

Well, I think we all agree that the combination of datatype and implicit knowledge of the operation is a bit confusing. Thus I thought to explicitly state the max considered bits per in/output. that way, the consumer knows all it can, and could do whatever the arch allows it to do.

> 3. We're proposing here that _operation parameters_ are added to the `parameters` field, specifically `size` parameter for `addcarryx` and `subborrowx`. Maybe `size` is too vague of a name and we should use `carry_position`, not sure?

Im not sure about `carry_position`  isn't that always `<specified bitwidth>+1` ? So in this case position 65? I don't quite understand what that value of carry_position then mean?

JasonGross:
> I'm fine with this field being missing when empty, which I think is the majority of cases. I'd go with that if I were you, but it doesn't matter too much either way.

Sure, I'll do that tomorrow

> When is the parameter.size undefined? 

It's always defined for addcarryx, subborrowx, mulx, and never defined for any other operation.

> How about (optionally also putting the widths of the inputs in there).

These are not really easily computed with any precision greater than "smallest integer type not smaller than parameters.size bits".  And any change we make here is probably going to end up being global, and require piping around a bunch of extra information, so I'm loathe to do it if it's not necessary.

mmcloughlin:
> These are not really easily computed with any precision greater than "smallest integer type not smaller than parameters.size bits". And any change we make here is probably going to end up being global, and require piping around a bunch of extra information, so I'm loathe to do it if it's not necessary.

I don't think this is necessary right now so we can leave it if it's difficult.

JasonGross:
I'll plan to merge this tonight or tomorrow, unless @mmcloughlin @dderjoel you want more time to review the latest version before I merge?

fiat-json/src/README.md
```diff
@@ -36,18 +36,26 @@ Let us define the following types:
 
 - `bound`, either `null` or a `number` or a list of `bound`s
 
+- `parameters`, a comma-separated list of key-value pairs separated by
+  `:`, where the keys are quoted string literals and the values are
+  either quoted string literals or `number`
```
mmcloughlin, line 41: This is not particularly important, but it seems odd to me to explain it in this level of detail, because the syntax is standard JSON object syntax.

So you could say "a JSON object with `string` keys and `string | number` values".

fiat-json/src/README.md
```diff
@@ -131,14 +146,29 @@ fiat_p256_subborrowx_u64(&x180, &x181, x179, x171, UINT64_C(0xffffffff00000001))
...
-"datatype": "u64",
+"datatype": "(auto)",
 "name": ["x180", "x181"],
 "operation": "subborrowx",
+"parameters": {"size": 64},
+"arguments": ["x179", "x171", "0xffffffff00000001"]
+}
+```
+also, we can infer from `subborrowx` which arguments have which meaning
+and which varibable in `"name"` contains the carry out.
```
mmcloughlin, line 157:
```suggestion
and which variable in `"name"` contains the carry out.
```

mmcloughlin (approved):

mmcloughlin:
> I'll plan to merge this tonight or tomorrow, unless @mmcloughlin @dderjoel you want more time to review the latest version before I merge?

Yeah LGTM. I left some very small comments about the readme. I also approved it but I don't think that actually carries any weight haha!

fiat-json/src/README.md
JasonGross:
```suggestion
- `parameters`, a JSON object with `string` keys and `string | number` values
```

fiat-json/src/README.md
```diff
@@ -36,18 +36,26 @@ Let us define the following types:
...
 ```json
 {
-"datatype" : datatype,
-"name"     : [name],
-"operation": operation,
-"arguments": [ name | expr | number ]
+"datatype"  : datatype,
+"name"      : [name],
+"operation" : operation,
+"parameters": { parameters }
```
JasonGross, line 51:
```suggestion
"parameters": parameters
```

dderjoel:
I am also happy to provide my types which look like this:

```typescript
export namespace Fiat {
  export type Operation =
    | "!"
    | "&"
    | "*"
    | "+"
    | "-"
    | "<<"
    | "="
    | ">>"
    | "|"
    | "~"
    | "addcarryx"
    | "cmovznz"
    | "mulx"
    | "static_cast"
    | "subborrowx";

  export type Datatype = "(auto)" | DTArray | DTScalar;
  export type DTScalar =
    | `i${1 | 8 | 32 | 64 | 128}`
    | `u${1 | 8 | 32 | 64 | 128}`;
  export type DTArray = `u${8 | 32 | 64}[${number}]`;
  export type ArgName = `arg${number}`;
  export type OutName = `out${number}`;
  export type OutVarName = `${OutName}[${number}]`;
  export type ArgVarName = `${ArgName}[${number}]`;
  export type VarName = `x${number}`;
  export type HexConstant = `0x${string}`;

  export type FiatArgument<N extends ArgName | OutName> = {
    name: N;
  } & (FiatArgumentScalar | FiatArgumentArray);

  export interface FiatArgumentArray {
    datatype: DTArray;
    lbound: HexConstant[];
    ubound: HexConstant[];
  }

  export interface FiatArgumentScalar {
    datatype: DTScalar;
    lbound: HexConstant;
    ubound: HexConstant;
  }

  export interface Parameters {
    size: number;
  }

  export type Argument =
    | {
        datatype: Datatype;
        name: Array<"_" | VarName> | OutVarName[] | OutName[] | never[];
        operation: Operation;
        arguments: Argument[];
        parameters?: Parameters;
      }
    | HexConstant /*0x12*/
    | ArgName /*arg1*/
    | ArgVarName /*arg1[2]*/
    | VarName /*x34*/
    | `${number}` /*"1" stringigfied number(eg. shift ops )*/;

  export interface FiatFunction {
    operation: string;
    arguments: FiatArgument<ArgName>[];
    returns: FiatArgument<OutName>[];
    body: Argument[];
  }
}
```

mmcloughlin:
@dderjoel Very nice 😍

PR #1066, opened 2021-12-03
Move mmio

andres-erbsen:


JasonGross:
The submodule does not build:
```
File "/Users/runner/work/fiat-crypto/fiat-crypto/rupicola/src/Rupicola/Lib/Notations.v", line 451, characters 14-30:
Error:
In environment
width : Z
BW : Bitwidth width
word : Interface.word width
mem : map.map word.rep byte
locals : map.map string word.rep
env : map.map string (list string * list string * Syntax.cmd)
ext_spec : Semantics.ExtSpec
word_ok : word.ok word
mem_ok : map.ok mem
locals_ok : map.ok locals
env_ok : map.ok env
ext_spec_ok : Semantics.ext_spec.ok ext_spec
field_parameters : FieldParameters
field_representaton : FieldRepresentation
field_representation_ok : FieldRepresentation_ok
ladderstep_body := ?Goal : bedrock_func
spec_constr := {|
                 _defn_name := "ladderstep";
                 _defn_sig :=
                   {|
                     _defn_args := ["X1"; "X2"; "Z2"; "X3"; "Z3"];
                     _defn_rets := []
                   |};
                 _defn_impl := ladderstep_body;
                 _defn_gallina := ladderstep_gallina;
                 _defn_calls := [mul; add; sub; square; scmula24]
               |} : _defn_spec
The term "cmd" has type
"(string ->
  Semantics.trace ->
  map.rep ->
  list word.rep ->
  (Semantics.trace -> map.rep -> list word.rep -> Prop) -> Prop) ->
 Syntax.cmd ->
 Semantics.trace ->
 map.rep ->
 map.rep -> (Semantics.trace -> map.rep -> map.rep -> Prop) -> Prop"
which should be Set, Prop or Type.

Command exited with non-zero status 1
```

andres-erbsen:
Rupicola builds in [rupicola CI](https://github.com/mit-plv/rupicola/runs/4411579287?check_suite_focus=true#step:7:40) but not [fiat-crypto CI](https://github.com/mit-plv/fiat-crypto/runs/4411612547?check_suite_focus=true#step:7:5503). Even [commit hash of rupicola on this branch](https://github.com/mit-plv/fiat-crypto/pull/1066/files#diff-6df978616e068113dfa0ee6b2a1567f7dac8fe7f7d9006c34f12d4c1adfb3f9e) matches [that tested on rupicola CI](https://github.com/mit-plv/rupicola/runs/4411579287?check_suite_focus=true#step:4:384).  
[commit hash of fiat-crypto](https://github.com/mit-plv/fiat-crypto/pull/1066/commits/b520cb6b19d4d2dde83446858683c027df438c1d) on this branch does not seem to match that [that tested by fiat-crypto CI](https://github.com/mit-plv/fiat-crypto/runs/4411612547?check_suite_focus=true#step:2:607) but the rupicola sumodule tested [still matches](https://github.com/mit-plv/fiat-crypto/runs/4411612547?check_suite_focus=true#step:2:509).

I'm confused and will choose sleep.

JasonGross:
> commit hash of fiat-crypto on this branch does not seem to match that that tested by fiat-crypto CI 

This, at least, is to be expected, because GH Actions tests a merge of the PR with master by using the /merge url as the reference.

JasonGross:
Next bit of the puzzle: even though the error message location claims rupicola, the error is actually in `src/Bedrock/Group/ScalarMult/LadderStep.v`.  We should report this as a bug, my guess (untested) is that if there's an error in tactics in terms inside notations, the error gets reported at the location of the notation definition rather than at the location of the notation use.  (I claim this is a bug rather than a design issue because I think the spec of error locations should be "the top-level line of Coq code", because that's what's needed for the bug minimizer to work.). I can try reporting this tomorrow.

andres-erbsen:
Ah, I think the type this notation returns changed from cmd to func.

JasonGross:
Issue reported as https://github.com/coq/coq/issues/15298

Are you able to reproduce the error locally?

andres-erbsen:
Yes; I think https://github.com/mit-plv/rupicola/pull/23 resolved it and now I am just accepting new build outputs and such

PR #1068, opened 2021-12-08
Compile x25519 and inversion

andres-erbsen:
@jadephilipoom @ashley-lin @DIJamner any objections to merging this if it happens to pass? I know the code here isn't exactly "ready", but it's probably easier to develop it in the tree rather than separately.


DIJamner:
Looks fine to me. I'm going to have felem_cswap done soon, but it's fine to use a stub for now and add it after.

jadephilipoom:
Fine by me!

PR #1073, opened 2021-12-09
`allow(non_camel_case_types)` should be inner attribute

brycx:
Changes `allow(non_camel_case_types)` to be inner attribute, instead of outer. Before this, it only applied to the item immediately after declaration (all items thereafter generated warnings). Inner attributes are instead module-level.

### DIFF
--- a/fiat-rust/src/curve25519_32.rs
+++ b/fiat-rust/src/curve25519_32.rs
@@ -13,7 +13,7 @@
 //!   balance = [0x7ffffda, 0x3fffffe, 0x7fffffe, 0x3fffffe, 0x7fffffe, 0x3fffffe, 0x7fffffe, 0x3fffffe, 0x7fffffe, 0x3fffffe]
 
 #![allow(unused_parens)]
-#[allow(non_camel_case_types)]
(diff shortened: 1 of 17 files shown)


JasonGross:
The generated files need to be updated too, with `make rust-files`.  If you prefer, I could also run the command and push to your branch

JasonGross:
Alternatively, if you rebase on top of #1074, we should be able to get the CI to generate the files for us.

brycx:
Rebased upon #1074, also for the sake of testing this cool addition. Action workflows need to be approved though.

JasonGross:
Generated files available [here](https://github.com/mit-plv/fiat-crypto/suites/4608441697/artifacts/124685422), you should be able to just extract them to overwrite the existing files in `fiat-*` folders and commit the changes.  (And please also rebase on latest master, since apparently github is sad about the "rebase and merge button" when used on PR dependencies)

brycx:
Thanks for providing the generated files @JasonGross. Rebased and included new Rust files.

JasonGross:
Windows fails due to OOM, unrelated to this PR

PR #1075, opened 2021-12-09
[rust] Add conditional `no_std` to Cargo.toml

### LINKED ISSUE #1067: rust: consider labeling library as `no_std`
As far as I can tell, there shouldn't be any issue with adding Rust's `#![no_std]` declaration to the generated library code. This would enable fiat-crypto to be used in embedded environments without an allocator or the full Rust standard library.

This change also wouldn't affect any existing code.

If you don't want to commit the library to never adding allocating code, there's also the option to feature-gate the allocating code behind a default feature (usually named `std`) with something like:

```rust
#![cfg_attr(not(std), no_std)]
```

The above only labels the library as `no_std` when the `std` feature is not turned on (and it would be turned on by default).

JasonGross:
Adds `#![cfg_attr(not(feature = "std"), no_std)]` to the Rust library,
as suggested in https://github.com/mit-plv/fiat-crypto/issues/1067

Fixes #1067

@vlmutolo @brycx Does this look good?

### DIFF
--- a/fiat-rust/Cargo.toml
+++ b/fiat-rust/Cargo.toml
@@ -12,3 +12,7 @@ license = "MIT OR Apache-2.0 OR BSD-1-Clause"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+
+[features]
+default = [ "std" ]
+std = []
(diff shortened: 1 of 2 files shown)


brycx (approved):
LGTM.

PR #1078, opened 2021-12-11
concretize and implement sctestbit + partial proof

andres-erbsen:
I removed a lot of indirection in the process because it didn't seem to be helping any; we can put it back if we need it.


andres-erbsen:
:tada: all CI checks started at once! But the one for coq master does not seem to show up in the list in the PR, although I can see it if I click on the Ubuntu one first.

JasonGross:
Master is the first one in the list for me:
image

PR #1085, opened 2021-12-17
Bump bedrock2: parsing, printing, LittleEndianList

andres-erbsen:
~~After https://github.com/mit-plv/rupicola/pull/32 .~~


andres-erbsen:
@dependabot merge would be too convenient to be true, right?

JasonGross:
If we want to protect the master branch and manually list out every single status check that has to pass, then we can use GitHub's automerge feature.  If we trust bots with more permissions, we can set up something like mergy a la https://github.com/isaacs/github/issues/586#issuecomment-729076296.  Or we can wait to get into the beta for merge queues that I signed us up for a month ago when I heard about them from the Coq devs (https://mobile.twitter.com/natfriedman/status/1453394825269571591, https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/using-a-merge-queue)

PR #1090, opened 2021-12-27
Attempt to fix windows build

JasonGross:
Since opam 2.1, depext is integrated in install, so we try disabling it.
See https://opam.ocaml.org/blog/opam-2-1-0/


JasonGross:
Seems to be an upstream issue: https://github.com/ocaml/setup-ocaml/issues/361 ...

PR #1091, opened 2021-12-27
Autogenerate fiat-rust/src/lib.rs

### LINKED ISSUE #1089: Rust: Exposing the Poly1305 modules
Is there a reason that the current `poly1305_32` and `poly1305_64` modules are not exposed publicly, such that they are available from the crate?

If not, I can create a quick PR to add this (or any other missing modules, if existent, for that matter).

JasonGross:
Fixes #1089, exposing poly1305


brycx (commented):
As mentioned, I know little about working with Makefiles, but FWIW it looks good to me.

PR #1094, opened 2022-01-12
Adapt to Coq PR #15434: even erasable subterms of fixpoints must be guarded

herbelin:
Hi,

The fixpoint `show_expr_lines_gen` in file `Stringification/Language.v` has the following form once the helpers it depends on are unfolded:
```coq
Fixpoint show_expr_lines_gen ... e ... :=
  match e with
  | expr.Abs s d f =>
      fix show_eta_abs_cps' ... e ... :=
        match e with
        | expr.Abs s d f => ψ (show_eta_abs_cps' ... f ...)
        | _ => show_expr_lines_gen ... e ...
        end (expr.Abs s d f)
  | _ => ...
  end
```

This includes in the default case of `show_eta_abs_cps'` an erasable recursive call to `show_expr_lines_gen` which is not guarded and which thus may loop for a specific reduction strategy that unfolds `show_eta_abs_cps'` one step, then reduces the expression in the default case of the `match (expr.Abs s d f) with  ... end` before contracting the `match` itself. PR coq/coq#15434 forbids it.

This fiat-crypto PR reformulates `show_expr_lines_gen` by forcing the evaluation of the `match (expr.Abs s d f) with  ... end` in advance as in:
```coq
Fixpoint show_expr_lines_gen ... e ... :=
  match e with
  | expr.Abs s d f =>
      ψ (fix show_eta_abs_cps' ... e ... :=
        match e with
        | expr.Abs s d f => ψ (show_eta_abs_cps' ... f ...)
        | _ => show_expr_lines_gen ... e ...
        end f)
  | _ => ...
  end
```

If this reformulation is ok to you, it can be merged as soon as now.

### DIFF
--- a/src/Stringification/Language.v
+++ b/src/Stringification/Language.v
@@ -703,32 +703,38 @@ Module Compilers.
           Context {var}
                   (of_string : forall t, string -> option (var t))
                   (k : forall t, @API.expr var t -> type.for_each_lhs_of_arrow (fun t => (Level -> string) * ZRange.type.option.interp t)%type t -> positive -> (positive * (Level -> list string)) * ZRange.type.base.option.interp (type.final_codomain t)).
+          Definition show_eta_abs_aux
+              (aux : forall t (idx : positive) (e : @API.expr var t),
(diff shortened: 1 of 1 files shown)


JasonGross:
Thanks!

PR #1103, opened 2022-01-15
Add `encode_word` and `copy` to PushButtonSynthesis

JasonGross:


JasonGross:
I'm confused how this breaks things :-(

PR #1113, opened 2022-02-02
Word-by-word Montgomery with new Bedrock2 Synthesis Pipeline

RasmusHoldsbjergCSAU:
So, I finally got around to this; using the new Bedrock2 synthesis pipeline for word-by-word Montgomery arithmetic, as discussed in #944.

I suggest the following small change for sharing as many parts of the pipeline between Montgomery and Solinas arithmetic:
Keeping the type of ``bounds``, previously ``list (option zrange)``, and the function used to evaluate if a ``list Z`` is in bounds, previously ``list_Z_bounded_by``, abstract in the first parts of the pipeline.
These are different for Montgomery and Solinas arithmetic, so may be injected at a later point, when the pipeline diverges.

A comprehensive list of changes:

In Bedrock/Specs/Field.v
- Added specs of to- and from Montgomery and selectznz
- feval_alt added to FieldRepresentation. This is an alternative field evaluation used for distinguishing between evaluating with or without pulling the argument out of the Montgomery domain. For Solinas arithmetic, this should be the same as feval

In Bedrock/Field/Interface/ Representation.v
- Bounds are kept abstract as noted above

In Bedrock/Field/Synthesis/New/Signature.v
- Bounds are still kept abstract, and a number of hypothesies are rephrased to take account of that. Proofs are also slightly adjusted for the same reason.
- Section SelectZnZ added

In Bedrock/Field/Synthesis/New/UnsaturatedSolinas.v
- FieldRepresentation Instantiated with ``bounds := (list (option zrange))``

Bedrock/Field/Synthesis/New/WordByWordMontgomery.v added
- Similar to New/UnsaturatedSolinas, but for Montgomery arithmetic.

A note on ``selectZnZ``; I found it more convenient to use a spec of ``selectZnZ``, that did not evaluate its argument, so added this in PushButtonSynthesis/WordByWordMontgomery.v and COpeartionSpecifications.v as ``selectznz_spec_alt``. To keep consistency with other parts of the project, I left the original one in place, although I do not see any reason to prefer it, as can be easily inferred from the alternative version (whereas the converse is not true).
Also, I did not add ``selectZnZ`` to UnsaturatedSolinas, but i guess the relevant lemmas can be reused.

In Bedrock/Group/AdditionChains.v
Added strings for ``from_mont``, ``to_mont`` and ``select_znz`` in Group/AdditionChains.v, since FieldParameters has changed.

An example of using the pipeline is included in p224_64_new.v. This is similar to End2End/X25519/Field25519.v.

Let me know of any comments and suggestions.


RasmusHoldsbjergCSAU:
I'm guessing @jadephilipoom will have some opinions of the suggested changes.

src/Bedrock/Field/Interface/Representation.v
```diff
@@ -23,31 +23,41 @@ Section Representation.
...
     fun ws =>
       F.of_Z _ (Positional.eval weight n (map word.unsigned ws)).
 
   Definition eval_bytes : list byte -> F M_pos :=
     fun bs =>
       F.of_Z _ (Positional.eval
                            (ModOps.weight 8 1)
-                           n_bytes
-                           (map byte.unsigned bs)).
+                           n_bytes (eval_bytes_transformation
```
jadephilipoom, line 49: Is the `eval_bytes_transformation` actually needed? It looks like it gets instantiated with `id`.

src/Bedrock/Specs/Field.v
```diff
@@ -43,6 +48,15 @@ Class FieldRepresentation
...
   { felem := list word;
     feval : felem -> F M_pos;
+    (*  
+        TODO: do we actually need this?
+        feval_alt is used only by word-by-word Montgomery,
+        as it needs to distinguish between evaulation with/without
+        pulling the argument out of Montgomery domain.
+        For Solinas, this is the same as feval
+    *)
+    feval_alt : felem -> F M_pos;
```
jadephilipoom, line 58: One potentially crazy suggestion I have to eliminate the need for `feval_alt`: we could use two separate field representations for the Montgomery domain and non-Montgomery values. Most arithmetic in `WordByWordMontgomery.v` would use the specs and representation from the Montgomery-domain representation, but `to_bytes` and `from_bytes` would use the non-Montgomery one. Additionally, the `to_mont` and `from_mont` would then just be special cases of non-Montgomery multiplication, so the `to_mont` and `from_mont` could be removed from the high-level field specification (avoiding a bit of a leaky abstraction).

I realize that it would be a lot of work to try this out, though, so I'm still fine with merging with two `feval`s. If we do keep the two `feval`s, though, I think a name change might be a good idea. To me, `feval_alt` sort of suggests that that one is the one for the "alternate" internal representation, i.e. Montgomery; it took me a while to figure out it was the other way around. Perhaps `feval_raw`, to indicate it shows the unmodified internal representation?

src/Bedrock/Specs/Field.v
```diff
@@ -241,6 +255,41 @@ Section FunctionSpecs.
...
+    Local Notation bit_range := {|ZRange.lower := 0; ZRange.upper := 1|}.
+
+    Instance spec_of_selectznz  : spec_of select_znz :=
+    fnspec! select_znz (pout pc px py : word) / out Rout Rx Ry x y,
+    {
+        requires tr mem :=
+        (*Are this list_Z_bounded_by's necessary? Can this information be inferred from FElem?*)
+        list_Z_bounded_by (@max_bounds width felem_size_in_words) (List.map word.unsigned x) /\
+        list_Z_bounded_by (@max_bounds width felem_size_in_words) (List.map word.unsigned y) /\
+        list_Z_bounded_by (@max_bounds width felem_size_in_words) (List.map word.unsigned out) /\
```
jadephilipoom, line 268: No, these shouldn't be needed; in fact, `list_Z_bounded_by (@max_bounds width felem_size_in_words) (List.map word.unsigned x)` should always be true, because `word.unsigned` always produces a value `< 2^width`, and `max_bounds` says that the number is `< 2^width`. It's definitely proven in a few places in the unsaturated solinas code, and I think there's a general-ish lemma somewhere; try Searching for `List.map` and `word.unsigned`, and possibly unfolding `max_bounds` a bit.

jadephilipoom (commented):
Really excited to see this work happen! Thanks for the hard work :slightly_smiling_face:

JasonGross:
> A note on `selectZnZ`; I found it more convenient to use a spec of `selectZnZ`, that did not evaluate its argument, so added this in PushButtonSynthesis/WordByWordMontgomery.v and COpeartionSpecifications.v as `selectznz_spec_alt`. To keep consistency with other parts of the project, I left the original one in place, although I do not see any reason to prefer it, as can be easily inferred from the alternative version (whereas the converse is not true).
> Also, I did not add `selectZnZ` to UnsaturatedSolinas, but i guess the relevant lemmas can be reused.

I don't see any reason to leave the original spec, and would be a fan of just removing the evals from the original spec.  @andres-erbsen @jadephilipoom does this seem like the right decision to both of you too?

(Also, btw, I didn't see any changes to PushButtonSynthesis, only Cooperation specifications; did you forget to commit that change?)

RasmusHoldsbjergCSAU:
Thanks for the quick response; I made some adjustments based on your comments:

As per Jade's suggestion, removed the feval_alt from FieldRepresentation to instead use two different representation for word-by-word Montgomery arithmetic; one with raw evaluation and one where the argument is pulled out of the Montgomery domain. These are `field_representation` and `field_representation_raw` in Synthesis/New/WordByWordMontgomery.v.

Removed strings `from_mont` and `to_mont` from FieldParameters and instead made these implicit parameters for `spec_of_from_mont` and `spec_of_to_mont`, which are still in Specs/Field.v. This does not really change anything, but removes the need to instantiate these strings when using the pipeline for Solinas arithmetic.

Changed the spec of `select_ZnZ` to not do evaluation. This change propagated only to PushButtonSynthesis/Primitives.v, where it was an easy fix.

And a number of minor cosmetic changes.


The project does compile with these changes on by machine, but I see that the CI failed some checks. I guess this could be because I forgot to commit the changes to PushButtonSynthesis?

Also I have not comitted any changes to _CoqProject or Makefile, but I am not really sure what the CI needs, so let me know if anything is missing.

src/PushButtonSynthesis/Primitives.v
```diff
@@ -1045,10 +1045,10 @@ Section __.
   Lemma Wf_cmovznz_by_mul s' res (Hres : cmovznz_by_mul s' = Success res) : Wf res.
   Proof using Type. prove_pipeline_wf (). Qed.
 
-  Lemma selectznz_correct limbwidth res
+  Lemma selectznz_correct res
         (Hres : selectznz = Success res)
-    : selectznz_correct (weight (Qnum limbwidth) (QDen limbwidth)) n saturated_bounds (Interp res).
-  Proof using Type. prove_correctness I. Qed.
+    : selectznz_correct saturated_bounds (Interp res).
+  Proof using Type. prove_correctness I. eapply select_eq; eauto. Qed.
```
JasonGross, line 1051: Please add select_eq to the `Hint Rewrite` above (I think that should be sufficient?) rather than using it manually here.

src/PushButtonSynthesis/WordByWordMontgomery.v
```diff
@@ -1032,10 +1032,13 @@ Section __.
...
 
   Local Opaque Pipeline.BoundsPipeline. (* need this or else [eapply Pipeline.BoundsPipeline_correct in Hres] takes forever *)
+
   Lemma selectznz_correct res
         (Hres : selectznz = Success res)
-    : selectznz_correct machine_wordsize n saturated_bounds (Interp res).
-  Proof using curve_good. Primitives.prove_correctness use_curve_good. Qed.
+    : selectznz_correct saturated_bounds (Interp res).
+  Proof using curve_good.
+    Primitives.prove_correctness use_curve_good. eapply select_eq; eauto.
```
JasonGross, line 1040: Again, please add `select_eq` to the `Hint Rewrite` above

JasonGross (changes requested):
When you add a new file, you need to run `make update-_CoqProject`.  I expect all the generated files will need to be regenerated, because the comment on selectznz will be different; if you rebase on master (to get b61ebb14caecd5d012bffcb414f1d3beb74a3377) then the CI will generate the updates for you and you can just download the updated files; otherwise you can run `make` and just wait a long time while it regenerates the files.  I expect that you'll also need to rebase once more on master; I'm expecting dependabot to make a PR updating the rewriter to be compatible with master soon, and I expect this won't build on Coq master until that PR is merged.  (I might be wrong though.)

src/PushButtonSynthesis/Primitives.v
RasmusHoldsbjergCSAU: Adding to `Hint Rewrite` didn't work, I instead added it to `prove_correctness'`.

src/PushButtonSynthesis/WordByWordMontgomery.v
```diff
@@ -1032,10 +1032,13 @@ Section __.
...
   Local Opaque Pipeline.BoundsPipeline. (* need this or else [eapply Pipeline.BoundsPipeline_correct in Hres] takes forever *)
+
   Lemma selectznz_correct res
         (Hres : selectznz = Success res)
-    : selectznz_correct machine_wordsize n saturated_bounds (Interp res).
-  Proof using curve_good. Primitives.prove_correctness use_curve_good. Qed.
+    : selectznz_correct saturated_bounds (Interp res).
+  Proof using curve_good.
+    Primitives.prove_correctness use_curve_good.
+  Qed.
```
JasonGross, line 1041: Slight preference for minimizing the diff here
```suggestion
  Proof using curve_good. Primitives.prove_correctness use_curve_good. Qed.
```

src/PushButtonSynthesis/Primitives.v
```diff
@@ -241,9 +241,10 @@ Ltac prove_correctness' should_not_clear use_curve_good :=
     repeat first [ reflexivity
                  | progress autorewrite with interp_extra interp_gen_cache
                  | progress autorewrite with push_eval
+                 | progress erewrite select_eq
```
JasonGross, line 244:
```suggestion
                 | erewrite select_eq
```
`progress` is redundant

src/PushButtonSynthesis/Primitives.v
```diff
@@ -241,9 +241,10 @@ Ltac prove_correctness' should_not_clear use_curve_good :=
...
                  | progress autounfold with push_eval
                  | progress autorewrite with distr_length in *
-                 | typeclasses eauto ]
+                 | typeclasses eauto | eauto with zarith ]
```
JasonGross, line 247: Slight preference for an extra newline here.  Strong preference for either `progress` or `solve`
```suggestion
                 | typeclasses eauto
                 | solve [ eauto with zarith ] ]
```

RasmusHoldsbjergCSAU:
As far as I can tell, dependabot made an update of Rewriter on february 10th, but this PR still won't build. Is there still a pending update that I should wait for?

On a related note; I noticed that the ```Bedrock/Specs/Field.v``` depends on ```PrimeFieldTheorems.v```, allowing only fields of prime order. Is there any reason to prefer this?

For me, it would be more convenient to have the specs depend on arbitrary finite fields, rather than just those of prime order.

I suggest the following changes:
Splitting the ```Bedrock/Specs/Field.v``` file in two; one that is similar to the current version, but using abstract fields from ```Algebra/Hierarchy.v``` instead of prime order fields from ```PrimeFieldTheorems.v```. The other file would specialize the specs from the first one to a field from ```PrimeFieldTheorems.v```, much like the current version, and be used in the synthesis pipeline.

If you are happy with these changes, I will gladly make them and include them in this PR, or make a new one, as you prefer.

spitters:
@JasonGross @jadephilipoom is there anything blocking this one, or is it just a glitch in dependabot?

JasonGross:
It looks like there's a merge conflict that needs to be resolved in `src/Bedrock/Field/Synthesis/New/UnsaturatedSolinas.v`, and according to the CI, you need to run `make accept-Crypto.Bedrock.Group.ScalarMult.LadderStep.ladderstep_body` because the body of ladderstep changed.

As far as I'm concerned, the changes to files outside of `Bedrock/` are good to go (and if you want to split them off, I'm even happy to merge them as soon as the CI passed); I'll defer to @jadephilipoom @andres-erbsen on the `Bedrock/` changes.

JasonGross:
I imagine, btw, that you'll also need to regenerate all the generated files, which will now have different documentation comments for `select`

jadephilipoom (approved):
I'm happy with this once CI passes.

RasmusHoldsbjergCSAU:
I resolved the conflict in ```Synthesis/New/UnsaturatedSolinas.v``` and ran ```make accept-Crypto.Bedrock.Group.ScalarMult.LadderStep.ladderstep_body```.
The latest commit also includes the generated files with the updated comments for select.

JasonGross:
Hm it looks like the CI never ran, let me try closing and reopening to get the CI to run

src/PushButtonSynthesis/Primitives.v
JasonGross: It looks like the change adding it to `prove_correctness'` got undone?

JasonGross (changes requested):

JasonGross:
(Sorry it's taking so long to get this merged)

JasonGross (approved):
The code outside of Bedrock/ LGTM again, let's hope CI passes and we can merge

src/Bedrock/Field/Synthesis/Examples/p224_64_new.v
```diff
@@ -0,0 +1,173 @@
...
+      (field_parameters_prefixed
+         M ((a - F.of_Z _ 2) / F.of_Z _ 4)%F prefix).
+  Defined.
+
+  Definition to_mont_string := prefix ++ "to_mont".
+  Definition from_mont_string := prefix ++ "from_mont".
+
+  (* Call fiat-crypto pipeline on all field operations *)
+  Instance p224_ops : @word_by_word_Montgomery_ops from_mont_string to_mont_string _ _ _ _ _ _ _ _ _ _ _ (WordByWordMontgomery.n m machine_wordsize) m.
+  Proof using Type. Time constructor; make_computed_op. Defined.
```
JasonGross, line 51: This proof seems to make coqchk / validate OOM.  Any idea what's going on here @RasmusHoldsbjergCSAU @jadephilipoom ?

PR #1123, opened 2022-02-10
adding `shl shlx` to mul * imm

dderjoel:
This adds two example files using different instructions to multiply by constants using lea / shl / shlx.
Also it enables the parser to parse those new instructions, but it is still unable to unify.
I think this is because it does not know that `a + a == 2*a == 2<<1` and `a * 19 == a + 2 * (8a + a)`, `a * 2 * 2 == a * 4` ...
Also I'd need some pointers to parse `MEM` with scaled index, i.e. `[ reg + N * reg ]`; N being 2, 4, 8.


JasonGross:
> Also I'd need some pointers to parse `MEM` with scaled index, i.e. `[ reg + N * reg ]`; N being 2, 4, 8.

First, you need to update the representation of memory addresses to support this at
https://github.com/mit-plv/fiat-crypto/blob/e612a3222ba85e088db996fdf4cd69fd84b638ae/src/Assembly/Syntax.v#L26
Presumably you want to change `mem_extra_reg` to be `option (Z * REG)`?

Then you want to change printing of this at
https://github.com/mit-plv/fiat-crypto/blob/e612a3222ba85e088db996fdf4cd69fd84b638ae/src/Assembly/Parse.v#L268-L286

Then you want to update the parser combinator expression at
https://github.com/mit-plv/fiat-crypto/blob/e612a3222ba85e088db996fdf4cd69fd84b638ae/src/Assembly/Parse.v#L173-L189
probably changing
https://github.com/mit-plv/fiat-crypto/blob/e612a3222ba85e088db996fdf4cd69fd84b638ae/src/Assembly/Parse.v#L187
to something like
```coq
((strip_whitespace_around ("+" ||->{id} "-") ;; (parse_Z_arith_strict ;;L strip_whitespace_around "*")? ;; parse_REG)?) ;; 
```
Note that it's important to not have overlapping options for parsing whitespace, otherwise we end up with exponentially many possible parses.  Roughly this line says "start with a + or - (and just give the `string + string` result transformed by `id`), then optionally (postfix `?`) take a `Z` arithmetic expression followed by a `*` with maybe whitespace (and drop the `*` and whitespace in the parse result; the L in `;;L` means 'only care about the left result'), and then parse a register".
Then you need to change
https://github.com/mit-plv/fiat-crypto/blob/e612a3222ba85e088db996fdf4cd69fd84b638ae/src/Assembly/Parse.v#L178
to pull out the + or - and the optional Z and the register and combine them into the right representation.

Let me know if you have any trouble with this.

JasonGross:
> I think this is because it does not know that `a + a == 2*a == 2<<1` and `a * 19 == a + 2 * (8a + a)`, `a * 2 * 2 == a * 4` ...

Do you mean `a << 1` instead of `2 << 1`?

@andres-erbsen do you have thoughts on encoding this into the canonicalizer?  Presumably we want a variant of
https://github.com/mit-plv/fiat-crypto/blob/77731c496cc901b71887374d4bee2cd1c3f1d844/src/Assembly/Symbolic.v#L1137-L1152
that uses something like [Haskell's `groupBy`](https://hoogle.haskell.org/?hoogle=groupBy) on node equality or something?  Maybe we can add a "groups_by" typeclass like associative or identity, which says for each associative commutative operation how to turn an n-fold operation on identical arguments into a single operation on `n` and the argument?  I guess we need a two-step version of grouping, where we decompose each argument that's a product into the number of copies that it is?

JasonGross:
> `2*a == a<<1`

This one should be easy, just add a pass that unconditionally rewrites shifts by constants into multiplications

> `a * 2 * 2 == a * 4` 

Are you sure this one isn't already handled?  I think it should be handled already.  (If not, what's the error message?)

dderjoel:
> > `a * 2 * 2 == a * 4`
> 
> Are you sure this one isn't already handled? I think it should be handled already. (If not, what's the error message?)

see commit 1ee484e; 
end of the error message is:
```
Unable to unify: [inr [2110, 2114, 2113, 2045, 2055, 2065, 2075, 2086, 2096]] == [inr [569, 574, 576, 485, 499, 513, 527, 541, 558]]
Could not unify the values at index 0: [2110, 2114, 2113, 2045, 2055, 2065, 2075, 2086, 2096] != [569, 574, 576, 485, 499, 513, 527, 541, 558]
index 0: 2110 != 569
(slice 0 58, [2103]) != (slice 0 58, [559])
index 0: 2103 != 559                                                                                                                                     nul
(add 64, [2013, 2100]) != (add 64, [251, 554])
index 0: 2013 != 251
(slice 0 58, [2008]) != (slice 0 58, [235])                                                                                                              nul
index 0: 2008 != 235
(add 64, [106, 126, 150, 214, 1969]) != (add 64, [90, 106, 126, 150, 214])
index 0: 106 != 90
(mul 64, [2, 5, 10, 10]) != (mul 64, [3, 4, 10, 10])
index 0: 2 != 3
(old 64 2, []) != (old 64 3, [])
(old 64 2, []) != (old 64 3, [])
Operation mismatch: old 64 2 != old 64 3
2 is a special value no longer present in the symbolic machine state at the end of execution.
3 is a special value no longer present in the symbolic machine state at the end of execution.

Fatal error: exception Failure("Synthesis failed")
```
I don't really understand the `old` part; is that some kind of helper ~thing~ operation?

> Do you mean `a << 1` instead of `2 << 1`?

yes, that's what I meant. My bad.

For the rest, I'll have a go on the Memory-adjustments, and for the `*2 == <<1` I think @andres-erbsen will help out, or is that something, that I would want to  look into, too ?

JasonGross:
> I don't really understand the `old` part; is that some kind of helper ~thing~ operation?

Could you rebase on top of https://github.com/mit-plv/fiat-crypto/pull/1125 and give me the error message after this?  The `old` part is a red herring; roughly this means that you're comparing non-equal input arguments, but the real issue with `(add 64, [106, 126, 150, 214, 1969]) != (add 64, [90, 106, 126, 150, 214])` is not that `90 != 106`, but that `1969 != 90`.  #1125 compares argument lists of commutative operations modulo commutativity, so we should get a better error message here.

JasonGross:
@dderjoel
File "./src/Assembly/WithBedrock/Semantics.v", line 158, characters 0-7741:
Error: Non exhaustive pattern-matching: no clause found for patterns shl, _

dderjoel:
> Could you rebase on top of #1125 and give me the error message after this? 
sure thing

```
Unable to unify: [inr [2110, 2114, 2113, 2045, 2055, 2065, 2075, 2086, 2096]] == [inr [569, 574, 576, 485, 499, 513, 527
Could not unify the values at index 0: [2110, 2114, 2113, 2045, 2055, 2065, 2075, 2086, 2096] != [569, 574, 576, 485, 49
index 0: 2110 != 569
(slice 0 58, [2103]) != (slice 0 58, [559])
index 0: 2103 != 559
(add 64, [2013, 2100]) != (add 64, [251, 554])
index 0: 2013 != 251
(slice 0 58, [2008]) != (slice 0 58, [235])
index 0: 2008 != 235
(add 64, [106, 126, 150, 214, 1969]) != (add 64, [90, 106, 126, 150, 214])
index 4: 1969 != 90
(mul 64, [3, 4, 1964]) != (mul 64, [3, 4, 10, 10])
index 2: 1964 != 10
(const 4, []) != (const 2, [])
(const 4, []) != (const 2, [])
Operation mismatch: const 4 != const 2

Fatal error: exception Failure("Synthesis failed")
```



> dderjoel
> File "./src/Assembly/WithBedrock/Semantics.v", line 158, characters 0-7741:
> Error: Non exhaustive pattern-matching: no clause found for patterns shl, _
> (Note that bedrock2-dependent files only build when you're running on a sufficiently new version of Coq)

I must admit that I have trouble with the coq-version and ocaml versions on my system (or had, Can't really remember) And I am just happy that I can compile with `SKIP_BEDROCK2=1 make standalone-ocaml`

But what that error means, is that the matches that I've put into the `Symbolic.v` I also need to put into `WithBedrock/Semantics.v`? (Which for me will mostly be copy-adjust-paste-hope-that-it'll-compile)

JasonGross:
> > `a * 2 * 2 == a * 4`
> 
> Are you sure this one isn't already handled? I think it should be handled already. (If not, what's the error message?)

Hrm, I guess we explicitly disable this equivalence
https://github.com/mit-plv/fiat-crypto/blob/e151e211da23827d2b2ce1f99fc6126c8d77d9ea/src/Assembly/Symbolic.v#L1140-L1141
@andres-erbsen Do you know what's up with this?

dderjoel:
> I've created #1127 to enable constfolding on mul, let's see how it works

i rebased on that one, and the test file (from commit 1ee484e77241983088850acb2d3c5db752dd9b36) can now be proven correct.

JasonGross:
> and for the `*2 == <<1` I think @andres-erbsen will help out, or is that something, that I would want to look into, too ?

I think both of us would prefer if you take a stab at it, but either of us can probably help out if need be.

dderjoel:
> Let me know if you have any trouble with this.

I think I have something for the syntax and parsing ; but I am struggling with the parser combinator expression.
I understand that with `; mem_extra_reg := r'` a `MEM` object is build and in this case it is a "mem with an extra register" (as apposed to `[rsp]` or `[rsp + offset]` and for that `mem_extra_reg` I need that second register named `r'`. 
And now I want this object also to have that `z` in there, but I don't know which part is declaring that variable, and how I'd access it, and where to put the data, and where to tell it that `4*RAX` means `(multiply rax 4)`.

When changing `mem_extra_reg` to `option (Z*REG)`, cocq complains in `Assembly/Equality.v`
```
COQC src/Assembly/Equality.v
File "./src/Assembly/Equality.v", line 52, characters 29-46:
Error:
In environment
x : MEM
y : MEM
The term "mem_extra_reg x" has type "option (Z * REG)"
while it is expected to have type "option REG".
```
The code there is: 
`      && (option_beq REG_beq x.(mem_extra_reg) y.(mem_extra_reg))`
and I assume `REG_beq` can check if two regs are the same, but since mem_exra_reg is not a single reg anymore, I need to also check that `z` is equal. A Similar check is done in the line after with `Z.eqb`, but first, how do I access the reg in that line from the `mem_extra_reg`-type?


> > `2*a == a<<1`
> 
> This one should be easy, just add a pass that unconditionally rewrites shifts by constants into multiplications

Where would I add such a pass? That is not in the Assembly-part I assume.

dderjoel:
> You want something like `prod_beq Z.eqb REG_beq` instead of just `REG_beq`, to get boolean equality on `Z * REG` instead of `REG`. You may need some extra underscores for type variables that are not implicit.

okay, using `&& (Prod.prod_beq _ _ Z.eqb REG_beq x.(mem_extra_reg) y.(mem_extra_reg))` cocq complains with 
```
> The term "mem_extra_reg x" has type "option (Z * REG)"
while it is expected to have type "(Z * REG)%type".
```
which, fair enough, makes sense. But the type cannot be non-option (i assume that mean optional) because there might not be an `Z`, let alone a `Z * REG`.
But as I couldn't find an `Prod.prod_option_beq`, I am thinking of getting the `Z` and `reg` and then falling back to `option_beq`.
     
       && (
           xr <- GetReg x.(mem_extra_reg);
           yr <- GetReg y.(mem_extra_reg);
          xz <- RevealConst x.(mem_extra_reg);
           yz <- RevealConst y.(mem_extra_reg);
           req <- (option_beq REG_beq xr yr);
          zeq <- (option_beq Z.beq xz yz);
           req && zeq
       )

But, probably unsurprisingly, that does not work.
`Error: Unknown interpretation for notation "_ <- _ ; _".`

All this syntax is so confusing to me, and since you suggested the Prod.prod_beq, I think that this long approach is not the way to go anyway.

> 
> > Where would I add such a pass? That is not in the Assembly-part I assume.
> 
> It should be added to the list
> 
> https://github.com/mit-plv/fiat-crypto/blob/af618d318907a1c1e0661f7f34bd95184dc93aa7/src/Assembly/Symbolic.v#L1370-L1394

uff. I don't even know where to start there. I am having a hard time understanding what is going on in those passes. Are they all rewriting? I can't find any that looks like anything I'd need for this shl->mul rewrite, But also I don't know what I am looking for. And does it make a difference it it is `shl` or `shlx` ? I think that this is on the `shlZ` level I think, and at that point it should maybe not matter?

JasonGross:
> And does it make a difference it it is `shl` or `shlx` ?

I think you're right that it probably does not matter at this level, but you might need to add something that deals with the flags.

dderjoel:
> to something like
> 
> ```coq
> ((strip_whitespace_around ("+" ||->{id} "-") ;; (parse_Z_arith_strict ;;L strip_whitespace_around "*")? ;; parse_REG)?) ;; 
> ```
> 
> Note that it's important to not have overlapping options for parsing whitespace, otherwise we end up with exponentially many possible parses. Roughly this line says "start with a + or - (and just give the `string + string` result transformed by `id`), then optionally (postfix `?`) take a `Z` arithmetic expression followed by a `*` with maybe whitespace (and drop the `*` and whitespace in the parse result; the L in `;;L` means 'only care about the left result'), and then parse a register". Then you need to change

I think I start to understand `Parse.v`. We define that function which takes arguments `(has_byte, (r, (r', maybe_pm_z)))` and we call that function instantly with the argument list below, separated by `;;`. And MEM is a Record (in c it would be a struct`)

Okay.  We have an issue then.
we need to be able to parse the following:

1. `[reg]` (for `mov reg, [rsp]`)
1. `[reg +/- z ]` (for `mov reg, [rsp +- 0x8]`, spills to memory)
1. `[reg + reg]` (for `lea reg, [reg + reg]`, to calculate *2)
1. `[z * reg]` (for `lea reg, [2/4/8 * reg]`, to calculate *2, *4, *8)
1. `[reg + z * reg]` (for `lea reg, [reg + 2*reg]` to calculate *3, *5, *9)

so the issue is, that the 4. makes `mem_reg` optional, but I am hesitant to change that in the record (I expect severe impacts on the rest of the code to then need to always check if `mem_reg` is actually there.

My question is, if a `parse_REG` for 4. would be able to parse the reg-part although it has that `z*` in front of it. But even then, reg would become `mem_reg` and I'd have no spot to store the `z` part.

So now I'm thinking to change the `Record MEM ` to `Record MEM := { mem_is_byte : bool ; mem_reg : REG ; mem_extra_reg : option REG ; mem_scale: option Z ; mem_offset : option Z }.` with the semantics, that when there is a scale, it always scales the `mem_reg` (to support 4.) Matter of fact, it does not need to be `option Z`, it can be `Z` and if there is nothing in the string it defaults to `1`.

But then I think all this ends up being a bigger change, right? Do you have a better Idea?
Oh and in case there are two `regs`, they do not necessarily need to be identical.

dderjoel:
> > And does it make a difference it it is `shl` or `shlx` ?
> 
> What's the difference between these two (in assembly land)?
> 


> > And does it make a difference it it is `shl` or `shlx` ?
> 
> I think you're right that it probably does not matter at this level, but you might need to add something that deals with the flags. The Ok proof will tell you whether or not you need to care; if you can prove Ok, then you're fine.

Just for completeness-sake:
they both shift left, but they have different syntax and affect the flags differently `shl r/m, imm`, operand 1 is source and destination, operand 2 is the shift count; and it affects the flags, in the sense that the first shifted bit is in the `CF` (or undefined in some cases) `OF` undefined for `imm != 1`, for `imm==1`: `OF= xor first-two-bits-of-original-value`  (but in our case It is enough to just kill both flags), and I don't use `shl m, imm`, I only use `shl r, imm` but I don't think that that matters much here.
`shlx ra, r/m, rb`. does not affect any flags. reads the shift count from rb, masks the least 5/6 bits (depends on the whether r's are 32/64 bit) and reads from `r/m`, shifts by shift count and stores in `ra`.

> > I think that this is on the `shlZ` level I think, and at that point it should maybe not matter?
> 
> What do you mean?

I am not too sure about what `shlZ` is tbh. I thought that from the parsing level (shl, shlx) it gets translated into a logical 'shift something by Z to the left'. Hence I should after the parsing we deal with `shlZ` which I'd parse both into. I mean the correctness-prove should not care if it shifts because of a `shl` or `shlx`, (as long as the Flags are updated accordingly of cause). Hence I thought the re-writer is at `shlZ`-level.

dderjoel:
> Try starting with this (untested):
> 
> ```coq
> Definition shift_to_mul :=
>   fun e => match e with
>     ExprApp ((shl | shlx | shlZ) as o, [e'; ExprApp (const v, [])]) =>
>       let o' := match o with shl | shlx => mul 64%N | shlZ => mulZ | _ => o (* impossible *) end in
>       if Z.eqb v 0
>       then e'
>       else if Z.ltb 0 v
>       then ExprApp (o', [e'; ExprApp (const (2^v)%Z, [])])
>       else e | _ => e end.
> Global Instance shift_to_mul_ok : Ok shift_to_mul_small.
> Proof. t. cbn in *. firstorder (lia + eauto). Qed.
> ```

just adding this errors in 
```
File "./src/Assembly/Symbolic.v", line 1373, characters 14-17:
Error: The constructor shl (in type op) is expected to be applied to
1 argument while it is actually applied to no arguments.
```
And the char's are the `shl` in the first parenthesis.
So what I do understand from this code is, that we are trying to match an operation of type `shl, shlx, shlz` with that below, match and rewrite `shl` and `shlx` to `mul to a 64-bit positive whole number`; and rewrite `shlZ` as `mulZ`. But all that only if the constant is `>0`. Otherwise , if `v==0`, (i.e. shift left by 0) we return the value itself (`then e'`) and if `v<0` we return the original expression `else e`...
Which seems correct to me, too. Is coqc complaining about that 'impossible'-case there, that from that definition, as we know it cant happen, because if `o` is from type `(shl|shlx|zhlZ)` we must have matched earlier. But does coq not know that and expects `o'` to possibly be `o` and  (i.e. possibly `shl`),  but even then in the `then ExprApp (o', [e'; ExprApp (const (2^v)%Z, [])]) ` we still give it arguments. So I dont really see, where we would construct `shl` with no arguments.

JasonGross:
> So now I'm thinking to change the `Record MEM ` to `Record MEM := { mem_is_byte : bool ; mem_reg : REG ; mem_extra_reg : option REG ; mem_scale: option Z ; mem_offset : option Z }.` with the semantics, that when there is a scale, it always scales the `mem_reg` (to support 4.) Matter of fact, it does not need to be `option Z`, it can be `Z` and if there is nothing in the string it defaults to `1`.
> 
> But then I think all this ends up being a bigger change, right? Do you have a better Idea?

This sounds good to me, and I don't have a better idea.  Perhaps the best way to do the parsing is to have five alternatives separated by `||->{id}`, one for each bullet point.  You could combine some of them if you want (making various parts optional), but make sure you don't have overlapping alternatives.

> I am not too sure about what `shlZ` is tbh.

`shlZ` is for a left-shift that doesn't truncate the output to any bit width, i.e., left-shift in Z (rather than, e.g., uint64)







> Is coqc complaining about that 'impossible'-case there, that from that definition, as we know it cant happen, because if `o` is from type `(shl|shlx|zhlZ)` we must have matched earlier.

It's saying that `shl` needs to be `shl bitwidth` because `shl` takes an argument.  Presumably `shlx` does as well, and so you want `(shl _|shlx _|zhlZ)` here and `match o with shl biteidth | shlx biteidth => mul bitwidth | shlZ => mulZ | _ => o (* impossible *) end` later on.  (Your analysis of what the entire function is doing is correct)

dderjoel:
okay, I removed the MEM-changes, I will create another PR for that, lets focus here on the shl-> mul rewrite.
I added that rewrite with the bit widths, but now it complains about an a missing  `shift_to_mul_small` (and I can confirm. It does not exist :) ). Looking at the other Proofs, there is none where the `Global Instance` line has `A` and then `A_small`, they all end with the same name as the rewrite itself is called. Replacing `shift_to_mul_small` with `shift_to_mul.` cocq complains with 

File "./src/Assembly/Symbolic.v", line 1381, characters 46-50:
Error:  (in proof shift_to_mul_ok): Attempt to save an incomplete proof

Also I cannot find any similar Proofs in the any other rewrites. None starts with `Proof. t.`, and all the other ones I could find just say `Qed.` right after it, but I assume we don't want that.
So, (1) I don't understand what the Global Instance line means, and (2) what we need to proof, and (3) how we are trying to achieve that with that line `Proof. t. cbn in *. firstorder (lia + eauto). Qed.` https://github.com/mit-plv/fiat-crypto/pull/1123/commits/6efb428b662a25b5cf35a5a61f3ee160351882e3

src/Assembly/Syntax.v
```diff
@@ -23,7 +23,7 @@ Inductive REG :=
 Definition CONST := Z.
 Coercion CONST_of_Z (x : Z) : CONST := x.
 
-Record MEM := { mem_is_byte : bool ; mem_reg : REG ; mem_extra_reg : option REG ; mem_offset : option Z }.
+Record MEM := { mem_is_byte : bool ; mem_reg : REG ; mem_extra_reg : option  REG  ; mem_offset : option Z }.
```
JasonGross, line 26:
```suggestion
Record MEM := { mem_is_byte : bool ; mem_reg : REG ; mem_extra_reg : option REG  ; mem_offset : option Z }.
```
If you're not changing the record, let's not add spaces

src/Assembly/Parse.v
```diff
@@ -184,7 +186,7 @@ Definition parse_MEM : ParserAction MEM
         (((strip_whitespace_after "byte ")?) ;;
          (strip_whitespace_after "[" ;;R
           parse_REG ;;
-          ((strip_whitespace_around "+" ;;R parse_REG)?) ;;
+          ((strip_whitespace_around "+" ;;R parse_REG)?) ;;
```
JasonGross, line 189:
```suggestion
          ((strip_whitespace_around "+" ;;R parse_REG)?) ;;
```
Also no trailing spaces here

JasonGross:
> but now it complains about an a missing `shift_to_mul_small` (and I can confirm. It does not exist :) ).

This was just me typoing the name, it should just be `shift_to_mul`.

src/Assembly/Symbolic.v
```diff
@@ -1367,6 +1368,18 @@ Proof using Type.
...
+Definition shift_to_mul :=
+  fun e => match e with
+    ExprApp ((shl _ | shlx _ | shlZ) as o, [e'; ExprApp (const v, [])]) =>
+      let o' := match o with shl bitwidth | shlx bitwidth => mul bitwidth | shlZ => mulZ | _ => o (* impossible *) end in
+      if Z.eqb v 0
+      then e'
+      else if Z.ltb 0 v
+      then ExprApp (o', [e'; ExprApp (const (2^v)%Z, [])])
+      else e | _ => e end.
+Global Instance shift_to_mul_ok : Ok shift_to_mul_small.
```
JasonGross, line 1380:
```suggestion
Global Instance shift_to_mul_ok : Ok shift_to_mul.
```

JasonGross:
> File "./src/Assembly/Symbolic.v", line 1381, characters 46-50: Error: (in proof shift_to_mul_ok): Attempt to save an incomplete proof
> 
> Also I cannot find any similar Proofs in the any other rewrites. None starts with `Proof. t.`, and all the other ones I could find just say `Qed.` right after it, but I assume we don't want that. So, (1) I don't understand what the Global Instance line means, and (2) what we need to proof, and (3) how we are trying to achieve that with that line `Proof. t. cbn in *. firstorder (lia + eauto). Qed.` [6efb428](https://github.com/mit-plv/fiat-crypto/commit/6efb428b662a25b5cf35a5a61f3ee160351882e3)

I've pushed a commit to this PR / your master branch which fixes the proof.  This proof is that the rewrite needs to preserve the semantics of the operation.  It turns out my suggested rewrite was wrong, because I forgot to truncate when, e.g., doing 64-bit `x << 0` .  So I've added `slice`.  You may want to move `shift_to_mul` up higher in the list.  In particular, it should probably come before `flatten_associative` so that multiplication gets appropriately flattened.  @andres-erbsen Is there a principled way of ordering the rewrites?

src/Assembly/Symbolic.v
```diff
@@ -1805,6 +1824,17 @@ Definition SymexNormalInstruction (instr : NormalInstruction) : M unit :=
...
+    let cnt := andZ@(cnt, (PreApp (const (Z.of_N s-1)%Z) nil)) in
+    v <- Symeval (shl s@(dst, cnt));
+    _ <- SetOperand dst v;
+    HavocFlags
+  | Syntax.shlx, [dst; src; cnt] =>
+    cnt <- GetOperand cnt;
+    cnt <- RevealConst cnt;
+    let cnt' := andZ@(cnt, (PreApp (const (Z.of_N s-1)%Z) nil)) in
+    v <- Symeval (shl s@(src, cnt));
+         SetOperand dst v
```
JasonGross, line 1837: Nit:
```suggestion
    SetOperand dst v
```

src/Assembly/Symbolic.v
```diff
@@ -1805,6 +1824,17 @@ Definition SymexNormalInstruction (instr : NormalInstruction) : M unit :=
...
+  | Syntax.shl, [dst; cnt] =>
+    let cnt := andZ@(cnt, (PreApp (const (Z.of_N s-1)%Z) nil)) in
+    v <- Symeval (shl s@(dst, cnt));
+    _ <- SetOperand dst v;
+    HavocFlags
+  | Syntax.shlx, [dst; src; cnt] =>
+    cnt <- GetOperand cnt;
+    cnt <- RevealConst cnt;
+    let cnt' := andZ@(cnt, (PreApp (const (Z.of_N s-1)%Z) nil)) in
+    v <- Symeval (shl s@(src, cnt));
```
JasonGross, line 1836: Are you deliberately assigning to `cnt'` and then using `cnt`, or is that a typo?

src/Assembly/WithBedrock/Semantics.v
```diff
@@ -256,6 +256,21 @@ Definition DenoteNormalInstruction (st : machine_state) (instr : NormalInstructi
...
+      let st := HavocFlagsFromResult s st v in
+      let st := if cnt =? 1 then SetFlag st OF (Z.xor (Z.testbit v1 (Z.of_N s-2)) (Z.testbit v1 (Z.of_N s-1))) else st in
+      let st := if cnt <? Z.of_N s then SetFlag st CF (Z.testbit v1 (65-cnt)) else st in
+      HavocFlag st AF)
+  | shlx, [dst; src; cnt] =>
+    v1 <- DenoteOperand sa s st src;
+    cnt <- DenoteOperand sa s st cnt;
+    let v := Z.land (Z.shiftl v1 (Z.land cnt (Z.of_N s-1))) (Z.ones (Z.of_N s)) in
+    st <- SetOperand sa s st dst v;
   | shr, [dst; cnt] =>
```
JasonGross, line 274: Trailing `;` is a syntax error, are you missing something on the last line?  Or do you mean to do
```suggestion
    SetOperand sa s st dst v
  | shr, [dst; cnt] =>
```
?

src/Assembly/Symbolic.v
dderjoel: I think that is a typo. 1835 wants to `and` the cnt with the size (64 bit) -1; which makes sure that we don't shift more than 64-bits.  If I understand that correctly. Then yes, it should be `cnt'`. I'll fix that

dderjoel:
I have amended the test files, such that the tests should pass.
In particular, I remove the new test files, and amended the existing ones to
- show case the const-folding 27fed0e160c21c0c432201d78012b69af4f31b5c
- use `shl`, `shlx` to achieve multiplication 667262da9e8e7c31f3d50a0515d66bbf58986891 and 3c253266fcbef96c022ae99992d018b2bef29660 (2 examples for `shlx`: read from mem and reg)
- use shlx-by-2 to multiply-by-4 to mul-by-2-twice 5911f73

src/Assembly/WithBedrock/Semantics.v
```diff
@@ -256,6 +256,21 @@ Definition DenoteNormalInstruction (st : machine_state) (instr : NormalInstructi
...
       let st := if cnt <? Z.of_N s then SetFlag st CF (Z.testbit v1 (cnt-1)) else st in
       HavocFlag st AF)
+  | shl, [dst; cnt] =>
+    v1 <- DenoteOperand sa s st dst;
+    cnt <- DenoteOperand sa s st cnt;
+    let v := Z.land (Z.shiftl v1 (Z.land cnt (Z.of_N s-1))) (Z.ones (Z.of_N s)) in
+    st <- SetOperand sa s st dst v;
+    Some (if cnt =? 0 then st else
+      let st := HavocFlagsFromResult s st v in
+      let st := if cnt =? 1 then SetFlag st OF (Z.xor (Z.testbit v1 (Z.of_N s-2)) (Z.testbit v1 (Z.of_N s-1))) else st in
```
JasonGross, line 266: I think you want
```suggestion
      let st := if cnt =? 1 then SetFlag st OF (Z.lxor (Z.testbit v1 (Z.of_N s-2)) (Z.testbit v1 (Z.of_N s-1))) else st in
```
Coq has weird names

JasonGross:
> src/Assembly/WithBedrock/Semantics.v#L158
> Non exhaustive pattern-matching: no clause found for patterns shl, []

You should add a `shl, _` error case to `DenoteNormalInstruction`

src/Assembly/WithBedrock/Semantics.v
```diff
@@ -256,6 +256,21 @@ Definition DenoteNormalInstruction (st : machine_state) (instr : NormalInstructi
...
       let st := if cnt <? Z.of_N s then SetFlag st CF (Z.testbit v1 (cnt-1)) else st in
       HavocFlag st AF)
+  | shl, [dst; cnt] =>
+    v1 <- DenoteOperand sa s st dst;
+    cnt <- DenoteOperand sa s st cnt;
+    let v := Z.land (Z.shiftl v1 (Z.land cnt (Z.of_N s-1))) (Z.ones (Z.of_N s)) in
+    st <- SetOperand sa s st dst v;
+    Some (if cnt =? 0 then st else
+      let st := HavocFlagsFromResult s st v in
+      let st := if cnt =? 1 then SetFlag st OF (Z.lxor (Z.testbit v1 (Z.of_N s-2)) (Z.testbit v1 (Z.of_N s-1))) else st in
```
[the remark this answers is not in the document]
JasonGross: No, you want to use `Z.b2z : bool -> Z` to convert.  `%Z` is about disambiguation notations (such as `-`), not about resolving symbols like `Z.testbit` differently.

dderjoel:
I feel stupid... but then also "Why, coq, why?"
`Error: The reference Z.z2b was not found in the current environment.`
or let me ask this: how to convert a number to a boolean ?
`if z =? 0 then false else true` ? or can I shorten that to ... wait, what is the `not`... maybe `Z.neq z 0`?

JasonGross:
> `if z =? 0 then false else true` ? or can I shorten that to ... wait, what is the `not`... maybe `Z.neq z 0`?

I'm not sure if `neq` is a thing, but `negb (z =? 0)` should work. ("b" for Boolean) What do you need this for?

> "Why, coq, why?"

Because Coq evolved ad-hoc over many decades without a real steering committee, and many of the decisions are historical accidents

dderjoel:
> What do you need this for?

What do you need this for?

src/Assembly/WithBedrock/Semantics.v
```diff
@@ -256,6 +256,21 @@ Definition DenoteNormalInstruction (st : machine_state) (instr : NormalInstructi
...
       let st := if cnt <? Z.of_N s then SetFlag st CF (Z.testbit v1 (cnt-1)) else st in
       HavocFlag st AF)
+  | shl, [dst; cnt] =>
+    v1 <- DenoteOperand sa s st dst;
+    cnt <- DenoteOperand sa s st cnt;
+    let v := Z.land (Z.shiftl v1 (Z.land cnt (Z.of_N s-1))) (Z.ones (Z.of_N s)) in
+    st <- SetOperand sa s st dst v;
+    Some (if cnt =? 0 then st else
+      let st := HavocFlagsFromResult s st v in
+      let st := if cnt =? 1 then SetFlag st OF (Z.negb 0 =? (Z.lxor (Z.b2z (Z.testbit v1 (Z.of_N s-2))) (Z.b2z (Z.testbit v1 (Z.of_N s-1))))) else st in
```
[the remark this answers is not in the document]
JasonGross: This is overly complicated, though, just use `xorb` for Boolean xor.  Or use an equality test on Booleans (I think either of `Bool.eqb` or `(x =? y)%bool` should work)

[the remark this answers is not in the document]
dderjoel: is there a `Bool.neqb`?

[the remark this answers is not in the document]
JasonGross: No, just xorb I think.

JasonGross:
I think that specification is the same as
https://github.com/mit-plv/fiat-crypto/blob/d394fb145bd30c37b61d9afd595d22e22beda49c/src/Assembly/WithBedrock/Semantics.v#L292-L294
Maybe you want to just copy that?

dderjoel:
maybe its the same spec, I don't know, and the Intel manual doesn't either. If you're okay with the current `xorb`, I'd just leave it with that.

src/Assembly/WithBedrock/Semantics.v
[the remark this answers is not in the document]
JasonGross: I think negb (xorb a b) is the same as (Bool.eqb a b)...

dderjoel:
Okay, it works on my machine. That is, I can build, and proof the new files correct.
However, I think there still a problem with the BEDROCK2 build (according to the CI build) https://github.com/mit-plv/fiat-crypto/runs/5243721314?check_suite_focus=true#step:7:5065
```
File "./src/Assembly/WithBedrock/SymbolicProofs.v", line 1240, characters 0-4:
Error: Some unresolved existential variables remain
``` 

And I don't know what that means.

JasonGross:
> ```
> File "./src/Assembly/WithBedrock/SymbolicProofs.v", line 1240, characters 0-4:
> Error: Some unresolved existential variables remain
> ```
> 
> 
>     
> 
>   
> 
> And I don't know what that means

If you run `Unshelve.` at the end of such a proof, you should see some remaining goals.  Often these will be things like `nat`, where you used, e.g., `eexists` but then your proof was independent of what number existed.  Other times they'll be side conditions.

dderjoel:
```
File "./src/Assembly/WithBedrock/SymbolicProofs.v", line 1241, characters 0-4:
Error:
 (in proof SymexNornalInstruction_R): Attempt to save an incomplete proof
```
Okay, apparently, only adding an `Unshelve.` is not what you meant by that. What would I need to do instead?

JasonGross:
> Okay, apparently, only adding an `Unshelve.` is not what you meant by that. What would I need to do instead?

`Unshelve` turns "unresolved existential variables" into "incomplete proof".  If you want to see the remaining goals on CI, do `Unshelve. Show.`. Are you having trouble with the file locally?

dderjoel:
> Are you having trouble with the file locally?
No. I'm fine, but also I skip the Bedrock-build. Hence I don't see the error locally.

src/Assembly/WithBedrock/SymbolicProofs.v
```diff
@@ -1237,6 +1237,7 @@ Proof using Type.
     3: enough (0 <= Z.land v3 (Z.of_N n - 1)) by lia; eapply Z.land_nonneg; right.
     1,2,3:pose_operation_size_cases; intuition (subst; cbn; clear; lia). }
   all : fail.
+  Unshelve. Show.
```
JasonGross, line 1240: Let's try something else here (that can stay in the final version).  Rebase this PR on current master, add `Require Import Crypto.Util.Tactics.WarnIfGoalsRemain.` to the top of the file, and then replace these two lines with
```suggestion
  Unshelve. all: shelve_unifiable.
  all: fail_if_goals_remain ().
```
This should display the failing goals on the command line.

JasonGross:
> but also I skip the Bedrock-build. Hence I don't see the error locally.

I'm happy to help you get the bedrock build working, so you can see the error locally, if you'd like.

dderjoel:
> I'm happy to help you get the bedrock build working, so you can see the error locally, if you'd like.

Not needed anymore. Apparently, system updated updated the right things, and it does no longer error out without the `SKIP_BEDROCK2=1`. 
Matter of fact, everything works on my machine now.
```bash 
make standalone-ocaml -j3
./src/ExtractionOCaml/unsaturated_solinas  --no-primitives --no-wide-int --shiftr-avoid-uint1 --output /dev/null --output-asm /dev/null 'p521' '64' '9' '2^521 - 1' carry_square --tight-bounds-mul-by 1.000001 --hints-file ./fiat-amd64/141.31_ratio10675_seed3277703234885713_square_p521.asm
./src/ExtractionOCaml/unsaturated_solinas  --no-primitives --no-wide-int --shiftr-avoid-uint1 --output /dev/null --output-asm /dev/null 'p521' '64' '9' '2^521 - 1' carry_square --tight-bounds-mul-by 1.000001 --hints-file ./fiat-amd64/90.44_ratio11062_seed2001410703584519_square_p521.asm
./src/ExtractionOCaml/unsaturated_solinas  --no-primitives --no-wide-int --shiftr-avoid-uint1 --output /dev/null --output-asm /dev/null 'p521' '64' '9' '2^521 - 1' carry_square --tight-bounds-mul-by 1.000001 --hints-file ./fiat-amd64/90.10_ratio11854_seed600995032503849_square_p521.asm
```
no errors whatsoever.
Lets see what the CI says. (Or do I need to run a different goal to `make standalone-ocaml` ?)

dderjoel:
okay running `make -j2` results in
```coq
COQC src/Bedrock/Field/Translation/Proofs/Cmd.v
COQC src/Bedrock/Field/Translation/Proofs/ValidComputable/Expr.v
WARNING: Remaining goal:
frame : (mem_state -> Prop)
G : (symbol -> option Z)
s : symbolic_state
m : machine_state
HR : (R s m)
s' : symbolic_state
n : N
a : ARG
a0 : ARG
a1 : ARG
E :
(Syntax.operation_size {| Syntax.op := Syntax.shlx; args := [a; a0; a1] |} =
 Some n)
cnt : idx
s0 : symbolic_state
z : Z
x : idx
s1 : symbolic_state
x0 : idx
s2 : symbolic_state
y : idx
s3 : symbolic_state
y0 : idx
s4 : symbolic_state
y1 : idx
s5 : symbolic_state
Hs0 : (R s0 m)
Hls0 : (s :< s0)
v : Z
Hcnt : (eval s0 cnt v)
Hv : (DenoteOperand 64 n m a1 = Some v)
Heqe : (eval s0 (ExprApp (const z, [])) v)
Hs1 : (R s1 m)
Hls1 : (s0 :< s1)
v0 : Z
Hx : (eval s1 x v0)
Hv0 : (DenoteOperand 64 n m a0 = Some v0)
Hs2 : (R s2 m)
Hls2 : (s1 :< s2)
v1 : Z
Hx0 : (eval s2 x0 v1)
Hv1 : (DenoteOperand 64 n m z = Some v1)
Hs3 : (R s3 m)
Hls3 : (s2 :< s3)
Hy : (eval s3 y (Z.of_N n - 1))
Hs4 : (R s4 m)
Hls4 : (s3 :< s4)
Hy0 : (eval s4 y0 (fold_right Z.land (-1) [v1; Z.of_N n - 1]))
Hs5 : (R s5 m)
Hls5 : (s4 :< s5)
Hy1 :
(eval s5 y1
   (Z.land (Z.shiftl v0 (fold_right Z.land (-1) [v1; Z.of_N n - 1]))
      (Z.ones (Z.of_N n))))
m0 : machine_state
Hm0 :
(SetOperand 64 n m a
   (Z.land (Z.shiftl v0 (fold_right Z.land (-1) [v1; Z.of_N n - 1]))
      (Z.ones (Z.of_N n))) = Some m0)
Hs' : (R s' m0)
Hls' : (s5 :< s')
============================
(Z.land (Z.shiftl v0 (Z.land v1 (Z.of_N n - 1))) (Z.ones (Z.of_N n)) =
 Z.land (Z.shiftl v0 (Z.land v (Z.of_N n - 1))) (Z.ones (Z.of_N n)))
File "./src/Assembly/WithBedrock/SymbolicProofs.v", line 1241, characters 7-30:
Error: Tactic failure: A goal remains.

make: *** [Makefile.coq:764: src/Assembly/WithBedrock/SymbolicProofs.vo] Error 1
make: *** Waiting for unfinished jobs....

```

JasonGross:
> Or do I need to run a different goal to `make standalone-ocaml` ?

You want `make coq standalone-ocaml`

JasonGross:
> Edit: All I can read is that it matches with `Syntax.shlx` and that there is a mismatch `v` !=`v1`. And the only place I can find a `v1` in that file is with `inc`/`dec` instructions. So what does this error mean ?

The error just means the current proof script doesn't know how to deal with this code.  We can use `Heqe` to identify v and z, Hv1 to identify v1 and z, and thus we can identify v and v1 and then use reflexivity on the goal.  Maybe @andres-erbsen or @jadephilipoom can suggest whether this goal should be taken care-of in an ad hoc way or if some bit of automation should be extended?

dderjoel:
so do we wait for input from the others? Because, I wouldn't know how I could help on the proof script.

JasonGross:
I'm going to force-push a rebase on current master to your branch, and then see if I can fix the proof

dderjoel:
Wonderful. I just built successfully from this commit, and I can confirm, that it does everything that I need. 
from my side, we can merge and close this.

PR #1127, opened 2022-02-10
Enable const folding for mul asm

JasonGross:
c.f. https://github.com/mit-plv/fiat-crypto/pull/1123#issuecomment-1035216542

Let's see what goes wrong


JasonGross:
I guess nothing goes wrong, let's merge.  Still interested in the historical explanation if anyone has it.  (It was added in https://github.com/mit-plv/fiat-crypto/commit/14bbae7d38ee3d53c0fd4844193b91447460fa12)

Issue #95, opened 2016-11-08
MxDHRepChange.v is broken in Coq 8.4


JasonGross:
The `f_equiv` tactic behaves differently in Coq 8.4 and Coq 8.5/8.6.  We should either roll our own, or stop using it.

JasonGross:
I fixed this a day or two ago

PR #1134, opened 2022-02-17
Mem extention

dderjoel:
this is the PR split from #1123.
I do get an Error in `Parse.v` 

File "./src/Assembly/Parse.v", line 190, characters 129-130:
Error: Syntax error: [term level 70] expected after ';;R' (in [term]).

which I don't understand. Generally, I think I need some feedback on what I am trying to do. I just can't get the syntax right with `(Z * REG)`

Edit: I skipped the Bedrock files for now, because I want to get the others compile correct first.

### DIFF
--- a/fiat-amd64/16232_ratio11549_seed1785685356_square_poly1305.asm
+++ b/fiat-amd64/16232_ratio11549_seed1785685356_square_poly1305.asm
@@ -50,7 +50,11 @@ add r14, rbx; could be done better, if r0 has been u8 as well
 adc r13, 0x0; add CF to r0's alloc
 mov r15, r14;x20, copying x19_0 here, cause x19_0 is needed in a reg. It has those deps: "x20, x21, size: 2", current hard deps: ""
 shrd r15, r13, 43; x20 <- x19_1||x19_0 >> 43
+
+; imul works
+; imul r10, r15, 5
 lea r10, [r15 + 4 * r15]; x22 <- x20 * 5 (by add to itself*4)
+
 lea r8, [ r8 + r10 ]
 bzhi rcx, r14, rbp; x21 <- x19_0 (only least 0x2b bits)
 mov rdx, r8;x24, copying x23 here, cause x23 is needed in a reg. It has those deps: "x24, x25, size: 2", current hard deps: ""
--- a/fiat-amd64/16232_ratio11549_seed1785685356_square_poly1305.asm
+++ b/fiat-amd64/16232_ratio11549_seed1785685356_square_poly1305.asm
@@ -1,0 +1,84 @@
+SECTION .text
+	GLOBAL square_poly1305
+square_poly1305:
+sub rsp, 0x30
+mov rax, 0x1 ; moving imm to reg
+shlx r10, [ rsi + 0x8 ], rax; x4 <- arg1[1] * 0x2 (shlx does not change the flags)
+mov rdx, [ rsi + 0x0 ]; arg1[0] to rdx
+mulx rcx, r11, rdx; x10_1, x10_0<- arg1[0]^2
+imul rdx, [ rsi + 0x10 ], 0x14; x10000 <- arg1[2] * 0x14
+mov r8, [ rsi + 0x10 ]; load m64 arg1[2] to register64
+lea r9, [r8 + 4 * r8]; x1 <- arg1[2] * 5 (by add to itself*4)
+mulx rax, r8, [ rsi + 0x8 ]; x6_1, x6_0<- arg1[1] * x10000 (_0*_0)
+mov rdx, [ rsi + 0x10 ]; arg1[2] to rdx
+mov [ rsp + 0x0 ], rbx; spilling calSv-rbx to mem
+mov [ rsp + 0x8 ], rbp; spilling calSv-rbp to mem
+mulx rbp, rbx, r9; x5_1, x5_0<- arg1[2] * x1 (_0*_0)
+mov rdx, [ rsi + 0x10 ]; load m64 arg1[2] to register64
+mov r9, rdx; load m64 x3 to register64
+shl r9, 0x1; x3 <- arg1[2] * 0x2
+mov rdx, [ rsi + 0x0 ]; arg1[0] to rdx
+mov [ rsp + 0x10 ], r12; spilling calSv-r12 to mem
+mov [ rsp + 0x18 ], r13; spilling calSv-r13 to mem
+mulx r13, r12, r9; x8_1, x8_0<- arg1[0] * x3 (_0*_0)
+mov rdx, [ rsi + 0x8 ]; load m64 arg1[1] to register64
+mov r9, rdx; load m64 x10001 to register64
+shl r9, 0x1; x10001 <- arg1[1] * 0x2
+mov rdx, [ rsi + 0x8 ]; arg1[1] to rdx
+mov [ rsp + 0x20 ], r14; spilling calSv-r14 to mem
+mov [ rsp + 0x28 ], r15; spilling calSv-r15 to mem
+mulx r15, r14, r9; x7_1, x7_0<- arg1[1] * x10001 (_0*_0)
+mov rdx, r10; x4 to rdx
+mulx r9, r10, [ rsi + 0x0 ]; x9_1, x9_0<- arg1[0] * x4 (_0*_0)
+add rbx, r10; could be done better, if r0 has been u8 as well
+adcx r9, rbp
+add r8, r11; could be done better, if r0 has been u8 as well
+adcx rcx, rax
+mov rdx, r8;x12, copying x11_0 here, cause x11_0 is needed in a reg. It has those deps: "x12, x13, size: 2", current hard deps: ""
+shrd rdx, rcx, 44; x12 <- x11_1||x11_0 >> 44
+add r14, r12; could be done better, if r0 has been u8 as well
+adcx r13, r15
+mov r11, 0xfffffffffff ; moving imm to reg
+and r8, r11; x13 <- x11_0&0xfffffffffff
+adox rbx, rdx
+mov rax, 0x0 ; moving imm to reg
+adox r9, rax
+mov rbp, 0x2b ; moving imm to reg
+bzhi r12, rbx, rbp; x18 <- x16_0 (only least 0x2b bits)
+shrd rbx, r9, 43; x17 <- x16_1||x16_0 >> 43
+add r14, rbx; could be done better, if r0 has been u8 as well
+adc r13, 0x0; add CF to r0's alloc
+mov r15, r14;x20, copying x19_0 here, cause x19_0 is needed in a reg. It has those deps: "x20, x21, size: 2", current hard deps: ""
+shrd r15, r13, 43; x20 <- x19_1||x19_0 >> 43
+lea r10, [r15 + 4 * r15]; x22 <- x20 * 5 (by add to itself*4)
+lea r8, [ r8 + r10 ]
+bzhi rcx, r14, rbp; x21 <- x19_0 (only least 0x2b bits)
+mov rdx, r8;x24, copying x23 here, cause x23 is needed in a reg. It has those deps: "x24, x25, size: 2", current hard deps: ""
+shr rdx, 44; x24 <- x23>> 44
+lea rdx, [ rdx + r12 ]
+mov r9, rdx;x27, copying x26 here, cause x26 is needed in a reg. It has those deps: "x27, x28, size: 2", current hard deps: ""
+shr r9, 43; x27 <- x26>> 43
+bzhi r12, rdx, rbp; x28 <- x26 (only least 0x2b bits)
+lea r9, [ r9 + rcx ]
+and r8, r11; x25 <- x23&0xfffffffffff
+mov [ rdi + 0x10 ], r9; out1[2] = x29
+mov [ rdi + 0x8 ], r12; out1[1] = x28
+mov [ rdi + 0x0 ], r8; out1[0] = x25
+mov rbx, [ rsp + 0x0 ] ; pop
+mov rbp, [ rsp + 0x8 ] ; pop
+mov r12, [ rsp + 0x10 ] ; pop
+mov r13, [ rsp + 0x18 ] ; pop
+mov r14, [ rsp + 0x20 ] ; pop
+mov r15, [ rsp + 0x28 ] ; pop
+add rsp, 0x30 
+ret
+; cpu Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
+; ratio 1.1549
+; seed 1785685356 
+; CC / CFLAGS gcc / -march=native -mtune=native -O3 
+; time needed: 441 ms / 50 evals=> 8.82ms/eval
+; Time spent for assembling and measureing (initial batch_size=166, initial num_batches=31): 45 ms
+; number of used evaluations: 50
+; Ratio (time for assembling + measure)/(total runtime for 50 evals): 0.10204081632653061
+; number reverted permutation/ tried permutation: 7 / 26 =26.923%
+; number reverted decision/ tried decision: 5 / 23 =21.739%
--- a/fiat-amd64/9789_ratio10114_seed1785685356_square_curve25519.asm
+++ b/fiat-amd64/9789_ratio10114_seed1785685356_square_curve25519.asm
@@ -1,0 +1,170 @@
+SECTION .text
+	GLOBAL square_curve25519
+square_curve25519:
+sub rsp, 0x98
+mov rax, [ rsi + 0x8 ]; load m64 arg1[1] to register64
+mov r10, rax; load m64 x8 to register64
+shl r10, 0x1; x8 <- arg1[1] * 0x2
+imul rax, [ rsi + 0x18 ], 0x26; x5 <- arg1[3] * 0x26
+mov r11, [ rsi + 0x10 ]; load m64 arg1[2] to register64
+lea rdx, [r11 + r11]; x7 <- arg1[2] * 2 
+mov r11, rdx; preserving value of x7 into a new reg
+mov rdx, [ rsi + 0x0 ]; saving arg1[0] in rdx.
+mulx r8, rcx, rdx; x23_1, x23_0<- arg1[0]^2
+mov rdx, [ rsi + 0x20 ]; load m64 arg1[4] to register64
+mov [ rsp + 0x0 ], rbx; spilling calSv-rbx to mem
+
+; imul rbx, rdx, 0x13
+
+lea r9, [rdx + 8 * rdx]; TMP <- arg1[4] * 9
+lea rbx, [rdx + 2 * r9]; x1 <- arg1[4]*19: arg1[4] + 2 * TMP = arg1[4] + 2 * 9 * arg1[4]
+
+mov rdx, rbx; x1 to rdx
+mulx rbx, r9, [ rsi + 0x20 ]; x9_1, x9_0<- arg1[4] * x1 (_0*_0)
+imul rdx, [ rsi + 0x18 ], 0x13; x4 <- arg1[3] * 0x13
+mov [ rsp + 0x8 ], rbp; spilling calSv-rbp to mem
+mov rbp, rdx; preserving value of x4 into a new reg
+mov rdx, [ rsi + 0x0 ]; saving arg1[0] in rdx.
+mov [ rsp + 0x10 ], r12; spilling calSv-r12 to mem
+mov [ rsp + 0x18 ], r13; spilling calSv-r13 to mem
+mulx r13, r12, r11; x21_1, x21_0<- arg1[0] * x7 (_0*_0)
+mov rdx, rbp; x4 to rdx
+mov [ rsp + 0x20 ], r14; spilling calSv-r14 to mem
+mulx r14, rbp, [ rsi + 0x18 ]; x11_1, x11_0<- arg1[3] * x4 (_0*_0)
+mov rdx, [ rsi + 0x8 ]; arg1[1] to rdx
+mov [ rsp + 0x28 ], r15; spilling calSv-r15 to mem
+mov [ rsp + 0x30 ], rdi; spilling out1 to mem
+mulx rdi, r15, rdx; x18_1, x18_0<- arg1[1]^2
+mov rdx, [ rsi + 0x20 ]; load m64 arg1[4] to register64
+mov [ rsp + 0x38 ], r14; spilling x11_1 to mem
+lea r14, [rdx + rdx]; x3 <- arg1[4] * 2 
+mov rdx, [ rsi + 0x8 ]; arg1[1] to rdx
+mov [ rsp + 0x40 ], rbp; spilling x11_0 to mem
+mov [ rsp + 0x48 ], r8; spilling x23_1 to mem
+mulx r8, rbp, r11; x17_1, x17_0<- arg1[1] * x7 (_0*_0)
+mov rdx, r10; x8 to rdx
+mulx r11, r10, [ rsi + 0x0 ]; x22_1, x22_0<- arg1[0] * x8 (_0*_0)
+xor rdx, rdx
+adox r9, rbp
+adox r8, rbx
+mov rbx, [ rsi + 0x18 ]; load m64 arg1[3] to register64
+mov rbp, rbx; load m64 x6 to register64
+shl rbp, 0x1; x6 <- arg1[3] * 0x2
+mov rdx, rbp; x6 to rdx
+mulx rbp, rbx, [ rsi + 0x0 ]; x20_1, x20_0<- arg1[0] * x6 (_0*_0)
+mov [ rsp + 0x50 ], r8; spilling x10002_1 to mem
+mov r8, rdx; preserving value of x6 into a new reg
+mov rdx, [ rsi + 0x0 ]; saving arg1[0] in rdx.
+mov [ rsp + 0x58 ], r9; spilling x10002_0 to mem
+mov [ rsp + 0x60 ], rbp; spilling x20_1 to mem
+mulx rbp, r9, r14; x19_1, x19_0<- arg1[0] * x3 (_0*_0)
+imul rdx, [ rsi + 0x20 ], 0x26; x2 <- arg1[4] * 0x26
+mov [ rsp + 0x68 ], rbx; spilling x20_0 to mem
+mulx rbx, r14, [ rsi + 0x18 ]; x10_1, x10_0<- arg1[3] * x2 (_0*_0)
+mov [ rsp + 0x70 ], r11; spilling x22_1 to mem
+mov [ rsp + 0x78 ], r10; spilling x22_0 to mem
+mulx r10, r11, [ rsi + 0x8 ]; x15_1, x15_0<- arg1[1] * x2 (_0*_0)
+mov [ rsp + 0x80 ], rcx; spilling x23_0 to mem
+mov [ rsp + 0x88 ], r10; spilling x15_1 to mem
+mulx r10, rcx, [ rsi + 0x10 ]; x12_1, x12_0<- arg1[2] * x2 (_0*_0)
+test al, al
+adox r14, r15
+adox rdi, rbx
+mov rdx, rax; x5 to rdx
+mulx r15, rax, [ rsi + 0x10 ]; x13_1, x13_0<- arg1[2] * x5 (_0*_0)
+mov rdx, [ rsi + 0x10 ]; arg1[2] to rdx
+mov [ rsp + 0x90 ], r10; spilling x12_1 to mem
+mulx r10, rbx, rdx; x14_1, x14_0<- arg1[2]^2
+adcx r14, r12
+adcx r13, rdi
+mov rdx, r8; x6 to rdx
+mulx r12, r8, [ rsi + 0x8 ]; x16_1, x16_0<- arg1[1] * x6 (_0*_0)
+test al, al
+adox rbx, r8
+adox r12, r10
+adcx rbx, r9
+adcx rbp, r12
+xor rdx, rdx
+adox rax, r11
+adox r15, [ rsp + 0x88 ]
+adcx rax, [ rsp + 0x80 ]
+adcx r15, [ rsp + 0x48 ]
+mov r9, rax;x25, copying x24_0 here, cause x24_0 is needed in a reg. It has those deps: "x25, x26, size: 2", current hard deps: ""
+shrd r9, r15, 51; x25 <- x24_1||x24_0 >> 51
+mov r11, rcx;x10004_0, copying x12_0 here, cause x12_0 is needed in a reg. It has those deps: "", current hard deps: "x12_0"
+add r11, [ rsp + 0x40 ]; could be done better, if r0 has been u8 as well
+mov rdi, [ rsp + 0x38 ]; load m64 x11_1 to register64
+adcx rdi, [ rsp + 0x90 ]
+xor rcx, rcx
+adox r11, [ rsp + 0x78 ]
+adox rdi, [ rsp + 0x70 ]
+adcx r11, r9
+adc rdi, 0x0; add CF to r0's alloc
+mov rdx, r11;x32, copying x31_0 here, cause x31_0 is needed in a reg. It has those deps: "x32, x33, size: 2", current hard deps: ""
+shrd rdx, rdi, 51; x32 <- x31_1||x31_0 >> 51
+test al, al
+adox r14, rdx
+adox r13, rcx
+mov r10, [ rsp + 0x58 ]; load m64 x10002_0 to register64
+adcx r10, [ rsp + 0x68 ]
+mov r8, [ rsp + 0x50 ]; load m64 x10002_1 to register64
+adcx r8, [ rsp + 0x60 ]
+mov r12, r14;x35, copying x34_0 here, cause x34_0 is needed in a reg. It has those deps: "x35, x36, size: 2", current hard deps: ""
+shrd r12, r13, 51; x35 <- x34_1||x34_0 >> 51
+xor r15, r15
+adox r10, r12
+adox r8, r15
+mov rcx, r10;x38, copying x37_0 here, cause x37_0 is needed in a reg. It has those deps: "x38, x39, size: 2", current hard deps: ""
+shrd rcx, r8, 51; x38 <- x37_1||x37_0 >> 51
+xor r9, r9
+adox rbx, rcx
+adox rbp, r9
+mov r15, 0x33 ; moving imm to reg
+bzhi rdi, r10, r15; x39 <- x37_0 (only least 0x33 bits)
+mov rdx, rbx;x41, copying x40_0 here, cause x40_0 is needed in a reg. It has those deps: "x41, x42, size: 2", current hard deps: ""
+shrd rdx, rbp, 51; x41 <- x40_1||x40_0 >> 51
+bzhi r13, rbx, r15; x42 <- x40_0 (only least 0x33 bits)
+bzhi r12, rax, r15; x26 <- x24_0 (only least 0x33 bits)
+mov rax, [ rsp + 0x30 ]; load m64 out1 to register64
+mov [ rax + 0x20 ], r13; out1[4] = x42
+
+; using imul works
+; imul r8, rdx, 0x13
+
+; this does not.
+lea r10, [rdx + 8 * rdx]; TMP <- x41 * 9
+lea r8, [rdx + 2 * r10]; x43 <- x41*19: x41 + 2 * TMP = x41 + 2 * 9 * x41
+
+
+lea r12, [ r12 + r8 ]
+bzhi r10, r11, r15; x33 <- x31_0 (only least 0x33 bits)
+bzhi r11, r12, r15; x46 <- x44 (only least 0x33 bits)
+shr r12, 51; x45 <- x44>> 51
+lea r12, [ r12 + r10 ]
+mov rcx, r12;x48, copying x47 here, cause x47 is needed in a reg. It has those deps: "x48, x49, size: 2", current hard deps: ""
+shr rcx, 51; x48 <- x47>> 51
+bzhi rbx, r14, r15; x36 <- x34_0 (only least 0x33 bits)
+mov [ rax + 0x0 ], r11; out1[0] = x46
+lea rcx, [ rcx + rbx ]
+mov [ rax + 0x10 ], rcx; out1[2] = x50
+bzhi r14, r12, r15; x49 <- x47 (only least 0x33 bits)
+mov [ rax + 0x18 ], rdi; out1[3] = x39
+mov [ rax + 0x8 ], r14; out1[1] = x49
+mov rbx, [ rsp + 0x0 ] ; pop
+mov rbp, [ rsp + 0x8 ] ; pop
+mov r12, [ rsp + 0x10 ] ; pop
+mov r13, [ rsp + 0x18 ] ; pop
+mov r14, [ rsp + 0x20 ] ; pop
+mov r15, [ rsp + 0x28 ] ; pop
+add rsp, 0x98 
+ret
+; cpu Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
+; ratio 1.0114
+; seed 1785685356 
+; CC / CFLAGS clang / -march=native -mtune=native -O3 
+; time needed: 1758 ms / 100 evals=> 17.58ms/eval
+; Time spent for assembling and measureing (initial batch_size=161, initial num_batches=31): 101 ms
+; number of used evaluations: 100
+; Ratio (time for assembling + measure)/(total runtime for 100 evals): 0.05745164960182025
+; number reverted permutation/ tried permutation: 15 / 51 =29.412%
+; number reverted decision/ tried decision: 15 / 48 =31.250%


src/Assembly/Parse.v
```diff
@@ -172,21 +172,25 @@ Definition parse_FLAG : ParserAction FLAG
...
+(*[rax           + 0x10]*) (parse_REG;; ((                                ((const 0)                                             * None      );; (strip_whitespace_before ("+" ||->{id} "-") ;; parse_Z_arith_strict)    )))||->{id}
+(*[rax +     rbx]       *) (parse_REG;; ((strip_whitespace_around "+" ;;R ((const 1)                                             * parse_REG );; ((const 0)                                                         )    )))||->{id}
+(*[rax + 2 * rbx]       *) (parse_REG;; ((strip_whitespace_around "+" ;;R (parse_Z_arith_strict (strip_whitespace_around "*";;R  * parse_REG));; ((const 0)                                                         )    )))||->{id}
+(*[      2 * rax]       *) (None;;      ((                                (parse_Z_arith_strict (strip_whitespace_around "*";;R  * parse_REG));; ((const 0)                                                         )    )))
+            ) ||->{id} parse_any_whitespace);;L
+            "]")).
```
JasonGross, line 193: Sequencing is `;;` not `*`, and there's no `const` (I think you would write `const v` as `parse_map (fun _ => v) ""`).  (Perhaps I should add one?)  Here's an alternate way to do it, where the `parse_map` is a bit more factored out.

Btw, if you have suggestions for more standard notation for parser combinator (w/ actions) libraries, I'd be happy to take advice/suggestions on how the notations should be set up.

```coq
Definition parse_MEM : ParserAction MEM
  := parse_map
       (fun '(has_byte, (br (*base reg*), sr (*scale reg, including z *), offset))
        => {| mem_is_byte := if has_byte:option _ then true else false
           ; mem_base_reg := br:option REG
           ; mem_scale_reg := sr:option (Z * REG)
           ; mem_offset := offset:option Z |})
       (((strip_whitespace_after "byte ")?) ;;
        (strip_whitespace_after "["
         ;;R
             (
               (*[rax]                 *)
               (parse_map
                  (fun r => (Some r, None, None))
                  parse_REG)
               || (*[rax           + 0x10]*)
                 (parse_map
                    (fun '(br, (pm, z)) => (Some br, None, Some match pm with
                                                                | inl _ (* plus *) => z
                                                                | inr _ (* minus *) => -z
                                                                end%Z))
                    (strip_whitespace_after parse_REG ;; ("+" ||->{id} "-") ;; strip_whitespace_before parse_Z_arith_strict))
               || (*[rax +     rbx]       *)
                 (parse_map
                    (fun '(br, sr) => (Some br, Some (1%Z, sr), None))
                    (strip_whitespace_after parse_REG ;; "+" ;;R strip_whitespace_before parse_REG))
               || (*[rax + 2 * rbx]       *)
                 (parse_map
                    (fun '(br, (z, sr)) => (Some br, Some (z, sr), None))
                    (strip_whitespace_after parse_REG ;; "+" ;;R strip_whitespace_around parse_Z_arith_strict ;; "*" ;;R strip_whitespace_before parse_REG))
               || (*[      2 * rax]       *)
                 (parse_map
                    (fun '(z, r) => (None, Some (z, r), None))
                    (strip_whitespace_after parse_Z_arith_strict ;; "*" ;;R strip_whitespace_before parse_REG))
             )
         ;;L strip_whitespace_before "]")).
```

dderjoel:
Using `*`, I was trying to give it what I defined in the Record, which there is `(Z * REG)` and I thought it'll then match those three tokens. But yeah, works now. I removed that later on in the print function, too, and it is at least compiling now.

Now, what is that [`Symbolic.v`](https://github.com/dderjoel/fiat-crypto/blob/bf5cd443ae3aee693a310ba023be7d2091e2682c/src/Assembly/Symbolic.v#L1634) trying to do, where The `Definition Address` is. What are we calculating? I don't get, what that `App` is and how those (i think different types) can be added.
Also I think I should use some multiplication instead of add to utilize the `z`.
But it is now failing with:

File "./src/Assembly/Symbolic.v", line 1634, characters 39-52:
Error:
In environment
sa : AddressSize
a : MEM
base : idx
p : Z * REG
z : Z
r : REG
The term "[z]" has type "list Z" while it is expected to have type
 "list (M idx)".

JasonGross:
> Are we calculating the result of whatever is calculated in the mem `[...]`?

Yes, it's the memory address.

dderjoel:
Okay, In that case, how can I multiply two numbers, or more, the 'value on r` times 'z'
  | Some (z, r) => App (mulZ, [(GetReg r); z])`
and why is there a `list z` ? 
I also tried to get the first element from apparent 'list z' with 
`           | Some (z, r) => App (mulZ, [(GetReg r); (fst z)])` but that error'ed even worse.

JasonGross:
`App` is about internalizing values in the e-graph dag.  You want to write something like
```coq
base_reg <- GetReg base_reg;
other_reg <- GetReg other_reg;
scale <- App (zconst sa scale, [])
offset <- App (zconst sa offset, []);
scaled_reg <- App (mul sa, [scale; other_reg]);
App (add sa, [base_reg; scaled_reg; offset])
```
except with more `match`es to deal with `option`s

src/Assembly/Symbolic.v
```diff
@@ -1626,11 +1626,14 @@ Definition SetReg r (v : idx) : M unit :=
...
 Class AddressSize := address_size : OperationSize.
 Definition Address {sa : AddressSize} (a : MEM) : M idx :=
-  base <- GetReg a.(mem_reg);
-  index <- match a.(mem_extra_reg) with
+  base <- match a.(mem_base_reg) with
            | Some r => GetReg r
            | None => App ((const 0), nil)
            end;
+  index <- match a.(mem_scale_reg) with
+           | Some (z, r) => App (mulZ, [(GetReg r); z])
```
JasonGross, line 1634: You could do something like
```suggestion
           | Some (z, r) => z <- App (zconst sa z, []); r <- GetReg r; App (mul sa, [r; z])
```
instead

JasonGross:
> why is there a `list z` ?

To explain the error message, when you write `[GetReg r; z]` this desugars to `GetReg r :: [z]`.  Since `GetReg r` has type `M idx`, Coq expects the tail to be a list of these things, but instead `[z]` is a `list Z`.  The "meaning" of this error message is that `GetReg r` and `z` have different types, so you can't put them in the same list.

src/Assembly/Parse.v
```diff
@@ -269,11 +288,14 @@ Global Instance show_lvl_MEM : ShowLevel MEM
...
-           => "[" ++ (show m.(mem_reg))
-                  ++ (match m.(mem_extra_reg) with
+           => "[" ++ (match m.(mem_base_reg) with
                       | None => ""
                       | Some r => " + " ++ show r
                       end)
+                  ++ (match m.(mem_scale_reg) with
+                      | None => ""
+                      | Some (z, r) => show z ++ " * " ++ show r 
+                      end)
```
JasonGross, line 298: You're missing the separator here, it'll print as something like `[+ RAX5 * RAX]` instead of `[+ RAX + 5 * RAX]`.
```suggestion
                      | Some (z, r) => (if z <? 0 then " - " else " + ") ++ (let z := Z.abs z in if z =? 1 then "" else Decimal.show_Z z ++ " * ") ++ show_REG r
                      end%Z)
```

Btw, is it fine to have things like `[+2*RAX]`, or does it need to be `[2*RAX]`?

src/Assembly/Parse.v
```diff
@@ -269,11 +288,14 @@ Global Instance show_lvl_MEM : ShowLevel MEM
   := fun m
      => (if m.(mem_is_byte) then show_lvl_app (fun 'tt => "byte") else show_lvl)
           (fun 'tt
-           => "[" ++ (show m.(mem_reg))
-                  ++ (match m.(mem_extra_reg) with
+           => "[" ++ (match m.(mem_base_reg) with
                       | None => ""
                       | Some r => " + " ++ show r
```
JasonGross, line 293: might as well make this
```suggestion
                      | Some r => " + " ++ show_REG r
```
while we're editing this code so that if this field later gets a scale factor, the code here will complain.

src/Assembly/Parse.v
dderjoel: when is that being printed anyway ? I thought this is more for debugging :o

dderjoel: Because I don't read anything from fiat-at the moment. I just give it the hints file and check the exit-code.
As long as we can parse `[2*reg]`, I should be fine

src/Assembly/Parse.v
dderjoel: so 'show' just prints whatever, and `shot_REG` will only print 'reg's? Then yes, sure, lets use the latter.

src/Assembly/Parse.v
JasonGross: I think we reprint the assembly to an output file.  We don't need to process assembly this way, but we should make sure this way works.

src/Assembly/Parse.v
JasonGross: > so 'show' just prints whatever, and `shot_REG` will only print 'reg's?

Yes

src/Assembly/Parse.v
dderjoel: yes, then I'm happy with `[+2*rax]`

dderjoel: but in the case of `z==1` would it print  `*rax` or `rax` ? I'm not sure about the associativity of `++` and `else`; Either way, `1*rax` would also be fine, so we could write: 
`| Some (z, r) => (if z <? 0 then " - " else " + ") ++ (let z := Z.abs z in Decimal.show_Z z ++ " * ") ++ show_REG r`

JasonGross: I believe `++` binds more tightly than `else`; feel free to to add more parentheses though.  I think not printing the `1 *` in the common case would be nice.

dderjoel:
so the current parser can parse
`[reg]` and `[reg+reg]` but cannot parse `[reg + Z]`, `[reg + Z * reg]`, `[Z*reg]`, (regardless of spaces). Is there any --debug -- verbose mode that I could use to debug, on what Coq tries to parse and where it fails to find a match ?

JasonGross:
> so the current parser can parse
> `[reg]` and `[reg+reg]` but cannot parse `[reg + Z]`, `[reg + Z * reg]`, `[Z*reg]`, (regardless of spaces).

Strange

> Is there any --debug -- verbose mode that I could use to debug, on what Coq tries to parse and where it fails to find a match ?

No, but you can do `Goal False.  pose (parse_mem "[RAX + 5]") as v. cbv [parse_mem] in v.` and then keep going to see how it computes interactively.  (If you want a very verbose mode, you can do something like:
```coq
Ltac reduce_first_thing_in body v :=
  lazymatch body with
  | ?f ?x => first [ reduce_first_thing_in f v | reduce_first_thing_in x v ]
  | ?f => progress unfold f in (value of v) at 1
  end.
Ltac reduce_first_in v :=
  first [ progress cbv beta in v
        | progress cbv iota in v
        | progress cbv zeta in v
        | let body := (eval cbv delta [v] in v) in progress reduce_first_thing_in body v ].
Goal False. pose (parse_mem "[RAX + 5]") as v. Set Ltac Debug. Set Ltac Batch Debug. repeat reduce_first_in v.
```
)

JasonGross:
> adding the `Goal False. pose (parse_MEM "[RAX + 5]") as v. cbv [parse_MEM] in v.` after the Definition of `parse_MEM` (I changed, `parse_mem` to `parse_MEM`) errors in

This is meant for interactive exploration in either CoqIDE or emacs/PG or vscoq, etc.  You'll have to add a bunch more `cbv` commands to manually unfold things.  I also missed a case in `reduce_first_thing_in`, I'll go back and edit my comment to add it.

> Edit: using that longer version I get _much_ More output... Which I cant parse myself - where would I start to read, it keeps printing the function but I cant read any differences. (also I don't know what the code does)

It should print a context including `v := ...` and every couple of steps the body should be updated.

dderjoel:
currently creating a logfile with that `v :=` blocks... 5GB, and counting. I'll try to dig through that once it's done, otherwise I'll get myself an ide and try the interactive approach.

JasonGross:
Here's a manual debugging session that led me to realize that we can parse `rax` but not `RAX`:
```coq
Goal False.
  pose (parse_MEM "[rax + 5]") as v.
  cbv [parse_MEM] in v.
  Set Printing Coercions.
  unfold parse_map at 1 in (value of v).
  unfold ";;" in (value of v).
  set (v' := (strip_whitespace_after (parse_str "byte ")?) "[rax + 5]") in (value of v).
  vm_compute in v'; subst v'.
  cbn [List.map List.flat_map List.app] in v.
  Ltac tac v :=
    let v' := fresh "v'" in
    match (eval cbv delta [v] in v) with
    | context[?e]
      => lazymatch e with
         | strip_whitespace_after ?f (String ?h ?t) => idtac
         | parse_map ?f ?p (String ?h ?t) => idtac
         end;
         idtac e;
         set (v' := e) in (value of v);
         vm_compute in v';
         let v'' := (eval cbv delta [v'] in v') in
         idtac v'';
         subst v'
    end;
    cbn [List.map List.flat_map List.app] in v.
  tac v.
  unfold "||" in v.
  unfold parse_alt_gen in v.
  tac v.
  set (v' := parse_map _ _ (String _ _)) in (value of v) at 1.
  unfold parse_map at 1 in (value of v').
  clear v.
  Compute strip_whitespace_after parse_REG "RAX".
```

JasonGross:
A bit more debugging:
```coq
Compute finalize parse_MEM "[rax]". (* Some ... *)
Compute finalize parse_MEM "[rax           + 0x10]". (* None *)
Compute finalize parse_MEM "[rax +     rbx]". (* Some ... *)
Compute finalize parse_MEM "[rax + 2 * rbx]". (* None *)
Compute finalize parse_MEM "[      2 * rax]". (* None *)
Compute parse_MEM "[rax]". (* one result *)
Compute parse_MEM "[rax           + 0x10]". (* two results *)
Compute parse_MEM "[rax +     rbx]". (* one result *)
Compute parse_MEM "[rax + 2 * rbx]". (* four results *)
Compute parse_MEM "[      2 * rax]". (* many results *)
```
shows that what's going wrong is that we have multiple ways to parse the expression.

JasonGross:
Ah, the issue is that `parse_Z_arith_strict` permits whitespace both before and after, so we were overlapping whitespace patterns.

I've pushed a commit just now that fixes this

dderjoel:
despite that Bedrock build failing, (works on my machine with `SKIP_BEDROCK`) It still fails to proof correct.
with `./ExtractionOCaml/unsaturated_solinas  --no-primitives --no-wide-int --shiftr-avoid-uint1 --tight-bounds-mul-by 1.000001 -o /dev/null --output-asm /dev/null --hints-file ../fiat-amd64/10_ratio11074_seed2040113663254706_square_poly1305.asm  'poly1305' '64' '3' '2^130 - 5' carry_square`
it fails with 
```
check_args
/* Autogenerated: './ExtractionOCaml/unsaturated_solinas' --no-primitives --no-wide-int --shiftr-avoid-uint1 --tight-bounds-mul-by 1.000001 -o '/dev/null' --output-asm '/dev/null' --hints-file '../fiat-amd64/10_ratio11074_seed2040113663254706_square_poly1305.asm' poly1305 64 3 '2^130 - 5' */
/* curve description: poly1305 */
/* requested operations: (all) */
/* n = 3 (from "3") */
/* s-c = 2^130 - [(1, 5)] (from "2^130 - 5") */
/* tight_bounds_multiplier = 1.000001 (from "1.000001") */
/*  */
/* Computed values: */
/*   carry_chain = [0, 1, 2, 0, 1] */
/*   eval z = z[0] + (z[1] << 44) + (z[2] << 87) */
/*   bytes_eval z = z[0] + (z[1] << 8) + (z[2] << 16) + (z[3] << 24) + (z[4] << 32) + (z[5] << 40) + (z[6] << 48) + (z[7] << 56) + (z[8] << 64) + (z[9] << 72) + (z[10] << 80) + (z[11] << 88) + (z[12] << 96) + (z[13] << 104) + (z[14] << 112) + (z[15] << 120) + (z[16] << 128) */
/*   balance = [0x1ffffffffff6, 0xffffffffffe, 0xffffffffffe] */

#include <stdint.h>
typedef unsigned char fiat_poly1305_uint1;
typedef signed char fiat_poly1305_int1;
#ifdef __GNUC__
#  define FIAT_POLY1305_FIAT_EXTENSION __extension__
#  define FIAT_POLY1305_FIAT_INLINE __inline__
#else
#  define FIAT_POLY1305_FIAT_EXTENSION
#  define FIAT_POLY1305_FIAT_INLINE
#endif

FIAT_POLY1305_FIAT_EXTENSION typedef signed __int128 fiat_poly1305_int128;
FIAT_POLY1305_FIAT_EXTENSION typedef unsigned __int128 fiat_poly1305_uint128;

/* The type fiat_poly1305_loose_field_element is a field element with loose bounds. */
/* Bounds: [[0x0 ~> 0x300003254e71], [0x0 ~> 0x18000192a73a], [0x0 ~> 0x18000192a73a]] */
typedef uint64_t fiat_poly1305_loose_field_element[3];

/* The type fiat_poly1305_tight_field_element is a field element with tight bounds. */
/* Bounds: [[0x0 ~> 0x1000010c6f7b], [0x0 ~> 0x800008637be], [0x0 ~> 0x800008637be]] */
typedef uint64_t fiat_poly1305_tight_field_element[3];

#if (-1 & 3) != 3
#error "This code only works on a two's complement system"
#endif



In parsing:
Error while parsing assembly:
Error while parsing assembly:
mov [ + rsp], rbx; saving to stack
mov [ + rsp + 0x08 * 1], rbp; saving to stack
mov [ + rsp + 0x08 * 2], r12; saving to stack
mov [ + rsp + 0x08 * 3], r13; saving to stack
mov [ + rsp + 0x08 * 4], r14; saving to stack
mov [ + rsp + 0x08 * 5], r15; saving to stack
mov rax, [ + rsi + 0x08 * 2]
lea rax, [ + rax + 4 * rax]
lea r10, [ + 2 * rax]
mov rdx, [ + rsi + 0x08 * 0]; arg1[0] to rdx
mulx r11, rbx, [ + rsi + 0x08 * 0]; x16, x15<- arg1[0] * arg1[0]
lea r10, [ + r10 + r10]
mov rdx, [ + rsi + 0x08 * 1]; arg1[1] to rdx
mov rdx, [ + rsi + 0x08 * 2]; arg1[2] to rdx
imul rdx, [ + rsi + 0x08 * 1], 2; x4 <- arg1[1] * 0x2
mulx rdx, rcx, [ + rsi + 0x08 * 0]; x14, x13<- arg1[0] * x4
imul r12, [ + rsi + 0x08 * 1], 2; x10001 <- arg1[1] * 0x2
imul rbx, [ + rsi + 0x08 * 2], 2; x3 <- arg1[2] * 0x2
mov rdx, [ + rsi + 0x08 * 1]; arg1[1] to rdx
mov rdx, [ + rsi + 0x08 * 0]; arg1[0] to rdx
lea r15, [ + r15 + r8]
lea r15, [ + r15 + r13]
mov [ + rdi + 0x08 * 1], rbp; out1[1] = x47
lea r15, [ + r15 + r10]
mov [ + rdi + 0x08 * 0], rcx; out1[0] = x44
mov [ + rdi + 0x08 * 2], r15; out1[2] = x48
mov rbx, [ + rsp + 0x08 * 0]; restoring from stack
mov rbp, [ + rsp + 0x08 * 1]; restoring from stack
mov r12, [ + rsp + 0x08 * 2]; restoring from stack
mov r13, [ + rsp + 0x08 * 3]; restoring from stack
mov r14, [ + rsp + 0x08 * 4]; restoring from stack
mov r15, [ + rsp + 0x08 * 5]; restoring from stack

New assembly being parsed:
SECTION .text
	GLOBAL square_poly1305
square_poly1305:
sub rsp, 48 ; last 0x30 (6) for Caller - save regs 

;instead of 
; mov [ rsp + 0x0 ], rbx; saving to stack

; we also want to use
mov [ + rsp], rbx; saving to stack

mov [ + rsp + 0x08 * 1], rbp; saving to stack
mov [ + rsp + 0x08 * 2], r12; saving to stack
mov [ + rsp + 0x08 * 3], r13; saving to stack
mov [ + rsp + 0x08 * 4], r14; saving to stack
mov [ + rsp + 0x08 * 5], r15; saving to stack

; instead of 
;imul rax, [ rsi + 0x10 ], 0x5; x1 <- arg1[2] * 0x5

; we want also to use 
mov rax, [ + rsi + 0x08 * 2]
lea rax, [ + rax + 4 * rax]

; instead of 
;imul r10, rax, 0x2; x2 <- x1 * 0x2

; we also want to use
lea r10, [ + 2 * rax]

mov rdx, [ + rsi + 0x08 * 0]; arg1[0] to rdx
mulx r11, rbx, [ + rsi + 0x08 * 0]; x16, x15<- arg1[0] * arg1[0]

; instead of 
; imul r10, r10, 0x2; x10000 <- x2 * 0x2

;we want to use 
lea r10, [ + r10 + r10]

; we also still need to support basic addressing with REG+off
mov rdx, [ + rsi + 0x08 * 1]; arg1[1] to rdx

mulx r10, rbp, r10; x8, x7<- arg1[1] * x10000
xor r12, r12
adox rbp, rbx
adox r11, r10
mov rdx, [ + rsi + 0x08 * 2]; arg1[2] to rdx
mulx rax, r13, rax; x6, x5<- arg1[2] * x1
mov r14, 17592186044415 ; moving imm to reg
mov r15, rbp; x22, copying x17 here, cause x17 is needed in a reg for other than x22, namely all: , x21, x22, size: 2
and r15, r14; x22 <- x17&0xfffffffffff
imul rdx, [ + rsi + 0x08 * 1], 2; x4 <- arg1[1] * 0x2
mulx rdx, rcx, [ + rsi + 0x08 * 0]; x14, x13<- arg1[0] * x4
shrd rbp, r11, 44; x21 <- x19||x17 >> 44
xor r8, r8
adox r13, rcx
adox rdx, rax
adcx r13, rbp
adc rdx, 0
imul r12, [ + rsi + 0x08 * 1], 2; x10001 <- arg1[1] * 0x2
mov r9, r13; x34, copying x31 here, cause x31 is needed in a reg for other than x34, namely all: , x34, x35, size: 2
shrd r9, rdx, 43; x34 <- x33||x31 >> 43
imul rbx, [ + rsi + 0x08 * 2], 2; x3 <- arg1[2] * 0x2
mov rdx, [ + rsi + 0x08 * 1]; arg1[1] to rdx
mulx r12, r10, r12; x10, x9<- arg1[1] * x10001
mov rdx, [ + rsi + 0x08 * 0]; arg1[0] to rdx
mulx rbx, r11, rbx; x12, x11<- arg1[0] * x3
xor rdx, rdx
adox r10, r11
adox rbx, r12
adcx r10, r9
adc rbx, 0
mov r8, r10; x39, copying x36 here, cause x36 is needed in a reg for other than x39, namely all: , x39, x40, size: 2
shrd r8, rbx, 43; x39 <- x38||x36 >> 43
imul r8, r8, 5; x41 <- x39 * 0x5
mov rax, 8796093022207 ; moving imm to reg
and r13, rax; x35 <- x31&0x7ffffffffff
lea r15, [ + r15 + r8]
mov rcx, r15; x44, copying x42 here, cause x42 is needed in a reg for other than x44, namely all: , x44, x43, size: 2
and rcx, r14; x44 <- x42&0xfffffffffff
shr r15, 44; x43 <- x42>> 44
lea r15, [ + r15 + r13]
mov rbp, r15; x47, copying x45 here, cause x45 is needed in a reg for other than x47, namely all: , x47, x46, size: 2
and rbp, rax; x47 <- x45&0x7ffffffffff
shr r15, 43; x46 <- x45>> 43
and r10, rax; x40 <- x36&0x7ffffffffff
mov [ + rdi + 0x08 * 1], rbp; out1[1] = x47
lea r15, [ + r15 + r10]
mov [ + rdi + 0x08 * 0], rcx; out1[0] = x44
mov [ + rdi + 0x08 * 2], r15; out1[2] = x48
mov rbx, [ + rsp + 0x08 * 0]; restoring from stack
mov rbp, [ + rsp + 0x08 * 1]; restoring from stack
mov r12, [ + rsp + 0x08 * 2]; restoring from stack
mov r13, [ + rsp + 0x08 * 3]; restoring from stack
mov r14, [ + rsp + 0x08 * 4]; restoring from stack
mov r15, [ + rsp + 0x08 * 5]; restoring from stack
add rsp, 48
ret
; cpu Intel(R) Core(TM) i7-10710U CPU @ 1.10GHz
; clocked at 1600 MHz
; first cyclecount 15.38, best 10.657863145258103, lastGood 10.66546329723225
; seed 2040113663254706 
; CC / CFLAGS clang / -march=native -mtune=native -O3 
; time needed: 15590 ms / 500 runs=> 31.18ms/run
; Time spent for assembling and measureing (initial batch_size=833, initial num_batches=101): 13489 ms
; Ratio (time for assembling + measure)/(total runtime for 500runs): 0.8652341244387428
; number reverted permutation/ tried permutation: 221 / 270 =81.852%
; number reverted decision/ tried decision: 169 / 231 =73.160%


In fiat_poly1305_from_bytes:
Computed bounds (Some [Some [0x0 ~> 0xfffffffffff], Some [0x0 ~> 0x7ffffffffff], Some [0x0 ~> 0x17ffffffffff]]) are not tight enough (expected bounds not looser than (Some [Some [0x0 ~> 0x1000010c6f7b], Some [0x0 ~> 0x800008637be], Some [0x0 ~> 0x800008637be]])).
The bounds [0x0 ~> 0x17ffffffffff] are looser than the expected bounds [0x0 ~> 0x800008637be]
The syntax tree:
(λ x1,
  let x2 := x1[16] << 41 (* : [0x0 ~> 0x60000000000] *) in
  let x3 := x1[15] << 33 (* : [0x0 ~> 0x1fe00000000] *) in
  let x4 := x1[14] << 25 (* : [0x0 ~> 0x1fe000000] *) in
  let x5 := x1[13] << 17 (* : [0x0 ~> 0x1fe0000] *) in
  let x6 := x1[12] << 9 (* : [0x0 ~> 0x1fe00] *) in
  let x7 := x1[11] * 2 (* : [0x0 ~> 0x1fe] *) in
  let x8 := x1[10] << 36 (* : [0x0 ~> 0xff000000000] *) in
  let x9 := x1[9] << 28 (* : [0x0 ~> 0xff0000000] *) in
  let x10 := x1[8] << 20 (* : [0x0 ~> 0xff00000] *) in
  let x11 := x1[7] << 12 (* : [0x0 ~> 0xff000] *) in
  let x12 := x1[6] << 4 (* : [0x0 ~> 0xff0] *) in
  let x13 := x1[5] << 40 (* : [0x0 ~> 0xff0000000000] *) in
  let x14 := x1[4] << 32 (* : [0x0 ~> 0xff00000000] *) in
  let x15 := x1[3] << 24 (* : [0x0 ~> 0xff000000] *) in
  let x16 := x1[2] << 16 (* : [0x0 ~> 0xff0000] *) in
  let x17 := x1[1] << 8 (* : [0x0 ~> 0xff00] *) in
  let x18 := x1[0] (* : uint8_t *) in
  let x19 := x17 + x18 (* : uint16_t *) in
  let x20 := x16 + x19 (* : uint24_t *) in
  let x21 := x15 + x20 (* : uint32_t *) in
  let x22 := x14 + x21 (* : uint40_t *) in
  let x23 := x13 + x22 (* : uint48_t *) in
  let x24 := x23 & (2^44-1) (* : uint44_t *) in
  let x25 := x23 >> 44 (* : uint4_t *) in
  let x26 := x12 + x25 (* : uint12_t *) in
  let x27 := x11 + x26 (* : uint20_t *) in
  let x28 := x10 + x27 (* : uint28_t *) in
  let x29 := x9 + x28 (* : uint36_t *) in
  let x30 := x8 + x29 (* : uint44_t *) in
  let x31 := x30 & (2^43-1) (* : uint43_t *) in
  let x32 := x30 >> 43 (* : [0x0 ~> 0x2] *) in
  let x33 := x7 + x32 (* : [0x0 ~> 0x200] *) in
  let x34 := x6 + x33 (* : [0x0 ~> 0x20000] *) in
  let x35 := x5 + x34 (* : [0x0 ~> 0x2000000] *) in
  let x36 := x4 + x35 (* : [0x0 ~> 0x200000000] *) in
  let x37 := x3 + x36 (* : [0x0 ~> 0x20000000000] *) in
  let x38 := x2 + x37 (* : [0x0 ~> 0x80000000000] *) in
  let x39 := x38 & (2^43-1) (* : uint43_t *) in
  let x40 := x38 >> 43 (* : [0x0 ~> 0x2] *) in
  let x41 := (x40 << 43) + x39 (* : [0x0 ~> 0x17ffffffffff] *) in
  x24 :: x31 :: x41 :: []
)

which can be pretty-printed as:
/*
 * Input Bounds:
 *   arg1: [[0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0xff], [0x0 ~> 0x3]]
 * Output Bounds:
 *   out1: None
 */
void f(uint64_t out1[3], const uint8_t arg1[17]) {
  uint64_t x1;
  uint64_t x2;
  uint64_t x3;
  uint32_t x4;
  uint32_t x5;
  uint16_t x6;
  uint64_t x7;
  uint64_t x8;
  uint32_t x9;
  uint32_t x10;
  uint8_t x11;
  uint64_t x12;
  uint64_t x13;
  uint32_t x14;
  uint32_t x15;
  uint8_t x16;
  uint8_t x17;
  uint16_t x18;
  uint32_t x19;
  uint32_t x20;
  uint64_t x21;
  uint64_t x22;
  uint64_t x23;
  uint4 x24;
  uint16_t x25;
  uint32_t x26;
  uint32_t x27;
  uint64_t x28;
  uint64_t x29;
  uint64_t x30;
  uint2 x31;
  uint16_t x32;
  uint32_t x33;
  uint32_t x34;
  uint64_t x35;
  uint64_t x36;
  uint64_t x37;
  uint64_t x38;
  uint2 x39;
  uint64_t x40;
  x1 = ((uint64_t)(uint2)(arg1[16]) << 41);
  x2 = ((uint64_t)(arg1[15]) << 33);
  x3 = ((uint64_t)(arg1[14]) << 25);
  x4 = ((uint32_t)(arg1[13]) << 17);
  x5 = ((uint32_t)(arg1[12]) << 9);
  x6 = (uint16_t)((arg1[11]) * 0x2);
  x7 = ((uint64_t)(arg1[10]) << 36);
  x8 = ((uint64_t)(arg1[9]) << 28);
  x9 = ((uint32_t)(arg1[8]) << 20);
  x10 = ((uint32_t)(arg1[7]) << 12);
  x11 = ((arg1[6]) << 4);
  x12 = ((uint64_t)(arg1[5]) << 40);
  x13 = ((uint64_t)(arg1[4]) << 32);
  x14 = ((uint32_t)(arg1[3]) << 24);
  x15 = ((uint32_t)(arg1[2]) << 16);
  x16 = ((arg1[1]) << 8);
  x17 = (arg1[0]);
  x18 = (uint16_t)(x16 + x17);
  x19 = (x15 + (uint32_t)x18);
  x20 = (x14 + x19);
  x21 = (x13 + x20);
  x22 = (x12 + x21);
  x23 = (x22 & UINT64_C(0xfffffffffff));
  x24 = (uint4)(x22 >> 44);
  x25 = (uint16_t)(x11 + x24);
  x26 = (x10 + (uint32_t)x25);
  x27 = (x9 + x26);
  x28 = (x8 + x27);
  x29 = (x7 + x28);
  x30 = (x29 & UINT64_C(0x7ffffffffff));
  x31 = (uint2)(x29 >> 43);
  x32 = (uint16_t)(x6 + x31);
  x33 = (x5 + (uint32_t)x32);
  x34 = (x4 + x33);
  x35 = (x3 + x34);
  x36 = (x2 + x35);
  x37 = (x1 + x36);
  x38 = (x37 & UINT64_C(0x7ffffffffff));
  x39 = (uint2)(x37 >> 43);
  x40 = (((uint64_t)x39 << 43) + x38);
  out1[0] = x23;
  out1[1] = x30;
  out1[2] = x40;
}

with input bounds (Some [Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0xff], Some [0x0 ~> 0x3]]).


Fatal error: exception Failure("Synthesis failed")
```
which I dont understand. Where is the problem now, It says `Error while parsing assembly:          ` and then prints a bunch of instructions with memory operands, but then prints them correctly. Different (`[ + rsp + 0x08 * 1]`) , but correctly (as far as I can see)

JasonGross:
> Edit: how can I provide long code in github-markdown in that lovely scroll-pane, or does that only work for referencing files?

I think if you tag it with a language (like ```` ```coq ````) it does syntax highlighting and scroll?

The issue here is that the pretty-printed assembly needs to be reparsable as the initial code, and the parsing code doesn't support leading + in memory addressing.  If you make the show code a bit more clever, so that it never prints a leading +, it should work fine.

dderjoel:
> I think if you tag it with a language (like `` ```coq ``) it does syntax highlighting and scroll?
It does not scrollify, only the syntax.

The parsing/printing seems to work now. I can compile (yey!).
Now it Is unable to unify
```coq
;
  symbolic_reg_state := [(rax, 82), (rcx, 378), (rdx, 4), (rbx, 112), (rsp, 184), (rbp, 114), (rsi, 133), (rdi, 128), (r8, 375), (r9, 364), (r10, 383), (r11, 33), (r12, 121), (r13, 122), (r14, 123), (r15, 124)];
  symbolic_flag_state := (*flag_state*)(CF=Some 385 PF=None AF=None ZF=None SF=None ZF=None OF=Some 386);
  symbolic_mem_state :=
[(324, 124), (321, 123), (318, 122), (315, 121), (312, 114), (309, 112), (306, 142), (303, 143), (300, 144), (297, 145), (294, 146), (291, 147), (288, 148), (285, 149), (282, 150), (279, 151), (276, 152), (273, 153), (270, 154), (267, 155), (264, 156), (261, 157), (258, 158), (255, 159), (252, 160), (249, 161), (246, 162), (243, 163), (240, 164), (237, 165), (234, 166), (231, 167), (228, 168), (225, 169), (222, 170), (219, 171), (216, 172), (213, 173), (210, 174), (207, 175), (205, 176), (202, 177), (199, 178), (196, 179), (193, 180), (190, 181), (188, 182), (186, 183), (135, 0), (134, 1), (133, 2), (132, 384), (130, 381), (128, 378)]
;
|}
Unable to unify: [inr [378, 381, 384]] == [inr [101, 106, 108]]
Could not unify the values at index 0: [378, 381, 384] != [101, 106, 108]
index 0: 378 != 101
(slice 0 44, [377]) != (slice 0 44, [98])
index 0: 377 != 98
(add 64, [345, 375]) != (add 64, [57, 96])
index 0: 345 != 57
(slice 0 44, [337]) != (slice 0 44, [44])
index 0: 337 != 44
(add 64, [41, 334]) != (add 64, [25, 41])
index 1: 334 != 25
(mul 64, [1, 331]) != (mul 64, [0, 1, 22])
index 1: 331 != 0
(add 64, [329, 329]) != (old 64 0, [])
(add 64, [(mul 64, [7, 328]), (mul 64, [7, 328])]) != (old 64 0, [])
Operation mismatch: add 64 != old 64 0
0 is a special value no longer present in the symbolic machine state at the end of execution.

Fatal error: exception Failure("Synthesis failed")
```
Full output here: [poly1305_sq.txt](https://github.com/mit-plv/fiat-crypto/files/8105196/poly1305_sq.txt)

dderjoel:
also the bedrock-build fails with 
```
 File "./src/Assembly/WithBedrock/Semantics.v", line 110, characters 46-92:
command time -f "src/Assembly/WithBedrock/Semantics.vo (real: %e, user: %U, sys: %S, mem: %M ko)" "coqc"  -time -q '-w' '+implicit-core-hint-db,+implicits-in-term,+non-reversible-notation,+deprecated-intros-until-0,+deprecated-focus,+unused-intro-pattern,+variable-collision,+omega-is-deprecated,+deprecated-instantiate-syntax,+non-recursive' -w -notation-overridden,-undeclared-scope,-deprecated-hint-rewrite-without-locality,-deprecated-hint-constr,-fragile-hint-constr,-native-compiler-disabled,-ambiguous-paths  "-w" "-deprecated-native-compiler-option" "-native-compiler" "ondemand" -R src Crypto src/Assembly/WithBedrock/Semantics.v > src/Assembly/WithBedrock/Semantics.v.timing
Error: Unknown interpretation for notation "_ <- _ ; _".
```

src/Assembly/WithBedrock/Semantics.v
```diff
@@ -106,8 +106,9 @@ Definition DenoteConst (sz : N) (a : CONST) : Z :=
 
 Definition DenoteAddress (sa : N) (st : machine_state) (a : MEM) : Z :=
   Z.land (
-    get_reg st (mem_reg a) +
-    match mem_extra_reg a with Some e => get_reg st e | _ => 0 end +
+    get_reg st (mem_base_reg a) +
+    match mem_scale_reg a with Some (z, r) => r <- get_reg st r; z<- DenoteConst sa z; r * z | _ => 0 end +
```
JasonGross, line 110:
```suggestion
    match mem_scale_reg a with Some (z, r) => r <- get_reg st r; z<- DenoteConst sa z; r * z | _ => 0 end%option +
```
Should fix the syntax error by telling it to interpret this in option scope.  But these aren't options, so I think you actually just want to do
```suggestion
    match mem_scale_reg a with Some (z, r) => get_reg st r * DenoteConst sa z | _ => 0 end +
```

JasonGross:
> ```coq
> (mul 64, [1, 331]) != (mul 64, [0, 1, 22])
> ```

This is the really interesting line.  I should probably make the behavior slightly different when either list is longer than length 1.  Maybe printing out successive inlinings is the right thing to do here?

We need to figure out why 331 is supposed to be the same as 0 * 22

src/Assembly/WithBedrock/Semantics.v
dderjoel: why does the latter fix it? as far as I can understand, this change only gets rid of the variable-reassignment.

dderjoel:
is `0` the number zero, or it that the variable #0?

dderjoel:
```coq
(*16*) (const 64, []);
(*17*) (mulZ, [0, 6]);
(*18*) (mul 64, [0, 0, 3]);
(*19*) (shrZ, [17, 16]);
(*20*) (shr 64, [17, 16]);
(*22*) (const 20, []);
```
hmm... so maybe it has to do with my `shl` <- `mul` change, which is not in that branch at the moment.

JasonGross:
> is `0` the number zero, or it that the variable #0?

The variable `#0`

There is also
```
(*326*) (const 4, []);
(*327*) (mul 64, [0, 326]);
(*328*) (add 64, [0, 327]);
(*329*) (mul 64, [7, 328]);
```

JasonGross:
It seems to want `2 * (#0 + #0 * 4) + 2 * (#0 + #0 * 4) == #0 * 20`.  So I think we need a rewrite rule that counts the occurrences of all the terms in a sum, where it will look two levels deeper to see if any terms are products of constants with a given term.  That is, we say `occurences(x * const v) := occurences(const v * x) := (x, v)` and otherwise `occurences(x) := (x, 1)`, and then we do something like `groupBy fst (sort (map occurences summands))` and then sum all the occurence counts and then recreate the sum.  @andres-erbsen @dderjoel does this sound like the right rewrite rule?

dderjoel:
yes, sounds right. So, how I generate the `mul (const)` is: If there is a `variable *const * const`,  I statically reduce that (But this is done already in [59600ae](https://github.com/mit-plv/fiat-crypto/commit/59600ae82eea35f7f0f8933be1beb38de27df4e7), see [here](https://github.com/mit-plv/fiat-crypto/pull/1123#issuecomment-1035624433), in the other issue. Then, for a couple known constants (like `5`) I have map (`5=1+4`) which is quickly computed with `lea`.

I don't exactly know, why this problem here (`5 = 1 + 4`) cannot be solved like that (`4 = 2*2`), too; is it because I 'break the constant up' myself, rather than 'combing them'?

JasonGross:
> I don't exactly know, why this problem here (`5 = 1 + 4`) cannot be solved like that (`4 = 2 * 2`), too. Is it because I 'break the constant up' myself here, rather than 'combing them' ?

The issue is that here we have `x + 4 * x` and we want to turn that into `5 * x`, whereas elsewhere we have `x * 2 * 2` and want to turn that into `x * 4`.  The rewriting pass is very syntactic.  https://github.com/mit-plv/fiat-crypto/commit/59600ae82eea35f7f0f8933be1beb38de27df4e7 says "if you see two constants in a multiplication list, fuse them".  There are not two constants in a multiplication list in `x + 4 * x`, so the rule doesn't apply.  We need a different syntactic code transformation here.

dderjoel:
I see. Thanks, that makes sense. If I should have an attempt with this, I'd need some pointers to similar rewrite rules or a code snipped, that I could try, as those rewrite-rules and proofs, are still a bit too intimidating for me to touch alone.

JasonGross:
This is actually a pretty tricky pass, unless you or @andres-erbsen has an idea about how to do it in a more algorithmically simple way.  Here's what I have so far:
```coq
Definition combines_to o := match o with add s => Some (mul s) | addZ => Some mulZ | _ => None end.

Fixpoint combine_to_consts_map (o : op) (args : list expr) (mapping : list (expr * list Z))
  : list (expr * list Z)
  := match args with
     | [] => mapping
     | arg :: args
       => let mapping
            := match combines_to o with
               | Some o'
                 => if commutative o' && associative o'
                    then
                      let '(e, vs)
                        := match arg with
                           | ExprApp (o'', args')
                             => if op_beq o' o''
                                then
                                  let csts_exprs := List.partition isCst args' in
                                  match interp0_expr (ExprApp (o', fst csts_exprs)) with
                                  | Some v => (ExprApp (o', snd csts_exprs), [v])
                                  | None => (arg, [])
                                  end
                                else (arg, [])
                           | _ => (arg, [])
                           end in
                      (n <- List.indexof (fun '(e', _) => expr_beq e e') mapping;
                       e_v' <- List.nth_error mapping n;
                       Some (set_nth n (e, vs ++ snd e_v') mapping))
                        ;;;
                        ((e, vs) :: mapping)
                    else (arg, []) :: mapping
               | None => (arg, []) :: mapping
               end in
          combine_to_consts_map o args mapping
     end%bool%option.

(* combines_to o = Some o' *)
Definition compress_consts_map (o o' : op) (mapping : list (expr * list Z)) : option expr
  := (args <-- (List.map
                  (fun '(e, vs) =>
                     match vs with
                     | [] => Some e
                     | _
                       => (v <- interp0_op o vs;
                           i <- identity o;
                           Some
                             (if (v =? i)%Z
                              then e
                              else ExprApp (o', [ExprApp (const v, []); e])))
                     end)
                  mapping);
      Some (ExprApp (o, args)))%option.

Definition combine_consts : expr -> expr :=
  fun e => match e with ExprApp (o, args) =>
    if commutative o then match combines_to o with
    | Some o' =>
        let mapping := combine_to_consts_map o args [] in
        match compress_consts_map o o' mapping with
        | Some e => e | None => e
        end
    | None => e end else e | _ => e end.
```

Now I need to come up with the right invariants and a proof strategy about how rearranging expressions like this is fine.

dderjoel:
I just had a thought, of restructuring `4*x + x` to `(x+x+x+x)+x` (one could do that maybe in the parsing pass) and then rewrite that as 5*x`;

JasonGross:
> I just had a thought, of restructuring `4*x + x` to `(x+x+x+x)+x` (one could do that maybe in the parsing pass) and then rewrite that as 5*x`;
> Edit: I don't know if that is helpful (at all, just a thought I had)

I'm concerned this would be too expensive, if, e.g., we every multiply by 2^10 or something, and end up making an addition of 2^10 identical terms.  We could do this only for small constants, but I think that'd be even worse.

Also, heads up, I'll be pretty busy for the next two weeks and might not get to this before then.  If you or @andres-erbsen wants to take a stab at it in the meantime, that'd be great.

dderjoel:
I'd help where I can, but I doubt I'd be any useful.
So adding only your code snippet to `Symbolic.v` and adding the `combine_const` to the list of rewrite rules, results in 
```
File "./src/Assembly/Symbolic.v", line 1469, characters 0-4:
Error: Some unresolved existential variables remain
```
And remembering from the other issue, I added
`Require Import Crypto.Util.Tactics.WarnIfGoalsRemain.` and very empty proofs, with just
```coq
Global Instance compress_consts_map_ok : Ok compress_consts_map.
Proof using Type.
  Unshelve. all: shelve_unifiable.
  all: fail_if_goals_remain ().
Qed.

Global Instance combine_consts_ok : Ok combine_consts.
Proof using Type.
  Unshelve. all: shelve_unifiable.
  all: fail_if_goals_remain ().
Qed.
```
after both of the `Definition`s. But the output does not help me (after adding both "Proofs"):
```
The term "compress_consts_map" has type
 "op -> op -> list (expr * list Z) -> option expr"
while it is expected to have type "expr -> expr" (cannot unify 
"expr" and "op").
```
and after only adding the latter proof, the compiler errors out with
```
WARNING: Remaining goal:
============================
(Ok combine_consts)
File "./src/Assembly/Symbolic.v", line 1436, characters 7-30:
Error: Tactic failure: A goal remains.
```
which does not surprise me, as there is no actual proof.

Also I do understand what we are trying to rewrite (high-level), but I don't understand the rewrite-code (like, I get the first line: we rewrite `add`s to mul`s, that's it), let alone could come up with any Proof myself.

I'm happy to wait, though.

JasonGross:
Here's a hoprefully-more-understandable version on top of #1140:
```coq
Definition combines_to o := match o with add s => Some (mul s) | addZ => Some mulZ | _ => None end.

(* o is like mul *)
(* invariant: Forall2 (fun x '(y, z) => eval (o x i) matches eval (o y z)) input output *)
Definition split_consts (o : op) (i : Z) : list expr -> list (expr * Z)
  := List.map
       (fun e
        => match e with
           | ExprApp (o', args)
             => if op_beq o' o
                then
                  let '(csts, exprs) :=
                    if commutative o' && associative o'
                    then let '(csts, exprs) := List.partition isCst args in
                         (interp0_expr (ExprApp (o', csts)), exprs)
                    else
                      match args with
                      | [arg; ExprApp ((const c), _)]
                        => (Some c, [arg])
                      | _ => (Some i, args)
                      end
                  in
                  match csts, exprs with
                  | None, args => (ExprApp (o', args), i)
                  | Some c, [arg] => (arg, c)
                  | Some c, args => (ExprApp (o', args), i)
                  end
                else (e, i)
           | _ => (e, i)
           end%bool).

(* invariant: input is a permutation of concat (List.map (fun '(e, zs) => List.map (pair e) zs) output) *)
Definition group_consts (ls : list (expr * Z)) : list (expr * list Z)
  := Option.List.map
       (fun xs => match xs with
                  | [] => None
                  | (e, z) :: xs => Some (e, z :: List.map snd xs)
                  end)
       (List.groupAllBy (fun x y => expr_beq (fst x) (fst y)) ls).

(* o is like add *)
(* spec: if interp0_op o zs is always Some _, then Forall2 (fun '(e, zs) '(e', z) => e = e' /\ interp0_op o zs = Some z) input output *)
Definition compress_consts (o : op) (ls : list (expr * list Z)) : list (expr * Z)
  := List.flat_map
       (fun '(e, zs) => match interp0_op o zs with
                        | None => List.map (pair e) zs
                        | Some z => [(e, z)]
                        end)
       ls.

(* o is like mul *)
(* spec is that Forall (fun '(e, z) e' => o (eval e) z matches eval e') inputs outputs *)
Definition app_consts (o : op) (ls : list (expr * Z)) : list expr
  := List.map (fun '(e, z) => let z := ExprApp (const z, []) in
                              match e with
                              | ExprApp (o', args)
                                => if op_beq o' o
                                   then ExprApp (o, args ++ [z])
                                   else ExprApp (o, [e; z])
                              | _ => ExprApp (o, [e; z])
                              end)
              ls.

Definition combine_consts : expr -> expr :=
  fun e => match e with ExprApp (o, args) =>
    if commutative o && associative o then match combines_to o with
    | Some o' => match identity o' with
    | Some idv =>
        ExprApp (o, app_consts o' (compress_consts o (group_consts (split_consts o' idv args))))
    | None => e end | None => e end else e | _ => e end%bool.

Global Instance combine_consts_0k : Ok combine_consts.
Proof using Type.
  repeat (step; eauto; []).
...
```

dderjoel:
okay, I Understand the last part, I think. we split `n * x` into `n` and `x`; then we group all those by `x`:`n`; then we compress those, i.e. `x`:`n`  and `x`:`m` -> `x`:`m+n`. But then what does that `app_consts` do ?

JasonGross:
> But then what does that `app_consts` do ?

It converts a pair of an expression and a Z back into an expression (e.g., (x, 5) -> x * 5).  It does a little bit more than this: if the expression is already a product, it just appends the constant to the list of things being multiplied, rather than having nested multiplications.

dderjoel:
yes, okay, that makes sense, and indeed is more straight-forward than the previous solution.
However, I still don't see myself helpful on attempting to proof any of those correct.

JasonGross:
I've force-pushed a rebase on current master, I'm going to take another stab at this

JasonGross:
@dderjoel I've finished the proofs about `combines_to`.  Now the errors are things like
```
File "./src/Assembly/WithBedrock/Semantics.v", line 109, characters 16-30:
Error:
In environment
sa : N
st : machine_state
a : MEM
The term "mem_base_reg a" has type "option REG"
while it is expected to have type "REG".
```

Would you take over again?

dderjoel:
the semantics are getting compiled now. But it errors out in the `SymbolicProofs.v` with:
```
File "./src/Assembly/WithBedrock/SymbolicProofs.v", line 486, characters 9-17:
Error: Unable to apply lemma of type
 "forall (s : symbolic_state) (m : machine_state),
  R s m ->
  forall (r : REG) (i : idx) (s' : symbolic_state),
  GetReg r s = Success (i, s') ->
  R s' m /\ s :< s' /\ eval s' i (get_reg m r)" on hypothesis of type
 "match
    Syntax.mem_base_reg
      {|
        mem_is_byte := mem_is_byte;
        mem_base_reg := mem_base_reg;
        mem_scale_reg := mem_scale_reg;
        mem_offset := mem_offset
      |}
  with
  | Some r => GetReg r
  | None => Symbolic.App (const 0, [])
  end s = Success (base, s0)".
```
which makes sense, because `mem_extra_reg` does not exist anymore. But I don't know what this part of the code is supposed to do. and thus don't know how to change it with the new syntax.

JasonGross:
It's actually failing on `eapply GetReg_R in HSbase`.  Can you paste the goal here right before the `eapply` on both master and this PR?

JasonGross:
Also, @andres-erbsen why is there no `step_GetReg` tactic, while there's a step for all the other things like https://github.com/mit-plv/fiat-crypto/blob/37cfe57e2b7b444ec2ec96029b0e10a34404cbad/src/Assembly/WithBedrock/SymbolicProofs.v#L502-L506

dderjoel:
> It's actually failing on `eapply GetReg_R in HSbase`. Can you paste the goal here right before the `eapply` on both master and this PR?

[master](https://github.com/mit-plv/fiat-crypto/blob/master/src/Assembly/WithBedrock/SymbolicProofs.v#L469-L500)
```coq
Lemma GetReg_R s m (HR: R s m) r i s'
  (H : @GetReg r s = Success (i, s'))
  : R s' m  /\ s :< s' /\ eval s' i (get_reg m r).
Proof using Type.
  cbv [GetReg GetReg64 bind some_or get_reg index_and_shift_and_bitcount_of_reg] in *.
  pose proof (get_reg_R s _ ltac:(eassumption) (reg_index r)) as Hr.
  destruct Symbolic.get_reg in *; [|inversion H]; cbn in H.
  specialize (Hr _ eq_refl); case Hr as (v&Hi0&Hv).
  rewrite Hv; clear Hv.
  step_symex; eauto.
  repeat (eauto || econstructor); cbn [interp_op].
Qed.

Lemma Address_R s m (HR : R s m) sa o a s' (H : @Symbolic.Address sa o s = Success (a, s'))
  : R s' m /\ s :< s' /\ exists v, eval s' a v /\ @DenoteAddress sa m o = v.
Proof using Type.
  destruct o as [? ? ? ?]; cbv [Address DenoteAddress] in *; repeat step_symex.
  eapply GetReg_R in HSbase; eauto; []; DestructHead.destruct_head'_and.
  destruct mem_extra_reg; cbn -[GetReg] in *.
  1: eapply GetReg_R in HSindex; eauto; []; DestructHead.destruct_head'_and.
  all: destruct mem_offset; cbn -[GetReg] in *.
  all: (step_symex; try solve [repeat (eauto || econstructor)]; []).
  all: (step_symex; try solve [repeat (eauto || econstructor)]; []).
  all: (step_symex; try solve [repeat (eauto || econstructor)]; []).
  3,4: (step_symex; try solve [repeat (eauto || econstructor)]; []).
  all : cbn in *.
  all : Tactics.ssplit; eauto 99 with nocore.
  all : eexists; split; eauto; [].
  all : cbv [DenoteConst].
  all : rewrite ?Z.add_0_r, ?Z.land_ones, ?Z.shiftr_div_pow2 by lia.
  all : push_Zmod; pull_Zmod; split; trivial.
Qed.

```
[PR](https://github.com/dderjoel/fiat-crypto/blob/mem-extention/src/Assembly/WithBedrock/SymbolicProofs.v#L469-L500):
```coq
Lemma GetReg_R s m (HR: R s m) r i s'
  (H : @GetReg r s = Success (i, s'))
  : R s' m  /\ s :< s' /\ eval s' i (get_reg m r).
Proof using Type.
  cbv [GetReg GetReg64 bind some_or get_reg index_and_shift_and_bitcount_of_reg] in *.
  pose proof (get_reg_R s _ ltac:(eassumption) (reg_index r)) as Hr.
  destruct Symbolic.get_reg in *; [|inversion H]; cbn in H.
  specialize (Hr _ eq_refl); case Hr as (v&Hi0&Hv).
  rewrite Hv; clear Hv.
  step_symex; eauto.
  repeat (eauto || econstructor); cbn [interp_op].
Qed.

Lemma Address_R s m (HR : R s m) sa o a s' (H : @Symbolic.Address sa o s = Success (a, s'))
  : R s' m /\ s :< s' /\ exists v, eval s' a v /\ @DenoteAddress sa m o = v.
Proof using Type.
  destruct o as [? ? ? ?]; cbv [Address DenoteAddress] in *; repeat step_symex.
  eapply GetReg_R in HSbase; eauto; []; DestructHead.destruct_head'_and.
  destruct mem_extra_reg; cbn -[GetReg] in *.
  1: eapply GetReg_R in HSindex; eauto; []; DestructHead.destruct_head'_and.
  all: destruct mem_offset; cbn -[GetReg] in *.
  all: (step_symex; try solve [repeat (eauto || econstructor)]; []).
  all: (step_symex; try solve [repeat (eauto || econstructor)]; []).
  all: (step_symex; try solve [repeat (eauto || econstructor)]; []).
  3,4: (step_symex; try solve [repeat (eauto || econstructor)]; []).
  all : cbn in *.
  all : Tactics.ssplit; eauto 99 with nocore.
  all : eexists; split; eauto; [].
  all : cbv [DenoteConst].
  all : rewrite ?Z.add_0_r, ?Z.land_ones, ?Z.shiftr_div_pow2 by lia.
  all : push_Zmod; pull_Zmod; split; trivial.
Qed.
```

JasonGross:
> Is that what you mean ?

Ah, no, I meant the output of sticking `Require Import Crypto.Util.PrintGoal. print_context_and_goal ().` right before the `eapply GetReg_R` line.

src/Assembly/WithBedrock/SymbolicProofs.v
```diff
@@ -483,6 +483,7 @@ Lemma Address_R s m (HR : R s m) sa o a s' (H : @Symbolic.Address sa o s = Succe
   : R s' m /\ s :< s' /\ exists v, eval s' a v /\ @DenoteAddress sa m o = v.
 Proof using Type.
   destruct o as [? ? ? ?]; cbv [Address DenoteAddress] in *; repeat step_symex.
+  Require Import Crypto.Util.PrintGoal. print_context_and_goal ().
```
JasonGross, line 486: Oops, sorry
```suggestion
  Require Import Crypto.Util.Tactics.PrintGoal. print_context_and_goal ().
```

JasonGross:
I've fixed the proofs, rebased on master, and force-pushed.  Hopefully everything should compile now

JasonGross:
Now the error message is:
```
Index 2063 loaded without being present.
2063 is (add 64, [#1088, #1179, #2061])
2061 is (mul 64, [#2057])
2057 is (const 18446744073709551288, [])
1179 is (mul 64, [#221, #1178])
1178 is (const 280, [])
1088 is (mul 64, [#221, #1086])
1086 is a special value no longer present in the symbolic machine state at the end of execution.
221 is (const 1, [])
```

dderjoel:
> Does the symbolic executor load from memory where is supposed to only compute the address or something?

Is that a question to me ? Afaik no. At least it should not. 

Where do we see that error ?
the only new file is https://github.com/mit-plv/fiat-crypto/pull/1134/files#diff-4b90f9bbb120f4e25888c642079f00f617a8e24ccda574d9216640086646fc19 and it not loads data from any uninitialized memory. It solely uses `lea` for those. 
the file above is a manually fabricated version of this one [here](https://github.com/mit-plv/fiat-crypto/blob/master/fiat-amd64/10.67_ratio11074_seed2040113663254706_square_poly1305.asm).

JasonGross:
The issue, I think, is that it's accessing the same memory via multiple different computations that the symbolic evaluator can't see are the same.  I've tried sprinkling in a couple extra passes in some places, but it doesn't seem to be enough.  I'm now inclined to run the rewriting twice, unless @andres-erbsen has other ideas for getting complete rewriting.  (My only other idea is implementing NBE-style rewriting, which adds a lot of overhead.)

dderjoel:
> I think, is that it's accessing the same memory via multiple different computations that the symbolic evaluator can't see are the same.

I don't quite understand that. Is, what you're saying, that from asm (`lea x3, [x1 + x2]`) and from the internal structure (`u64 x3 = x1 + x2`), we're 'accessing' `x1 + x2` and the evaluator doesn't see they're the same? I mean we're not accessing the memory at `x1 + x2`, right? 

I mean, is the flow evaluate the expression in `[]`, then access the memory, and in case of `lea` return the address? Then it makes sense that we're 'reading' uninitialized memory (like as a intermediate step). Could we maybe do it the other way around? calculate the address and if !lea, return the data from that address? Or abuse the `MEM` structure in case of a `lea` just as a means to specify  arithmetic of certain structures (that is `x + z * y + z'`)?

JasonGross:
Ah, I figured out the issue.  The subsequent rewriting passes occur only at the top level of the output of the previous one.  I just need to be a bit more careful generating the output, so there aren't further simplification opportunities.

JasonGross:
Roughly the issue is that `app_consts` leaves over spurious `x * 1` that don't get removed by subsequent passes.

JasonGross:
Issue fixed, all assembly tests now pass on my machine

dderjoel:
nice, Thank you!
I can also build and 'many' files that I generate get proven correct.
But there is some interesting behavior. I've added another file for curve25519; 
It has to do with the 'multiply by 19' in with `lea`s

There is actually two occurrences oft that. [Lines 17-20](https://github.com/mit-plv/fiat-crypto/pull/1134/files#diff-c6ab46bff21ea598bb6f427d2e3ab625a9fee22ef8dad0712368c734bec77636R17-R20) and [lines 131-136](https://github.com/mit-plv/fiat-crypto/pull/1134/files#diff-c6ab46bff21ea598bb6f427d2e3ab625a9fee22ef8dad0712368c734bec77636R131-R137)

The first mul can be split up into `lea`s but for the second one it fails with
`../src/ExtractionOCaml/unsaturated_solinas  --no-primitives --no-wide-int --shiftr-avoid-uint1 --output /dev/null --output-asm /dev/null --tight-bounds-mul-by 1.000001 'curve25519' '64' '5' '2^255 - 19' carry_square --hints-file ./9789_ratio10114_seed1785685356_square_curve25519.asm`
```
Unable to unify: [inr [890, 895, 894, 189, 200]] == [inr [207, 212, 214, 189, 200]]
Could not unify the values at index 0: [#890, #895, #894, #189, #200] != [#207, #212, #214, #189, #200]
index 0: #890 != #207
(slice 0 51, [#889]) != (slice 0 51, [#204])
index 0: #889 != #204
(add 64, [#108, #199, #887]) != (add 64, [#108, #202])
(add 64, [#199, #887]) != (add 64, [#202])
(add 64, [#199, (mul 64, [#9, #886])]) != (add 64, [(mul 64, [#5, #199])])
(add 64, [#199, (mul 64, [#9, (add 64, [#199, #885])])]) != (add 64, [(mul 64, [#5, #199])])
(add 64, [#199, (mul 64, [#9, (add 64, [#199, (mul 64, [#199, #237])])])]) != (add 64, [(mul 64, [#5, #199])])
(add 64, [#199, (mul 64, [#9, (add 64, [#199, (mul 64, [#199, (const 8, [])])])])]) != (add 64, [(mul 64, [#5, #199])])
(add 64, [(or 64, [#195, #197]), (mul 64, [(const 2, []), (add 64, [(or 64, [#195, #197]), (mul 64, [(or 64, [#195, #197]), (const 8, [])])])])]) != (add 64, [(mul 64, [(const 19, []), (or 64, [#195, #197])])])
```

full output in attached file.
[err.txt](https://github.com/mit-plv/fiat-crypto/files/8321022/err.txt)

dderjoel:
another test case in the `poly1305` case.
this one is unable to unify 
```Unable to unify: [inr [351, 349, 350]] == [inr [103, 108, 110]]
Could not unify the values at index 0: [#351, #349, #350] != [#103, #108, #110]
index 0: #351 != #103
(slice 0 44, [#345]) != (slice 0 44, [#100])
index 0: #345 != #100
(add 64, [#58, #95, #343]) != (add 64, [#58, #98])
(add 64, [#95, #343]) != (add 64, [#98])
(add 64, [#95, (mul 64, [#95, #331])]) != (add 64, [(mul 64, [#3, #95])])
(add 64, [#95, (mul 64, [#95, (const 4, [])])]) != (add 64, [(mul 64, [#3, #95])])
(add 64, [(or 64, [#91, #93]), (mul 64, [(or 64, [#91, #93]), (const 4, [])])]) != (add 64, [(mul 64, [(const 5, []), (or 64, [#91, #93])])])
```
where I guess the problem is the `mul * 5`
```
(add 64, [
    (or 64, [#91, #93]), 
    (mul 64, [(or 64, [#91, #93]), (const 4, [])])
])
!=
(add 64, [
    (mul 64, [(const 5, []), 
    (or 64, [#91, #93])])
])
```

dderjoel:
I think both of the cases have the same cause; but I wonder, why it works, in [lines 19-20](https://github.com/mit-plv/fiat-crypto/pull/1134/files#diff-c6ab46bff21ea598bb6f427d2e3ab625a9fee22ef8dad0712368c734bec77636R19-R20). The only difference I see, is that there the argument is an argument from the input.

JasonGross:
Unfortunately rewriting is hard to debug.  Could you restore this PR to it's previous (succeeding) state, and open a new one adding the failing files?  (Alternatively you could make a new branch and a new PR, and I can force-push to this branch to restore it to passing.)  I'd like to merge this tomorrow to avoid merge conflicts with my other work in progress.

As far as debugging the issue, could you try (in the new PR) adding a second copy of the `combine_consts` pass at the end of the list in `Definition expr` in Symbolic.v?  If this changes nothing, it means there is an issue with the rewriting pass; if it changes something, it means that one of the later passes is exposing an opportunity that gets passed up.

> Is an empty summand treated as `0`? Or another way of asking: Why is there an addition in the first place?

Where do you see an empty summand?  But, yes, `(add 64, [])` is 0, and `(add 64, [x])` is `x mod 2^64`.  The reason we see `add 64` with a single argument here is an artifact of the explanation process; I strip out common indices from the top-level addition. (This is what happens in going from `(add 64, [#58, #95, #343]) != (add 64, [#58, #98])` to 
`(add 64, [#95, #343]) != (add 64, [#98])`.)  Maybe this is too confusing and I should instead move them to the end or something?

At a glance, your analysis seems correct to me, btw.

JasonGross:
I've merged #1166, which contains all but the new examples of this PR

JasonGross:
My current guess is that the code turns `add 64, [#95, #343]` into `add 64, [(or 64, [#91, #93]), (mul 64, [#95, #331])]` and thereafter misses the fact that `(or 64, [#91, #93])` and `#95` are the same.  Instead I guess it should only reveal as needed, or else it should reveal an extra level when it finds multiplications.

JasonGross:
I've implemented this in #1167, which makes all the tests here pass

dderjoel:
So this now contains all the changes, right? I'm building from this then and check from my side.

JasonGross:
I've rebased and force-pushed atop the now-merged #1167, and I will merge this once the CI passes

JasonGross:
All non-windows tests build, and the windows one seems to build other than maybe hanging at the very end, so I'm going to merge this.

PR #1136, opened 2022-02-21
Better explanation of unification error messages

JasonGross:
On top of #1135

Hopefully this will help with the error messages from #1134


dderjoel:
```
[(add 64, [
    (mul 64, [
        (const 2, []),
        (add 64, [
            #0,
            (mul 64, [
                #0,
                (const 4, [])
            ])
         ])
    ]),
    (mul 64, [
        (const 2, []),
        (add 64, [
            #0,
            (mul 64, [
                #0,
                (const 4, [])
            ])
        ])
    ])
])] != [#0, (const 20, [])]
```

would such a formatting be too bloated, or does it help to read (I am really not sure myself :D)?
Also, where is the operation on the right side? I can only see the arguments `[#0, (const 20, [])]`. For me, now, its clear, that `#0` shall be `mul`tiplied with `20`, but I don't see that in that resolve- chain.

Also I *really* like the addition of `#0` to be able to distinguish easily from `0`

JasonGross:
> Also, where is the operation on the right side? I can only see the arguments `[#0, (const 20, [])]`. For me, now, its clear, that `#0` shall be `mul`tiplied with `20`, but I don't see that in that resolve- chain.

That's a good point, I'll add the top-level operation on both sides.

> would such a formatting be too bloated, or does it help to read (I am really not sure myself :D)?

I think this is too bloated, and I'm also disinclined to figure out the indentation code.

PR #1165, opened 2022-03-22
Inline all functions in C, even in clang

### LINKED ISSUE #1164: Unused functions in fiat-c trip -Wunused-function
GCC and Clang's -Wunused-function seem to trip unused functions if the consumer doesn't use every one of them. I'm currently looking at wrapping the include in pragmas, but this is kind of annoying. Some ideas...

1. Tag all the generated functions with `__attribute__((unused))`, or...
2. Mark every function `static inline`, which reliably silences the warning in GCC and Clang. One detail to keep in mind, since fiat's currently' doing `__inline__` with a `__GNUC__` check: while Clang usually defines `__GNUC__`, it doesn't when building as `clang-cl` (MSVC-like Clang). I usually write `defined(__GNUC__) || defined(__clang__)` to check for GCC extension support. Or you could just use plain `static inline` which should work on remotely recent compilers, or...
3. Stick some macro in front of every function, so the consumer can customize it.

JasonGross:
Fixes #1164

Maybe we should do #820?

Let's see if the CI will generate the C files for us


JasonGross:
Success, the CI has generated the files for us

davidben (approved):
Thanks! This seems to avoid the warnings for us.

PR #1167, opened 2022-03-22
Add reveal_at_least, a more clever form of reveal

JasonGross:
Depth determines which indices get expanded, but all references to the
same index get expanded if they appear in the output.

This is because Joel's latest examples require a rewriting pass that
needs either uneven revealing or the ability to check expression
equality modulo the dag in the middle of rewriting.
https://github.com/mit-plv/fiat-crypto/pull/1134#issuecomment-1075320548
Roughly the issue is that if we want to turn `a + 4 * a` into `5 * a`,
we need to reveal enough structure to see `4 * a`, but we need to see
that the two instances of `a` are the same (e.g., if `a` is an ExprRef
pointing to `b|c`, and we reveal uniformly, then we need to recognize
`b|c + 4 * a`)


JasonGross:
Windows CI failed with the periodic “Terminate Batch Job (Y/N)?”, see e.g., https://github.community/t/windows-github-action-randomly-freezes/16690/3
Everything seems to have built fine other than that, though.

Issue #1170, opened 2022-03-23
Windows CI is too slow


JasonGross:
Since we added a couple more AMD64 tests, the Windows build has been timing out.

Current CI Build times are:

Stage \ Version | [Windows (8.12.0, no bedrock)](https://github.com/mit-plv/fiat-crypto/runs/5656129903?check_suite_focus=true) | [MacOS (8.14.0, with bedrock)](https://github.com/mit-plv/fiat-crypto/runs/5656130024?check_suite_focus=true) | [Ubuntu LTS (8.11.0, no bedrock)](https://github.com/mit-plv/fiat-crypto/runs/5652949675?check_suite_focus=true) | [Ubuntu (master, with bedrock)](https://github.com/mit-plv/fiat-crypto/runs/5652949651?check_suite_focus=true)
-|-|-|-|-
Install OCaml | 2m45s | 2m58s | - | -
Install Coq | 15m10s | 9m53s | 1m28s | 1m27s
Build deps | 17m56s | 24m32s | 9m3s | 20m27s
Build all | 3h9m48s | 1h52m30s | 1h25m47s | 1h35m10s
Test AMD 64 files | > 2h13m44s | 1h17m9s | 1h1m27s | 1h2m58s
Build Haskell binaries | - | - | 22m16s | 26m7s
validate | - | 1h37m20s | 48m33s | 1h22m48s
**Total** | > 6h | 5h26m27s | 3h49m48s | 4h50m44s

Part of why Windows is so much slower is that we need to build single-threaded to not run out of RAM.  I guess maybe I can multithread the AMD64 tests?  But we should maybe think about eliminating some of the slower AMD64 tests?  Or performance-optimizing the equivalence checker?  cc @dderjoel @andres-erbsen

dderjoel:
1. Could we put multiple hint-files per invocation, such that the binary doesn't need to start from scratch synthesizing the IR that the amd file need to match? 

1. I think the files that were there before the last couple extensions were just a random selection. We could definitely scrap some of them, or merge them such that all the constructs we support are in one (fabricated) file.

I tend to do both actually, I don't know how much work the first is and if it'd be worth it. (It'd would help my project a bit too. I could then check all emitted files at the end once, rather than checking every time I write a file, but that is super-low priority).

And finally, testing in parallel makes sense. Is -j2 already enough ?

JasonGross:
> Is -j2 already enough ?

[...]

PR #1172, opened 2022-03-23
[CI] [Windows] Use -j2

JasonGross:
Maybe this will fix the issue in #1170


JasonGross:
Let's try this for now

PR #1191, opened 2022-04-13
Add some operations to equivalence checker

JasonGross:
We now support `mul`, `push`, `pop`.  <strike>On top of #1197.</strike>

@andres-erbsen @davidben do either of you want to review my implementation of push, pop, etc?


JasonGross:
@andres-erbsen @davidben Could you review at least the changes in Semantics.v and Symbolic.v (i.e., the non-proof changes)?

davidben (commented):
I don't think I understand this code enough to usefully review it, but it seems a plausible description of push/pop/mul.

PR #1196, opened 2022-04-16
Add support for TIMED=1 in make test-amd64-files

JasonGross:
Also make only-test-amd64-files actually not rebuild/depend on the
binaries, and make Haskell error messages line up with OCaml ones
(always printed to stdout).


JasonGross:
Windows build failure is OOM, I guess we have to adjust some options

PR #1205, opened 2022-04-19
Use `perf record -g` on CI and save perf data for later analysis

JasonGross:
On top of #1202


JasonGross:
I've asked about this error message [on stack exchange] (https://unix.stackexchange.com/questions/699708/how-do-i-debug-perfs-permission-error-mapping-pages-when-running-perf-on); let's make this draft for now.

Issue #1215, opened 2022-04-25
Abstract Field Specs for Bedrock2 functions


RasmusHoldsbjergCSAU:
Continuing on the discussion here #1150 

I have made modifications to the Bedrock2 specs for field operations that allow for arbitrary finite fields, as opposed to the current versions that have a field modulus hard-coded, and thus allowing only for prime order fields.

The changes can be found in [this](https://github.com/RasmusHoldsbjergCSAU/fiat-crypto) fork of fiat-crypto.

The main changes are to the ```Bedrock/Specs/Field.v``` file, which has been split into two files; [```Bedrock/Specs/AbstractField.v```](https://github.com/RasmusHoldsbjergCSAU/fiat-crypto/blob/master/src/Bedrock/Specs/AbstractField.v) and [```Bedrock/Specs/PrimeField.v```](https://github.com/RasmusHoldsbjergCSAU/fiat-crypto/blob/master/src/Bedrock/Specs/PrimeField.v)

```AbstractFields.v``` is similar to the current ```Field.v```, except that it does not require the field to have prime order.
```PrimeField.v``` constructs instances of prime-order fields given a prime modulus. These instances should be similar to the ones in the current ```Field.v```.

These changes propagate outwards, but adapting the pipeline for new specs has been mostly a matter of small superficial changes.

If there is interest, I will gladly open a pull request to have this integrated into fiat-crypto.

There are a few problems, however, that perhaps should be dealt with first; namely the interaction with Rupicola is not as smooth as one could hope for.

Notice, for instance, the compiler hints [here](https://github.com/RasmusHoldsbjergCSAU/fiat-crypto/blob/master/src/Bedrock/Group/AdditionChains.v#L433) for squaring and multiplying field elements. These are highly adapted to the particular use in this particular derivation, in contrast to the [current versions](https://github.com/mit-plv/fiat-crypto/blob/master/src/Bedrock/Field/Interface/Compilation2.v#L342), which are more general. Clearly one would prefer the more general hints.

For some reason, the compiler gets stuck when I try to use the current versions of compiler hints with the new Bedrock2 specs. If anyone with insight, @andres-erbsen, @jadephilipoom or even @cpitclaudel, could comment on this, it would be much appreciated.

andres-erbsen:
Thank you for contributing. I think everybody on our end has been unusually busy of late. I do intend to get back to this and I think we are interested.

jadephilipoom:
[...]

PR #1220, opened 2022-04-27
Abstract Field specs for Bedrock2 functions

RasmusHoldsbjergCSAU:
Opening a pull request for the changes suggested in #1215


jadephilipoom (commented):
Finally had a chance to review this in detail! A lot of minor comments, but only one that is blocking for me.

It looks like typeclass inference is really struggling to use `prime_field_parameters` as an `AbstractField` instance. This means a lot of manually plugging in e.g. `@mul prime_field_parameters` everywhere, and also is (I suspect) behind the need to re-declare `spec_of` instances in several places. I'd like to figure this out and resolve it before merging, since typeclass inference struggles are annoying to debug and impose a large cost on future code.

Another, more minor concern for me is that `PrimeFieldParameters` are used in a lot of cases where I'd expect `FieldParameters`. I think we discussed how some lemmas from `PrimeFieldTheorems` are needed -- but `PrimeFieldTheorems` basically is not relevant to anything other than `inv` when using Fermat's Little Theorem. So ideally, I would like to use `FieldParameters` for nearly everything and then only introduce the prime-order special case for `inv` -- is there a reason that's not feasible?

RasmusHoldsbjergCSAU:
I have now made a commit based on the requests.

I have not been able to resolve the issues with typeclass inference; so these have been left as is.

Otherwise a number of minor changes and comments added as appropriate

jadephilipoom:
I spent a while today trying to rebase this code and resolve the typeclass inference problem, but I couldn't really do it in a satisfactory way because of the general amount of tech debt in this part of the codebase -- my fault, not yours, and I'm sorry for the pain. With some refactors to make the interfaces cleaner and simpler this would work. However, I don't currently have the time to undertake those, and I think unfortunately adding this change on top of the interfaces we have would create too much of a barrier for future code.

You can see my (still not quite working) version here in case you want to use the rebase: https://github.com/jadephilipoom/fiat-crypto/tree/rasmus-abstract-field

A note on that code: one of the issues I ran into while rebasing is that the `to_bytes` compilation lemma (which didn't yet exist in the version of the code the PR is based on) is not really provable in the abstract field because there's no way to know how a field element should be broken down into bytes without the correspondence to `Z`. I started trying to resolve this by adding an extra field to the representation before I ran out of evening.

I think perhaps for now we should close this PR, and try to resurrect it once someone has the time to do a more fundamental refactor. At that point, what would really be ideal is to keep the single `Field.v` and ensure that it doesn't require the modulus to be prime in the first place; this should really only come into play for certain inversion algorithms.

I appreciate all the work that you put into this PR, and I'm very sorry about this.

andres-erbsen:
Yeah, ++ everything Jade said. This is a feature that we'd in principle want, but the pre-existing level of technical debt in that class hierarchy makes it too perilous to add it in the current form.

I'd also like to add that there's now a plan to add extension-field support to Arithmetic and Curves code built on top of fields, I'm not sure what that will look like but we'll be looking into your relevant code as a potential building-block or inspiration.

andres-erbsen:
Some work was done and I am unsure how far it got; I am following up with the implementers.

PR #1225, opened 2022-04-27
Table-based multiplication

talkon:
Based on Section 3.3 in https://eprint.iacr.org/2012/309.pdf


andres-erbsen:
rebuilding to see logs...

andres-erbsen:
Let's see whether it builds after https://github.com/mit-plv/fiat-crypto/commit/8c2a2e171cc639bd915a7134ad13ff79b68e6ac7 ...

PR #1238, opened 2022-05-18
Automatically skip bedrock2 on older versions of Coq

### LINKED ISSUE #1212: I cloned ed2105e0bab3ca36e95921c1b2856baa3e15b1df and running gmake fails
The error is:

```
gmake -C /home/hannes/fiat-crypto/rupicola/bedrock2/bedrock2 noex
Generating Makefile.coq.noex
rm -f .coqdeps.d
gmake -f Makefile.coq.noex
gmake[4]: *** No rule to make target '/home/hannes/fiat-crypto/rupicola/bedrock2/deps/coqutil/src/coqutil/Word/Properties.vo', needed by '/home/hannes/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2/SepLogAddrArith.vo'.  Stop.
gmake[3]: *** [Makefile.coq.noex:343: all] Error 2
gmake[2]: *** [Makefile:42: noex] Error 2
gmake[1]: *** [Makefile:102: bedrock2_noex] Error 2
gmake: *** No rule to make target 'bedrock2', needed by 'src/StandaloneOCamlMain.vo'.  Stop.
```

```
> coqc --version
The Coq Proof Assistant, version 8.13.1 (March 2021)
compiled on Mar 4 2021 19:25:57 with OCaml 4.12.0
```

```
> gmake --version
GNU Make 4.3
Built for amd64-portbld-freebsd12.1
Copyright (C) 1988-2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

JasonGross:
It's confusing to have `make` not work by default on older versions of
Coq, I think, so instead when `SKIP_BEDROCK2` is unset we will set it by
default on old versions of Coq.

Fixes #1212

### DIFF
--- a/.github/workflows/coq-windows.yml
+++ b/.github/workflows/coq-windows.yml
@@ -19,7 +19,6 @@ jobs:
     env:
       NJOBS: "2"
       COQ_VERSION: "8.12.0" # https://packages.debian.org/stable/coq
-      SKIP_BEDROCK2: "1"
       COQEXTRAFLAGS: "-async-proofs-j 1"
       COQCHKEXTRAFLAGS: ""
       OPAMYES: "true"
--- a/.github/workflows/coq.yml
+++ b/.github/workflows/coq.yml
@@ -15,9 +15,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - env: { COQ_VERSION: "master"    , COQ_PACKAGE: "coq libcoq-ocaml-dev", SKIP_BEDROCK2: "" , SKIP_VALIDATE: "", COQCHKEXTRAFLAGS: "-bytecode-compiler yes", PPA: "ppa:jgross-h/coq-master-daily", EXTRA_GH_REPORTIFY: "--warnings" }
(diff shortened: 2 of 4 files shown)


JasonGross:
> I'm still wondering shouldn't `make standalone-ocaml` (the target I use) be doing the least amount of work to get the command line utilities up and running? IIUC `bedrock2` is not required for the cli utilities -- isn't there an opportunity to adapt/minimize the dependencies (or should I use a different target)?

Two parts to the answer:
1. `bedrock2` is actually required for the cli utilities that can output bedrock2 code.
2. The dependencies (rewriter, bedrock2, coqutil, rupicola, coqprime) are treated as external projects.  So we require complete compilation of these dependencies before we begin compiling the files specific to fiat-crypto.  This allows us to support other modes of depending on these projects, for example assuming that they are already installed rather than building the ones in the submodules.  (Perhaps if we switched the build system to dune, we could support both modes of dependency and also have finer-grained incremental compilation?)

hannesm:
great, thanks for your answer.

PR #1259, opened 2022-05-26
Generate code for the scalar field of standard curves

jedisct1:
`fiat-crypto` is already used for arithmetic on scalars, at least by the Zig standard library and in Rust (`p384_rs`, being merged into the `p384` crate as we speak).

So it may be useful to pregenerate and test this in `fiat-crypto`.

This change adds support for the curve25519, p256, secp256k1, p384 and p521 scalar fields.


JasonGross:
This is nice, but I'm afraid fiat-crypto is too slow at larger curves to support this.  6+ hours of CI time is really too much.  Do you have thoughts on how to support this better?  Options I can think of:
1. Making this PR about adding 64-bit versions only (with the possible exception of curve25519)
2. Having slow code gen be tested in a separate CI job that runs after the main ones, possibly only on master, possibly pushing any newly generated files that it makes?

Also, this PR is missing Java files.

jedisct1:
CI is indeed a major issue.

Testing slow and fast cases in distinct jobs may be the way to go, especially since we already have such distinction with the "lite" files.

jedisct1:
> Also, this PR is missing Java files.

The Java files are included, but they were generated with funny names:

```text
FiatuLucurve2u5u5u1u9u.java
FiatuLucurve2u5u5u1u9uuscalar.java
```

This is a distinct issue, that also happens on `master`. This is on macOS; it probably doesn't happen on Linux.

tarcieri:
> Having slow code gen be tested in a separate CI job that runs after the main ones, possibly only on master, possibly pushing any newly generated files that it makes?

This seems reasonable to me

tarcieri:
Alternatively, if the scalar fields could be included in the `fiat-crypto` Rust crate releases, that would work just as well: https://crates.io/crates/fiat-crypto

jedisct1:
Removing P521 code generation makes it more reasonable. Due to its size, Java cannot compile the P521 scalar code anyway.

spitters:
https://github.com/ejgallego/coq-universe
tries to improve CI by better caching with the use of Dune. Could this be helpful here?

PR #1260, opened 2022-05-28
Generate Java files with the expected file names on all platforms

jedisct1:
This replaces the sed script that depends on GNU `sed` with portable `awk` code.

Fixes generation of Java files on macOS and BSD.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -226,8 +226,7 @@ ZIG_DIR := fiat-zig/src/
 
 # Java only really supports 32-bit builds, because we have neither 64x64->64x64 multiplication, nor uint128
 # Java also requires that class names match file names
-# from https://stackoverflow.com/q/42925485/377022
-to_title_case = $(shell echo '$(1)' | sed 's/.*/\L&/; s/[a-z]*/\u&/g')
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -226,8 +226,7 @@ ZIG_DIR := fiat-zig/src/
 
 # Java only really supports 32-bit builds, because we have neither 64x64->64x64 multiplication, nor uint128
 # Java also requires that class names match file names
-# from https://stackoverflow.com/q/42925485/377022
-to_title_case = $(shell echo '$(1)' | sed 's/.*/\L&/; s/[a-z]*/\u&/g')
+to_title_case = $(shell echo '$(1)' | awk '{split($0,w,"");u=1;for(i=1;i<=length(w);i++){c=tolower(w[i]);if(u)c=toupper(c);u=0;if(c~/[a-zA-Z0-9]/)printf("%s",c);if(c~/[^a-zA-Z]/)u=1;}}')
```
JasonGross, line 229: I assume you need to split `$$0` in bash and not `$0` in GNU Make?
```suggestion
to_title_case = $(shell echo '$(1)' | awk '{split($$0,w,"");u=1;for(i=1;i<=length(w);i++){c=tolower(w[i]);if(u)c=toupper(c);u=0;if(c~/[a-zA-Z0-9]/)printf("%s",c);if(c~/[^a-zA-Z]/)u=1;}}')
```

JasonGross (approved):
I might want to make this a bit more concise later, but it seems to work, thanks!

PR #1262, opened 2022-05-28
Add 1-level broadcasting-style array operation compilation and use for poly1305

DIJamner:


DIJamner:
Establishes the lemmas necessary for compiling poly1305 up until the main loop. A number of side-conditions and some separation logic goals are admitted to reach that point.

PR #1265, opened 2022-06-01
Fix some structures (missing Orig/Both)

JasonGross:


JasonGross:
Already merged as #1264

PR #1280, opened 2022-06-03
Friendlier GH Actions Queue

JasonGross:
We now cancel previous queued and in progress actions runs on PRs when a
PR is updated. Because [`github.head_ref` is only defined for PRs and
`github.run_id` is
unique-per-workflow](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value-1),
we won't cancel previous jobs on branches, meaning that we'll still see
a log of which commits succeeded and which ones failed.

### DIFF
--- a/.github/workflows/c.yml
+++ b/.github/workflows/c.yml
@@ -11,6 +11,10 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    concurrency:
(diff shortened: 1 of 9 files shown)


JasonGross:
Seems to work, and I want to use it in new PRs, so I'm going to merge it

PR #1296, opened 2022-06-12
Revert "Generate .mli files: `{ => Recursive} Extraction` (#1292)"

JasonGross:
This reverts commit d9f123fd330dab3fab551d69cd144a15dd0a77c5

It did not work for fixing the extraction, so no need to slow down
compilation.


JasonGross:
It does help with debugging type errors, though, so maybe I'll leave it

PR #1301, opened 2022-06-14
bump rupicola and adapt to bedrock2 compiler changes

samuelgruetter:
This should fix the failures observed in https://github.com/mit-plv/fiat-crypto/pull/1299.


andres-erbsen:
Thank you!

PR #1339, opened 2022-07-16
Overlay for Coq #16308 (Wildcards in _CoqProject)

jfehrle:


JasonGross (changes requested):
Is this necessary?  I'd prefer to have the files listed explicitly

JasonGross:
I expect this is also not backwards compatible

jfehrle:
Perhaps you didn't see my comment at https://github.com/coq/coq/pull/16308#issuecomment-1186089482?

> I'd prefer to have the files listed explicitly

Why do you prefer that?  I don't see how it's helpful.

PR #1309, opened 2022-07-01
Make hint globality explicit

JasonGross:
This is mostly an adaptation for coq/coq#16004

We also make some other warnings into errors at the same time.

PR #1343, opened 2022-07-25
Adapt w.r.t. coq/coq#16004.

ppedrot:
We should also bump the submodules, but this is a good first step.

Closes #1309

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,7 +114,7 @@ GREP_EXCLUDE_SPECIAL_VOFILES := grep -v '^src/Extraction\(OCaml\|Haskell\)/'
 ifneq (,$(wildcard .git/))
 SORT_COQPROJECT = sed 's,[^/]*/,~&,g' | env LC_COLLATE=C sort | sed 's,~,,g'
 EXISTING_COQPROJECT_CONTENTS_SORTED:=$(shell cat _CoqProject 2>&1 | $(SORT_COQPROJECT))
-WARNINGS_PLUS := +implicit-core-hint-db,+implicits-in-term,+non-reversible-notation,+deprecated-intros-until-0,+deprecated-focus,+unused-intro-pattern,+variable-collision,+unexpected-implicit-declaration,+omega-is-deprecated,+deprecated-instantiate-syntax,+non-recursive,+undeclared-scope#,+deprecated-hint-rewrite-without-locality,+deprecated-hint-without-locality,+deprecated-instance-without-locality,+deprecated-typeclasses-transparency-without-locality
+WARNINGS_PLUS := +implicit-core-hint-db,+implicits-in-term,+non-reversible-notation,+deprecated-intros-until-0,+deprecated-focus,+unused-intro-pattern,+variable-collision,+unexpected-implicit-declaration,+omega-is-deprecated,+deprecated-instantiate-syntax,+non-recursive,+undeclared-scope,+deprecated-hint-rewrite-without-locality,+deprecated-hint-without-locality,+deprecated-instance-without-locality,+deprecated-typeclasses-transparency-without-locality
 # Remove unsupported-attributes once we stop supporting < 8.14
 WARNINGS := $(WARNINGS_PLUS),unsupported-attributes
 COQPROJECT_CMD:=(echo '-R $(SRC_DIR) $(MOD_NAME)'; echo '-arg -w -arg $(WARNINGS)'; echo '-arg -native-compiler -arg ondemand'; ((echo "$(sort $(VERSION_DEPENDENT_FILES) $(SPECIAL_VERSION_DEPENDENT_FILES))" | tr ' ' '\n'; git ls-files 'src/*.v' | $(GREP_EXCLUDE_SPECIAL_VOFILES)) | $(SORT_COQPROJECT)))
--- a/_CoqProject
+++ b/_CoqProject
@@ -1,5 +1,5 @@
 -R src Crypto
--arg -w -arg +implicit-core-hint-db,+implicits-in-term,+non-reversible-notation,+deprecated-intros-until-0,+deprecated-focus,+unused-intro-pattern,+variable-collision,+unexpected-implicit-declaration,+omega-is-deprecated,+deprecated-instantiate-syntax,+non-recursive,+undeclared-scope,unsupported-attributes
+-arg -w -arg +implicit-core-hint-db,+implicits-in-term,+non-reversible-notation,+deprecated-intros-until-0,+deprecated-focus,+unused-intro-pattern,+variable-collision,+unexpected-implicit-declaration,+omega-is-deprecated,+deprecated-instantiate-syntax,+non-recursive,+undeclared-scope,+deprecated-hint-rewrite-without-locality,+deprecated-hint-without-locality,+deprecated-instance-without-locality,+deprecated-typeclasses-transparency-without-locality,unsupported-attributes
 -arg -native-compiler -arg ondemand
 src/BoundsPipeline.v
 src/CLI.v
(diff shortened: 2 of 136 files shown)


JasonGross:
Thank you!

Might be nice to add `,+deprecated-hint-rewrite-without-locality` to https://github.com/mit-plv/fiat-crypto/blob/eeedafeb6b53a1a00ebec1b0f7602e664684f050/Makefile#L117

JasonGross:
> We should also bump the submodules

dependabot will create the PRs to bump the submodules for us

ppedrot:
I've added the warnings to the makefile. What is the oldest version you're trying to support exactly? This PR will only work with Coq >= 8.14 with the warnings set up this way.

JasonGross:
We support 8.11.  Doesn't Coq ignore unknown warnings, though?  What's the issue with older versions?

ppedrot:
Shouldn't you get an unsupported attribute, given that Hint Rewrite locality is only available from 8.14 onwards?

JasonGross:
That's what https://github.com/mit-plv/fiat-crypto/blob/eeedafeb6b53a1a00ebec1b0f7602e664684f050/Makefile#L119 is for

Looks like the error message is that `_CoqProject` also needs to be updated (you can just do `make _CoqProject`)

Makefile
```diff
@@ -114,8 +114,7 @@ GREP_EXCLUDE_SPECIAL_VOFILES := grep -v '^src/Extraction\(OCaml\|Haskell\)/'
...
 EXISTING_COQPROJECT_CONTENTS_SORTED:=$(shell cat _CoqProject 2>&1 | $(SORT_COQPROJECT))
-WARNINGS_PLUS := +implicit-core-hint-db,+implicits-in-term,+non-reversible-notation,+deprecated-intros-until-0,+deprecated-focus,+unused-intro-pattern,+variable-collision,+unexpected-implicit-declaration,+omega-is-deprecated,+deprecated-instantiate-syntax,+non-recursive,+undeclared-scope#,+deprecated-hint-rewrite-without-locality,+deprecated-hint-without-locality,+deprecated-instance-without-locality,+deprecated-typeclasses-transparency-without-locality
-# Remove unsupported-attributes once we stop supporting < 8.14
+WARNINGS_PLUS := +implicit-core-hint-db,+implicits-in-term,+non-reversible-notation,+deprecated-intros-until-0,+deprecated-focus,+unused-intro-pattern,+variable-collision,+unexpected-implicit-declaration,+omega-is-deprecated,+deprecated-instantiate-syntax,+non-recursive,+undeclared-scope,+deprecated-hint-rewrite-without-locality,+deprecated-hint-without-locality,+deprecated-instance-without-locality,+deprecated-typeclasses-transparency-without-locality,# Remove unsupported-attributes once we stop supporting < 8.14
```
JasonGross, line 117: No need to remove the newline and introduce a useless `,`
```suggestion
WARNINGS_PLUS := +implicit-core-hint-db,+implicits-in-term,+non-reversible-notation,+deprecated-intros-until-0,+deprecated-focus,+unused-intro-pattern,+variable-collision,+unexpected-implicit-declaration,+omega-is-deprecated,+deprecated-instantiate-syntax,+non-recursive,+undeclared-scope,+deprecated-hint-rewrite-without-locality,+deprecated-hint-without-locality,+deprecated-instance-without-locality,+deprecated-typeclasses-transparency-without-locality
# Remove unsupported-attributes once we stop supporting < 8.14
```

_CoqProject
```diff
@@ -1,5 +1,5 @@
 -R src Crypto
--arg -w -arg +implicit-core-hint-db,+implicits-in-term,+non-reversible-notation,+deprecated-intros-until-0,+deprecated-focus,+unused-intro-pattern,+variable-collision,+unexpected-implicit-declaration,+omega-is-deprecated,+deprecated-instantiate-syntax,+non-recursive,+undeclared-scope,unsupported-attributes
+-arg -w -arg +implicit-core-hint-db,+implicits-in-term,+non-reversible-notation,+deprecated-intros-until-0,+deprecated-focus,+unused-intro-pattern,+variable-collision,+unexpected-implicit-declaration,+omega-is-deprecated,+deprecated-instantiate-syntax,+non-recursive,+undeclared-scope,+deprecated-hint-rewrite-without-locality,+deprecated-hint-without-locality,+deprecated-instance-without-locality,+deprecated-typeclasses-transparency-without-locality,,unsupported-attributes
```
JasonGross, line 2:
```suggestion
-arg -w -arg +implicit-core-hint-db,+implicits-in-term,+non-reversible-notation,+deprecated-intros-until-0,+deprecated-focus,+unused-intro-pattern,+variable-collision,+unexpected-implicit-declaration,+omega-is-deprecated,+deprecated-instantiate-syntax,+non-recursive,+undeclared-scope,+deprecated-hint-rewrite-without-locality,+deprecated-hint-without-locality,+deprecated-instance-without-locality,+deprecated-typeclasses-transparency-without-locality,unsupported-attributes
```

src/Util/PrimitiveProd.v
```diff
@@ -76,6 +76,7 @@ Local Arguments f_equal {_ _} _ {_ _} _.
 
 Definition fst_pair {A B} (a:A) (b:B) : fst (a,b) = a := eq_refl.
 Definition snd_pair {A B} (a:A) (b:B) : snd (a,b) = b := eq_refl.
+#[global]
 Create HintDb cancel_primpair discriminated. Hint Rewrite @fst_pair @snd_pair : cancel_primpair.
```
JasonGross, line 80:
```suggestion
Create HintDb cancel_primpair discriminated. #[global] Hint Rewrite @fst_pair @snd_pair : cancel_primpair.
```
Looks like the script can't handle multiple commands on the same line?

src/Util/PrimitiveSigma.v
```diff
@@ -99,6 +99,7 @@ Local Arguments f_equal {_ _} _ {_ _} _.
 
 Definition projT1_existT {A B} (a:A) (b:B a) : projT1 (existT B a b) = a := eq_refl.
 Definition projT2_existT {A B} (a:A) (b:B a) : projT2 (existT B a b) = b := eq_refl.
+#[global]
 Create HintDb cancel_primsig discriminated. Hint Rewrite @projT1_existT @projT2_existT : cancel_primsig.
```
JasonGross, line 103:
```suggestion
Create HintDb cancel_primsig discriminated. #[global] Hint Rewrite @projT1_existT @projT2_existT : cancel_primsig.
```

src/Util/PrimitiveProd.v
ppedrot: Indeed it assumes that the hint command is on a single line. That said, I don't think that the create db command should take an attribute. AFAIK it doesn't trigger any warning without one, and it fails with an export one. I've fixed this quirk, it should hopefully work now.

src/Util/CPSUtil.v
```diff
@@ -426,6 +450,7 @@ Module Tuple.
   Proof.
     induction n; simpl map_cps; intros; try destruct t;
     [|rewrite IHn, <-map_append,<-subst_append]; reflexivity.
+#[global]
   Qed. Hint Rewrite @map_cps_correct : uncps.
```
JasonGross, line 454: Your script also missed this one (and a couple more below it)
```suggestion
  Qed. #[global] Hint Rewrite @map_cps_correct : uncps.
```

JasonGross:
There seems to be an issue with an instance at
src/Bedrock/Group/ScalarMult/LadderStep.v", line 138, characters 0-37:

ppedrot:
It seems like the Examples subfolder of Bedrock was somehow missed, I've added the attributes by hand. Hopefully that's the last set of missing attributes...

PR #1353, opened 2022-08-05
Adapt w.r.t. coq/coq#16004.

ppedrot:
This is 100% backwards compatible [w.r.t. Coq 8.16], but is lacking a submodule bump for bbv. The current commit of the latter is still incompatible with coq/coq#16004.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ endif
 
 .DEFAULT_GOAL := coq
 
-WARNINGS := -deprecated-hint-rewrite-without-locality,+deprecated-hint-without-locality,+deprecated-instance-without-locality
+WARNINGS := -deprecated-hint-rewrite-without-locality,+deprecated-hint-without-locality,+deprecated-instance-without-locality,unsupported-attributes
 
 update-_CoqProject::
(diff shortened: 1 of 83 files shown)


ppedrot:
This will need the same kind of CI tweaks as #1343. cc @JasonGross

ppedrot:
Not sure why this is failing, I don't understand the logs. They're basically diffs with the added locality attributes.

JasonGross:
This is failing because we check that the checked-in files are not modified by building, and I added a line at
https://github.com/mit-plv/fiat-crypto/blob/a0e7c050690f6f0611df8d1294a4e77921ea135d/.github/workflows/coq.yml#L59-L61
to remove the global attributes on old versions of Coq. I will fix this shortly.

JasonGross (changes requested):
Please rebase on top of https://github.com/mit-plv/fiat-crypto/pull/1354 and https://github.com/mit-plv/fiat-crypto/pull/1355.  We don't need to drop support for versions earlier than 8.11.

ppedrot:
CI green. AFAIU this is ready.

JasonGross:
Merged

Issue #1351, opened 2022-08-01
feature request: concise annotated assembly output for assembly-checker


andres-erbsen:
Currently, the assembly-checker generates annotated output like the following:

`        add rcx, rax    ; #572, #574, #576, #578, #583, #584, #585, #589, #590, #591, #595, #596, #597, #601, #602, #603, #607, #608, #612, #619, #620, #621, #625, #626, #627, #631,
#632, #633, #637, #638, #639, #679, #680, #681, #685, #686, #687, #691, #692, #693, #735, #736, #737, #741, #742, #743`

This is overwhelmingly verbose and I am not sure what a list this long might even mean.

Would it be possible to generate output like

`add rcx, rax     ; rcx <- #123=(add 64 #100 #105); CF <- #127=addcarry(addcarry 64 #100 #105); created #100..104`

Only changed registers should be printed, and it might be ideal for  rhs expressions to be either they were created during symbolic execution of this line of assembly code or perhaps (even better?) if they have not been printed in an annotation before. Even fancier would be to also print PHOAS variable names for indices that have them, or do the same for PHOAS output...

Do you think this would be useful and how reasonable would (some variant of) this be to implement? @JasonGross @ChuyueSun

andres-erbsen:
Jason tells me that the overwhelmingly long list is the bug. Getting information by PHOAS variables seems difficult given current code structure. For the rest: Wrap SymexNormalInstruction, based on old and new symbolic states generate the printout while folding through instructions. All this might need to be in a thunk or something to not spend time computing it on successful runs, or maybe it doesn't matter.

PR #1357, opened 2022-08-18
fixed annotations on assembly output for assembly-checker

OwenConoly:
Fixed the bug that caused the "overwhelmingly long list" mentioned in https://github.com/mit-plv/fiat-crypto/issues/1351.

The issue was caused by the assembly checker not recognizing that identical (separate) lines of assembly code were not in fact the same line.  The annotations from identical lines were all lumped together, resulting in the "overwhelmingly long" annotations.

### DIFF
--- a/src/Assembly/Parse.v
+++ b/src/Assembly/Parse.v
@@ -354,7 +354,7 @@ Definition parse_RawLine : ParserAction RawLine
                       parsed_prefix
                end.
 
-Definition parse_Line n : ParserAction Line
+Definition parse_Line (line_num : N) : ParserAction Line
   := fun s
      => let '(indentv, rest_linev) := take_while_drop_while Ascii.is_whitespace s in
         let '(precommentv, commentv)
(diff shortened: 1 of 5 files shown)


src/Assembly/Parse.v
```diff
@@ -367,23 +367,23 @@ Definition parse_Line : ParserAction Line
         let rawlinev := String.rev rev_rawlinev in
         let trailing_whitespacev := String.rev rev_trailing_whitespacev in
         List.map
-          (fun '(r, rem) => ({| indent := indentv ; rawline := r ; pre_comment_whitespace := trailing_whitespacev ; comment := commentv |}, rem))
+          (fun '(r, rem) => ({| indent := indentv ; rawline := r ; pre_comment_whitespace := trailing_whitespacev ; comment := commentv ; line_number := n |}, rem))
           (parse_RawLine rawlinev).
 
 (* the error is the unparsable lines *)
-Fixpoint parse_Lines' (l : list string) : ErrorT (list string) Lines
+Fixpoint parse_Lines' (l : list (string*nat)) : ErrorT (list string) Lines
```
JasonGross, line 374: Might be nice to include line numbers in parse errors, now that the information is around

src/Assembly/Parse.v
```diff
@@ -367,23 +367,23 @@ Definition parse_Line : ParserAction Line
...
+       => match finalize (parse_Line n) l, parse_Lines' ls with
           | None, Error ls => Error (l :: ls)
           | None, Success _ => Error (l :: nil)
           | Some _, Error ls => Error ls
           | Some l, Success ls => Success (l :: ls)
           end
      end.
 
-Definition parse_Lines (l : list string) := parse_Lines' (String.split_newlines l).
+Definition parse_Lines (l : list string) := parse_Lines' (List.combine (String.split_newlines l) (seq 1 (length l))).
```
JasonGross, line 386: I think we have List.enumerate for this (you might have to add 1 somewhere though)

src/Assembly/Syntax.v
```diff
@@ -112,7 +112,7 @@ Inductive RawLine :=
 | INSTR (instr : NormalInstruction)
 .
 Coercion INSTR : NormalInstruction >-> RawLine.
-Record Line := { indent : string ; rawline :> RawLine ; pre_comment_whitespace : string ; comment : option string }.
+Record Line := { indent : string ; rawline :> RawLine ; pre_comment_whitespace : string ; comment : option string ; line_number : nat}.
```
JasonGross, line 115: Please use `N` rather than `nat` so we get better asymptotics.

JasonGross (changes requested):
Looks mostly good, thank you!

src/Assembly/Parse.v
OwenConoly: I restructured this code a bit, so that the enumeration is no longer necessary.

src/Assembly/Parse.v
OwenConoly: I made this change

src/Assembly/Syntax.v
OwenConoly: I made this change as well

JasonGross (approved):

PR #1358, opened 2022-08-22
Saturated Solinas Reduction

samuel-tian:
Saturated implementation of Solinas reduction that currently only supports machine_wordsize=64.

There were multiple places in the template where I had to extract bounds information for proofs. This is accomplished by the if-statement below:
```
Definition reduce1 base s c n m (p : list Z) :=
      let p_a := Positional.to_associational weight n p in
      let r_a := sat_reduce base s c n p_a in
      let r_rows := Saturated.Rows.from_associational weight m r_a in
      let r_flat := Saturated.Rows.flatten weight m r_rows in
      let bound := (0, 2^machine_wordsize - 1) in
      if (is_bounded_by (repeat bound n) p) then
        fst r_flat
      else
        add_to_nth 0 (weight (m) * snd r_flat) (fst r_flat).
```
This change made reification fail when calling `reduce1` (and other functions that used a similar trick) inside another if-statement without inlining. Inlining all of these functions causes the code length to dramatically increase, leading to long local build times (~1hr 40m for the reification cache).


JasonGross:
> This change made reification fail when calling `reduce1` (and other functions that used a similar trick) inside another if-statement without inlining.

Seems like a bug in reification.  What's the error message?  And what happens if you do `Ltac reify_debug_level ::= constr:(5%nat)`? (you might need to qualify it)

src/PushButtonSynthesis/SolinasReductionReificationCache.v
```diff
@@ -0,0 +1,87 @@
...
+                 Crypto.Arithmetic.Saturated.Rows.max_column_size
+                 Crypto.Arithmetic.Saturated.Rows.sum_rows
+                 Crypto.Arithmetic.ModOps.weight
+                 Coq.Init.Decimal.revapp
+                 Crypto.Arithmetic.Saturated.Associational.sat_mul_const
+                 Coq.Lists.List.hd
+                 Coq.ZArith.BinInt.Z.to_num_int
+                 Crypto.Arithmetic.Saturated.Rows.from_associational
+                 Coq.PArith.BinPos.Pos.to_uint
+                 Rewriter.Util.LetIn.Let_In
```
JasonGross, line 69:
```suggestion
```
This should absolutely not be unfolded early, as it will blow up Qed times

src/Rewriter/Rules.v
```diff
@@ -345,6 +347,10 @@ Definition arith_with_casts_rewrite_rulesT (adc_no_carry_to_add : bool) : list (
                   -1 ∈ rlor -> -1 ∈ rm1
                   -> cstZ rlor (Z.lor (cstZ rm1 ('-1)) (cstZ rv v)) = cstZ rm1 ('-1))
 
+            ; (* edit here *)
+            (forall rx x ry y, upper (n rx) <= lower (n ry) -> (cstZ rx x <=? cstZ ry y)%Z = true)
+            ; (forall rx x ry y, lower rx <= upper rx -> lower ry <= upper ry -> upper rx <= lower ry -> (cstZ rx x <=? cstZ ry y)%Z = true)
```
JasonGross, line 352: This seems redundant with the above?

src/Rewriter/RulesProofs.v
```diff
@@ -553,6 +553,34 @@ Proof using Type.
...
+  unfold ZRange.goodb in *.
+  unfold is_bounded_by_bool in *.
+  repeat match goal with
+         | [ H : _ && _ = true |- _ ] => apply Datatypes.andb_prop in H
+         end.
+  intuition.
+  repeat match goal with
+         | [ H : context[(_ <=? _) = true] |- _ ] => rewrite Z.leb_le in H
+         end.
+  rewrite H3 in H6, H7; auto.
```
JasonGross, line 580: Please don't rely on autogenerated names.

src/PushButtonSynthesis/SolinasReduction.v
```diff
@@ -0,0 +1,257 @@
...
+          correctness)
+         (only parsing, at level 10, summary at next level, correctness at next level).
+
+  Definition mulmod
+    := Pipeline.BoundsPipeline
+         false (* subst01 *)
+         None (* fancy *)
+         possible_values
+         (reified_solmul_gen
+            @GallinaReify.Reify base
```
JasonGross, line 190:
```suggestion
            @ GallinaReify.Reify base
```
Without the space, this reads like you're using `@` as a prefix (for disabling implicit insertion), when it's actually an infix.

samuel-tian:
> > This change made reification fail when calling `reduce1` (and other functions that used a similar trick) inside another if-statement without inlining.
> 
> Seems like a bug in reification. What's the error message? And what happens if you do `Ltac reify_debug_level ::= constr:(5%nat)`? (you might need to qualify it)

I looked more into this and realized that the issue is probably with the another part of the synthesis pipeline, not the reification itself. Here's the end of the debug trace when synthesizing non-inlined reduce_full: [no_inline_reify_reduce_full.txt](https://github.com/mit-plv/fiat-crypto/files/9399682/no_inline_reify_reduce_full.txt). The segfault occurs after Reify completes, perhaps during the `exact` call on the large amount of reified code?

JasonGross:
This might be caused by you including `Let_In` in the Strategy list.  What's the memory usage like?

samuel-tian:
Around 3GB for synthesis. I just repeatedly used `ps`; is there a better way of tracking memory?

JasonGross:
`make TIMED=1` will report peak memory usage

samuel-tian:
I switched to an Ubuntu machine (with Coq v8.11) and `make TIMED=1` exits with stack overflow instead of segfault, so it gives more useful information:
```
src/Arithmetic/SolinasReduction.vo (real: 136.25, user: 135.18, sys: 1.05, mem: 3526432 ko)
```
I am currently locally building the PR fixes that I made, so I pushed them to another [branch](https://github.com/mit-plv/fiat-crypto/commit/70ac065f5c2b1ee93e36df95ae7045153af7490a). I removed Let_In from the strategy list and reification cache Qed took ~2900s (from ~3000s). Did I do this correctly, and is there anything else I can do to improve the runtime?

Also, the strategy list of identifiers generated on v8.11 differs from v8.15. In particular, it seems that v8.11 did not support hexadecimal types while v8.15 does. Will this cause any issues?

JasonGross:
> In particular, it seems that v8.11 did not support hexadecimal types while v8.15 does. Will this cause any issues?

It should not, unless you're making use of hexadecimal conversations in, e.g., generating fuel for recursive functions or something.  (I'm curious, though, where do you use them?  And how do they end up getting reified?)

andres-erbsen:
So in principle we can drop support for coq <8.15 at the cost of Debian users needing to install coq from their testing repo. But let's consider whether to do that when other points here are resolved.

JasonGross:
I've begun porting reification to Ltac2 at https://github.com/mit-plv/rewriter/pull/57 and https://github.com/mit-plv/fiat-crypto/pull/1361.  I'm not sure I'll find time to finish debugging it, and there's still a bit of work left to do after that.  In particular, we still need to:
- Finish making reification work on all the examples
- Dig through the commits to see what performance regressions were introduced (most profiling is available in the commit messages of https://github.com/mit-plv/fiat-crypto/pull/1361), and figure out how to fix any unexpected performance regressions
- Finish porting the relevant surrounding wrapper code from Ltac to Ltac2 so that refication doesn't roundtrip between Ltac and Ltac2 on every step (the roundtripping incurs massive overhead)
- Implement caching at the cbv-delta step: the way this will work is adding an argument of type `((constr * ((constr * (ident * constr)) list)) list) ref` which will be an associative list mapping constants to maps from `template_ctx` to a pair of a fresh identifier and the reified expression.  (We should also pass around a `Ltac2.Fresh.Free.t` that tracks used identifiers.)  We can then check the cache at every unfolding step and update it when reification succeeds, plausibly doing some sort of groundedness check to ensure that there are no free de Bruijn indices.  At the very end of reification, we can use `Constr.Unsafe.closenl` to add reified terms as let binders; we should add them one-by-one in order so that newer terms at the head of the list can reference older terms in the tail.
- Misc performance optimizations (possibly requring https://github.com/coq/coq/issues/10111)

@samuel-tian Do you feel up to reviewing https://github.com/mit-plv/fiat-crypto/pull/1361 / https://github.com/mit-plv/rewriter/pull/57, asking questions about what you don't understand (or think is wrong), and possibly taking on some of this work?  (Or rewriting reification from scratch in Ltac2, or even coq-elpi or MetaCoq?)

JasonGross:
@samuel-tian Can you try out this PR with #1361 and see how the performance is?

JasonGross:
I've finished porting reification to Ltac2 and adding caching of identifiers.  If I'm understanding correctly, though, this won't actually solve the Qed problem; even though the size of the reified term will now be managable (rather than duplicating the reification of things like `Positional.to_associational` at every location, it'll instead let-bind the reified form of `Positional.to_associational` at top level and use that everywhere), the conversion algorithm doesn't let us find a small intermediate term that is cheaply convertible to both the interpretation of the reified form and the original term.  I'm not entirely sure what to do here, other than to ask the coq devs to implement some form of congruence closure, or to make the rewrite rules depend on arithmetic and add all of the relevent definitions as primitive identifiers.  I'll open an issue for figuring this out.

JasonGross:
We should also check whether the problem is in proving the interp equality or in proving wf.  I suspect interp, but it would be nice to get a breakout of the two (actually three) different operations, by replacing
https://github.com/mit-plv/fiat-crypto/blob/master/src/PushButtonSynthesis/ReificationCache.v#L38-L48
with something like
```coq
Ltac cache_reify _ :=
  split;
  [ intros;
    etransitivity;
    [
    | time "repeat apply f_equal" (repeat match goal with |- _ = ?f' ?x => is_var x; apply (f_equal (fun f => f _)) end);
      time "Reify_rhs" (Reify_rhs ());
      lazymatch goal with |- ?LHS = ?RHS => time "unify1" unify LHS RHS end;
      time "abstract1" abstract (time "reflexivity1" reflexivity) ];
    subst_evars;
    lazymatch goal with |- ?LHS = ?RHS => time "unify2" unify LHS RHS end;
    time "abstract2" abstract (time "reflexivity2" reflexivity)
  | time "abstract3" abstract (time "prove_Wf" (prove_Wf ())) ].
```

JasonGross:
I rebased on current master at https://github.com/JasonGross/fiat-crypto/tree/saturated-solinas if you're ineteresed

JasonGross:
Timing on the rebased version:
```
COQC src/PushButtonSynthesis/SolinasReductionReificationCache.v
Tactic call repeat apply f_equal ran for 0.001 secs (0.001u,0.s) (success)
Tactic call Reify_rhs ran for 6459.956 secs (6450.624u,0.489s) (success)
Tactic call reflexivity1 ran for 0.028 secs (0.028u,0.s) (success)
Tactic call subst_evars1 ran for 0.022 secs (0.022u,0.s) (success)
Tactic call exact_no_check1 ran for 0. secs (0.u,0.s) (success)
Tactic call abstract1 ran for 0.607 secs (0.606u,0.s) (success)
Tactic call subst_evars1.5 ran for 0.012 secs (0.012u,0.s) (success)
Tactic call reflexivity2 ran for 0.024 secs (0.024u,0.s) (success)
Tactic call subst_evars2 ran for 0. secs (0.u,0.s) (success)
Tactic call exact_no_check2 ran for 0. secs (0.u,0.s) (success)
Tactic call abstract2 ran for 0.524 secs (0.523u,0.s) (success)
Tactic call prove_Wf ran for 0.029 secs (0.029u,0.s) (success)
Tactic call subst_evars3 ran for 0. secs (0.u,0.s) (success)
Tactic call exact_no_check3 ran for 0. secs (0.u,0.s) (success)
Tactic call abstract3 ran for 0.494 secs (0.493u,0.s) (success)
Finished transaction in 6461.72 secs (6452.385u,0.489s) (successful)
Finished transaction in 6288.894 secs (6280.258u,0.079s) (successful)
src/PushButtonSynthesis/SolinasReductionReificationCache.vo (real: 12753.59, user: 12735.20, sys: 0.98, mem: 1945172 ko)
```

More timing experiments coming up soon, hopefully

JasonGross:
Further timing experiements:
```
COQC src/PushButtonSynthesis/SolinasReductionReificationCache.v
Tactic call repeat apply f_equal ran for 0.001 secs (0.001u,0.s) (success)
Tactic call reify_in_context_opt ran for 7.349 secs (7.328u,0.009s) (success)
Tactic call reify_let_bind_cache ran for 4.725 secs (4.708u,0.009s) (success)
Tactic call _Reify_rhs._Reify ran for 12.24 secs (12.202u,0.019s) (success)
Tactic call _Reify_rhs.transitivity ran for 0.114 secs (0.113u,0.s) (success)
Tactic call
_Reify_rhs.reflexivity ran for 6447.651 secs (6438.072u,0.359s)
(success)
Tactic call Reify_rhs ran for 6460.007 secs (6450.389u,0.379s) (success)
Tactic call reflexivity1 ran for 0.028 secs (0.028u,0.s) (success)
Tactic call subst_evars1 ran for 0.022 secs (0.022u,0.s) (success)
Tactic call exact_no_check1 ran for 0. secs (0.u,0.s) (success)
Tactic call abstract1 ran for 0.63 secs (0.629u,0.s) (success)
Tactic call subst_evars1.5 ran for 0.012 secs (0.012u,0.s) (success)
Tactic call reflexivity2 ran for 0.024 secs (0.024u,0.s) (success)
Tactic call subst_evars2 ran for 0. secs (0.u,0.s) (success)
Tactic call exact_no_check2 ran for 0. secs (0.u,0.s) (success)
Tactic call abstract2 ran for 0.53 secs (0.53u,0.s) (success)
Tactic call prove_Wf ran for 0.028 secs (0.028u,0.s) (success)
Tactic call subst_evars3 ran for 0. secs (0.u,0.s) (success)
Tactic call exact_no_check3 ran for 0. secs (0.u,0.s) (success)
Tactic call abstract3 ran for 0.493 secs (0.493u,0.s) (success)
Finished transaction in 6461.798 secs (6452.179u,0.379s) (successful)
```
This shows that indeed reification on this is now relatively fast (about 12 seconds to reify), but `reflexivity` converting between the interpretation of the reified term and the original term is quite slow and takes up the bulk of the time.  There's no reason to repeat this work in Qed, though, so https://github.com/mit-plv/rewriter/pull/69 should speed up things up by a factor of 2 (which is still 2x slower than what you reported. apparently?)

JasonGross:
With any luck, #1413 will fix the performance problems in reification

JasonGross:
Luck smiles upon us.  #1413 fixes the performance problems.
```
COQC src/PushButtonSynthesis/SolinasReductionReificationCache.v
Tactic call repeat apply f_equal ran for 0.006 secs (0.006u,0.s) (success)
Tactic call _Reify_rhs.make intermediate ran for 0. secs (0.u,0.s) (success)
Tactic call
reify_in_context_opt ran for 13.386 secs (13.347u,0.039s)
(success)
Tactic call
reify_UnderLets_bind_cache ran for 0.034 secs (0.034u,0.s)
(success)
Tactic call
_Reify_rhs.reify_and_cache ran for 13.422 secs (13.382u,0.039s)
(success)
Tactic call
_Reify_rhs.refine with reified ran for 0.61 secs (0.61u,0.s)
(success)
Tactic call
_Reify_rhs.reify and refine ran for 14.112 secs (14.072u,0.039s)
(success)
Tactic call _Reify_rhs.transitivity ran for 0.378 secs (0.378u,0.s) (success)
Tactic call
_Reify_rhs.make proof unsafe ran for 0.004 secs (0.004u,0.s)
(success)
Tactic call
_Reify_rhs.make proof check for universes ran for 3.936 secs (3.936u,0.s)
(success)
Tactic call
_Reify_rhs.abstract exact_no_check UnderLets.cached_interp_related_impl... ran for 1.205 secs (1.198u,0.007s)
(success)
Tactic call Reify_rhs ran for 19.91 secs (19.862u,0.046s) (success)
Tactic call reflexivity1 ran for 0.137 secs (0.137u,0.s) (success)
Tactic call subst_evars1 ran for 0.126 secs (0.126u,0.s) (success)
Tactic call exact_no_check1 ran for 0.003 secs (0.003u,0.s) (success)
Tactic call abstract1 ran for 1.855 secs (1.855u,0.s) (success)
Tactic call subst_evars1.5 ran for 0.048 secs (0.048u,0.s) (success)
Tactic call reflexivity2 ran for 0.109 secs (0.109u,0.s) (success)
Tactic call subst_evars2 ran for 0. secs (0.u,0.s) (success)
Tactic call exact_no_check2 ran for 0.003 secs (0.003u,0.s) (success)
Tactic call abstract2 ran for 1.532 secs (1.532u,0.s) (success)
Tactic call prove_Wf ran for 0.149 secs (0.149u,0.s) (success)
Tactic call subst_evars3 ran for 0. secs (0.u,0.s) (success)
Tactic call exact_no_check3 ran for 0. secs (0.u,0.s) (success)
Tactic call abstract3 ran for 1.199 secs (1.199u,0.s) (success)
Finished transaction in 25.082 secs (25.034u,0.047s) (successful)
Finished transaction in 10.823 secs (10.823u,0.s) (successful)
src/PushButtonSynthesis/SolinasReductionReificationCache.vo (real: 44.17, user: 43.56, sys: 0.41, mem: 1001296 ko)
```

Makefile
```diff
@@ -271,8 +271,10 @@ endef
 UNSATURATED_SOLINAS_FUNCTIONS := carry_mul carry_square carry add sub opp selectznz to_bytes from_bytes relax
 FUNCTIONS_FOR_25519 := $(UNSATURATED_SOLINAS_FUNCTIONS) carry_scmul121666
 WORD_BY_WORD_MONTGOMERY_FUNCTIONS := mul square add sub opp from_montgomery to_montgomery nonzero selectznz to_bytes from_bytes one msat divstep divstep_precomp
+SOLINAS_REDUCTION_FUNCTIONS := mulmod
```
JasonGross, line 274:
```suggestion
SOLINAS_REDUCTION_FUNCTIONS := mul
```

Makefile
```diff
@@ -308,8 +310,11 @@ LITE_BASE_FILES := curve25519_64 poly1305_64 poly1305_32 p256_64 secp256k1_64 p3
 
 EXTRA_C_FILES := inversion/c/*_test.c
 
-ALL_C_FILES := $(patsubst %,$(C_DIR)%.c,$(ALL_BASE_FILES))
 ALL_BEDROCK2_FILES := $(patsubst %,$(BEDROCK2_DIR)%.c,$(filter-out $(BASE_FILES_NEEDING_INT128),$(ALL_BASE_FILES)))
+
+$(foreach bw,64,$(eval $(call add_curve_keys,curve25519_solinas_$(bw),SOLINAS_REDUCTION,'curve25519_solinas',$(bw),'2^255 - 19',$(SOLINAS_REDUCTION_FUNCTIONS),SOLINAS_REDUCTION)))
```
JasonGross, line 315: Please move this back up and fix bedrock2 compilation.  The issue is that the generated PHOAS code includes
```
  let x188 := Z.mul_split((2^64), (38, (uint64_t)x187)) in                                                                    let x189 := x170 + x188₁ (* : uint64_t *) in
  let x190 := Z.ltz((uint64_t)x189, (uint64_t)x170) (* : uint64_t *) in                                                       let x191 := x190 + x174 (* : uint64_t *) in                                                                                 let x192 := Z.ltz((uint64_t)x191, (uint64_t)x174) (* : uint64_t *) in                                                       let x193 := x192 + x179 (* : uint64_t *) in                                                                                 let x194 := Z.ltz((uint64_t)x193, (uint64_t)x179) (* : uint64_t *) in                                                       let x195 := x194 + x184 (* : uint64_t *) in                                                                                 let x196 := Z.ltz((uint64_t)x195, (uint64_t)x184) (* : uint64_t *) in                                                       let x197 := Z.mul_split((2^64), (38, (uint64_t)x196)) in                                                                    let x198 := x189 + x197₁ (* : uint64_t *) in       
```
and bedrock2 does not handle `Z.mul_split`.  This is supposed to be handled by
https://github.com/mit-plv/fiat-crypto/blob/b1e324529b5e65492ee14be553405d96d5f82240/src/Rewriter/Rules.v#L1080-L1085
but I believe the ranges on constants are narrowed to be singleton (or perhaps they are stripped all together?  I don't quite recall).  So I think we want to replace it with a rule like
```coq
                ; (forall rx x ry y rpow2_bitwidth,
                      0 <= lgcarrymax <= bitwidth -> 2^bitwidth ∈ rpow2_bitwidth -> (rx <= singlewidth_range)%zrange -> (ry <= singlewidth_range)%zrange
                      -> pairsinglewidth (Z.mul_split (cstZ rpow2_bitwidth ('(2^bitwidth))) (cstZ rx x) (cstZ ry y))
                         = (dlet low := singlewidth (cstZ rx x * cstZ ry y) in
                                dlet high := singlewidth (Z.mul_high (cstZ rpow2_bitwidth ('(2^bitwidth))) (cstZ rx x) (cstZ ry y)) in
                                (singlewidth low, singlewidth high)))
```

Except we already have https://github.com/mit-plv/fiat-crypto/blob/b1e324529b5e65492ee14be553405d96d5f82240/src/Rewriter/Rules.v#L1045-L1047
so I'm not sure what's going wrong...

Maybe we should look at JSON output to see all the casts?  Or adjust printing of bedrock2 errors so that all the casts get inserted?

src/Bedrock/Everything.v
```diff
@@ -218,6 +221,7 @@ Crypto.Rewriter.RulesProofs
 Crypto.Rewriter.TestRules
 Crypto.Rewriter.TestRulesProofs
 Crypto.SlowPrimeSynthesisExamples
+Crypto.SlowPrimeSynthesisExamplesCopy
```
JasonGross, line 224: Please move whatever debugging you're doing here back into the Examples file

src/Bedrock/Standalone/StandaloneHaskellMain.v
```diff
@@ -29,6 +29,11 @@ Module Bedrock2First.
       := main_gen ForExtraction.SaturatedSolinas.PipelineMain.
   End SaturatedSolinas.
 
+  (* Module SolinasReduction. *)
+  (*   Definition main : IO_unit *)
+  (*     := main_gen ForExtraction.SolinasReduction.PipelineMain. *)
+  (* End SolinasReduction. *)
+
```
JasonGross, line 36: Why is this commented out?

src/Bedrock/Standalone/StandaloneHaskellMain.v
```diff
@@ -58,6 +63,11 @@ Module Bedrock2Later.
...
```
JasonGross, line 70: Why is this commented out?

src/COperationSpecifications.v
```diff
@@ -494,6 +494,26 @@ Module SaturatedSolinas.
...
+            (length_saturated_bouns : length saturated_bounds = n).
+    Local Notation eval := (Positional.eval wt n).
+
+    Definition mulmod_correct
+               (mulmod : list Z -> list Z -> list Z) :=
+      forall x y,
+        list_Z_bounded_by saturated_bounds x ->
+        list_Z_bounded_by saturated_bounds y ->
+        ((eval (mulmod x y)) mod m = (eval x * eval y) mod m) /\
+          (list_Z_bounded_by saturated_bounds (mulmod x y)).
```
JasonGross, line 512:
```suggestion
    Definition mul_correct
               (mul : list Z -> list Z -> list Z) :=
      forall x y,
        list_Z_bounded_by saturated_bounds x ->
        list_Z_bounded_by saturated_bounds y ->
        ((eval (mul x y)) mod m = (eval x * eval y) mod m) /\
          (list_Z_bounded_by saturated_bounds (mul x y)).
```
The synthesized operations should not be `mulmod`, that is only what we're using in `Arithmetic/`; `Arithmetic/` gets to see underneath the field abstraction, and so it makes sense to distinguish between things that reduce modulo and things that don't; the synthesized code doesn't get to see this distinction, it only gets to see things like "bounds"; everything is evaluated only up to its value modulo.  So unsaturated solinas we have `mul` and `carry_mul` for "mul with loose output bounds" and "mul with tight output bounds", but nothing about modular reduction.

src/PushButtonSynthesis/SolinasReduction.v
```diff
@@ -0,0 +1,257 @@
...
+Import Associational Positional.
+Import SolinasReduction.
+
+Local Coercion Z.of_nat : nat >-> Z.
+Local Coercion QArith_base.inject_Z : Z >-> Q.
+Local Coercion Z.pos : positive >-> Z.
+
+Local Set Keyed Unification. (* needed for making [autorewrite] fast, c.f. COQBUG(https://github.com/coq/coq/issues/9283) *)
+
+Local Opaque reified_solmul_gen. (* needed for making [autorewrite] not take a very long time *)
```
JasonGross, line 60: I think it should just be
```suggestion
Local Opaque reified_mul_gen. (* needed for making [autorewrite] not take a very long time *)
```

src/PushButtonSynthesis/SolinasReduction.v
```diff
@@ -0,0 +1,257 @@
...
+             weight "weight"
+             CorrectnessStringification.dyn_context.nil))%string.
+  Local Notation "'docstring_with_summary_from_lemma!' summary correctness"
+    := (docstring_with_summary_from_lemma_with_ctx!
+          notations_for_docstring
+          summary
+          correctness)
+         (only parsing, at level 10, summary at next level, correctness at next level).
+
+  Definition mulmod
```
JasonGross, line 184:
```suggestion
  Definition mul
```

src/PushButtonSynthesis/SolinasReduction.v
```diff
@@ -0,0 +1,257 @@
...
+    := Pipeline.BoundsPipeline
+         false (* subst01 *)
+         None (* fancy *)
+         possible_values
+         (reified_solmul_gen
```
JasonGross, line 189:
```suggestion
         (reified_mul_gen
```

src/PushButtonSynthesis/SolinasReduction.v
```diff
@@ -0,0 +1,257 @@
...
+            @GallinaReify.Reify base
+            @GallinaReify.Reify s
+            @GallinaReify.Reify c
+            @GallinaReify.Reify n )
+         (Some boundsn, (Some boundsn, tt))
+         (Some boundsn).
+
+  Definition smulmod (prefix : string)
```
JasonGross, line 197:
```suggestion
  Definition smul (prefix : string)
```

src/PushButtonSynthesis/SolinasReduction.v
```diff
@@ -0,0 +1,257 @@
...
+    : string * (Pipeline.ErrorT (Pipeline.ExtendedSynthesisResult _))
+    := Eval cbv beta in
+        FromPipelineToString!
+          machine_wordsize prefix "mulmod" mulmod
+          (docstring_with_summary_from_lemma!
+             (fun fname : string => [text_before_function_name ++ fname ++ " multiplies two field elements."]%string)
+             (mulmod_correct weight n m boundsn)).
```
JasonGross, line 204:
```suggestion
             (mul_correct weight n m boundsn)).
```

src/PushButtonSynthesis/SolinasReduction.v
```diff
@@ -0,0 +1,257 @@
...
+
+  Local Ltac solve_extra_bounds_side_conditions :=
+    cbn [lower upper fst snd] in *; Bool.split_andb; Z.ltb_to_lt; lia.
+
+  Hint Rewrite
+       (fun pf => @SolinasReduction.SolinasReduction.mulmod_correct (@wprops _ _ pf)) using solve [ auto with zarith | congruence | solve_extra_bounds_side_conditions ] : push_eval.
+  Hint Unfold mulmod : push_eval.
```
JasonGross, line 211:
```suggestion
  Hint Unfold mul : push_eval.
```

src/PushButtonSynthesis/SolinasReduction.v
```diff
@@ -0,0 +1,257 @@
...
+
+  Local Ltac prove_correctness _ := Primitives.prove_correctness use_curve_good.
+
+  Lemma mulmod_correct res
+        (Hres : mulmod = Success res)
+    : mulmod_correct weight n m boundsn (Interp res).
```
JasonGross, line 217:
```suggestion
  Lemma mul_correct res
        (Hres : mul = Success res)
    : mul_correct weight n m boundsn (Interp res).
```

src/PushButtonSynthesis/SolinasReduction.v
```diff
@@ -0,0 +1,257 @@
...
+  Proof using curve_good. prove_correctness (). Qed.
+
+  Lemma Wf_mulmod res (Hres : mulmod = Success res) : Wf res.
```
JasonGross, line 220:
```suggestion
  Lemma Wf_mul res (Hres : mulmod = Success res) : Wf res.
```

src/PushButtonSynthesis/SolinasReduction.v
```diff
@@ -0,0 +1,257 @@
...
+  Proof using Type. prove_pipeline_wf (). Qed.
+
+  Section for_stringification.
+    Local Open Scope string_scope.
+    Local Open Scope list_scope.
+
+    Definition known_functions
+      := [("mulmod", wrap_s smulmod)].
```
JasonGross, line 228:
```suggestion
      := [("mul", wrap_s smul)].
```

src/PushButtonSynthesis/SolinasReduction.v
```diff
@@ -0,0 +1,257 @@
...
+                     "Computed values:";
+                     ""]%string)))
+           function_name_prefix requests.
+  End for_stringification.
+End __.
+
+Module Export Hints.
+#[global]
+  Hint Opaque
+       mulmod
```
JasonGross, line 251:
```suggestion
       mul
```

src/PushButtonSynthesis/SolinasReduction.v
```diff
@@ -0,0 +1,257 @@
...
+  : wf_op_cache.
+#[global]
+  Hint Immediate
+       Wf_mulmod
```
JasonGross, line 255:
```suggestion
       Wf_mul
```

src/PushButtonSynthesis/SolinasReductionReificationCache.v
```diff
@@ -0,0 +1,87 @@
...
+Local Open Scope Z_scope.
+
+Local Set Keyed Unification. (* needed for making [autorewrite] fast, c.f. COQBUG(https://github.com/coq/coq/issues/9283) *)
+
+Import
+  Language.API.Compilers
+  Language.Wf.Compilers.
+Import SolinasReduction.SolinasReduction.
+
+Strategy -500 [Crypto.Arithmetic.SolinasReduction.SolinasReduction.is_bounded_by
```
JasonGross, line 19: With the new reification strategy, this entire list/command is probably unneeded now.

src/PushButtonSynthesis/SolinasReductionReificationCache.v
```diff
@@ -0,0 +1,87 @@
...
+                 Rewriter.Util.LetIn.Let_In
+                 Crypto.Arithmetic.Core.Positional.zeros
+                 Coq.PArith.BinPos.Pos.to_hex_uint
+                 Coq.ZArith.BinInt.Z.to_num_hex_int
+                 Coq.Init.Hexadecimal.rev
+                 Crypto.Arithmetic.Saturated.Rows.flatten].
+
+Derive reified_solmul_gen
+       SuchThat (is_reification_of reified_solmul_gen mulmod)
+       As reified_solmul_gen_correct.
```
JasonGross, line 78:
```suggestion
Derive reified_mul_gen
       SuchThat (is_reification_of reified_mul_gen mulmod)
       As reified_mul_gen_correct.
```

src/PushButtonSynthesis/SolinasReductionReificationCache.v
```diff
@@ -0,0 +1,87 @@
...
+Proof. Time cache_reify (). Time Qed.
+
+#[global]
+ Hint Extern 1 (_ = _) => apply_cached_reification mulmod (proj1 reified_solmul_gen) : reify_cache_gen.
```
JasonGross, line 82:
```suggestion
 Hint Extern 1 (_ = _) => apply_cached_reification mulmod (proj1 reified_mul_gen) : reify_cache_gen.
```

src/PushButtonSynthesis/SolinasReductionReificationCache.v
```diff
@@ -0,0 +1,87 @@
...
+#[global]
+ Hint Immediate (proj2 reified_solmul_gen_correct) : wf_gen_cache.
```
JasonGross, line 84:
```suggestion
 Hint Immediate (proj2 reified_mul_gen_correct) : wf_gen_cache.
```

src/PushButtonSynthesis/SolinasReductionReificationCache.v
```diff
@@ -0,0 +1,87 @@
...
+#[global]
+ Hint Rewrite (proj1 reified_solmul_gen_correct) : interp_gen_cache.
+Local Opaque reified_solmul_gen. (* needed for making [autorewrite] not take a very long time *)
```
JasonGross, line 87:
```suggestion
 Hint Rewrite (proj1 reified_mul_gen_correct) : interp_gen_cache.
Local Opaque reified_mul_gen. (* needed for making [autorewrite] not take a very long time *)
```

src/Rewriter/RulesProofs.v
```diff
@@ -553,6 +553,34 @@ Proof using Type.
...
+  rewrite H4 in H2, H8; auto.
+  etransitivity; eauto.
+  etransitivity; eauto.
```
JasonGross, line 583: Please automate all this a bit more (or one of us can take a stab at it)

src/SlowPrimeSynthesisExamplesCopy.v
```diff
@@ -0,0 +1,533 @@
...
+Local Instance : ignore_unique_asm_names_opt := false.
+Local Instance : tight_upperbound_fraction_opt := default_tight_upperbound_fraction.
+Local Existing Instance default_language_naming_conventions.
+Local Existing Instance default_documentation_options.
+Local Instance : package_name_opt := None.
+Local Instance : class_name_opt := None.
+
+Local Existing Instance default_output_options.
+
+Module compiling_mul.
```
JasonGross, line 54: Please merge this into src/SlowPrimeSynthesisExamples

src/SlowPrimeSynthesisExamplesCopy.v
```diff
@@ -0,0 +1,533 @@
...
+                 (fun _ _ => [])
+                 (Some boundsn, (Some boundsn, tt))
+                 (Some boundsn, None)
+                 (None, (None, tt))
+                 (None, None)
+         : Pipeline.ErrorT _).
+
+End compiling_mul.
+
+Module compiling_reduce_flatten.
```
JasonGross, line 114: This one too

src/SlowPrimeSynthesisExamplesCopy.v
```diff
@@ -0,0 +1,533 @@
...
+  out1[6] = 0x0;
+  out1[7] = 0x0;
+}"", {| bitwidths_used := [uint1, uint64] ; addcarryx_lg_splits := [64] ; mulx_lg_splits := [64] ; cmovznz_bitwidths := [] ; value_barrier_bitwidths := [] ; typedefs_used := [] |})"
+     : string
+Finished transaction in 4.122 secs (4.07u,0.05s) (successful)
+*)
+
+End compiling_reduce_flatten.
+
+Module compiling_red.
```
JasonGross, line 352: And this

src/StandaloneOCamlMain.v
```diff
@@ -15,6 +15,7 @@ Global Unset Extraction Optimize.
 
 (** Work around COQBUG(https://github.com/coq/coq/issues/4875) / COQBUG(https://github.com/coq/coq/issues/7954) / COQBUG(https://github.com/coq/coq/issues/7954) / https://discuss.ocaml.org/t/why-wont-ocaml-specialize-weak-type-variables-in-dead-code/7776 *)
 Extraction Inline Show.ShowLevel_of_Show.
+Extraction Inline SolinasReductionReificationCache.reified_solmul_gen.
```
JasonGross, line 18: Why do we need this?

JasonGross (changes requested):
Please do not add `fiat-bedrock2/src/curve25519_solinas_64.c.tmp`.  Also please make sure none of the intermediate commits add the compiled binaries nor extracted OCaml files, etc (or squash when merging); these are pretty big and will make `.git` much bigger

src/Arithmetic/SolinasReduction.v
```diff
@@ -0,0 +1,2155 @@
...
+Require Import Crypto.Util.ZRange.
+Require Import Crypto.Arithmetic.Core.
+Require Import Crypto.Arithmetic.ModOps.
+Require Import Crypto.Arithmetic.Partition.
+Require Import Crypto.Arithmetic.UniformWeight.
+Require Import Crypto.Arithmetic.Saturated.
+Require Crypto.Stringification.C.
+Require Crypto.Stringification.Go.
+Require Crypto.Stringification.Java.
+Require Import Crypto.BoundsPipeline.
```
JasonGross, line 16:
```suggestion
```
Arithmetic files should not depend on the pipeline

JasonGross (changes requested):

src/Arithmetic/SolinasReduction.v
```diff
@@ -0,0 +1,2155 @@
...
+Local Open Scope string_scope.
+Local Open Scope list_scope.
+Import ListNotations. Local Open Scope Z_scope.
+
+Import
+  AbstractInterpretation.Compilers
+  Language.Compilers
+  Language.API.Compilers.
+
+Import Language.API.Compilers.API.
```
JasonGross, line 53:
```suggestion
```

Makefile
JasonGross: After making use of https://github.com/mit-plv/fiat-crypto/pull/1464, the offending line before rewriting is
```
let x218 := (uint64_t, uint0_t)(Z.mul_split(([0x10000000000000000 ~> 0x10000000000000000])(2^64), ((uint64_t)38, (uint64_t)(uint64_t)(uint64_t)x217))) (* : uint64_t, uint0_t *) in
```
which is not handled becuase
https://github.com/mit-plv/fiat-crypto/blob/73c467e0309f7dea6b854ec9e5311a8c7e13b014/src/Rewriter/Rules.v#L938
does not include `Z.mul_split` where one side or the other is guaranteed to be zero by bounds analysis.  So we should add some more rules to handle the cases where exactly one of the returns is `uint0_t`.  (We don't need to handle both returns being `uint0_t`, because other rules will replace downstream uses of this with `0`, and then dead code elimination will handle the case for us.)

andres-erbsen: I believe the issue here is just that a mul_split in the template ends up applied to arguments that are small enough that splitting is not necessary and a normal `*` would do. If it's easy to change the template to use `*`, that'd fix it. Doing the same replacement in a post-bounds-analysis rewrite rule should work as well.

JasonGross: https://github.com/mit-plv/fiat-crypto/pull/1474 should fix this (but reification of the rewrite rule fails for some reason, and https://github.com/coq/coq/issues/16716 is making the reason hard to diagnose)

JasonGross: #1474 now builds locally, and with that change, https://github.com/JasonGross/fiat-crypto/tree/saturated-solinas successfully builds bedrock2 files

samuel-tian:
I implemented squaremod and modified the Solinas reduction algorithm to yield better performance post-CryptOpt optimization. Also, the assembly equivalence checker now supports the `cmovo` instruction.

The new proofs in `src/Arithmetic/SolinasReduction.v` are quite messy and I will clean them up shortly.

Makefile-coq
```diff
@@ -0,0 +1,924 @@
+##########################################################################
```
JasonGross, line 1: This file should not be checked in.  (Please also make sure the history doesn't contain any large generated files that are checked in.  If possible, it'd be nice to have this branch cleanly rebased on master rather than having a bunch of merge commits.  If you want, I can do a rebase.)

JasonGross (changes requested):
Looking good!

Makefile-coq
JasonGross: (I can also force-push to your master if you'd like)

JasonGross: I've rebased https://github.com/samuel-tian/fiat-crypto/pull/1 / https://github.com/JasonGross/fiat-crypto/tree/saturated-solinas .  Do you want me to force push?  The diff with this branch is (ignoring whitespace changes via `diff -w` and ignoring `Makefile-coq`):
```diff
diff --git a/src/Arithmetic/SolinasReduction.v b/src/Arithmetic/SolinasReduction.v
index 3e46fa9ec..993dd8125 100644
--- a/src/Arithmetic/SolinasReduction.v
+++ b/src/Arithmetic/SolinasReduction.v
@@ -322,7 +322,7 @@ Module SolinasReduction.
 
     Lemma fold_right_add : forall l x,
         fold_right Z.add x l = x + fold_right Z.add 0 l.
-    Proof.
+    Proof using Type.
       intros l x.
       induction l as [ | l' IHl ]; cbn; try rewrite IHl; lia.
     Qed.
@@ -333,7 +333,7 @@ Module SolinasReduction.
 
     Lemma eval_weight_S' : forall p,
         eval_weight_P p.
-    Proof.
+    Proof using Type.
       apply (ListAux.list_length_induction Z).
       unfold eval_weight_P.
       intros l1 H n b.
@@ -382,7 +382,7 @@ Module SolinasReduction.
     Lemma eval_weight_S p n:
       eval (fun i : nat => weight (S i)) n p =
         (eval weight n p) * weight 1.
-    Proof.
+    Proof using Type.
       cbv [eval to_associational].
       rewrite eval_weight_S'.
       lia.
@@ -392,7 +392,7 @@ Module SolinasReduction.
     Lemma eval_weight_S_gen p a b :
       Associational.eval (combine (map (fun x0 : nat => weight (S x0)) (seq a b)) p) =
         weight 1 * Associational.eval (combine (map weight (seq a b)) p).
-    Proof.
+    Proof using Type.
       apply eval_weight_S'.
     Qed.
     Hint Rewrite eval_weight_S_gen : push_eval.
@@ -527,7 +527,7 @@ Module SolinasReduction.
     Lemma is_bounded_by_cons1 : forall b bounds p' p,
         is_bounded_by (b :: bounds) (p' :: p) = true ->
         is_bounded_by bounds p = true.
-    Proof.
+    Proof using Type.
       intros; repeat autounfold in *; match goal with | H : _ |- _ => push' H end.
     Qed.
     Hint Resolve is_bounded_by_cons1 : ibb.
@@ -535,7 +535,7 @@ Module SolinasReduction.
     Lemma is_bounded_by_cons2 : forall b bounds p' p,
         is_bounded_by (b :: bounds) (p' :: p) = true ->
         fst b <= p' <= snd b.
-    Proof.
+    Proof using Type.
       intros; repeat autounfold in *; match goal with | H : _ |- _ => push' H end.
     Qed.
     Hint Resolve is_bounded_by_cons2 : ibb.
@@ -544,7 +544,7 @@ Module SolinasReduction.
         is_bounded_by (b :: bounds) (p' :: p) = true ->
         is_bounded_by bounds p = true /\
           fst b <= p' <= snd b.
-    Proof.
+    Proof using Type.
       intros; repeat autounfold in *; match goal with | H : _ |- _ => push' H end.
     Qed.
     Hint Resolve is_bounded_by_cons : ibb.
@@ -554,7 +554,7 @@ Module SolinasReduction.
         (n < length p)%nat ->
         (n < length bounds)%nat ->
         fst (nth_default (0,0) bounds n) <= nth_default 0 p n <= snd (nth_default (0,0) bounds n).
-    Proof.
+    Proof using Type.
       intros p bounds H H0 H1.
       generalize dependent n.
       generalize dependent p.
@@ -575,7 +575,7 @@ Module SolinasReduction.
         is_bounded_by (bound1 ++ bound2) (l1 ++ l2) = true ->
         length bound1 = length l1 ->
         is_bounded_by bound1 l1 = true.
-    Proof.
+    Proof using Type.
       intros b1 b2 l1 l2 H H1.
       generalize dependent b1.
       generalize dependent b2.
@@ -598,7 +598,7 @@ Module SolinasReduction.
 
     Lemma fold_right_andb_default : forall d l,
         fold_right andb d l = true -> d = true.
-    Proof.
+    Proof using Type.
       intros; induction l;
         repeat multimatch goal with
                | H : context[fold_right _ _ _] |- _ => push' H
@@ -610,7 +610,7 @@ Module SolinasReduction.
         is_bounded_by (bound1 ++ bound2) (l1 ++ l2) = true ->
         length bound1 = length l1 ->
         is_bounded_by bound2 l2 = true.
-    Proof.
+    Proof using Type.
       intros b1 b2 l1 l2 H H1.
       generalize dependent b1.
       generalize dependent b2.
@@ -635,7 +635,7 @@ Module SolinasReduction.
         is_bounded_by bound1 l = true ->
         fold_andb_map' (fun x y => (fst y <=? fst x) && (snd x <=? snd y)) bound1 bound2 = true ->
         is_bounded_by bound2 l = true.
-    Proof.
+    Proof using Type.
       intros l bound1 bound2 H H0 H1.
       generalize dependent bound1.
       generalize dependent bound2.
@@ -658,7 +658,7 @@ Module SolinasReduction.
 
     Lemma bounds_same : forall b,
         fold_andb_map' (fun x y => (fst y <=? fst x) && (snd x <=? snd y)) b b = true.
-    Proof.
+    Proof using Type.
       intros b.
       repeat autounfold.
       induction b;
@@ -796,7 +796,7 @@ Module SolinasReduction.
     Lemma adjust_s_finished' fuel s w (s_nz:s<>0) :
       Rows.adjust_s weight fuel s = (w, true) ->
       Rows.adjust_s weight (S fuel) s = (w, true).
-    Proof.
+    Proof using Type.
       cbv [Rows.adjust_s].
       rewrite !fold_right_map.
       replace (rev (seq 0 (S fuel))) with (fuel :: rev (seq 0 fuel)).
@@ -814,7 +814,7 @@ Module SolinasReduction.
       (fuel' > fuel)%nat ->
       Saturated.Rows.adjust_s weight fuel s = (w, true) ->
       Saturated.Rows.adjust_s weight fuel' s = (w, true).
-    Proof.
+    Proof using Type.
       induction 1; intros; apply adjust_s_finished'; auto.
     Qed.
 
@@ -913,7 +913,7 @@ Module SolinasReduction.
 
     Lemma reduce1_cps_ok {T} base s c n m (f : list Z -> T) : forall p,
         reduce1_cps base s c n m p f = f (reduce1 base s c n m p).
-    Proof.
+    Proof using Type.
       intros.
       cbv [reduce1 reduce1_cps].
       break_match; reflexivity.
@@ -957,7 +957,7 @@ Module SolinasReduction.
 
     Lemma reduce_full_cps_ok {T} base s c n (f : list Z -> T) : forall p,
         reduce_full_cps base s c n p f = f (reduce_full base s c n p).
-    Proof.
+    Proof using Type.
       intros.
       cbv [reduce_full reduce_full_cps].
       repeat (rewrite reduce1_cps_ok ||
@@ -983,7 +983,7 @@ Module SolinasReduction.
 
     Lemma mul_no_reduce_cps_ok {T} base n (f : list Z -> T) : forall p q,
         mul_no_reduce_cps base n p q f = f (mul_no_reduce base n p q).
-    Proof.
+    Proof using Type.
       intros.
       cbv [mul_no_reduce mul_no_reduce_cps].
       break_match; reflexivity.
@@ -995,7 +995,7 @@ Module SolinasReduction.
 
     Lemma mulmod_cps_ok {T} base s c n (f : list Z -> T) : forall p q,
         mulmod_cps base s c n p q f = f (mulmod' base s c n p q).
-    Proof.
+    Proof using Type.
       intros.
       cbv [mulmod' mulmod_cps].
       rewrite mul_no_reduce_cps_ok, reduce_full_cps_ok.
@@ -1008,7 +1008,7 @@ Module SolinasReduction.
 
     Lemma mulmod_unfold base s c n : forall p q,
         mulmod' base s c n p q = mulmod_cps base s c n p q id.
-    Proof.
+    Proof using Type.
       intros.
       rewrite mulmod_cps_ok.
       reflexivity.
@@ -1016,7 +1016,7 @@ Module SolinasReduction.
 
     Lemma mulmod_cps_conv base s c n : forall p q,
       mulmod base s c n p q = mulmod' base s c n p q.
-    Proof.
+    Proof using Type.
       intros.
       rewrite mulmod_unfold.
       reflexivity.
@@ -1028,7 +1028,7 @@ Module SolinasReduction.
     Lemma split_lt w l1 l2:
       (forall x, In x l1 -> 0 < x < w) ->
       split w (combine l1 l2) = (combine l1 l2, []).
-    Proof.
+    Proof using Type.
       intros H.
       generalize dependent l2.
       induction l1 as [| ? ? IHl1]; intros l2; destruct l2; push;
@@ -1051,7 +1051,7 @@ Module SolinasReduction.
     Lemma split_gt w l1 l2:
       (forall x, In x l1 -> x mod w = 0) ->
       split w (combine l1 l2) = ([], combine (map (fun t => t / w) l1) l2).
-    Proof.
+    Proof using Type.
       intros H.
       generalize dependent l2.
       induction l1 as [| ? ? IHl1]; intros l2; destruct l2; push;
@@ -1074,7 +1074,7 @@ Module SolinasReduction.
 
     Lemma weight_mono' x :
       weight x < weight (S x).
-    Proof.
+    Proof using Type.
       weight_comp.
       rewrite Zred_factor0 at 1.
       rewrite Z.mul_comm.
@@ -1086,7 +1086,7 @@ Module SolinasReduction.
     Lemma weight_mono'' x1 x2 :
       (x2 > 0)%nat
       -> weight x1 < weight (x2 + x1).
-    Proof.
+    Proof using Type.
       intros H.
       induction H as [| ? ? IHle];
         repeat match goal with
@@ -1099,7 +1099,7 @@ Module SolinasReduction.
     Lemma weight_mono x1 x2 :
       (x1 < x2)%nat ->
       weight x1 < weight x2.
-    Proof.
+    Proof using Type.
       intros.
       replace x2%nat with ((x2 - x1) + x1)%nat by lia.
       apply weight_mono''; lia.
@@ -1108,7 +1108,7 @@ Module SolinasReduction.
     Lemma weight_mono_le x1 x2 :
       (x1 <= x2)%nat ->
       weight x1 <= weight x2.
-    Proof.
+    Proof using Type.
       intros H.
       apply le_lt_or_eq in H.
       intuition.
@@ -1119,7 +1119,7 @@ Module SolinasReduction.
     Lemma map_seq_start : forall a b,
         map weight (seq a b) =
           map (fun t => t * weight a) (map weight (seq 0 b)).
-    Proof.
+    Proof using Type.
       intros a b.
       induction b as [| ? IHb];
         repeat multimatch goal with
@@ -1155,7 +1155,7 @@ Module SolinasReduction.
 
     Lemma weight_dif_mono' : forall n,
         weight (S n) - weight n < weight (S (S n)) - weight (S n).
-    Proof.
+    Proof using Type.
       intros n.
       induction n.
       weight_comp; lia.
@@ -1171,7 +1171,7 @@ Module SolinasReduction.
     Lemma weight_dif_mono : forall n m,
         (n < m)%nat ->
         weight (S n) - weight n < weight (S m) - weight m.
-    Proof.
+    Proof using Type.
       intros n m H.
       induction H;
         repeat multimatch goal with
diff --git a/src/Assembly/Symbolic.v b/src/Assembly/Symbolic.v
index 026dca504..333d641b4 100644
--- a/src/Assembly/Symbolic.v
+++ b/src/Assembly/Symbolic.v
@@ -3299,10 +3299,6 @@ Definition SymexNormalInstruction {descr:description} (instr : NormalInstruction
     =>
     v <- Symeval (selectznz@(CF, dst, src));
     SetOperand dst v
-  | cmovo, [dst; src]
-    =>
-    v <- Symeval (selectznz@(OF, dst, src));
-    SetOperand dst v  
   | cmovnz, [dst; src] =>
     v <- Symeval (selectznz@(ZF, src, dst));
     SetOperand dst v
diff --git a/src/Assembly/WithBedrock/SymbolicProofs.v b/src/Assembly/WithBedrock/SymbolicProofs.v
index 795b4fd3d..7aaa4ec14 100644
--- a/src/Assembly/WithBedrock/SymbolicProofs.v
+++ b/src/Assembly/WithBedrock/SymbolicProofs.v
@@ -1264,18 +1264,13 @@ Proof using Type.
   Unshelve. all : match goal with H : context[Syntax.adox] |- _ => idtac | _ => shelve end.
   { cbn [fold_right] in *; rewrite ?Z.bit0_odd, ?Z.add_0_r, ?Z.add_assoc in *; assumption. }
 
-  Unshelve. all : match goal with H : context[Syntax.cmovc] |- _ => idtac |  H : context[Syntax.cmovb] |- _ => idtac | _ => shelve end.
+  Unshelve. all : match goal with H : context[Syntax.cmovc] |- _ => idtac |  H : context[Syntax.cmovb] |- _ => idtac |  H : context[Syntax.cmovo] |- _=> idtac | _ => shelve end.
   (* cmovc / cmovb / cmovo *)
   all: destruct vCF; cbn [negb Z.b2z Z.eqb] in *; eauto 9; [].
   all: enough (m = m0) by (subst; eauto 9).
   all: clear -Hm0 Hv frame G ; eauto using SetOperand_same.
   all: fail.
 
-  Unshelve. all: match goal with H : context[Syntax.cmovo] |- _=> idtac | _ => shelve end.
-  destruct vOF; cbn [negb Z.b2z Z.eqb] in *; eauto 9; [].
-  enough (m = m0) by (subst; eauto 9).
-  clear -Hm0 Hv frame G ; eauto using SetOperand_same.
-
   Unshelve. all : match goal with H : context[Syntax.cmovnz] |- _ => idtac | _ => shelve end.
   { (* cmovnz *)
     destruct vZF; cbn [negb Z.b2z Z.eqb] in *; eauto 9; [].
diff --git a/src/PushButtonSynthesis/SolinasReduction.v b/src/PushButtonSynthesis/SolinasReduction.v
index 68bd52013..3066dbe35 100644
--- a/src/PushButtonSynthesis/SolinasReduction.v
+++ b/src/PushButtonSynthesis/SolinasReduction.v
@@ -146,11 +146,7 @@ Section __.
     { lazymatch goal with
       | |- ?x = _ => rewrite surjective_pairing with (p:=x)
       end.
-      repeat match goal with
-             | H : ?x = _ |- (?x, _) = _ => rewrite H
-             | H : ?x = _ |- (_, ?x) = _ => rewrite H
-             end.
-      auto. }
+      congruence. }
   Qed.
 
   Local Notation evalf := (eval weight n).
```

JasonGross: I can also open a new PR here on fiat-crypto with this branch if you'd like

andres-erbsen:
```
-  | cmovo, [dst; src]
-    =>
-    v <- Symeval (selectznz@(OF, dst, src));
-    SetOperand dst v
```

why this?

JasonGross: I thought it was a conflict with your addition to the semantics on master

JasonGross: Split off into https://github.com/mit-plv/fiat-crypto/pull/1491

PR #1361, opened 2022-08-28
Port reification to Ltac2

JasonGross:
Closes #1350 

To be reviewed and merged in tandem with https://github.com/mit-plv/rewriter/pull/57

Note for future reference: timing done with
```
export branch=ltac2-reify; export n="$(git log 9af638cfe53c76f7bf8e38b8c22464af2e587cb6..origin/ltac2-reify --oneline  | wc -l)"; for i in $(seq $((n-1)) -1 0); do spec="$branch$(for j in $(seq 1 $i); do echo -n ^; done)"; git checkout "$spec" && if ! git log -1 | grep -q 'Timing Diff'; then ./etc/coq-scripts/timing/make-pretty-timed-diff-tip.sh -j5 pre-standalone SKIP_BEDROCK2=1 && ((git log -1 --pretty=%B; echo; echo '<details><summary>Timing Diff</summary>'; echo '<p>'; echo; echo '```'; cat ./time-of-build-both.log; echo; echo '```'; echo '</p>'; echo '</details>') | git commit --amend -F - ) && export sha="$(git log -1 --pretty=%H)"; git checkout $branch && git rebase $sha; git push origin $branch:$branch -f; fi; done; git checkout $branch; git push origin $branch:$branch -f
```
Bumps done with
```
export shas="$(cd rewriter; git log jgross/ltac2-reify-incremental-001..jgross/ltac2-reify-incremental-010 --format=%H | tac)"; for sha in $shas; do (cd rewriter/; descr="Bump rewriter to$(git log $sha -1 --format="%C(auto) %h %s")"; git checkout $sha; cd ..; git commit -am "$descr"); done
```

PR #1369, opened 2022-09-05
Garagedoor opens

andres-erbsen:


andres-erbsen:
I just merged the dependencies; this is ready to merge if it passes.

andres-erbsen:
`Error: Cannot find a physical path bound to logical path bedrock2Examples.LAN9250.`
Probably the thing to do here is to create another build category that depends on bedrock2 and rupicola examples...

andres-erbsen:
8.14:
```
File "/Users/runner/work/fiat-crypto/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/memmove.v", line 190, characters 12-63:
Error: No such bound variable m (possible names are: k, v1, v2 and m1).
```

andres-erbsen:
@JasonGross please consider whether it's worth getting this to build on 8.14 or whether we're likely enough to drop it anyway. My recollection of build discussion:

- [x] `make clean` does not remove `.vo` files from submodules (probably because these `.v` files are not listed in `_CoqProject`)
- [x] `.v` files not wanted in `_CoqProject` can be outside `src`
- [ ] `_CoqProject` generation (and the rest of the `Makefile`) should work without `.git`
- [ ] we want to be able to check locally that `.v`-files in `_CoqProject` are in git to avoid CI surprises, and enable this check by default. I think this should be a separate phony target from `_CoqProject` generation.

I'm willing to implement these if you expect you'd find the result acceptable. wdyt?

JasonGross:
I think we're going to drop 8.14, to get adequate Ltac2 support if nothing else, so you should just cherry-pick the first two commits of https://github.com/mit-plv/fiat-crypto/pull/1361

I think actually the Makefile already works without .git by falling back to `find`, though I could be misremembering.  I continue to be slightly sad about, e.g., having to pick a different directory to run the bug minimizer in.  If we already fall back to find, feel free to stop checking in the generated files.  Otherwise I'm happy to add a find fallback.

.dir-locals.el
```diff
@@ -1,23 +0,0 @@
-((coq-mode . ((eval . (let* ((project-root (locate-dominating-file buffer-file-name "_CoqProject"))
```
JasonGross, line 1: Why delete this?

JasonGross:
Hmmm, I guess we don't fallback to `find`.  Thoughts on using git when it's available and otherwise `find`?

.dir-locals.el
andres-erbsen: It's redundant as `Makefile`-generated `_CoqProject` already passes `-Q` with these directories.

andres-erbsen:
`make: *** No rule to make target 'rupicola/src/Rupicola/Lib/Tactics.vo', needed by 'src/Assembly/WithBedrock/Proofs.vo'.  Stop.`
I vaguely remember @samuelgruetter  had to do some hacks to get `-Q` to work on Windows...

samuelgruetter:
> I vaguely remember @samuelgruetter had to do some hacks to get `-Q` to work on Windows...

I don't remember inventing or using any hacks for Windows myself, but here's one by @JasonGross that might be relevant:

https://github.com/mit-plv/fiat-crypto/blob/64aa176060d7575426c3161d39c2f3e9713682d0/Makefile#L438-L442

andres-erbsen:
Yeah, looks possible that it worked before I just broke it when cleaning up the Makefile...

Makefile
```diff
@@ -140,22 +119,6 @@ BEDROCK2_FILES_PATTERN := \
...
-else
-ifeq ($(SKIP_BEDROCK2),)
-SKIP_BEDROCK2=1
-else
-$(error Cannot build bedrock2! Pass FORCE_BEDROCK2=1 to override this error and build anyway, or pass SKIP_BEDROCK2=1 (instead of SKIP_BEDROCK2=$(SKIP_BEDROCK2)) to skip bedrock2)
-endif
-endif
-endif
-endif
-
```
JasonGross, line 158: Why remove this?  It seems useful to be able to easily desync the bedrock2 minimal version from the fiat-crypto minimal version, no?

Makefile
```diff
@@ -580,7 +542,7 @@ endif
 
 Makefile.coq: Makefile _CoqProject
 	$(SHOW)'COQ_MAKEFILE -f _CoqProject > $@'
-	$(HIDE)$(COQBIN)coq_makefile -f _CoqProject INSTALLDEFAULTROOT = $(INSTALLDEFAULTROOT) -o Makefile-coq && cat Makefile-coq | sed 's/^printenv:/printenv::/g; s/^printenv:::/printenv::/g; s/^all:/all-old:/g; s/^validate:/validate-vo:/g; s/^.PHONY: validate/.PHONY: validate-vo/g' > $@ && rm -f Makefile-coq
+	$(HIDE)$(COQBIN)coq_makefile -f _CoqProject INSTALLDEFAULTROOT = $(INSTALLDEFAULTROOT) -o $@
```
JasonGross, line 545: Note that this will result in duplicated targets for `Makefile.coq`.  idk how bad this is

JasonGross:
> `make: *** No rule to make target 'rupicola/src/Rupicola/Lib/Tactics.vo', needed by 'src/Assembly/WithBedrock/Proofs.vo'. Stop.` I vaguely remember @samuelgruetter had to do some hacks to get `-Q` to work on Windows...

This is still an issue.  I think the problem is that deps is for some reason not building this .vo file.

(What's the reason to update the Makefile logic in the same PR as updating code?)

Makefile
```diff
@@ -106,27 +103,9 @@ endif
...
-
-ifneq ($(EXISTING_COQPROJECT_CONTENTS_SORTED),$(NEW_COQPROJECT_CONTENTS_SORTED))
-.PHONY: _CoqProject
-_CoqProject:
-	$(SHOW)'ECHO > $@'
-	$(HIDE)$(COQPROJECT_CMD) > $@
-endif
-endif
+	src/ExtractionHaskell/%.vo \
+	src/Rewriter/PerfTesting/Specific/generated/%.v
```
JasonGross, line 107:
```suggestion
	src/Rewriter/PerfTesting/Specific/generated/%.vo
```
This is SPECIAL_VOFILES

PR #1376, opened 2022-09-16
bump rewriter

andres-erbsen:


andres-erbsen:
```
File "src/Rewriter/Util/plugins/inductive_from_elim.ml", line 43, characters 31-53:
43 |       EConstr.to_constr sigma (EConstr.Vars.subst_var sigma name cty)
                                    ^^^^^^^^^^^^^^^^^^^^^^
Error: This function has type Names.Id.t -> EConstr.t -> EConstr.t
       It is applied to too many arguments; maybe you forgot a `;'.
```

JasonGross:
If the master PPA is out of date, maybe we should switch the master job to use the docker image (I've been less on top of maintaining ppa breakages lately)

PR #1391, opened 2022-09-30
Bump rupicola

JasonGross:


JasonGross:
I guess this is not going to work:
```
File "./src/Bedrock/End2End/X25519/MontgomeryLadder.v", line 64, characters 10-27:
Error:
In environment
montladder_compiler_result := ?Goal0 : ?Goal
The term "map.of_list funcs" has type "map.rep"
while it is expected to have type
 "list (string * (list string * list string * cmd))".
```
so I guess we should just merge #1369 all at once

PR #1393, opened 2022-10-01
Makefile fixups

andres-erbsen:
I tested that build succeeded locally with SKIP_BEDROCK2=1 and then running make again started building bedrock2 files.


JasonGross (approved):
Thanks!

PR #1400, opened 2022-10-02
Only run validate on Coq master / dev

JasonGross:
It seems to be timing out on Ubuntu LTS or something, and we don't really need to see the validate output from multiple versions.  Insofar as we want to stress coqchk, we sort-of only care about stressing the newest version, I think.  Note that validate already only runs on commits to master, not on PRs, so this won't speed up PR CI timing at all.


JasonGross:
This is not going to break anything, and I want to tag a release, so I'm going to merge

PR #1407, opened 2022-10-04
Add stacksize warnings to avoid confusion

JasonGross:
c.f. https://github.com/coq/platform/issues/178#issuecomment-1265444555

@andres-erbsen Does this seem like a good idea to you?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -589,10 +589,12 @@ $(STANDALONE_HASKELL:%=src/ExtractionHaskell/%.hs) : %.hs : %.v src/haskell.sed
 # unix package needed for Unix.gettimeofday for the perf_* binaries
 $(STANDALONE_OCAML:%=src/ExtractionOCaml/%.cmi) : %.cmi : %.ml
 	$(SHOW)'$(CAMLOPT_PERF_SHOW) $*.mli'
+	$(HIDE)etc/check_stack_limit.sh || true
 	$(HIDE)$(TIMER) $(CAMLOPT_PERF) -package unix -w -20 -g $*.mli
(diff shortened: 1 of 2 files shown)


andres-erbsen:
Sure. Please make the shell script use `-eu` so reading I can tell which commands' failure you're expecting to handle when reading it in order. In particular we should not break the build when `ulimit` is not available or errors on the request.

Also, could we just recommend `unlimited` instead of a large number?

JasonGross:
> Sure. Please make the shell script use `-eu` so reading I can tell which commands' failure you're expecting to handle when reading it in order.

@andres-erbsen  Updated.  Is the current version better?

> In particular we should not break the build when `ulimit` is not available or errors on the request.

Subshell errors don't quit bash.  And anyway there's an `|| true` in the `Makefile`, so it's not going to break the build.  But I've made this more explicit in the script.

> Also, could we just recommend `unlimited` instead of a large number?

I want to not print this warning if the stack size is big enough.  Based on Coq's CI, the suggested number is big enough.  What's the benefit of recommending unlimited?

andres-erbsen:
Sure, the script is now better: now I can see that you in fact want nonzero exit status when the desired ulimit setting is not detected. With the new factoring you might as well use printf instead of echo and variable substitution, but that's fine. I don't feel strongly about reocmmending unlimited; just any stack limit smaller than total memory limit seems silly for functional programming.

JasonGross:
Note that this is the stack size required to compile our code with ocamlopt, not the stack required to run our code (which is less, I believe).

PR #1424, opened 2022-10-08
Adjust etc/machine.sh to work better on Arch

JasonGross:
Old version prints
```
./etc/machine.sh: line 21: hostname: command not found
-noht-tb-nops-schedutil-haswell
option:
```
because
```
$ gcc -march=native -Q --help=target|grep march
  -march=                               haswell
  Known valid arguments for -march= option:
```
New version prints
```
transparent-rooster-noht-tb-nops-schedutil-haswell
```

### DIFF
--- a/etc/machine.sh
+++ b/etc/machine.sh
@@ -18,7 +18,14 @@ online_governors() {
   fi
 }
 
-printf "$(hostname)"
+# hostname is not on arch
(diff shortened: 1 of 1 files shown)


andres-erbsen:
Sure. I'm not sure how to prioritize hostname, hostnamectl, and /etc/hostname, but that seems like a fine fix for the concrete case.

PR #1428, opened 2022-10-13
Optimize Proof while proving riscv verify

andres-erbsen:
I hope this might reduce memory usage sufficiently to pass Coq CI


JasonGross:
Helps a little but not enough:

Coq Version | Mem Before this PR | Mem After this PR
--|--|--
Ubuntu LTS | 6202932 ko | 4875708 ko
master | 5430044 ko | 4368000 ko

PR #1429, opened 2022-10-13
use vm_compute instead of Ltac to check instruction bounds

samuelgruetter:
Should improve https://github.com/mit-plv/fiat-crypto/issues/1427

Closes #1428

### DIFF
--- a/src/Bedrock/End2End/X25519/GarageDoor.v
+++ b/src/Bedrock/End2End/X25519/GarageDoor.v
@@ -166,7 +166,7 @@ Definition garagedoor_iteration : state -> list (lightbulb_spec.OP _) -> state -
     lightbulb_spec.lan9250_recv_packet_too_long _ ioh \/
     TracePredicate.concat TracePredicate.any (lightbulb_spec.spi_timeout _) ioh) \/
   (exists incoming, lightbulb_spec.lan9250_recv _ incoming ioh /\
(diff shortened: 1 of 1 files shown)


src/Bedrock/End2End/X25519/GarageDoor.v
```diff
@@ -966,36 +966,12 @@ Require Import compiler.CompilerInvariant.
...
-  Time do 10 p 50%nat.
-  match goal with |- Forall _ ?l => replace l with (@nil Decode.Instruction) end.
-  { trivial. }
-  { vm_compute. trivial. }
-Time Qed.
-
-Definition good_trace s t s' := 
-  exists ioh, SPI.mmio_trace_abstraction_relation ioh t /\ 
+  bverify.bvalidInstructions Decode.RV32IM garagedoor_insns = true.
+Proof. vm_compute. reflexivity. Qed.
```
JasonGross, line 971: Why not
```suggestion
Proof. vm_cast_no_check (eq_refl true). Qed.
```
?  (Is it slow enough to matter?)

andres-erbsen: https://github.com/mit-plv/fiat-crypto/pull/1431

samuelgruetter: > ? (Is it slow enough to matter?)

It's instantaneous, so the question I asked myself was whether to keep it as a separate lemma at all, not how to optimize it 😉

PR #1430, opened 2022-10-13
Automatically increase stack size

JasonGross:
Requested at https://coq.zulipchat.com/#narrow/stream/247791-fiat-crypto/topic/Fiat.20crypto.20and.20flambda/near/303701926


Makefile.standalone
```diff
@@ -26,17 +26,23 @@ endif
 	standalone-haskell install-standalone-haskell uninstall-standalone-haskell \
 	#
 
+ifneq ($(SET_ULIMIT_STACK),1)
+CHECK_STACK_LIMIT:=etc/check_stack_limit.sh || true
+else
+CHECK_STACK_LIMIT:=etc/check_stack_limit.sh 2>/dev/null || ulimit -S -s 32768 || etc/check_stack_limit.sh || true
```
andres-erbsen, line 32: rename to `STACK_LIMIT`?

maybe pass the value of `SET_ULIMIT_STACK` to ulimit?

JasonGross: > maybe pass the value of `SET_ULIMIT_STACK` to ulimit?

That sort-of defeats the purpose of users not having to think about what stack value they need to pass

andres-erbsen: if we don't want to allow configurability, perhaps we should do this unconditionally?

JasonGross: Good idea.  I've updated it, does it look good now?

PR #1478, opened 2022-11-03
Func notation

andres-erbsen:


andres-erbsen:
Merged as https://github.com/mit-plv/fiat-crypto/commit/b6dae11c838d026745f1886a93dc40e524a94b24 after a history-only rebase in rupicola.

PR #1342, opened 2022-07-18
More efficient datastructures for asm dag

JasonGross:
This reverts commit de47c096b0f0bbee3763ab6293ed8e8e41413147.

They are not actually more efficient, though; I'm creating this PR as a placeholder to remind me to work on this / #1293

PR #1187, opened 2022-04-12
Add some assembly tests from BoringSSL p256

JasonGross:
Towards better equivalence checking

cc @davidben @andres-erbsen 

We don't support these instructions yet:
```
$ src/ExtractionOCaml/word_by_word_montgomery p256 64 '2^256 - 2^224 + 2^192 + 2^96 - 1' mul --no-wide-int --shiftr-avoid-uint1 --hints-file fiat-amd64/boringssl_intel_manual_mul_p256.asm
[...]
In parsing:
Error while parsing assembly:
mul    r9
mul    r10
mul    r11
mul    r12
mul    r15
mov    rax,QWORD PTR [rbx+0x8]
mul    QWORD PTR [rsi]
mul    QWORD PTR [rsi+0x8]
mul    QWORD PTR [rsi+0x10]
mul    QWORD PTR [rsi+0x18]
mul    r15
mov    rax,QWORD PTR [rbx+0x10]
mul    QWORD PTR [rsi]
mul    QWORD PTR [rsi+0x8]
mul    QWORD PTR [rsi+0x10]
mul    QWORD PTR [rsi+0x18]
mul    r15
mov    rax,QWORD PTR [rbx+0x18]
mul    QWORD PTR [rsi]
mul    QWORD PTR [rsi+0x8]
mul    QWORD PTR [rsi+0x10]
mul    QWORD PTR [rsi+0x18]
mul    r15
cmovb  r12,rcx
cmovb  r13,rbp
mov    QWORD PTR [rdi],r12
cmovb  r8,rbx
mov    QWORD PTR [rdi+0x8],r13
cmovb  r9,rdx
mov    QWORD PTR [rdi+0x10],r8
mov    QWORD PTR [rdi+0x18],r9
repz ret

Fatal error: exception Failure("Synthesis failed")
```

PR #1481, opened 2022-11-07
Added dettman multiplication algorithm, made changes to equivalence checker

OwenConoly:
Two main things that I did here:
1. Adding the coq implementation of the "dettman multiplication" algorithm.
2. Changes to the assembly equivalence checker:
    a. A canonicalization of sums of addcarries, which effectively makes the equivalence checker aware that double-wide (e.g., 128-bit) addition is associative.  This is analogous to the preexisting canonicalization of commutative operations, which just sorted the operands.
    b. A (new, in-some-ways better) implementation of bounds on integer values corresponding to nodes in the dag.  Instead of just revealing a few layers of the dag, and using that to bound an expression, I changed the data type of the dag, so that as nodes are inserted into the dag, their bounds are inserted along with the node and its description.  The bound on each node can be calculated in terms of the bounds on its operands.  So now, the bounds can "look to arbitrary depth", using this dynamic-programming type of strategy.  
     c. A new rewriting rule, which replaces or operations with add operations in some cases.  When we have a node of the form (orZ, [x; y]), for example, and x is a multiple of 16, and y is upper-bounded by 15, we can rewrite this as (addZ, [x; y]).  This uses the bounds that I implemented.
     d. A new cli optional argument, "--extra-rewrite-rule", which allows optional rewrite rules (which are off by default) to be "turned on" when the equivalence checker is invoked.  Currently, the only available optional rewrite rule is the or-to-add rule I mentioned in part (c), which can be turned on by writing "--extra-rewrite-rule or-to-add".  It would be easy to add more optional rewrites.  I found it necessary to add this option because the or-to-add rule was necessary for some dettman verification, but it was breaking some solinas verification.


andres-erbsen:
please rebase, perhaps `git fetch -a -v && git pull --rebase origin master`

OwenConoly:
> please rebase, perhaps `git fetch -a -v && git pull --rebase origin master`

Ok!

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -0,0 +1,212 @@
...
+Proof.
+  cbv [mulmod_general carry' reduce'].
+  repeat (try rewrite Positional.eval_from_associational;
+          try rewrite Positional.eval_to_associational; 
+          try rewrite Associational.eval_carry; 
+          try rewrite eval_reduce_one;
+          try rewrite eval_carry_down;
+          try rewrite eval_dedup_weights;
+          try rewrite eval_loop').
+  rewrite Associational.eval_mul. repeat rewrite Positional.eval_to_associational. reflexivity.
```
JasonGross, line 172: I think we have a `push_eval` rewrite database that you can use `Hint Rewrite` with, and then `autorewrite with push_eval` or something?

src/Assembly/Symbolic.v
```diff
@@ -511,7 +516,330 @@ Definition node_beq {A : Set} (arg_eqb : A -> A -> bool) : node A -> node A -> b
 Global Instance reflect_node_beq {A : Set} {arg_eqb} {H : reflect_rel (@eq A) arg_eqb}
   : reflect_rel eq (@node_beq A arg_eqb) | 10 := _.
 
+Definition keep n x := Z.land x (Z.ones (Z.of_N n)).
+
+Local Open Scope Z_scope.
+
+Definition in_bounds (x : Z) (r : option (Z*Z)) :=
```
JasonGross, line 523: I think it'd be nicer to do
```suggestion
Definition in_bounds (x : Z) (r : option zrange) :=
```
here and elsewhere

src/Assembly/Symbolic.v
```diff
@@ -511,7 +516,330 @@ Definition node_beq {A : Set} (arg_eqb : A -> A -> bool) : node A -> node A -> b
...
+  apply Z_mod_lt with (a := v) in H. lia.
+Qed.
+
+Lemma div_mul_le x y : 0 < y -> x / y * y <= x.
+Proof.
+  intros H. assert (H': y <> 0) by lia. apply Z.mod_bound_or with (a := x) in H'.
+  destruct H' as [H'|H'].
+  - rewrite Zmod_eq in H' by lia. destruct H' as [H' _]. lia. 
+  - lia. 
+Qed.
```
JasonGross, line 558: Please move these to the appropriate ZUtil files

src/Assembly/Symbolic.v
```diff
@@ -511,7 +516,330 @@ Definition node_beq {A : Set} (arg_eqb : A -> A -> bool) : node A -> node A -> b
...
+    + apply Z.le_trans with (m := arg * min2).
+      -- apply Z.mul_le_mono_nonneg_r; lia.
+      -- apply Z.mul_le_mono_nonneg_l; lia.
+    + apply Z.le_trans with (m := arg * max2).
+      -- apply Z.mul_le_mono_nonneg_l; lia.
+      -- apply Z.mul_le_mono_nonneg_r; lia.
+Qed.
+
+(* input bounds should be nonnegative.  if lower bound is less than zero, returns None *)
+Definition bound_node (o : op) (arg_bounds : list (option (Z*Z))) : option (Z*Z) :=
```
JasonGross, line 742: Some of this might be duplicative with https://github.com/mit-plv/fiat-crypto/pull/1187, which has a decent amount of work on reusing the PHOAS bounds analysis for assembly operations (which is hopefully more future-proof than redoing all the bounding work in two different ways).  Might be worth considering polishing / using some of that code?

src/Assembly/Symbolic.v
```diff
@@ -511,7 +516,330 @@ Definition node_beq {A : Set} (arg_eqb : A -> A -> bool) : node A -> node A -> b
...
+  | addZ, Some actual_bounds => Some (accumulate Z.add (0, 0) actual_bounds)
+  | mulZ, Some actual_bounds => if bounds_positive actual_bounds then
+                                  Some (accumulate Z.mul (1, 1) actual_bounds)
+                                else
+                                  None
+  | (old s _ | slice _ s | mul s | shl s | shr s | sar s | neg s | and s | or s | xor s), _ => Some (0, Z.ones (Z.of_N s))
+  | _, _ => None
+  end.
+
+Ltac thing:= match goal with
```
JasonGross, line 768: Just use `inversion_head @Forall2` or similar, in `Crypto.Util.Tactics.InversionHead` (or `Crypto.Util.Tactics.DestructHead`; I don't recall)

src/Assembly/Symbolic.v
```diff
@@ -525,13 +853,18 @@ Existing Class node_ok.
 Hint Extern 1 (node_ok ?i ?n) => exact (@new_node_ok n I i) : typeclass_instances.
 Module Old.
 Module dag.
-  Definition t : Type := list (node idx * description).
+  Definition t : Type := list (node idx * description * (option (Z*Z))).
```
JasonGross, line 856:
```suggestion
  Definition t : Type := list (node idx * description * option zrange).
```

JasonGross: Please also update `Module New` and ensure that nothing below breaks when replacing `Export Old.` with `Export New.`; I'd like to make it easy to profile alternate dag datastructures and hopefully switch to a more efficient one at some point.

src/Assembly/Symbolic.v
```diff
@@ -525,13 +853,18 @@ Existing Class node_ok.
...
   Definition empty : t := nil.
   Definition size (d : t) : N := N.of_nat (List.length d).
   Definition lookup (d : t) (i : idx) : option (node idx)
-    := option_map fst (List.nth_error d (N.to_nat i)).
+    := option_map (fun x => fst (fst x)) (List.nth_error d (N.to_nat i)).
+  Definition lookup_bounds (d : t) (i : idx) : option (Z*Z)
```
JasonGross, line 861:
```suggestion
  Definition lookup_bounds (d : t) (i : idx) : option zrange
```

src/Assembly/Symbolic.v
```diff
@@ -525,13 +853,18 @@ Existing Class node_ok.
...
+    :=  match (List.nth_error d (N.to_nat i)) with
+        | Some thing => snd thing
```
JasonGross, line 863:
```suggestion
        | Some (_, bounds) => bounds
```

src/Assembly/Symbolic.v
```diff
@@ -1282,17 +1615,131 @@ End WithDag.
 Definition merge_node {descr : description} (n : node idx) : dag.M idx
   := dag.merge_node n.
 
-Fixpoint merge {descr : description} (e : expr) (d : dag) : idx * dag :=
+Require Import coqutil.Tactics.autoforward coqutil.Decidable coqutil.Tactics.Tactics.
+Global Set Default Goal Selector "1".
```
JasonGross, line 1619: Please don't set things globally like this in random files.  Do
```suggestion
Local Set Default Goal Selector "1".
```
instead

src/Assembly/Symbolic.v
```diff
@@ -1282,17 +1615,131 @@ End WithDag.
...
+                                  (Z.ldiff (Z.ones (Z.succ (Z.log2 a))) (Z.ones (Z.of_N w))))
+      | _, _ => None
+      end
+  | ExprApp ((old s _ | slice _ s | mul s | shl s | shr s | sar s | neg s | and s | or s | xor s), _) => Some (Z.ones (Z.of_N s))
+  | ExprApp ((addcarry _ | subborrow _ | addoverflow _ | iszero), _) => Some 1
+  | _ => None
+  end%Z.
+
+Import coqutil.Tactics.Tactics.
+Ltac t:= match goal with
```
JasonGross, line 1649:
```suggestion
Local Ltac t:= match goal with
```

src/Assembly/Symbolic.v
```diff
@@ -1282,17 +1615,131 @@ End WithDag.
...
+  induction He; cbn in *; repeat t.
+  { cbv [fold_right]; Lia.lia. }
+  destruct (bound_expr _) eqn:? in *; cbn in *; repeat t.
+  destruct (fold_right (B:=option _) _) eqn:? in *; cbn in *; repeat t.
+  specialize (IHHe _ ltac:(eassumption) _ ltac:(eassumption)); cbn.
+  specialize (H _ ltac:(exact eq_refl) _ _ ltac:(eassumption)).
+  Lia.lia.
+Qed.
+
+Require Import Util.ZRange.LandLorBounds.
```
JasonGross, line 1687: Please move this up to the top

src/Assembly/Symbolic.v
```diff
@@ -1282,17 +1615,131 @@ End WithDag.
...
+Lemma eval_bound_expr G e b : bound_expr e = Some b ->
+  forall d v, eval G d e v -> (0 <= v <= b)%Z.
+Proof using Type.
+  revert b; induction e; simpl bound_expr; BreakMatch.break_match;
```
JasonGross, line 1691: Please `Require Import Crypto.Util.Tactics.BreakMatch` rather than qualifying this.

src/Assembly/Symbolic.v
```diff
@@ -1282,17 +1615,131 @@ End WithDag.
...
+    inversion 2; intros; inversion_option; subst;
+    try match goal with H : context [set_slice] |- _ => shelve end;
+    cbv [interp_op] in *;
+    BreakMatch.break_match_hyps; inversion_option; subst;
+    rewrite ?Z.ldiff_ones_r, ?Z.land_ones, ?Z.ones_equiv;
+    cbv [Z.b2z];
+    try match goal with |- context [(?a mod ?b)%Z] => unshelve epose proof Z.mod_pos_bound a b ltac:(eapply Z.pow_pos_nonneg; Lia.lia) end;
+    repeat t;
+    try (Z.div_mod_to_equations; Lia.lia).
+  { clear dependent args'0.
```
JasonGross, line 1701: Please don't use autogenerated names

src/Assembly/Symbolic.v
```diff
@@ -1282,17 +1615,131 @@ End WithDag.
...
+    epose proof bound_sum' _ ltac:(eassumption) _ ltac:(eassumption) _ ltac:(eassumption) _ ltac:(eassumption).
+    split; try Lia.lia.
```
JasonGross, line 1703: Please add `Require Import Coq.micromega.Lia.` instead of qualifying this

src/Assembly/Symbolic.v
```diff
@@ -1338,9 +1785,22 @@ Proof using Type.
...
 Qed.
 
+Lemma eval_permute_commutative ctx d o args1 args2 n :
+  commutative o = true ->
+  Permutation.Permutation args1 args2 ->
+  eval ctx d (ExprApp (o, args1)) n ->
+  eval ctx d (ExprApp (o, args2)) n.
+Proof.
+  intros. inv H1. apply (Permutation_Forall2 H0) in H4.
+  destruct H4 as [args2' [H4 H5] ]. apply EApp with (args' := args2').
```
JasonGross, line 1795: No autogenerated names please

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
-  cbv [gensym_dag_ok dag_ok gensym_ok].
-  repeat match goal with |- _ /\ _ => split end; try exact _; intros *;
-    rewrite ?dag.lookup_empty; try congruence.
+Lemma empty_gensym_dag_ok : gensym_dag_ok (fun _ => None) dag.empty.
+Proof using Type.
+  cbv [gensym_dag_ok dag_ok gensym_ok].
+  repeat match goal with |- _ /\ _ => split end; try exact _; intros *;
+    rewrite ?dag.lookup_empty; try congruence.
+
+  cbv [bounds_ok]. intros. inv H. rewrite dag.lookup_empty in H1. discriminate H1.
```
JasonGross, line 1821: No autogenerated names please

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
+  destruct (i <? dag.size d1)%N eqn:E. 
+  - cbv [dag.lookup]. rewrite nth_error_app. destruct (lt_dec _ _).
+    + reflexivity.
+    + apply N.ltb_lt in E. exfalso. apply n. cbv [dag.size] in E. lia.
+  - cbv [dag.lookup]. rewrite nth_error_app. destruct (lt_dec _ _).
+    + apply N.ltb_nlt in E. exfalso. apply E. cbv [dag.size]. lia.
+    + cbv [dag.size]. replace (N.to_nat (i - N.of_nat (Datatypes.length d1))) with (N.to_nat i - Datatypes.length d1)%nat by lia. reflexivity.
+Qed.
+
+Lemma something ctx G d1 d2 i v n0 :
```
JasonGross, line 1848: Please pick a better name

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
+        by (rewrite dag.fst_merge_node; break_innermost_match; congruence)
+    end.
+    eexists; econstructor;
+      [ rewrite dag.lookup_merge_node' by assumption; reflexivity
+      | eauto using Forall2_weaken, eval_weaken_merge_node .. ]. }
+Qed.
+
+(* begin addcarry stuff *)
+
+Print bound_node.
```
JasonGross, line 1993:
```suggestion
```

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
+
+Fixpoint bound_expr' d e : option (Z * Z) :=
+  match e with
+  | ExprApp (o, args) => bound_node o (map (bound_expr' d) args)
+  | ExprRef i => dag.lookup_bounds d i
+  end.
+
+Search bound_node.
+Check eval.
```
JasonGross, line 2002:
```suggestion
```

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
+  destruct E1 as [m1 E1]. destruct E2 as [m2 E2].
+  exists m1. exists m2. split.
+  - simpl. f_equal. rewrite Z.add_0_r. simpl in H4. injection H4 as H4. simpl in E1. simpl in E2. subst.
+    injection E1 as E1. injection E2 as E2. subst. rewrite <- break_addZ_small. reflexivity.
+  - split.
+    + apply EApp with (args' := l1'); assumption.
+    + apply EApp with (args' := l2'); assumption.
+Qed.
+
+Lemma app_is_cons {X} (x : X) (l : list X):
```
JasonGross, line 2133: Please move to ListUtil

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
+  x :: l = [x] ++ l.
+Proof. reflexivity. Qed.
+
+Lemma and_pos : forall x y,
```
JasonGross, line 2137: Please move to some ZUtil file

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
+  Z.le 0 y -> Z.le 0 (Z.land x y).
+Proof.
+  intros. cbv [Z.land]. destruct x as [| a| a]; destruct y as [| b| b]; lia.
+Qed.
+
+Lemma le_and_ones y n:
```
JasonGross, line 2143: Please move to some ZUtil file

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
+           apply (mod_addZ_eq'''_backwards) in H6.
+           destruct H6 as [n2' [H6 H6'] ].
+           apply (break_add (n1' mod 2 ^ Z.of_N s) (n2' mod (2 ^ Z.of_N s))).
+           --- simpl. f_equal. repeat rewrite Z_land_ones'. repeat rewrite Z.add_0_r. rewrite H5'. rewrite <- H6'. rewrite <- Zplus_mod.
+               reflexivity. 
+           --- apply IHf'. apply H5.
+           --- apply IHargs1. apply H6.
+Qed.
+
+Fixpoint withadc_without (d : dag) (s : OperationSize) (args : list idx) : (list (list (list idx))) * (list idx) :=
```
JasonGross, line 2365: What does this do?  Please add a comment or pick a more informative name

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
+                ---- constructor.
+              +++ simpl. reflexivity.
+           --- assumption.
+      -- apply Lemma. rewrite <- E. destruct op; reflexivity.
+    + apply Lemma. rewrite E. reflexivity.
+Qed.
+
+Local Open Scope nat_scope.
+
+Require Import Coq.Sorting.Mergesort.
```
JasonGross, line 2493: Please move to the top

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
+      | None => None
+      end
+  end.
+
+Lemma lists_equal (l1 l2 : list N) :
+  l1 = l2 <->
+  list_beq N N.eqb l1 l2 = true.
+Proof.
+  intros. apply Bool.reflect_iff. apply reflect_eq_list.
+Qed.
```
JasonGross, line 2535: I don't think this deserves its own lemma.  Presumably you want some variant of the `reflect_hyps` tactic?

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
+  Theorem leb_total : forall x y, x <=? y = true \/ y <=? x = true.
+  Proof.
+    intros x. induction x as [| x0 x' IHx'].
+    - intros y. left. reflexivity.
+    - intros y. destruct y as [|y0 y'].
+      + right. reflexivity.
+      + simpl. rewrite <- (N.eqb_sym x0 y0). destruct (x0 =? y0)%N. 
+        -- apply IHx'.
+        --  unfold leb. repeat rewrite N.leb_le. lia. Qed.
+End lexOrder.
```
JasonGross, line 2879: We already have this:
https://github.com/mit-plv/fiat-crypto/blob/b6dae11c838d026745f1886a93dc40e524a94b24/src/Util/Structures/Orders/List.v#L505
https://github.com/mit-plv/fiat-crypto/blob/b6dae11c838d026745f1886a93dc40e524a94b24/src/Util/Structures/Orders/List.v#L201-L208

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
+  - intros. simpl in E. inv E. auto.
+  - intros. simpl in E. destruct (fold_right _ _ l') as [idxs1 d1]. simpl in E.
+    assert (H': (idxs1, d1) = (idxs1, d1)) by reflexivity.
+    apply IHl' in H'. clear IHl'. destruct H' as [H2 [H_ok1 H_eval1] ].
+    apply (eval_standardize_adc_sum _ _ s _ _ _ H H_ok1) in H2. rewrite E in H2. clear E H H_ok1.
+    destruct H2 as [H2 [H3 H4] ]. auto.
 Qed.
 
-Require Import coqutil.Tactics.autoforward coqutil.Decidable coqutil.Tactics.Tactics.
-Global Set Default Goal Selector "1".
```
JasonGross, line 1399: Ew, we already had this?  @andres-erbsen Do you know why this was here and global?

src/Assembly/Symbolic.v
```diff
@@ -2380,6 +4095,8 @@ Definition shift_to_mul :=
 Global Instance shift_to_mul_ok : Ok shift_to_mul.
 Proof. t; cbn in *; rewrite ?Z.shiftl_mul_pow2, ?Z.land_0_r by lia; repeat (lia + f_equal). Qed.
 
+Print eval.
+
```
JasonGross, line 4099:
```suggestion
```

src/Assembly/Symbolic.v
```diff
@@ -2454,19 +4171,53 @@ Definition combine_consts_pre : expr -> expr :=
...
+Definition g := ExprApp (add 64%N, [a; b; c]).
+Definition h := ExprApp (mul 64%N, [a; b; c]).
+
+Compute f.
+Compute (combine_consts_pre f).
+Compute (split_consts (mul 64%N) 1 [ExprRef 1%N; ExprRef 2%N; b; c]).
+Compute (group_consts (split_consts (mul 64%N) 1 [c; b; b; c])).
+Compute (compress_consts (add 64%N) (group_consts (split_consts (mul 64%N) 1 [b; b; c]))).
+Compute (app_consts (mul 64%N) (compress_consts (add 64%N) (group_consts (split_consts (mul 64%N) 1 [b; b; c])))).
+
```
JasonGross, line 4204:
```suggestion
```

src/Assembly/Symbolic.v
```diff
@@ -2454,19 +4171,53 @@ Definition combine_consts_pre : expr -> expr :=
...
 Definition cleanup_combine_consts : expr -> expr :=
   let simp_outside := List.fold_left (fun e f => f e) [flatten_associative] in
   let simp_inside := List.fold_left (fun e f => f e) [constprop;drop_identity;unary_truncate;truncate_small] in
   fun e => simp_outside match e with ExprApp (o, args)  =>
     ExprApp (o, List.map simp_inside args)
                    | _ => e end.
+Compute (cleanup_combine_consts (combine_consts_pre f)).
```
JasonGross, line 4211:
```suggestion
```

src/Assembly/Symbolic.v
```diff
@@ -2868,46 +4619,397 @@ Qed.
...
+  ;(addcarry_small, None)
+  ;(addoverflow_small, None)
+  ;(addbyte_small, None)
+  ;(xor_same, None)
+  ]%string.
+
+Search (map (filter _)).
+
+Check List.fold_right.
+
```
JasonGross, line 4964:
```suggestion
```

src/Assembly/Symbolic.v
```diff
@@ -2868,46 +4619,397 @@ Qed.
...
+  intros. cbv [or_to_add]. destruct e as [i| [o args] ]; try apply H0. destruct o; try apply H0.
+  - destruct (or_is_add _ _) eqn:E; try apply H0. apply eval_ors_is_adds; assumption.
+  - destruct (or_is_add _ _) eqn:E; try apply H0. apply eval_orZ_is_addZ; assumption.
+Qed.
+
+(* end or-to-add stuff *)
+
+Check in_dec.
+Search (string -> string -> bool).
+
```
JasonGross, line 4932:
```suggestion
```

src/Assembly/Symbolic.v
```diff
@@ -2868,46 +4619,397 @@ Qed.
...
+(* "named" rewrite rules are off by default and can be turned on using the commmand-line option*)
```
JasonGross, line 4933: Can we make an inductive enum for the names?  Then add `show` and `parse` instances, so that we can have the CLI parse just the valid ones?

src/Assembly/Symbolic.v
```diff
@@ -2868,46 +4619,397 @@ Qed.
...
+
+Lemma property_fold_left {X} (P : X -> Prop) fs :
+  Forall (fun f => (forall e : X, P e -> P (f e))) fs ->
+  forall e : X,
+  P e -> P (fold_left (fun e' f => f e') fs e).
+Proof.
+  intros H. induction fs as [|f fs' IHfs'].
+  - simpl. intros. assumption.
+  - simpl. intros. inv H. apply IHfs'; auto.
+Qed.
```
JasonGross, line 4982: Belongs in ListUtil or some file in ListUtil/

src/Assembly/Symbolic.v
```diff
@@ -2998,15 +5100,17 @@ Global Instance show_flag_state : Show flag_state :=
...
-                                             then match dag.get_eager_description_description descr with
+    ++List.map (fun '(i, v, descr, bounds) =>"(*"++show i ++"*) " ++ show v++";"
+                                         ++ "(((" ++ show bounds ++ ")))" ++ 
+                                            ((* if dag.get_eager_description_always_show descr
+                                             then *) match dag.get_eager_description_description descr with
                                                   | Some descr => " " ++ String.Tab ++ "(*" ++ descr ++ "*)"
                                                   | None => ""
                                                   end
-                                             else ""))%string (dag.eager.force d)
+                                             (* else "" *)))%string (dag.eager.force d)
```
JasonGross, line 5113: Why?

src/CLI.v
```diff
@@ -613,6 +666,10 @@ Module ForExtraction.
   (* We follow the standard convention of giving precedence to the final option passed *)
   Definition choose_one_of_many {A} (args : list A) : option A
     := List.nth_error (List.rev args) 0.
+  Compute (Arg.keyed_spec_list_data [asm_extra_rewrite_rules_spec; asm_extra_rewrite_rules_spec]).
+  Compute (Arg.keyed_spec_list_data common_optional_options).
+  Print unit.
+
```
JasonGross, line 672:
```suggestion
```

src/CLI.v
```diff
@@ -1045,6 +1104,71 @@ Module ForExtraction.
...
+      match limbwidth with
+      | ActualNat limbwidth' => 
+        let suggestion := Z.to_nat (Qceiling ((e + 1) / limbwidth')) in
+        (suggestion, limbwidth')
+      | AutoNat =>
+        let n' := Z.to_nat (Qceiling ((e + 1) / machine_wordsize)) in
+        let limbwidth' := Z.to_nat (Qceiling((e + 1) / n')) in
+        (n', limbwidth')
+      end
+    end.
```
JasonGross, line 1131: Please either merge this logic with the logic in `UnsaturatedSolinasHeuristics.v`, or move it to `DettmanMultiplicationHeuristics.v`

src/CLI.v
```diff
@@ -332,6 +364,20 @@ Module ForExtraction.
     := ("n",
         Arg.Custom (parse_string_and parse_n) "an ℕ or the literal '(auto)' or '(autoN)' for a non-negative number N",
         ["The number of limbs, or the literal '(auto)' or '(autoN)' for a non-negative number N, to automatically guess the number of limbs"]).
+  Definition n_maybenat_spec : anon_argT
+    := ("n",
+        Arg.Custom (parse_string_and parse_maybenat) "ℕ",
+        ["The number of limbs, or the literal '(auto)' or '(autoN)' for a non-negative number N, to automatically guess the number of limbs"]).
```
JasonGross, line 370: You don't seem to actually support things like `(auto5)`, though?  This syntax is for choosing from amongst multiple possible suggested choices, when there are multiple ones that might make sense.  Is this a sensible thing for Dettman multiplication?  (How do the heuristics work?)  If so, you should reuse what's in UnsaturatedSolinasHeuristics.  Otherwise you should drop the `(autoN)`.

src/CLI.v
```diff
@@ -1112,4 +1236,4 @@ Module ForExtraction.
       : A
       := Parameterized.PipelineMain argv.
   End BaseConversion.
-End ForExtraction.
+End ForExtraction.
```
JasonGross, line 1239:
```suggestion
End ForExtraction.
```

src/ExtractionHaskell/bedrock2_dettman_multiplication.v
```diff
@@ -0,0 +1,5 @@
+Require Import Crypto.Bedrock.Standalone.StandaloneHaskellMain.
+Import Bedrock2First.
+
+(*Redirect "/tmp/bedrock2_dettman_multiplication.hs"*) Recursive Extraction DettmanMultiplication.main.
+(* cat /tmp/bedrock2_dettman_multiplication.hs.out | sed -f haskell.sed  > ../../bedrock2_dettman_multiplication.hs *)
```
JasonGross, line 5: Please fix files missing a newline at the end of the file.  It's possible that `git rebase master --whitespace=fix` will do this.

src/CLI.v
```diff
@@ -425,6 +471,10 @@ Module ForExtraction.
     := ([Arg.long_key "asm-reg-rtl"],
         Arg.Unit,
         ["By default, registers are assumed to be assigned to function arguments from left to right in the hints file.  This flag reverses that convention to be right-to-left.  Note that this flag interacts with --asm-input-first, which determines whether the output pointers are to the left or to the right of the input arguments."]).
+  Definition asm_extra_rewrite_rules_spec : named_argT
+    := ([Arg.long_key "extra-rewrite-rule"],
```
JasonGross, line 475: Please make it clear this is specific to the equivalence checker, e.g.,
```suggestion
    := ([Arg.long_key "asm-extra-rewrite-rule"],
```
or
```suggestion
    := ([Arg.long_key "asm-extra-rewrite-rules"],
```

andres-erbsen:
IIRC Owen put in quite a bit of effort to figure out a confluent rewrite system that works for all of our examples and concluded that there isn't anything reasonably easy and elegant. There exist generic techniques for dealing with this by transforming rewrite systems, but having a concrete and human-readable rewrite system is a huge benefit for troubleshooting. Unless you have a concrete suggestion for how to fix the issue, I'd suggest either merging the checker changes as-is or not merging them at all, depending on how much we value the optimization the optional rules verify. (we may still write about it in a paper, though).

I agree that Dettman multiplication itself is a great addition and should be fast-tracked if we take time figuring out the checker stuff, but I kinda wish we didn't.

JasonGross:
> Unless you have a concrete suggestion for how to fix the issue

I don't, though if there's a way to include the insight as a comment, I think that'd be great.

> I'd suggest either merging the checker changes as-is or not merging them at all, depending on how much we value the optimization the optional rules verify.

I think merging them is better than not, and having them be conditional is fine.  By "as is" do you mean "literally as is" or "without blocking on figuring out a better strategy"?  Most of my requests are things like "don't use autogenerated names", "don't leave in print statements", "move utility code to Util", "use informative names", and some slightly broader suggestions like "use zrange instead of `Z * Z`" and "see if there's code-deduplication opportunities in this particular way around bounds computation".

andres-erbsen:
I meant merging a change based on the current strategy; I have not looked at the detailed feedback.

OwenConoly:
I'm sorry you had to look through the print statements and such things; I thought that I had already made a pass through to get rid of that sort of garbage, but in any case I should've checked again before making a pull request.

And yeah, like Andres said, I tried to come up with a way to avoid the conditional rewriting thing, but I couldn't find a good way to do it.  The way to fix it, I think, would've been to make the combine_consts function be much more general than it currently is.  Rewriting ors into adds was causing the adds to get sucked up into the combine_consts function, and they got combined in ways we don't want them to.

One idea I had to fix that was be to try some sort of neat canonicalization strategy like I did with the addcarries; but that sounded pretty complicated, so I didn't think about it for very long.  The other idea I had was the simplest possible canonicalization of expressions involving multiplication and addition: just distribute multiplication across addition.  I experimented with this, but it led to absurd runtime increases for the assembly checker, as I suppose I should have expected.  So, I gave up and did the conditional rewriting thing.

JasonGross:
> Rewriting ors into adds was causing the adds to get sucked up into the combine_consts function, and they got combined in ways we don't want them to.

Two thoughts/questions:
1. What was going wrong with how they were combined?
2. Very questionable idea: would it work to put this pass after combine_consts, so that combine_consts can't slurp up the adds?  This would make it more fragile, but we already don't have a great story for confluence / ordering.

OwenConoly:
The combine_consts function knows how to do this:
x + x*4     ===>     x*5.
But it isn't smart enough to do this:
x + y + (x + y)*4    ===>   (x + y)*5.

This means that if we don't activate the or-to-add rewriting thing, then combine_consts will do this:
(x | y) + (x | y)*4     ===>     (x | y)*5.
But if we activate the or-to-add rewriting thing, then those ors become adds, and then combine_consts doesn't know to do this:
x + y + (x + y)*4    ===>   (x + y)*5.

Then it seems like we have an easy fix: just augment combine_consts so it knows how to do that.  But then it's not obvious what combine_consts should do with an expression like
x + y + z + (x + y)*4 + (x + z)*4.

Should that become
z + (x + y)*5 + (x + z)*4,
or should it become 
y + (x + y)*4 + (x + z)*5?

And then things could potentially go badly if we make inconsistent choices for different expressions.  At this point I got the idea, why not just distribute everything, and then combine terms?  But that led to the runtime issues that I mentioned.

I had the "questionable idea" as well, and I tried it out.  It doesn't work :(

Here's an example showing that it doesn't work.
If we merge (x | y) into the dag (and label it with the number a), then it gets rewritten as 
(*a*) x + y.
Then suppose we merge (x | y) *4, and it gets rewritten as 
(*b*) (x + y)*4.
Then suppose we merge (a + b) into the dag.  It doesn't matter in which order we apply the or-to-add and combine_consts rules, since the operands a and b have already had both rules applied to them.  So, when we merge (a + b), we get
(*c*) x + y + (x + y)*4,
instead of the desired result.

OwenConoly:
> please rebase, perhaps `git fetch -a -v && git pull --rebase origin master`

Would 'master' in this context refer to the master branch of my fork?  I think I want it to be the master branch of the mit-plv repository, instead of my forked repository.  Based on a couple google searches, this is my best guess for what to do:

`git remote add original_repo https://github.com/mit-plv/fiat-crypto.git`
`git fetch -a -v original_repo`
`git pull --rebase origin upstream/master`

Would that do what we want it to do?

src/Assembly/Equivalence.v
```diff
@@ -673,7 +673,7 @@ Global Instance show_EquivalenceCheckingError : Show EquivalenceCheckingError
 Definition merge_fresh_symbol {descr:description} : dag.M idx := fun d => merge_node (dag.gensym 64%N d) d.
 Definition merge_literal {descr:description} (l:Z) : dag.M idx := merge_node ((const l, nil)).
 
-Definition SetRegFresh {descr:description} (r : REG) : M idx :=
+Definition SetRegFresh {descr:description} {errules : extra_rewrite_rules} (r : REG) : M idx :=
```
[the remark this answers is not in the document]
OwenConoly: It's an implicit argument for SetReg, just like descr

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -0,0 +1,212 @@
...
+Require Import Crypto.Util.ZUtil.Modulo.PullPush.
+Local Open Scope list_scope.
+
+Import Associational Positional.
+Import ListNotations. Local Open Scope Z_scope.
+
+Local Coercion Z.of_nat : nat >-> Z.
+Local Coercion Z.pos : positive >-> Z.
+
+Section Stuff.
```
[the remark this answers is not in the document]
OwenConoly: Do you know why the single underscore results in a syntax error?

src/Arithmetic/DettmanMultiplication.v
OwenConoly: That worked very nicely.  I cleaned up some other things in this file, too.

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -0,0 +1,212 @@
...
+      -- lia.
+      -- lia.
+  - apply div_nz.
+    + lia.
+    + rewrite <- Z.pow_mul_r. rewrite <- Nat2Z.inj_mul. rewrite <- Z.pow_le_mono_r_iff.
+      -- remember e_big as H. lia.
+      -- lia.
+      -- lia.
+      -- lia.
+      -- lia.
```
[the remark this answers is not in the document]
OwenConoly: Nope!  In fact, there was a lot of unnecessary stuff here.  It's a lot nicer now.

src/Assembly/Equivalence.v
```diff
@@ -689,7 +689,7 @@ Bind Scope symex_scope with symexM.
 Notation "'slet' x .. y <- X ; B"  := (symex_bind X (fun x => .. (fun y => B%symex) .. )) : symex_scope.
 Notation "A <- X ; B" := (symex_bind X (fun A => B%symex)) : symex_scope.
 (* light alias *)
-Definition App {descr:description} (e : Symbolic.node idx) : symexM idx := fun st => Success (merge (simplify st e) st).
+Definition App {descr:description} {errules : extra_rewrite_rules} (e : Symbolic.node idx) : symexM idx := fun st => Success (merge (simplify st e) st).
```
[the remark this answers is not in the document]
OwenConoly: It's an implicit parameter for simplify

src/Assembly/Equivalence.v
```diff
@@ -1269,7 +1269,10 @@ Definition LoadOutputs {descr:description} {dereference_scalar:bool} (outputaddr
            else
              Success (Error (Internal_error_LoadOutputs_length_mismatch outputaddrs output_types), s).
 
+Check SymexLines.
```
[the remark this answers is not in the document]
OwenConoly: Oops, sorry about that.

andres-erbsen:
yeah, that seems reasonable,  though I think you want to say `upstream` instead of `original_repo`. If you get merge conflicts, you may want to first squash some commits using `git rebase -i --keep-base upstream/master` so that you have to resolve conflicts fewer times, but please don't squash implementation of dettman multiplication with the checker changes.

JasonGross:
> At this point I got the idea, why not just distribute everything, and then combine terms? But that led to the runtime issues that I mentioned.

Did you distribute constants always, or only when the operand of the inner expression matched the operand of the outer expression?  That is, what if we perform the following distribution transformations:
1. If we have `x * c1 + y * c2` and `g := gcd(c1, c2)` is not the multiplicative identity, then we turn this into `(x * (c1/g) + y * (c2 / g)) * g`, dropping multiplications by 1 and conventionally factoring out a minus sign iff every constant is negative.
2. If, after doing (1), we still have `x + (y + z) * c`, we turn that into `x + y * c + z * c`.

This should never require more than 4 dag layers (which may already be too many?  See also my experiments with trying to make dag lookups more efficient), should maintain the property that expressions have a canonical form, and should not require duplicative work of distributing and then combining (recall that combining is quadratic in the number of addends), though we may have to run this pair of passes both before and after combining expressions.

Alternatively / additionally, the performance of combining can be improved from quadratic to n log n, I believe, by inserting the expressions into a trie map.  There's actually a trie map for expressions lying around in the New module in this file, which I coded up when trying to optimize the dag representation in #1342.  Unfortunately using them for the whole dag made everything much slower when running the equivalence checker, for reasons I haven't found time to debug (help very much welcome here), so if you decide to try this (definitely not a prerequisite for this PR), make sure to do it in its own commit and profile the result.

JasonGross:
If you're interested in some help splitting the PR and rebasing and stuff, I've made https://github.com/mit-plv/fiat-crypto/tree/dettman-mul-only-arith and https://github.com/mit-plv/fiat-crypto/tree/dettman-mul-partial-split; the former contains only the changes to the Arithmetic directory, should be pretty easy to rebase on master with `git rebase upstream/master --whitespace=fix`, and should probably be ready for merge as soon as you rename the `stuff` section and make sure there aren't autogenerated names.  The latter change is rebased on top of this and still needs more work.  I think it's probably worth splitting that branch into one that adds dettman mul to the CLI, and a separate one for the equiv checker.  Let me know if you want any further help with the splitting or rebasing.

src/Assembly/Symbolic.v
```diff
@@ -1406,62 +3252,52 @@ Lemma eval_merge {descr:description} G :
...
-    set A as y;
-    let m' := eval cbv beta in (B y) in
-    change m' in (value of m)
-  end.
-
-  inversion H1; clear H1 ; subst.
-
-  cbn [fst snd] in *.
-  assert (gensym_dag_ok G (snd idxs_d) /\
+  intros e. induction e as [i| [o args] IH] (* writing "eqn:E" here screws things up. why? *); intros.
```
JasonGross, line 3255: > writing "eqn:E" here screws things up. why?

Because you can't write an induction principle that gives you any equality between the current thing and the thing you started with.  e.g., if you do induction on `nat`s, you don't get that every intermediate `nat` is equal to the `nat` you started with.

src/Assembly/Symbolic.v
```diff
@@ -1352,51 +1812,1437 @@ Lemma gensym_ok_merge_node G d {descr:description} n
...
+
+Fixpoint withadc_without (d : dag) (s : OperationSize) (args : list idx) : (list (list (list idx))) * (list idx) :=
+  match args with
+  | [] => ([], [])
+  | i :: args' =>
+    let (withadc', without') := withadc_without d s args' in
+    match dag.lookup d i with
+    | Some (o, args'') =>
+      match o, args'' with
+      | addcarry s', (x :: y :: []) =>
```
JasonGross, line 2373: addcarry is not specific to two operands; why do you only consider two operands here?

JasonGross:
> 2\. a. A canonicalization of sums of addcarries, which effectively makes the equivalence checker aware that double-wide (e.g., 128-bit) addition is associative.  This is analogous to the preexisting canonicalization of commutative operations, which just sorted the operands.

Do we currently have any rewrites that talk about `addcarry` vs `shr`?  It seems plausible to me that we actually want to canonicalize all `addcarry` into `shr` of `addZ` (at least for powers of 2), add a rule that says that `(x >> a) + (y >> b)` is `(x + (y >> (b - a))) >> a` whenever `b >= a`, let the sorting happen by associativity of `addZ`, and make sure we have `x >> 0 = x`.  Any thoughts on this strategy vs what you're currently doing?

OwenConoly:
> Do we currently have any rewrites that talk about `addcarry` vs `shr`? It seems plausible to me that we actually want to canonicalize all `addcarry` into `shr` of `addZ` (at least for powers of 2), add a rule that says that `(x >> a) + (y >> b)` is `(x + (y >> (b - a))) >> a` whenever `b >= a`, let the sorting happen by associativity of `addZ`, and make sure we have `x >> 0 = x`. Any thoughts on this strategy vs what you're currently doing?

I like that strategy much better!  It's more general, and easier to work with as well.  What I'm currently doing is essentially combining things into the addZ and shr, and then splitting them apart again into addcarries.  The splitting apart is ugly and difficult to implement.  It would be much nicer to just write it as the shr of an addZ.

I'm not sure what you mean by "at least for powers of 2".  Isn't a power of two always involved?

OwenConoly:
> Did you distribute constants always, or only when the operand of the inner expression matched the operand of the outer expression?

I just distributed indiscriminately.

src/Assembly/Symbolic.v
OwenConoly: I can't think of a nice way to generalize the thing that I'm doing to more than two operands.  I can't even think of a way to generalize the more general strategy that you mentioned (with the addZ and shr) to more than two operands.

My rewriting strategy relies on the fact that `(addcarry s, [x; y])` is the same thing as `(shr s, [(addZ, [x; y])])`.  This is nice, but it's not true for more than two operands.  That is, it is not always the case that `(addcarry s, [x; y; z])` is the same thing as `(shr s, [(addZ, [x; y; z])])`.  For example, make x, y, and z as large as possible, and then the addcarry is equal to 0, while the shr of the addZ is equal to 2.

OwenConoly:
> If you're interested in some help splitting the PR and rebasing and stuff, I've made https://github.com/mit-plv/fiat-crypto/tree/dettman-mul-only-arith and https://github.com/mit-plv/fiat-crypto/tree/dettman-mul-partial-split;

Ok, I will start with those.

src/Assembly/Symbolic.v
JasonGross: Oh, huh, I'd forgotten that was how addcarry worked.  I think the general strategy is that if addZ of the operands can be bounded to <= 2^(s+1)-1, then you can replace addcarry with shr of addZ.  Furthermore, any number of shrs can be combined.

Actually, we can do better.  The implementation of addcarry is
https://github.com/mit-plv/fiat-crypto/blob/957dc226d222943cdb2028deca31981a4f9e7429/src/Assembly/Symbolic.v#L488-L449
so we can just do shr of addZ and then mask whenever the result cannot be shown to be bounded.

JasonGross:
> > Did you distribute constants always, or only when the operand of the inner expression matched the operand of the outer expression?
> 
> I just distributed indiscriminately

It seems at least plausible that more restricted distribution would be performant enough?

JasonGross:
> > If you're interested in some help splitting the PR and rebasing and stuff, I've made https://github.com/mit-plv/fiat-crypto/tree/dettman-mul-only-arith and https://github.com/mit-plv/fiat-crypto/tree/dettman-mul-partial-split;
> 
> Ok, I will start with those.

I'd suggest taking the dettman-mul-only-arith branch, renaming the section that I commented on in this PR (I think it's called "stuff"?), rebasing it on master, and preparing a PR from that branch (or equivalent), with only the arithmetic template; Andres can review that and hopefully it can be merged quickly. 

Then I'd suggest pulling out the dettman binary/CLI into its own branch on top of that (maybe just copy the files and do it in one commit?).  Note that I drastically reduced the number of typeclass arguments about a month ago, and the Makefile changes now go in Makefile.examples and Makefile.config, so there's some merge conflicts to fix.  That PR should also be quickly mergable.

I'd also prefer to have the four changes to the equivalence checker separated out.  I've reimplemented the ability to have conditional passes at #1498 (note that all passes now need to be listed in an enum and have descriptions right below their definition), and I've moved the bounds definitions up in #1499, which should make it much easier to write independent passes in a way that minimizes merge conflicts.  (I've also adjusted bounds checking to give back a zrange in #1485, and integrated some of the other smaller changes in #1486, #1487, #1490, #1492, #1494.)

I think the pass that rewrites or to add should be quickly mergable, if you separate it out (is the PHOAS-based bounds analysis powerful enough for this, or did I miss some refinement?).

I expect that the "combine shr" pass should be mostly self-contained and easily mergable.

The "turn addcarry into shr" pass is probably going to impact a bunch of the other addcarry passes, and might even significantly simplify some of them?

Finally, the "memoize bounds into the dag" change will need some work on top of what's already in this PR, in particular to respect the abstraction barriers of the dag module.  I started looking into this a bit, but an not sure how much more time I'll find for it.  In any case I think you should save this PR for last.

What do you think of this plan for splitting up the PRs?

src/Assembly/Symbolic.v
JasonGross: (plausibly we could just always mask and let a subsequent pass remove the masking when the bounds allow us to)

JasonGross:
I guess we maybe also want another (preliminary?) PR that gives all rewriting passes access to the dag, and gives their Ok proofs access to dag_ok or w/e.

And it would be nice to have separate PRs (or at least commits) for adding bounds info to the dag vs making use of that bounds info and changing the inspection depth, so that we can profile these separately.

OwenConoly:
> What do you think of this plan for splitting up the PRs?

It sounds good!  I am unable to push changes to the dettman-mul-only-arith branch though.  I get a "permission denied" error.  Can you give me permission to do that?

OwenConoly:
> It seems at least plausible that more restricted distribution would be performant enough?

Yeah, very plausible.

JasonGross:
> > What do you think of this plan for splitting up the PRs?
> 
> It sounds good! I am unable to push changes to the dettman-mul-only-arith branch though. I get a "permission denied" error. Can you give me permission to do that?

Done!  You should have an invite on GitHub

JasonGross:
> I guess we maybe also want another (preliminary?) PR that gives all rewriting passes access to the dag, and gives their Ok proofs access to dag_ok or w/e.
> 
> And it would be nice to have separate PRs (or at least commits) for adding bounds info to the dag vs making use of that bounds info and changing the inspection depth, so that we can profile these separately.

@OwenConoly I've done these things in #1503 and #1505.  You should base any changes to the equivalence checker on top of these PRs, so that they don't conflict with them.  I'm optimistic that the remaining planned changes to equivalence checker should not introduce more conflicts with each other.

Rayyy369:
I'm sorry guys.

Rayyy369:
I didn't mean to frustrate everyone. I frustrate myself for not being on the same page as y'all. Just know I do appreciate you guys efforts.

Rayyy369:
Everytime I attempted to verify your corrections i get hit wit a 404

JasonGross:
@OwenConoly I've prepared a series of PRs (#1508, #1509, #1510, #1511) which implements the dynamic programming bounds checking algorithm.  It doesn't change the revealed depth (insofar as that's possible, could you prepare a PR that does just that?  And then one of us can get a timing profile on transparent-rooster).

I expect to not do much more on the equivalence checker anytime soon (sorry for all the conflicts with your work), and will leave you to implement add-to-or, the combine shr, and the turn-addcarry-into-shr passes. (Let me know if you have any questions about the existing passes / what they do / how to adapt them to handle shr instead of addcarry.)

OwenConoly:
> It seems plausible to me that we actually want to canonicalize all `addcarry` into `shr` of `addZ` (at least for powers of 2), add a rule that says that `(x >> a) + (y >> b)` is `(x + (y >> (b - a))) >> a` whenever `b >= a`, let the sorting happen by associativity of `addZ`, and make sure we have `x >> 0 = x`.

I think I'll just implement this rule for the case that `b = a`.

I thought about doing it for the case that `b > a`, but I didn't really like the idea of taking a simple sum like 
`(x >> a) + (y >> b) + (z >> c)`
and turning it into something arbitrarily deeply nested like
`(((x >> (a - b)) + y) >> (b - c)) + z) >> c`.
Then we would have to 'dig up'  arbitrarily many layers to ensure that we keep the canonical representation when a new addend is added to the sum.  This sounds a bit complicated--and it looks like it could take time that's quadratic in the number of addends (I'm not sure how much that matters).

OwenConoly:
Oops, there's a bigger problem here.  This equality that I'm trying to use isn't true.  For example,
`(7 >> 3) + (7 >> 3) = 0`,
while
`(7 + 7) >> 3 = 1`

I'll have to try a bit harder to remember how my addcarry rewriting thing actually worked.

OwenConoly:
Ah, I see.  We want a rewrite rule that turns
`((x + y) >> a) + ((y + z) >> a)`
into
`(x + y + z) >> a`,
where the pluses are `addZ`.

OwenConoly:
But then I'm not sure that I really like this approach, because it's not clear what the right thing is to do with this expression:
`((x + y) >> a) + ((((x + y) mod 2^a) + z1) >> a) + ((((x + y) mod 2^a) + z2) >> a)`.

Should that be rewritten as
`((x + y + z1) >> a) + ((((x + y) mod 2^a) + z2) >> a)`,
or as
`((x + y + z2) >> a) + ((((x + y) mod 2^a) + z1) >> a)`?

JasonGross:
What about rewriting `(x >> a) + (y >> a)` into `((x + y - (x mod 2^a) - (y mod 2^a)) >> a` and then doing some sort of cancellation on this expression.  The invariant here is that the normal form of `((x - (x mod 2^a)) >> a` should be the same as the normal form of `x >> a`, and if you can accomplish that, then you can freely combine right shifts (as long as the shift amount is the same).

JasonGross:
(The other rule you'd want here is that `(x + (y mod b)) mod b` becomes `(x mod b) + (y mod b)`, and then I think the rest follows from rules like `x - x = 0`, which we should already have.)

OwenConoly:
Ok, neat!  I'll do that, and I'll rewrite `(x + (y mod b)) mod b` into `(x + y) mod b`.

andres-erbsen:
@JasonGross alerted me that this change is also needed for the artifact. @OwenConoly would you be willing to do a quick call with me today to see if I can help get this merged?

andres-erbsen:
>  a. A canonicalization of sums of addcarries, which effectively makes the equivalence checker aware that double-wide (e.g., 128-bit) addition is associative. This is analogous to the preexisting canonicalization of commutative operations, which just sorted the operands.

I cannot figure out what part of the PR does this.

> b. A (new, in-some-ways better) implementation of bounds on integer values corresponding to nodes in the dag.

This seems to be https://github.com/mit-plv/fiat-crypto/pull/1511

> c. A new rewriting rule, which replaces or operations with add operations in some cases.

https://github.com/mit-plv/fiat-crypto/pull/1481/files#diff-b7349927aba2f511b65b58306adb1b775a3d0aa3944d1baab4ac274639fd28efR4911-R4926; this doesn't seem to be merged yet but probably wouldn't be to difficult. 

> d. A new cli optional argument, "--extra-rewrite-rule", which allows optional rewrite rules (which are off by default) to be "turned on" when the equivalence checker is invoked. 

Is this https://github.com/mit-plv/fiat-crypto/pull/1498 ?

Is there anything else we still need?

JasonGross:
> > canonicalization of sums of addcarries, which effectively makes the equivalence checker aware that double-wide (e.g., 128-bit) addition is associative. This is analogous to the preexisting canonicalization of commutative operations, which just sorted the operands.
> 
> I cannot figure out what part of the PR does this.

I think it's standardize_adc_sum



> > b. A (new, in-some-ways better) implementation of bounds on integer values corresponding to nodes in the dag.
> 
> This seems to be #1511

I think that's right

> > d. A new cli optional argument, "--extra-rewrite-rule", which allows optional rewrite rules (which are off by default) to be "turned on" when the equivalence checker is invoked.
> 
> Is this #1498 ?

Yes

OwenConoly:
> @JasonGross alerted me that this change is also needed for the artifact. @OwenConoly would you be willing to do a quick call with me today to see if I can help get this merged?

I am just seeing this now.  Sorry about that; I thought GitHub was supposed to email me when I was tagged, but perhaps the email got lost.

I assume this is no longer relevant

PR #1484, opened 2022-11-08
Bump bedrock2

andres-erbsen:


JasonGross:
Did you mean to enable auto merge here?

andres-erbsen:
I didn't think about it, but yes, it seems like a good idea. Thanks!

PR #1485, opened 2022-11-09
Reuse PHOAS bounds analysis for asm bounds

JasonGross:
This should allow more fine-grained bounds and easier extensibility.  This might be duplicative with some work in #1481, but I think this strategy will allow for better code reuse going forward.

Thoughts @andres-erbsen @OwenConoly ?

### DIFF
--- a/src/Assembly/Symbolic.v
+++ b/src/Assembly/Symbolic.v
@@ -16,10 +16,13 @@ Require Crypto.Util.Tuple.
 Require Import Util.OptionList.
 Require Import Crypto.Util.ErrorT.
 Require Import Crypto.Util.ZUtil.Tactics.PullPush.Modulo.
+Require Import Crypto.Util.ZUtil.Tactics.LtbToLt.
 Require Import Crypto.Util.ZUtil.Testbit.
 Require Import Crypto.Util.ZUtil.Hints.ZArith.
 Require Import Crypto.Util.ZUtil.Land.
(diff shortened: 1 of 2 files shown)


src/Assembly/Symbolic.v
```diff
@@ -466,24 +489,23 @@ Section WithContext.
...
     | iszero, [a] => Some (Z.b2z (Z.eqb a 0))
-    | addZ, args => Some (List.fold_right Z.add 0 args)
-    | mulZ, args => Some (List.fold_right Z.mul 1 args)
     | negZ, [a] => Some (Z.opp a)
     | shlZ, [a; b] => Some (Z.shiftl a b)
     | shrZ, [a; b] => Some (Z.shiftr a b)
-    | andZ, args => Some (List.fold_right Z.land (-1) args)
-    | orZ, args => Some (List.fold_right Z.lor 0 args)
-    | xorZ, args => Some (List.fold_right Z.lxor 0 args)
     | addcarryZ s, args => Some (Z.shiftr (List.fold_right Z.add 0 args) (Z.of_N s))
```
JasonGross, line 509: @andres-erbsen Thoughts on this factoring of `interp_op`?

andres-erbsen: I remember thinking that some factoring would be desirable; no opinion on this particular one.

JasonGross:
Going to merge based on conversations with Andres OOB.  Happy to help with any trouble rebasing on top of this.

PR #1500, opened 2022-11-13
Dettman Multiplication Arithmetic

OwenConoly:
This is a part of #1481.  It just adds the Dettman multiplication algorithm to the Arithmetic directory.


JasonGross (approved):

src/Arithmetic/Core.v
```diff
@@ -486,6 +526,59 @@ Module Associational.
...
+  Proof.
+    intros H. induction a as [| x a' IHa'].
+    - reflexivity.
+    - simpl. destruct (d =? fst x) eqn:E.
+      + exfalso. apply H. simpl. left. apply Z.eqb_eq in E. symmetry. apply E.
+      + apply IHa'. intros H'. apply H. simpl. right. apply H'.
+  Qed.
+
+  Definition dedup_weights a :=
+    map (fun d => (d, value_at_weight a d)) (just_once Z.eqb (firsts a)).
```
andres-erbsen, line 554: is `just_once` the same as standard-library `nodup`?

src/Arithmetic/Core.v
```diff
@@ -486,6 +526,59 @@ Module Associational.
...
+  Proof.
+    induction a as [| a0 a' IHa'].
+    - cbv [Associational.eval]. simpl. lia.
+    - simpl. destruct (d =? fst a0) eqn:E.
+      + rewrite Associational.eval_cons. rewrite <- IHa'. apply Z.eqb_eq in E. rewrite E. lia.
+      + apply IHa'.
+  Qed.
+
+  Definition firsts (a : list (Z*Z)) :=
+    map (fun p => fst p) a.
```
andres-erbsen, line 542: Inline as `map fst`?

src/Arithmetic/Core.v
```diff
@@ -486,6 +526,59 @@ Module Associational.
       push; [|rewrite IHp]; reflexivity.
   Qed.
 
+  Definition value_at_weight (a : list (Z * Z)) (d : Z) :=
+    fold_right (fun p sum => if (d =? fst p) then (sum + snd p) else sum) 0 a.
```
andres-erbsen, line 530: `fold_right Z.add (filter (fun p => fst p =? d))`? In other words, and impmentation closer to the next lemma might just be nicer.

src/Arithmetic/Core.v
```diff
@@ -511,6 +604,36 @@ Module Associational.
       eval (carry w fw p) = eval p.
     Proof using Type*. cbv [carry]; induction p; push; nsatz. Qed.
     Hint Rewrite eval_carry using auto : push_eval.
+
+  (* 'carrying down', aka borrowing *)
+  Definition carryterm_down (w fw:Z) (t:Z * Z) :=
```
andres-erbsen, line 609: `borrowterm`?

src/Arithmetic/Core.v
```diff
@@ -511,6 +604,36 @@ Module Associational.
...
+  Lemma eval_carryterm_down w fw (t:Z * Z) (fw_nz:fw<>0) (w_fw:w mod fw = 0) :
+        Associational.eval (carryterm_down w fw t) = Associational.eval [t].
+  Proof using Type*.
+    cbv [carryterm_down Let_In]; break_match; push; [|trivial].
+    pose proof (Z.div_mod (snd t) fw fw_nz).
+    rewrite Z.eqb_eq in *.
+    ring_simplify. rewrite Z.mul_comm. rewrite Z.mul_assoc. rewrite <- Z_div_exact_full_2; lia.
+  Qed.
+
+  Definition carry_down (w fw:Z) (p:list (Z*Z)) :=
```
andres-erbsen, line 624: `borrow`?

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -0,0 +1,161 @@
+Require Import Crypto.Arithmetic.Core.
+Require Import Coq.ZArith.ZArith Coq.micromega.Lia.
+Require Import Coq.Lists.List.
+Require Import Crypto.Util.ZUtil.Modulo.PullPush.
+Local Open Scope list_scope.
+
+Import Associational Positional.
+Import ListNotations. Local Open Scope Z_scope.
+
+Local Coercion Z.of_nat : nat >-> Z.
```
andres-erbsen, line 10: Using non-injective functions as coercions can make goals difficult to read. Do you actually want it, and might a notation do the job instead?

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -0,0 +1,161 @@
...
+
+Let base := (2 ^ limb_size).
+
+Lemma base_nz : base <> 0.
+Proof. cbv [base]. apply Z.pow_nonzero; lia. Qed.
+
+Lemma s_nz : s <> 0.
+Proof. cbv [s]. apply Z.pow_nonzero; lia. Qed.
+
+Let weight (n : nat) := base ^ n.
```
andres-erbsen, line 37: https://github.com/mit-plv/fiat-crypto/blob/0cc0199d0d12c7c34bd01ee8d72a8fc35495ac03/src/Arithmetic/UniformWeight.v looks like it might have some of the following theory already

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -0,0 +1,161 @@
...
+  - apply weight_nz.
+  - cbv [weight]. apply mod_is_zero.
+    + apply base_nz.
+    + lia.
+  - cbv [weight]. apply limbs_mod_s_0.
+  - apply p_nz.
+  - apply weight_nz.
+Qed.
+
+Definition from_n_to_one n :=
```
andres-erbsen, line 97: `rev` of `seq`?

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -0,0 +1,161 @@
...
+
+Lemma eval_loop' start :
+  ((Associational.eval (loop' start)) mod (s - c) = (Associational.eval start) mod (s - c))%Z.
+Proof.
+  cbv [loop']. induction (from_n_to_one (limbs - 2 - 1)) as [| i l' IHl'].
+  - reflexivity.
+  - simpl. rewrite eval_loop_body. apply IHl'.
+Qed.
+
+Definition mulmod_general a b :=
```
andres-erbsen, line 111: maybe drop `_general`?

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -0,0 +1,161 @@
...
+Proof.
+  cbv [weight base s]. rewrite <- Z.pow_mul_r by lia. rewrite <- Nat2Z.inj_mul. apply mod_is_zero; lia.
+Qed.
+
+Local Open Scope nat_scope.
+
+Definition reduce' x1 x2 x3 x4 x5 := dedup_weights (reduce_one x1 x2 x3 x4 x5).
+Definition carry' x1 x2 x3 := dedup_weights (Associational.carry x1 x2 x3).
+
+Definition loop_body i before :=
```
andres-erbsen, line 73: If a descriptive name for what this function does (rather than how it is used) came handy, it'd be nice to use that instead.

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -0,0 +1,161 @@
...
+    + lia.
+  - cbv [weight]. apply limbs_mod_s_0.
+  - apply p_nz.
+  - apply weight_nz.
+Qed.
+
+Definition from_n_to_one n :=
+  fold_right (fun (x :nat) l' => (length l' + 1) :: l') [] (repeat 0 n).
+
+Definition loop' start :=
```
andres-erbsen, line 100: maybe inline this function?

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -0,0 +1,161 @@
...
+  fold_right loop_body start (from_n_to_one (limbs - 2 - 1)).
+
+Lemma eval_loop' start :
+  ((Associational.eval (loop' start)) mod (s - c) = (Associational.eval start) mod (s - c))%Z.
```
andres-erbsen, line 104: it seems that the specific list being folded over does not matter here; if you end up inlining `loop'` you could generalize over the list.

andres-erbsen (approved):
:tada: 

All the comments are truly optional, of the kind that I'd purse when trying to make my code look nice but not worry about if they got merged.

src/Arithmetic/Core.v
JasonGross: I think this is about being reifiable, cf https://github.com/mit-plv/fiat-crypto/pull/1490

src/Arithmetic/Core.v
JasonGross: Does `map fst` work, or do you need `map (@fst _ _)`?

src/Arithmetic/Core.v
JasonGross: Though plausibly we want to add reifications of the corresponding definitions and rewrite rules to unfold them?  *shrugs*

src/Arithmetic/DettmanMultiplication.v
OwenConoly: I removed the `pos >-> Z` coercion, since apparently it wasn't being used for anything.  The `nat >-> Z` coercion is useful, though.  I'm not sure how I would do this as a notation.

Isn't Z.of_nat an injective function?

src/Arithmetic/Core.v
OwenConoly: Yeah, I think it does the same thing.  But I remember that trying to use the standard library functions led to reification errors.

src/Arithmetic/Core.v
OwenConoly: `map fst` works!

JasonGross: I wonder if this is because we're somehow transitively `Require`ing `Coq.Classes.RelationPairs` or `Coq.Program.Syntax`?  (That seems to be the only place where the `Arguments` to `fst` are globally(?) set)

src/Arithmetic/Core.v
OwenConoly: Yeah, that sounds nice

src/Arithmetic/Core.v
OwenConoly: That sounds good

src/Arithmetic/DettmanMultiplication.v
OwenConoly: I can't think of one really; it's a carry, then a reduce, then another carry

src/Arithmetic/DettmanMultiplication.v
OwenConoly: I kinda like it separate, since it's conceptually special--the only part of the algorithm that I could describe as a loop.  It's also nice to have its proof `eval_loop` separate, I think

JasonGross:
```
COQC src/Arithmetic/DettmanMultiplication.v
Error: The reference carry_down was not found in the current environment.
```

OwenConoly:
oops
I fixed that error, and I tested it on my machine before pushing this time

src/Arithmetic/Core.v
andres-erbsen: If this is just about reification, please name the function `nodup` and add a comment explaining why we have our own version.

OwenConoly: The other standard-library function that I sort-of duplicated is `in_dec`.  The name of the function I wrote is `is_in`.  My function returns a boolean, unlike the standard-library version.  It would be weird to name my function `in_dec`, given the return type.  But would `in` be a better name (more conventional) than `is_in`?

OwenConoly: Also, maybe I should mention that my `just_once` function doesn't function totally identically to `nodup`.  This is the type of `nodup`:

`forall A : Type, (forall x y : A, {x = y} + {x <> y}) -> list A -> list A`,

whereas Coq says that the type of `just_once` is

`(?X -> ?X -> bool) -> list ?X -> list ?X
where
?X : [ |- Type]`.

The relevant difference is that my function deals with booleans.  Should I rewrite `just_once` so it has the same type as `nodup`?  Or would that lead to reification issues?

OwenConoly: On a subject less related to the issue at hand, I was expecting Coq to tell me that the type of `just_once` is 
`forall X : Type, (X -> X -> bool) -> list X -> list X`,
but instead I got the printout that I wrote above.  Do you know why this happened?  Does it have something to do with the fact that I defined `just_once` within a `Context`?

andres-erbsen: If it's not an exact match, feel free to leave it as-is.

`Check` and `About` differ in whether they instantiate maximally-inserted implicit arguments or not.

JasonGross: > The other standard-library function that I sort-of duplicated is `in_dec`. The name of the function I wrote is `is_in`. My function returns a boolean, unlike the standard-library version. It would be weird to name my function `in_dec`, given the return type. But would `in` be a better name (more conventional) than `is_in`?

I think the closest standard library match for this is `List.existsb`, which is a bit more general.  Might be with making a variant of that instead.  Also might be worth changing `Section Reifiable ` into `Module Reifiable ` and then shadowing the relevant names?

JasonGross: > What do you think of renaming `just_once` into `nodupb`, to make clear the similarity?

Sounds good

src/Util/ListUtil.v
```diff
@@ -3421,86 +3421,84 @@ Lemma fold_left_rev_higher_order A B f a ls
...
+  Section __.
+    Context {X : Type}
+            (eqb : X -> X -> bool)
+            (eqb_eq : forall x1 x2, eqb x1 x2 = true <-> x1 = x2).
 
-Context {X : Type}
-        (eqb : X -> X -> bool)
-        (eqb_eq : forall x1 x2, eqb x1 x2 = true <-> x1 = x2).
+    Definition existsb (f : X -> bool) (l : list X) : bool :=
+      fold_right (fun x found => orb (f x)found) false l.
```
JasonGross, line 3433: Minor:
```suggestion
      fold_right (fun x found => orb (f x) found) false l.
```

src/Util/ListUtil.v
```diff
@@ -3421,86 +3421,84 @@ Lemma fold_left_rev_higher_order A B f a ls
...
-    + simpl in *. destruct (eqb x' x) eqn:E.
-      -- apply eqb_eq in E. left. apply E.
-      -- simpl in H. right. rewrite <- IHl'. apply H.
-    + destruct H as [H | H].
-      -- simpl. apply eqb_eq in H. rewrite H. reflexivity.
-      -- simpl. rewrite <- IHl' in H. rewrite H. destruct (eqb x' x); reflexivity.
-Qed.
+    Lemma reifiable_existsb_is_existsb : forall (f : X -> bool) (l : list X),
+      existsb f l = List.existsb f l.
+    Proof. reflexivity. Qed.
```
JasonGross, line 3437: How does this end up being provable by reflexivity?

src/Util/ListUtil.v
```diff
@@ -3421,86 +3421,84 @@ Lemma fold_left_rev_higher_order A B f a ls
...
 
-Lemma is_in_false_iff : forall x l, is_in x l = false <-> ~ In x l.
-Proof.
-  intros. rewrite <- is_in_true_iff. split.
-  - intros H. rewrite H. auto.
-  - intros H. destruct (is_in x l).
-    + exfalso. apply H. reflexivity.
-    + reflexivity.
-Qed.
+    Lemma existsb_eqb_true_iff : forall x l, existsb (eqb x) l = true <-> In x l.
```
JasonGross, line 3439:
```suggestion
    Lemma existsb_eqb_true_iff : forall x l, existsb (eqb x) l = true <-> In x l.
    Proof.
```

src/Util/ListUtil.v
OwenConoly: I wasn't expecting it to be!  I don't fully understand how Coq represents `fixpoint`s, but apparently it's very good at simplifying them.

JasonGross: I guess it's probably the case that `List.existsb` is set up to be `fix existsb (A : Type) (l : list A) : bool`, and `fold_right` probably simplifies to something similar.  You can check with `Print`.

JasonGross:
I've rebased.  Is this ready for merge?

OwenConoly:
Yes!

PR #1510, opened 2022-11-15
Adjust a lemma for use in dag bounds

JasonGross:
On top of #1508, #1509


JasonGross:
This previously passed on the CI, so I am merging

PR #1514, opened 2022-11-18
add classification of Curve25519 points

andres-erbsen:


andres-erbsen:
Succeeded at https://github.com/mit-plv/fiat-crypto/actions/runs/3501087460/jobs/5864370724#step:7:2033 and then failed due to https://github.com/coq/coq/issues/16854

PR #1532, opened 2022-12-02
Montgomery reduction without loop unrolling

nathan-sheffield:


andres-erbsen:
I think this code was written on a pre-`func!` checkout and needs to be updated to work with master. Snippet from CI logs:

```coq
...})))" has type
"(String.string * (list String.string * list ?A * cmd))%type"
while it is expected to have type "func" (cannot unify 
"String.string" and "(list String.string * list String.string)%type").
```

Issue #1512, opened 2022-11-15
Bedrock/End2End/RupicolaCrypto/Derive.v is too-memory heavy


JasonGross:
This is preventing us from being tested on Coq's CI, see https://github.com/coq/coq/pull/16801.  Last I checked, it takes about a minute to compile and about *4 GB* of RAM (4384020 ko, as of 70f832b7df754b4a41f286a969493a5a9438bdcf).  This is up from 20s and about 1GB of RAM (1169688 ko as of 42cace10dde6471b6422ca282c585894dd16978b)

@andres-erbsen @DIJamner can one of you take care of this?

andres-erbsen:
My guess would be https://github.com/mit-plv/fiat-crypto/pull/1482/files#diff-076269a2427c8322a666f8b6d8d554d51d53706a7785ff1d7873c6282e30355bR173 for the cause. As for what to do about it, idk.

JasonGross:
The timing and memory log is
```
.607830900s +397672kb: Chars 0 - 32 [Require~Import~Rupicola.Lib.Api.] 0.494 secs (0.231u,0.193s)
.016281200s +13060kb: Chars 33 - 68 [Require~Import~Rupicola.Lib.Ar...] 0.008 secs (0.008u,0.s)
.004501300s +2308kb: Chars 69 - 103 [Require~Import~Rupicola.Lib.Lo...] 0. secs (0.u,0.s)
.004524400s +1844kb: Chars 104 - 139 [Require~Import~Rupicola.Lib.Ge...] 0. secs (0.u,0.s)
.004777000s +2048kb: Chars 140 - 185 [Require~Import~coqutil.Word.Li...] 0. secs (0.u,0.s)
2.400876900s +649160kb: Chars 186 - 243 [Require~Import~Crypto.Bedrock....] 2.42 secs (2.272u,0.147s)
.005078100s +0kb: Chars 244 - 286 [Require~Import~bedrock2.BasicC...] 0. secs (0.u,0.s)
.005360700s +0kb: Chars 287 - 350 [Require~Import~Crypto.Bedrock....] 0. secs (0.u,0.s)
.005389200s +0kb: Chars 352 - 369 [Section~Bedrock2.] 0. secs (0.u,0.s)
.004914000s +0kb: Chars 373 - 682 [Instance~spec_of_quarter~:~(sp...] 0.005 secs (0.005u,0.s)
[...]

PR #1533, opened 2022-12-03
(Almost) finish the chacha20 derivation

DIJamner:
Fills in admits and fixes supporting lemmas to finish the chacha20 derivation. Note that the derivation is still aborted because Qed takes an unreasonable amount of time (it's unconfirmed whether it finishes).  This PR should hopefully help address #1512 as well.


src/Bedrock/End2End/RupicolaCrypto/Broadcast.v
```diff
@@ -427,6 +428,7 @@ Section with_parameters.
     (*Issue: in loop invariant lp from -> lp from'*)
     let x := open_constr:(fun from' lst tr mem locals =>
                             (lst$@a_ptr ⋆ R) mem
+                            /\ tr = t
```
andres-erbsen, line 431: :+1:

PR #1534, opened 2022-12-04
Try Newer Go

JasonGross:
As per @davidben


andres-erbsen:
Are you sure we want to be version-pinning Go at all? I know we use it both through BoringSSL directly, but I think neither use has a particularly good reason to prefer a fixed version to the latest.

JasonGross:
I'm not sure if the action works without a particular version of Go, let's see

PR #1541, opened 2022-12-10
Compatibility with coq/coq#16756

andres-erbsen:
untested


src/Bedrock/End2End/RupicolaCrypto/Broadcast.v
```diff
@@ -340,7 +340,7 @@ Section with_parameters.
     : map f (upd l i a) = upd (map f l) i (f a).
   Proof.
     eapply nth_error_ext;
-      intros.
+      intros i0.
```
JasonGross, line 343: Are you sure you don't mean
```suggestion
      intros i0 **.
```
or
```suggestion
      intros i0; intros.
```
or
```suggestion
      intros i0 ?.
```

PR #1550, opened 2022-12-22
CLI stuff for dettman multiplication function

OwenConoly:


src/PushButtonSynthesis/DettmanMultiplication.v
```diff
@@ -0,0 +1,226 @@
...
+Section __.
+  Context {output_language_api : ToString.OutputLanguageAPI}
+          {pipeline_opts : PipelineOptions}
+          {pipeline_to_string_opts : PipelineToStringOptions}
+          {synthesis_opts : SynthesisOptions}
+          (machine_wordsize : machine_wordsize_opt)
+          (s_ : Z)
+          (c_ : list (Z*Z))
+          (n : nat) (* number of limbs *)
+          (limb_size : nat)
```
JasonGross, line 71: What is limb size?  Why do we need it here but not in unsaturated solinas?

JasonGross (changes requested):
Please amend the commits to remove the binary files, and add them to .gitignore

(This is a partial review, I'll look at the rest later)

src/Arithmetic/Core.v
```diff
@@ -528,23 +528,23 @@ Module Associational.
...
-    - cbv [value_at_weight]. simpl. destruct (fst a0 =? d) eqn:E.
-      + rewrite Associational.eval_cons. simpl. rewrite <- IHa'. cbv [value_at_weight]. lia.
+    - cbv [Associational.eval]. simpl. lia.
+    - simpl. destruct (d =? fst a0) eqn:E.
+      + rewrite Associational.eval_cons. rewrite <- IHa'. apply Z.eqb_eq in E. rewrite E. lia.
       + apply IHa'.
   Qed.
 
-  Lemma not_in_value_0 a d : ~ In d (map fst a) -> value_at_weight a d = 0.
+ Lemma not_in_value_0 a d : ~ In d (map fst a) -> value_at_weight a d = 0.
```
JasonGross, line 542:
```suggestion
  Lemma not_in_value_0 a d : ~ In d (map fst a) -> value_at_weight a d = 0.
```

JasonGross, line 542: Did you deliberately change the indentation?
```suggestion
  Lemma not_in_value_0 a d : ~ In d (map fst a) -> value_at_weight a d = 0.
```

OwenConoly: No, that was an accident

src/PushButtonSynthesis/DettmanMultiplication.v
OwenConoly: Limb size is the thing referred to as "limbwidth" in the unsaturated solinas file.  I'll rename it into "limbwidth" in the dettman code.

The unsaturated solinas code just calculates the correct limbwidth as a function of the given prime and number of limbs, but the dettman code takes the limbwidth as a parameter.  I decided to do this because there is an easy generalization of the dettman algorithm to work with different limbwidths, and I don't know, in general, how to calculate the optimal limbwidth for the dettman algorithm (given a modulus and a number of limbs).

To calculate the limbwidth, the solinas code just divides the number of bits in the modulus by the number of limbs, but I don't know that that's always the right thing to do with the dettman algorithm.  The dettman algorithm takes advantage of the least significant limb having a significantly smaller size than the rest.  For example, with the bitcoin prime, we have four limbs with 52 bits, and then a last limb with 47 bits.  (That might not be quite right, but it's something like that.)  So for instance, if we wanted to implement the algorithm with 5 limbs and a modulus of size about 2^259, then having a limbwidth of 52 might work out poorly, and we might want it to be 53 instead--I'm not sure.

src/PushButtonSynthesis/DettmanMultiplication.v
```diff
@@ -0,0 +1,226 @@
...
+  Local Notation s := (2 ^ e).
+  Local Notation c := (Associational.eval c_).
+  Definition m := s - c.
+
+  Local Notation possible_values := possible_values_of_machine_wordsize.
+
+  Definition output_magnitude : nat := input_magnitude / 2 + 1.
+  Definition last_limb_width : nat := Z.to_nat (Qceiling (Z.log2_up s) - ((n - 1) * limbwidth)).
+  Definition input_bounds : list (ZRange.type.option.interp base.type.Z)
+    := (repeat (Some r[0 ~> 2 * input_magnitude * (2^limbwidth - 1)]%zrange) 4) ++ [Some r[0 ~> 2 * input_magnitude * (2^last_limb_width - 1)]%zrange].
```
JasonGross, line 92: Why 4?  Doesn't this hardcode things to 5 limbs?

src/PushButtonSynthesis/DettmanMultiplication.v
JasonGross: Wouldn't it be better then to take in last limb width as an argument (guaranteed to be a nat), and distribute the other bits evenly across the other limbs?  (And this would allow fractional bits, as we have in unsaturated solinas, too.)

src/CLI.v
```diff
@@ -394,6 +404,20 @@ Module ForExtraction.
...
         ["The number of limbs, or the literal '(auto)' or '(autoN)' for a non-negative number N, to automatically guess the number of limbs"]).
+  Definition n_nat_spec : anon_argT
+    := ("n",
+        Arg.Custom (parse_string_and parse_nat) "ℕ",
+        ["The number of limbs"]).
+  Definition limbwidth_spec : anon_argT
+    := ("limbwidth",
+        Arg.Custom (parse_string_and parse_nat) "ℕ",
+        ["The limb width"]).
+  Definition input_magnitude_spec : anon_argT
```
JasonGross, line 415: This sounds very similar to `inbounds_multiplier_spec`.  Is it different?  If not, please reuse `inbounds_multiplier_spec`

src/PushButtonSynthesis/DettmanMultiplication.v
JasonGross: > I don't know, in general, how to calculate the optimal limbwidth for the dettman algorithm

Is there a way to compute the valid range of bitwidths for the final limb?  E.g., is it always the case that you need exactly one extra bit of space in all the limbs but the last one, and you always need at least some minimum number of bits in the last limb?

src/Arithmetic/Core.v
```diff
@@ -528,23 +528,23 @@ Module Associational.
   Qed.
 
   Definition value_at_weight (a : list (Z * Z)) (d : Z) :=
-    fold_right Z.add 0 (map snd (filter (fun p => fst p =? d) a)).
+    fold_right (fun p sum => if (d =? fst p) then (sum + snd p) else sum) 0 a.
```
JasonGross, line 531: What's the reason for this change?  (And is it supposed to be in this PR, or another one?). It seems like it shouldn't change the generated code?

JasonGross: Oh, is `filter` not reifiable?

JasonGross: Should we just add support for reifying and rewriting `filter` by adding lines at
https://github.com/mit-plv/fiat-crypto/blob/6e6809be8290a7d7a9243b9491c09ec899ecb15b/src/Language/IdentifierParameters.v#L128
and
https://github.com/mit-plv/fiat-crypto/blob/6e6809be8290a7d7a9243b9491c09ec899ecb15b/src/Rewriter/Rules.v#L156-L163
?

JasonGross: (I think I'd prefer the change that adds more reification support, since it builds inventory towards supporting more algorithms more easily.)

src/PushButtonSynthesis/DettmanMultiplication.v
OwenConoly: Oops, I overlooked that.  I'll change it to (n -1), for both input_bounds and output_bounds.

src/CLI.v
OwenConoly: It is slightly different, but `inbounds_multiplier_spec` is better - it's Q instead of nat, and it's a named argument instead of being anonymous.  I'll use `inbounds_multiplier_spec`.

src/Arithmetic/Core.v
OwenConoly: Right, filter isn't reifiable.  I'll make it reifiable, and then revert the `value_at_weight` implementation.

OwenConoly: On second thought, I think I'll just change `value_at_weight` to use `partition` instead of `filter`.  Unless I was looking at things wrong, there would be some non-trivial work required to make `filter` reifiable.  In particular, I think I would need to write something analogous to the `interp_related_List_partition` lemma in [src/AbstractInterpretation/ZRangeProofs.v](https://github.com/mit-plv/fiat-crypto/blob/93f79b646576c767c58b54e5ec58957046aff490/src/AbstractInterpretation/ZRangeProofs.v#L139).

And I suppose we don't really need `filter` for anything.  It's easy enough, if slightly ugly, to replace it by `partition`.

src/PushButtonSynthesis/DettmanMultiplication.v
OwenConoly: > Wouldn't it be better then to take in last limb width as an argument (guaranteed to be a nat), and distribute the other bits evenly across the other limbs? (And this would allow fractional bits, as we have in unsaturated solinas, too.)

I don't think that having fractional bits would be worth the effort.  I wrote Arithmetic/DettmanMultiplication.v with the assumption that the limbwidths weren't fractional.   I'm not completely sure how I would generalize certain things to account for fractional limbwidths.  And more importantly, I don't think that having fractional limbwidths (instead of just rounding up to the nearest integer) would allow for any significant tightening of the output bounds.  Certainly it would make calculating the output bounds more complicated.  I didn't actually calculate the output bounds myself, by the way---I copied them from [here](https://github.com/bitcoin-core/secp256k1/blob/0eb3000417fcf996e3805d0eb00f0f32b8849315/src/field_5x52_impl.h#L501).

And if we do keep the limbwidth non-fractional, then taking in the limbwidth of the first limbs as input (instead of the last limb) is a natural way to enforce that the limbwidth of the first limbs has to be a natural number.

OwenConoly: > > I don't know, in general, how to calculate the optimal limbwidth for the dettman algorithm
> 
> Is there a way to compute the valid range of bitwidths for the final limb? E.g., is it always the case that you need exactly one extra bit of space in all the limbs but the last one, and you always need at least some minimum number of bits in the last limb?

Maybe there is an easy way to do that?  I'm not sure.

src/Arithmetic/Core.v
JasonGross: > In particular, I think I would need to write something analogous to the `interp_related_List_partition` lemma in [src/AbstractInterpretation/ZRangeProofs.v](https://github.com/mit-plv/fiat-crypto/blob/93f79b646576c767c58b54e5ec58957046aff490/src/AbstractInterpretation/ZRangeProofs.v#L139).

That's not necessary.  Just say that the bounds on the output of `filter` are `None` (`ZRange.ident.option.interp.None` or something?), which will make that proof trivial.

Roughly this means that bounds analysis won't work until filtering has been unfolded, but I think that's fine, and the pipeline hopefully shouldn't require any changes.

I believe the only lemma you need to prove is that `filter` is equal to some expression in terms of `list_rect`, to be used in the rewriting rules around
https://github.com/mit-plv/fiat-crypto/blob/93f79b646576c767c58b54e5ec58957046aff490/src/Rewriter/Rules.v#L144-L177

JasonGross:
Zig target failing with
```
src/main.zig:77:28: error: unable to load 'src/secp256k1_64.zig': FileNotFound
    try testVector(@import("secp256k1_64.zig"), expected);
```
Windows failing with
```
make: Nothing to be done for 'c-files'.
make: *** No rule to make target 'fiat-c/src/secp256k1_64.c', needed by 'lite-c-files'.  Stop.
```

src/PushButtonSynthesis/DettmanMultiplication.v
JasonGross: > I'm not completely sure how I would generalize certain things to account for fractional limbwidths.

I'm curious what breaks if you parameterize over an arbitrary weight function that satisfies
https://github.com/mit-plv/fiat-crypto/blob/93f79b646576c767c58b54e5ec58957046aff490/src/Arithmetic/Core.v#L1220-L1226.
(This shouldn't be considered to be blocking this PR though)

src/Arithmetic/Core.v
OwenConoly: Oh, ok!  I didn't know I could do that.  I'll add reification support for filter, then.  I suppose the bounds on filter can be tightened later if necessary.

src/PushButtonSynthesis/DettmanMultiplication.v
OwenConoly: Yeah, I could definitely parameterize over an arbitrary weight function, not change anything else, and the proof would still go through as before.  I was more wondering if there would be other things we'd want to modify if we want to allow for fractional bitwidths.

For instance, suppose we have a modular reduction from one position to another, but now (because the bitwidth is fractional) the number in the former position is too wide to fit (or bigger than we want) in the latter position, do we still want that reduction to be done as before?  Or should the algorithm be modified in some way?

Or should I just not worry about things like this, and just parameterize over the arbitrary weight function?  I'm not saying things will break if we allow for an arbitrary weight function, just that the algorithm might be somehow suboptimal.

JasonGross: My intuition says "don't worry about this here, just parameterize over an arbitrary weight function".

Here's some slightly rambly explanation of my thinking:

AIUI, this is why unsaturated solinas needs to reduce multiple times / have a nontrivial carry chain?

I'm also lacking vision on the "most general form" of the algorithm, so a bit hesitant to make confident claims here, but my inclination is that if we do sometimes want a nonuniform weight, and the proof works fine with an arbitrary weight function, then we should parameterize the algorithm over an arbitrary weight function and leave it to the higher layers to specify the reasonable class of weight functions.  And if we find that we want to generalize the algorithm further, we can do that; I don't expect further generalizations to be hampered by parameterizing over the weight function.
Said another way, I don't expect further generalizations to overlap only partially with a generic weight function; I don't expect there to be an algorithmic generalization that can only be phrased at all if we don't generalize over the weight function.
(Example of further generalization: Like, right now auto bounds analysis takes care of ensuring bounds are tight enough everywhere, maybe at some point we'll prove that the current algorithm is sufficient boundswise for some class of weight functions, and this will inform what we expose on the CLI.)

FWIW, IIRC, the way we get bounds guesses for unsaturated solinas (in the Heuristics file) is by actually running the algorithm for the inputs that we think will saturate the bounds, skipping the carrying step, and then use that to determine how much we need to carry / how much headspace we have.  Maybe there's some similar algorithm here?

In any case, I'd be happy to see a change (either in this PR or in another one) that parameterizes the algorithm over a weight function, and then has the CLI take in whatever arguments are most fundamental / most aligned with the parameters we might want to fiddle with / most aligned with how you think about the algorithm.

src/Arithmetic/Core.v
OwenConoly: I've made this work, but I needed to make a change to src/Rewriter/Util/ListUtil.v in the rewriter submodule.

I don't really know how submodules work.  Should I make a separate pull request in the rewriter repository?  If so, can you give me permission to push to the rewriter repository?  Currently I get a 403 error.

JasonGross: What change do you need to make to the rewriter?  (And are you sure it's necessary?) If you're adding a lemma to ListUtil, you should just add it to Fiat Crypto's copy.  In any case, yes, you should fork the rewriter project, and make a pull request.  Probably the easiest way to do this is to clone it from scratch and copy the change you made in the submodule to the fresh clone of the rewriter, and push it to your fork.  (You can also do this directly in the submodule with `git remote add `)

OwenConoly: Yeah, I'm just adding a couple lemmas to ListUtil.  The lemmas prove that filter is "Proper".  I don't know what that means, but it seems to be kinda related to functional extensionality.  And I think the lemmas are necessary, because building was failing without them.

OwenConoly: Here's the error message that happens when I try to build without the ListUtil lemmas:

WARNING: Could not automatically prove Proper lemma
(Morphisms.Proper
   (Morphisms.respectful (Morphisms.respectful eq eq)
      (Morphisms.respectful eq eq))
   (List.filter (A:=Compilers.base.interp base_interp t)))
  Please ensure this goal can be solved by typeclasses eauto
File "./src/Language/IdentifiersBasicGENERATED.v", line 18, characters 9-28:
Error: Cannot infer an existential variable of type
"Morphisms.Proper
   (Morphisms.respectful (Morphisms.respectful eq eq)
      (Morphisms.respectful eq eq))
   (List.filter (A:=Compilers.base.interp base_interp t))" (no type class
instance found) in environment:
t : Compilers.base.type base

JasonGross: Yeah, you should just add `Instance Proper_filter_eq {A} : Proper ((eq ==> eq) ==> eq ==> eq) (@filter A)` with a proof to to https://github.com/mit-plv/fiat-crypto/blob/master/src/Util/ListUtil/Filter.v, and add a `Require Import` in `IdentifiersBasicGENERATED.v`.  (Add `Require Import Coq.Classes.Morphisms.` if the notation fails to parse; you may also need `Coq.Relations.Relations`, `Coq.Classes.RelationClasses`, and/or `Coq.Setoids.Setoid`; I can never remember which of these is required for which things.)

To explain `Proper`: `Proper (Morphisms.respectful P Q) f` (generally written `Proper (P ==> Q) f`) means "if the inputs of `f` are related by `P`, then the outputs are related by `Q`".  Technically, `Proper R x` is just defined as `R x x`, and `respectful P Q f g` is `forall x y, P x y -> Q (f x) (g y)`.

OwenConoly: Ok, so I should delete this pull request: https://github.com/mit-plv/rewriter/pull/99?

src/AbstractInterpretation/ZRangeProofs.v
```diff
@@ -194,6 +194,57 @@ Module Compilers.
...
+                apply IHl1'' in IHhyp. clear IHl1''. destruct IHhyp as [IH1 IH2].
+                assert (H2': ZRange.type.base.option.is_bounded_by a a' = true).
+                { apply (H2 (pair a a')). simpl. left. reflexivity. }
+                clear H2. remember (H0 _ _ H2') as H0' eqn:clearMe. clear H0 clearMe.
+                destruct (f a) as [r|]; try discriminate E'. injection E' as E'. subst.
+                simpl in H0'. destruct (g a') as [|] eqn:E'''; try discriminate H0'.
+                clear H0'. simpl. rewrite E'''. constructor.
+                -- f_equal. assumption.
+                -- intros v H. apply IH2. assumption.
+          Qed.
```
JasonGross, line 246: I see you went for the fancier bounds analysis? :-)

src/Arithmetic/Core.v
OwenConoly: And just make the change in https://github.com/mit-plv/fiat-crypto/blob/master/src/Util/ListUtil/Filter.v instead?

src/Arithmetic/Core.v
```diff
@@ -528,22 +528,22 @@ Module Associational.
...
+  Lemma value_at_weight_works a d : d * (value_at_weight a d) = Associational.eval (filter (fun p => d =? fst p) a).
   Proof.
-    induction a as [| a0 a' IHa'].
-    - cbv [Associational.eval value_at_weight]. simpl. lia.
-    - cbv [value_at_weight]. simpl. destruct (fst a0 =? d) eqn:E.
-      + rewrite Associational.eval_cons. simpl. rewrite <- IHa'. cbv [value_at_weight]. lia.
+     cbv [value_at_weight]. induction a as [| a0 a' IHa'].
+    - cbv [Associational.eval]. simpl. lia.
+    - simpl. destruct (d =? fst a0) eqn:E.
+      + rewrite Associational.eval_cons. rewrite <- IHa'. apply Z.eqb_eq in E. rewrite E. simpl. lia.
```
JasonGross, line 538: These changes can be reverted now, right?

src/Arithmetic/Core.v
```diff
@@ -528,22 +528,22 @@ Module Associational.
...
       + apply IHa'.
   Qed.
 
   Lemma not_in_value_0 a d : ~ In d (map fst a) -> value_at_weight a d = 0.
   Proof.
     intros H. induction a as [| x a' IHa'].
     - reflexivity.
-    - cbv [value_at_weight]. simpl. destruct (fst x =? d) eqn:E.
+    - cbv [value_at_weight]. simpl. destruct (d =? fst x) eqn:E.
```
JasonGross, line 546: This can also be reverted?

src/Arithmetic/Core.v
```diff
@@ -556,28 +556,30 @@ Module Associational.
...
+         repeat rewrite map_app. rewrite (map_ext_in _ _ l1 (funs_same l1 a0 a' H2)).
+         rewrite (map_ext_in _ _ l2 (funs_same l2 a0 a' H3)). repeat rewrite Associational.eval_app. simpl.
+         repeat rewrite Associational.eval_cons. simpl. rewrite <- IHa'. simpl. rewrite Associational.eval_nil. 
+         cbv [dedup_weights]. rewrite H1. repeat rewrite map_app. repeat rewrite Associational.eval_app.
+         cbv [value_at_weight]. simpl. rewrite Z.eqb_refl. simpl. cbv [Associational.eval]. simpl. lia.
+       + simpl. apply (existsb_eqb_false_iff Z.eqb Z.eqb_eq) in E. rewrite (map_ext_in _ _ _ (funs_same _ _ _ E)).
+         repeat rewrite Associational.eval_cons. simpl. rewrite <- IHa'. cbv [dedup_weights]. f_equal. f_equal.
+         rewrite <- (nodupb_in_iff Z.eqb Z.eqb_eq) in E. cbv [value_at_weight]. simpl. rewrite Z.eqb_refl.
+         apply not_in_value_0 in E. cbv [value_at_weight] in E. simpl. rewrite E. lia.
+Qed.
```
JasonGross, line 583: And this can also be reverted, right?

src/Arithmetic/Core.v
JasonGross: > Ok, so I should delete this pull request: [mit-plv/rewriter#99](https://github.com/mit-plv/rewriter/pull/99)?
> And just make the change in https://github.com/mit-plv/fiat-crypto/blob/master/src/Util/ListUtil/Filter.v instead?

Yes, please

src/Arithmetic/Core.v
OwenConoly: Yes!

OwenConoly: This pull request now does not include any changes to Core.v

src/AbstractInterpretation/ZRangeProofs.v
OwenConoly: Yes!  That was fun.

src/Language/IdentifiersBasicGENERATED.v
```diff
@@ -3,6 +3,7 @@ Require Import Rewriter.Language.Language.
 Require Import Crypto.Language.IdentifierParameters.
 Require Import Rewriter.Language.IdentifiersBasicGenerate.
 Require Import Rewriter.Util.plugins.RewriterBuild.
+Require Import Util.ListUtil.Filter.
```
JasonGross, line 6: Minor:
```suggestion
Require Import Crypto.Util.ListUtil.Filter.
```

src/Util/ListUtil/Filter.v
```diff
@@ -34,4 +35,15 @@ Module List.
     induction 1; cbn; [ constructor | ].
     destruct f, g; try constructor; auto; try (exfalso; assumption).
   Qed.
+
+  Lemma filter_Proper {A} : Proper (pointwise_relation _ eq ==> eq ==> eq) (@List.filter A).
```
JasonGross, line 39: Minor: I don't see any reason to not make this an instance too, as it'll allow a bit more setoid rewriting 
```suggestion
  #[export] Instance filter_Proper {A} : Proper (pointwise_relation _ eq ==> eq ==> eq) (@List.filter A).
```

src/PushButtonSynthesis/DettmanMultiplication.v
```diff
@@ -0,0 +1,228 @@
...
+  Definition e : nat := Z.abs_nat (Z.log2_up s_).
+  Local Notation s := (2 ^ e).
+  Local Notation c := (Associational.eval c_).
+  Definition m := s - c.
+
+  Local Notation possible_values := possible_values_of_machine_wordsize.
+
+  Definition input_magnitude := Option.value inbounds_multiplier 1.
+  Definition output_magnitude_first_limbs : Q := input_magnitude / 2 + 1 / 2.
+  Definition output_magnitude_last_limb : Q := input_magnitude / 2 + 1 / 4. (* Where these bounds came from: https://github.com/bitcoin-core/secp256k1/blob/0eb3000417fcf996e3805d0eb00f0f32b8849315/src/field_5x52_impl.h#L545 *)
```
JasonGross, line 91: Plausibly this should be moved to a Heuristics file (a la UnsaturatedSolinasHeuristics), though it's fine to leave that for later.

JasonGross (approved):
Looking good!  I'm happy to have this merged once it's rebased and the CI builds fine.

src/PushButtonSynthesis/DettmanMultiplication.v
OwenConoly: I'm trying to do this now, and I've gotten stuck right here: https://github.com/mit-plv/fiat-crypto/blob/3322f772f999f0941b5e1633fe433bfaa4ba957d/src/PushButtonSynthesis/DettmanMultiplication.v#L183-L190

Coq complains about the last argument on line 188.  The issue is that `reified_mul_gen` wants a weight function as input---the type should be whatever the "reified equivalent" of `nat -> Z` is.  I don't know how to reify a function.  Can I do this, or do I need to refactor things somehow?

If the answer is that I need to refactor things, I was thinking of changing the following line to take a couple of integers (a numerator and denominator) as parameters, instead of a function.
https://github.com/mit-plv/fiat-crypto/blob/3322f772f999f0941b5e1633fe433bfaa4ba957d/src/Arithmetic/DettmanMultiplication.v#L18

JasonGross: > If the answer is that I need to refactor things, I was thinking of changing the following line to take a couple of integers (a numerator and denominator) as parameters, instead of a function.

Yes, that's the solution we use at https://github.com/mit-plv/fiat-crypto/blob/990bf2c3fccc13b747647834f3c7b853fa1da09d/src/Arithmetic/ModOps.v#L16-L19 and https://github.com/mit-plv/fiat-crypto/blob/990bf2c3fccc13b747647834f3c7b853fa1da09d/src/Arithmetic/Freeze.v#L143-L146

I'd suggest adding a `mod_ops` section at the bottom like we do for `Freeze`, defining a version that is specified to the numerator, denominator pair, and then reify that.

OwenConoly: Ah, ok.  I like the idea of keeping the algorithm in terms of the weight function and it’s hypotheses, so adding in the section at the bottom seems like a very nice solution.

andres-erbsen:
I like the direction this is going. We also should make sure we can put together an artifact tomorrow, even if that means merging a version we intend to improve later. Where do you think we currently stand wrt that?

OwenConoly:
> I like the direction this is going. We also should make sure we can put together an artifact tomorrow, even if that means merging a version we intend to improve later. Where do you think we currently stand wrt that?

I think it'll be in a mergeable state in a couple hours!

src/PushButtonSynthesis/DettmanMultiplication.v
OwenConoly: I think I'll leave that for later.

src/PushButtonSynthesis/DettmanMultiplication.v
OwenConoly: Ok, I think I've made the fractional bitwidths work properly.  Thanks for the helpful thoughts!

JasonGross (changes requested):
Zig still fails, you need to update https://github.com/mit-plv/fiat-crypto/blob/990bf2c3fccc13b747647834f3c7b853fa1da09d/fiat-zig/src/main.zig#L75 with the rename (and ideally add a test for dettmann)

OwenConoly:
I fixed the zig thing!  I didn't add a dettman test, mostly because I don't understand much of what is going on in the main.zig file.  Perhaps Andres can help me figure that out at some point---and I should probably add tests for the other languages sometime, too.

PR #1563, opened 2023-02-22
adapt to coq/coq#17281

andres-erbsen:


andres-erbsen:
`all` has passed, merging

PR #1564, opened 2023-02-23
Adapt to coq/coq#17274 (Eval prints evar info like Check)

SkySkimmer:


andres-erbsen:
@JasonGross @jadephilipoom Now that we test that montladder compiles to RISC-V, perhaps it makes sense to just remove this file?

JasonGross:
> perhaps it makes sense to just remove this file?

That seems reasonable.  The other backwards-compatible change is to adapt the `sed` script in the Makefile to remove `^where$$` and `^?.*`

ppedrot:
If you guys are going to remove it upstream, please close this overlay when that happens.

ppedrot:
Please merge now.

andres-erbsen:
I created https://github.com/mit-plv/fiat-crypto/pull/1579 and set it to merge upon CI success. Let's not merge this PR as it does not pass CI.

PR #1570, opened 2023-03-08
bump rupicola

samuelgruetter:


samuelgruetter:
I believe bumping the Coq version from 8.15 to 8.16 would solve these build failures. Would that be possible @JasonGross ?

ppedrot:
FTR this is needed for coq/coq#17322.

andres-erbsen:
(If @JasonGross responds to this, feel free to ignore my response in favor of his.)

In principle, I don't see a good reason to support 8.15 specifically. However, the Ubuntu-LTS CI using 8.15 is currently the only test that uses independently-maintained packages, which I very much would like to keep. Do you have an idea how to achieve that without 8.15?

Additionally, regardless of how this issue resolved, I think it is high time we find a systematic way to avoid this kind of issues in the future. The build breakage here came up in early January in https://github.com/mit-plv/fiat-crypto/pull/1553 at which point it would presumably been easy to revert the bedrock2 change, but you didn't, even though @JasonGross clearly objected to removing 8.15 at that time. Even if we remove 8.15 now, I don't want to end up forced into changes like this due to Coq PRs being indirectly blocked on them. There could  be a number of solutions here, but for a start: Would you be willing to call fiat-crypto CI in bedrock2 CI and mark the master branch as protected?

PR #1576, opened 2023-03-19
Add the dettman square operation

OwenConoly:
Turns out that the squaring algorithm is essentially identical to the multiplication algorithm.  To transform the mul function into the square function, I just had to replace `Associational.mul` with `Associational.square`.


OwenConoly:
I benchmarked the generated C implementation for square using the tools in https://github.com/bitcoin-core/secp256k1/.  It's significantly faster than the mul operation, which is reassuring.  I wasn't able to discern a significant difference between the performance of the generated C implementation vs. that of the handwritten assembly version.

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -101,21 +98,46 @@ Module DettmanMultiplication.
...
       - apply div_nz; try assumption. remember (weight_positive wprops (limbs - 1)). lia.
       - apply Divide.Z.mod_divide_full in weight_limbs_mod_s_eq_0. destruct weight_limbs_mod_s_eq_0 as [x H].
         rewrite H. rewrite Z_div_mult; try apply s_positive. rewrite Z.mul_comm. rewrite Z_mod_mult. lia.
     Qed.
+
+    Theorem eval_mulmod a b :
+      (Positional.eval weight limbs (mulmod a b)) mod (s - c) =
+      (Positional.eval weight limbs a * Positional.eval weight limbs b) mod (s - c).
+    Proof.
+      cbv [mulmod]. rewrite eval_reduce_carry_borrow. autorewrite with push_eval. reflexivity.
```
JasonGross, line 132: You should `Hint Rewrite eval_reduce_carry_borrow : push_eval` instead of manually rewriting

src/PushButtonSynthesis/DettmanMultiplication.v
```diff
@@ -168,10 +180,22 @@ Section __.
...
+  Definition ssquare (prefix : string)
+    : string * (Pipeline.M (Pipeline.ExtendedSynthesisResult _))
+    := Eval cbv beta in
+      FromPipelineToString!
+        machine_wordsize prefix "square" square
+        (docstring_with_summary_from_lemma!
+           (fun fname : string => [text_before_function_name ++ fname ++ " squares a field element."]%string)
+           (square_correct weightf n m input_bounds output_bounds)).
+
+  Hint Rewrite eval_mulmod using solve [ auto with zarith | congruence ] : push_eval.
```
JasonGross, line 192: I think this should live in the file where `eval_mulmod` is defined (outside of any sections, probably outside the modules too? Or just `Export (hints) mod_name.`)

src/PushButtonSynthesis/DettmanMultiplication.v
```diff
@@ -168,10 +180,22 @@ Section __.
...
+  Hint Rewrite eval_squaremod using solve [ auto with zarith | congruence ] : push_eval.
   Hint Unfold dettman_multiplication_mod_ops.mulmod : push_eval.
+  Hint Unfold dettman_multiplication_mod_ops.squaremod : push_eval.
```
JasonGross, line 195: This seems actively bad?  What breaks if you remove both `Hint Unfold`?

JasonGross (changes requested):
Seems mostly good!  Here are a couple suggestions that I think will make the organization slightly cleaner

src/PushButtonSynthesis/DettmanMultiplication.v
OwenConoly: It works without them!  I don't remember what led me to put them there originally

JasonGross:
What's up with the
```
fatal: detected dubious ownership in repository at '/github/workspace/rewriter'
To add an exception for this directory, call:

	git config --global --add safe.directory /github/workspace/rewriter
```
errors?  (Does the branch just need to be rebased, or is something bigger going on?)

OwenConoly:
I don't know what could've caused that.

JasonGross:
Seems to be failing on master now? https://github.com/mit-plv/fiat-crypto/actions/workflows/docker-coq.yml?query=branch%3Amaster

JasonGross:
I've asked about this at https://coq.zulipchat.com/#narrow/stream/237656-Coq-devs-.26-plugin-devs/topic/docker.20dev.2Fmaster.20change.3F

src/Arithmetic/DettmanMultiplication.v
JasonGross: You marked this resolved but the code hasn't changed?  I guess you'll push an updated version soon?  (I want to make sure you didn't accidently overwrite changes with a force push)

OwenConoly: Yes!  I’ll push an updated version in a bit.  I was just marking things resolved as I went along

PR #1587, opened 2023-04-12
Update fiat-amd64 files

dderjoel:
I wanted to replace the current test files because they felt too cluttered to me.
So in this PR I've removed old assembly files, which contained (possibly confusing) debug information and replaced them with new files which
- are cleaned from (debug) comments
- follow unified naming schemes (symbol is the same as created by the respective cli-tool) and are in a dedicated folder
- include the new implementation strategies for `secp256k1_dettman` and the saturated `curve25519_solinas` implementation
- are optimized with CryptOpt for many architectures, so people may even choose to use them (They contain information on which CPU they were optimized, and how much faster than clang they are.)

Additionally, I've added a little documentation.

Those new files are generally using the red-zone of the stack. This seems to cause trouble in many verifications. 
For now, I've added `--asm-stack-size 80` to all commands, but this is just a random number and only a workaround such that small functions get verified. I am not sure what is causing this. Technically, everything regarding the stack size should be implicit in the file.

In the two commands here,
```bash
src/ExtractionOCaml/dettman_multiplication secp256k1_dettman 64 5 48 '2^256 - 4294968273' mul --no-wide-int --shiftr-avoid-uint1  --hints-file fiat-amd64/fiat_secp256k1_dettman_mul/seed0000000451748095_ratio12970.asm --asm-stack-size 10
src/ExtractionOCaml/dettman_multiplication secp256k1_dettman 64 5 48 '2^256 - 4294968273' mul --no-wide-int --shiftr-avoid-uint1  --hints-file fiat-amd64/fiat_secp256k1_dettman_mul/seed0000000451748095_ratio12970.asm --asm-stack-size 20
```

 the first one fails with 

```text
Symbolic execution failed:
In mem state:
[(451, 418), (448, 419), (446, 420), (443, 421), (441, 422), (438, 423), (436, 424), (434, 425), (432, 426), (430, 427), (417, 395), (416, 396), (415, 397), (414, 398), (413, 399), (412, 5), (411, 6), (410, 7), (409, 8), (408, 9), (407, 0), (405, 1), (403, 2), (402, 3), (400, 4)]
Index 453 loaded without being present.
453 is (add 64, [#428, #452])
452 is (const 18446744073709551488, [])
428 is the value of the register rsp.
```
whereas the second one succeeds (only difference is the stack size parameter). Note that this file is not changing the stack-pointer and only writes to the red zone from `[ rsp - 0x80 ]` to `[ rsp - 0x58 ]`. I'm not sure why why `--asm-stack-size 20` makes the validation succeed.


dderjoel:
Its just a draft because I've added the `--asm-stack-size 80`, so I assume that the CI fails, as I don't think that all checks go though.

That reminds me, that we should probably also add failing test cases... (but somehow CI-only and not in the project, so people don't accidentally use them).

Also, I am a python-n00b, feel free to give me hints on how to do better in the `gentest.py`.

JasonGross:
Stack size is inferred with
https://github.com/mit-plv/fiat-crypto/blob/59965517c88e79881112893d6b93fd03dd9a4f36/src/Assembly/Equivalence.v#L74-L116
I guess according to https://en.m.wikipedia.org/wiki/Red_zone_(computing) we should parameterize the stack size defaults on the calling convention a la https://github.com/mit-plv/fiat-crypto/blob/59965517c88e79881112893d6b93fd03dd9a4f36/src/Assembly/Equivalence.v#L40 to include the red zone by replacing `extra_ss` in https://github.com/mit-plv/fiat-crypto/blob/59965517c88e79881112893d6b93fd03dd9a4f36/src/Assembly/Equivalence.v#L115 with `red_zone` where `red_zone` is to be computed from a new parameter `{calling_convention : assembly_callee_saved_registers_opt}`.

Can you see if this code works?
```coq
(** According to https://en.m.wikipedia.org/wiki/Red_zone_(computing)

Whether a red zone is present depends on the calling convention. x86-64 systems that use the System V AMD64 ABI (including Linux and macOS) use a 128-byte red zone that begins directly under the current value of the stack pointer.[1][2] The OpenRISC toolchain assumes a 128-byte red zone.[3] Microsoft Windows does not have the concept of a red zone on x86.[4] In fact, the ABI explicitly states that the memory beyond the stack pointer is volatile and may be overwritten by debuggers or interrupt handlers.[5][6] However, Microsoft Windows has a red zone of 16 bytes on IA-64, 8 bytes on AArch32, and 16 bytes on AArch64.[6]
*)
Definition microsoft_x64_assembly_stack_red_zone : N := 0.
Definition system_v_amd64_assembly_stack_red_zone : N := 128.
Definition assembly_stack_red_zone {calling_convention : assembly_callee_saved_registers_opt} : N
  := match calling_convention with
     | Microsoft_x64 => microsoft_x64_assembly_stack_red_zone
     | System_V_AMD64 => system_v_amd64_assembly_stack_red_zone
     | explicit_registers _ => 0%N
     end.
Definition assembly_stack_size {calling_convention : assembly_callee_saved_registers_opt} {v : assembly_stack_size_opt} (asm : _) : N
  := match v with
     | Some v => v
     | None => let let red_zone := Z.of_N assembly_stack_red_zone in
               Z.to_N (fold_right Z.max red_zone (assembly_stack_size_at red_zone asm))
     end.
```
(and remove the definition of `extra_assembly_stack_size`)

Does this seem right to you?

dderjoel:
1. Why did the CI pass earlier? When I tun the `./fiat-amd64/test.sh` locally, It fails at p434 mul (as I'd expect)
2. Now, for your code, I hope I did that correct. I've replaced the `let let` with `let`. I'll be building and checking on my side, too.
3. I'll change the `./gentest.py` to remove the `--asm-stack-size` param.

fiat-amd64/Readme.md
```diff
@@ -0,0 +1,24 @@
...
+## Meta
+- `Readme.md`  This file
+- `gentest.py` Test generation script to create commands to formally verify the correctness of the assembly files.
+- `test.sh`    Helper script to manually call and run the tests. Uses `./gentest.py`.
+
+## Assembly files
+
+The structure is `./fiat_<CURVE>_<METHOD>/seed<SEED>_ratio<RATIO>.asm`.
+Every folder contains optimized and proven correct implementations of the respective function using the CryptOpt Framework.
+The `<CURVE>` is one of Fiat Cryptography's main supported curves: `curve25519_carry` `curve25519_solinas` `p224` `p256` `p384` `p434` `p448_solinas_carry` `p521_carry` `poly1305_carry` `secp256k1_dettman` `secp256k1_montgomery`
```
JasonGross, line 14: The `_carry` is not part of the `<CURVE>`; `carry_` is an optional prefix on the `<METHOD>`

fiat-amd64/Readme.md
```diff
@@ -0,0 +1,24 @@
...
+The `<METHOD>` is one of the respective `multiply` or `square` (carry) if it is based on the solinas primes.
+
+The `<SEED>` value is the seed that CryptOpt was called with to create this file, the `<RATIO>` is an indication of the measured performance during optimization.
+Specifically, a `<RATIO>`-value of `abbbb` corresponds to a performance ratio of `a.bbbb`x compared to the assembly that `clang-15 -O3 -march=native -mtune=native` would have produces on the system.
+Hence, the higher the value the faster the code.
+Every `asm` file contains a footer comment with some meta data.
+On which CPU it was produced.
```
JasonGross, line 21:
```suggestion
Every `asm` file contains a footer comment with some metadata on which CPU it was produced.
```
?

fiat-amd64/gentest.py
```diff
@@ -28,32 +32,58 @@ def removeprefix(s, prefix):
...
 output_makefile = ('--makefile' in sys.argv[1:])
-asm_files = tuple(i for i in sys.argv[1:] if i not in ('--makefile',))
+directories = tuple(i for i in sys.argv[1:] if i not in ('--makefile',))
 
 asm_op_names = OrderedDict()
-for fname in asm_files:
-    op, name = removesuffix(fname, '.asm').replace('_solinas','').split('_')[-2:]
-    asm_op_names.setdefault((name, op), []).append(fname)
+
+regex = re.compile(r'fiat_(?P<name>[^_]+(_((solinas)|(montgomery)|(dettman)))?)_(?P<op>(carry_)?((square)|(mul)))')
```
JasonGross, line 49: Optional: you don't need so many parentheses:
```suggestion
regex = re.compile(r'fiat_(?P<name>[^_]+(_(solinas|montgomery|dettman))?)_(?P<op>(carry_)?(square|mul))')
```

fiat-amd64/Readme.md
dderjoel: don't know how that happened ;D

fiat-amd64/gentest.py
dderjoel: Beauty of Python. (unrelated, I also always believed one would need them in JS but I just double checked. Seems alright in this instance...)

fiat-amd64/gentest.py
```diff
@@ -28,32 +32,58 @@ def removeprefix(s, prefix):
...
+for dirname in directories:
+    m = regex.match(os.path.basename(dirname)).groupdict(None)
```
JasonGross, line 51: Is `.groupdict(None)` different from `.groupdict()`?

fiat-amd64/gentest.py
```diff
@@ -28,32 +32,58 @@ def removeprefix(s, prefix):
...
+    if m is None:
+        continue
+    for fname in os.listdir(dirname):
+        asm_op_names.setdefault((m.get('name'), m.get('op')), []).append(os.path.join(dirname,fname))
```
JasonGross, line 55: Minor: I would write:
```suggestion
    if m:
        for fname in os.listdir(dirname):
            asm_op_names.setdefault((m.get('name'), m.get('op')), []).append(os.path.join(dirname, fname))
```

fiat-amd64/gentest.py
dderjoel: Not sure; say it does not match, does it then default to `None` such that the `if m is None:` works?

dderjoel: also, I'm not 100% sure what the 'kind = ` are for. I guessed they are for the sorting (line 89) itself? But is it then useful to have this order?

fiat-amd64/gentest.py
dderjoel: I also found it rather verbose to use `(m.get('name'), m.get('op'))` as the key. Could that not just be `m` in this case?

dderjoel:
The Coc build error'ed out on my machine with 
```
Finished transaction in 9.609 secs (9.609u,0.s) (successful)
COQC src/PushButtonSynthesis/SolinasReduction.v
COQC src/CLI.v
File "./src/CLI.v", line 486, characters 111-136:
Error: The reference extra_assembly_stack_size was not found in the current
environment.

make: *** [Makefile.coq:764: src/CLI.vo] Error 1
```

fiat-amd64/gentest.py
JasonGross: According to https://docs.python.org/3/library/re.html#re.Match.groupdict, `.groupdict()` is the same as `.groupdict(None)`.  (The `=None` in the doc is the default value of the argument.)

> also, I'm not 100% sure what the 'kind = ` are for. I guessed they are for the sorting (line 89) itself? But is it then useful to have this order?

Yes, for the sorting.  Generally solinas runs much faster than montgomery, IIRC, and I wanted the order that the tests are run in via `make` to be roughly based on timing (either to allow for better parallelism, by running the slowest ones first, or to allow for quicker failures, by running the fastest ones first).

fiat-amd64/gentest.py
JasonGross: `m` is a dictionary, so you don't want to use it as a key.  You could plausibly do something like `tuple(sorted(m.items()))`, but that seems needlessly arcane.  You can shorten it by doing `(m['name'], m['op'])`, though, I think.  Or you can use another line, do `key = (m['name'], m['op'])` and then do `.setdefault(key, []).append(...)`

JasonGross:
> Feels like this for the `--asm-stack-size` parameter. Would that be needed anymore then anyway?

This is just for the documentation of the `--asm-stack-size` argument.  Maybe replace
```coq
" ++ show (extra_assembly_stack_size:N) ++ "
```
with
```coq
the conventional size of the red zone (" ++ show (system_v_amd64_assembly_stack_red_zone:N) ++ " for --asm-callee-saved-registers " ++ show System_V_AMD64 ++ ", or " ++ show (microsoft_x64_assembly_stack_red_zone:N) ++ " for  --asm-callee-saved-registers " ++ show Microsoft_x64 ++ ", or " ++ show (@assembly_stack_red_zone (explicit_registers []):N) ++ " for an explicit list of registers given to --asm-callee-saved-registers)
```

JasonGross:
Oh, wait, no, that's wrong now.

dderjoel:
Wait, what do you mean is wrong? Should I use that snippet? It sounds correct to me.

JasonGross:
> Wait, what do you mean is wrong? Should I use that snippet? It sounds correct to me.

I thought it was wrong, but then reread it and realized it was fine, and deleted my comment that it was wrong.  Sorry for the confusion.  Yes, use the snippet.

dderjoel:
So the build succeeds now on my machine. 

The output in the Makefile is  not particularly useful, atm. It only prints the folder (i.e. `fiat_p521_carry`) ... I've fixed that locally already...

But the `make -j2 only-test-amd64-files SLOWEST_FIRST=1` checks fail.
When I execute the `./test.sh` it works for many methods already (including some (p256) with `word_by_word_montgomery`) but fails at 
```
src/ExtractionOCaml/word_by_word_montgomery p434 64 '2^216 * 3^137 - 1' mul --no-wide-int --shiftr-avoid-uint1 --hints-file fiat-amd64/fiat_p434_mul/seed0700010143014235_ratio13987.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000217524044_ratio17587.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000003350093_ratio15806.asm --hints-file fiat-amd64/fiat_p434_mul/seed2001662480271177_ratio14065.asm --hints-file fiat-amd64/fiat_p434_mul/seed1533427187598446_ratio13382.asm --hints-file fiat-amd64/fiat_p434_mul/seed1150715237846205_ratio15519.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000359819836_ratio18852.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000012268247_ratio14900.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000721769283_ratio13918.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000298855511_ratio19688.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000208758041_ratio20048.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000190604786_ratio11531.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000516314061_ratio17208.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000661934233_ratio18760.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000666693375_ratio18016.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000213147076_ratio17611.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000657779768_ratio19658.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000128560180_ratio14471.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000007811155_ratio14854.asm --hints-file fiat-amd64/fiat_p434_mul/seed0355724693291898_ratio13119.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000336868808_ratio18387.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000689842234_ratio11579.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000548862452_ratio17266.asm --hints-file fiat-amd64/fiat_p434_mul/seed2128444020844024_ratio13153.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000538470079_ratio17952.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000134942015_ratio14237.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000668518600_ratio11548.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000179753512_ratio12000.asm --hints-file fiat-amd64/fiat_p434_mul/seed0000000185516258_ratio11667.asm --hints-file fiat-amd64/fiat_p434_mul/seed3705239160001646_ratio13879.asm -o /dev/null --output-asm /dev/null
```

I'll try to get the error but p434 takes ages

dderjoel:
Okay, So running the `./test.sh` file locally results in the two failed tests for p434. (and they are terminating rather rapidly)
But running the commands themselves eventually succeeds.

I wonder why this happens, maybe this gives clues, but I am not understanding why that is:
https://github.com/mit-plv/fiat-crypto/actions/runs/4685439720/jobs/8302548799#step:11:1813

JasonGross:
> Okay, So running the `./test.sh` file locally results in the two failed tests for p434. (and they are terminating rather rapidly) But running the commands themselves eventually succeeds.

Is the failure a memkill?

> I wonder why this happens, maybe this gives clues, but I am not understanding why that is: https://github.com/mit-plv/fiat-crypto/actions/runs/4685439720/jobs/8302548799#step:11:1813

This looks like maybe the Makefile is set up wrong?  Is `fiat-amd64/p448_solinas-carry_square.only-status` supposed to have the `_solinas` in there?  What happens if you run `make fiat-amd64/p448_solinas-carry_square.only-status` locally?

dderjoel:
> > 
> 
> Is the failure a memkill?

No it exits instantly. I think p434 is the only prime with a '*' in it. I believe bash is expanding it in this particular use case. Hence It does succeed when pasted otherwise. Let me try to figure out how to fix it.

dderjoel:
> This looks like maybe the Makefile is set up wrong? Is `fiat-amd64/p448_solinas-carry_square.only-status` supposed to have the `_solinas` in there? What happens if you run `make fiat-amd64/p448_solinas-carry_square.only-status` locally?

```
make[1]: 'src/Coqprime/PrimalityTest/PocklingtonCertificat.vo' is up to date.
make: *** No rule to make target 'fiat-amd64/p448_solinas-carry_square.only-status'.  Stop.
```
Now  I see why there is the `replace('_solinas','')`. I've done that to not have clashes with `curve25519` and `curve25519_solinas`. I'll see what I can do

dderjoel:
It seems to be only a timeout issue now. 
running `time make -j only-test-amd64-files SLOWEST_FIRST=1` on my machine (Intel i7-11700KF in Docker) results in 

```
ALL 22 AMD64 ASM TESTS PASSED

real    134m15.276s
user    320m53.672s
sys     0m8.186s
```
I'm not too familiar with the GitHub-CI just yet. As far as I can see, we have two options now
1. We create a dedicated workflow for testing the amd64 files, which depends on the ExtractionOCaml binaries (as artifacts), and run on its own.
2. We skip checking the heavy files.

.github/workflows/amd64.yml
```diff
@@ -0,0 +1,23 @@
+name: Test Generated Go
```
JasonGross, line 1:
```suggestion
name: Equivalence Check AMD x64 Assembly Files
```

.github/workflows/amd64.yml
```diff
@@ -0,0 +1,23 @@
...
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 0 1 * *'
+
+jobs:
+  test-go:
```
JasonGross, line 10:
```suggestion
  test-amd64:
```

.github/workflows/amd64.yml
JasonGross, line 1: I think you want a separate job in the existing file, not a new file.  Maybe use https://github.com/coq-community/coq-performance-tests/blob/4aaef74e5742fe3ad87c5d18d735e82b1284ecc6/.github/workflows/coq.yml#L114-L140 as a reference?

.github/workflows/coq.yml
```diff
@@ -83,3 +83,19 @@ jobs:
     - name: validate
       run: make TIMED=1 validate COQCHKFLAGS="-o ${COQCHKEXTRAFLAGS}"
       if: matrix.env.SKIP_VALIDATE == '' && github.event_name != 'pull_request'
+  test-amd64:
+
+    runs-on: ubuntu-latest
+
```
JasonGross, line 89:
```suggestion

    concurrency:
      group: ${{ github.workflow }}-test-amd64-${{ github.head_ref || github.run_id }}
      cancel-in-progress: true
```

.github/workflows/coq-windows.yml
```diff
@@ -103,3 +99,28 @@ jobs:
...
+      with:
+        submodules: recursive
+    - name: Download a Build Artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: ExtractionOCaml
+        path: src/ExtractionOCaml
+    - name: Display structure of downloaded files
+      run: ls -R
+    - name: only-test-amd64-files
```
JasonGross, line 123:
```suggestion
    - name: only-test-amd64-files-lite
```

dderjoel:
how can we address 
https://github.com/mit-plv/fiat-crypto/actions/runs/4737616235/jobs/8413216495#step:5:3
in the windows pipeline, where it failed with 
```
Run %CYGWIN_ROOT%\bin\bash.exe -l -c 'cd "%cd%"; ./etc/ci/github-actions-make.sh -j${NJOBS} only-test-amd64-files-lite SLOWEST_FIRST=1'
  %CYGWIN_ROOT%\bin\bash.exe -l -c 'cd "%cd%"; ./etc/ci/github-actions-make.sh -j${NJOBS} only-test-amd64-files-lite SLOWEST_FIRST=1'
  shell: C:\Windows\system32\cmd.EXE /D /E:ON /V:OFF /S /C "CALL "{0}""
The system cannot find the path specified.
Error: Process completed with exit code 1.
```

I've included the python setups and opam. Is the problem that CYGWIN is not there? I'm unsure which system is yielding that message.

JasonGross:
> I've included the python setups and opam. Is the problem that CYGWIN is not there? I'm unsure which system is yielding that message

Probably yes, the problem is missing cygwin, which is incidentally installed by setup-ocaml.  Note that you haven't set up the env in the windows job, though, so you may get mem-killed from running with -j, or the OCaml step might fail from missing opam_yes, etc

JasonGross:
> Do I need the Python for OPAM?

I don't think so.  You probably don't actually need the OCaml job either, just use setup-cygwin or something.  Or we can just skip equivalence checking on Mac and Windows, or use the lite targets there and keep them in the main job, that seems maybe better?

dderjoel:
Not sure I understand this lite-target thing. If I understand correctly, It is just a subset of the curves, which can be checked quickly, right? Then We'd still need to get all the dependencies in, right?

I'm more than happy to leave the heavy verification of all files on the one CI Coq job, and skip entirely on the others. I mean the files wont change frequently, And I believe its rather unlikely that a change in the Fiat-Code makes a file unverifiable only on non-linux machines.

JasonGross:
> Not sure I understand this lite-target thing. If I understand correctly, It is just a subset of the curves, which can be checked quickly, right?

Yes

> Then We'd still need to get all the dependencies in, right?

What dependencies?  The asm checks are all completely independent, right?  (There are lite targets for the main build, which is separate; I'm talking here about the lite asm target, which has the same binary/Coq dependencies as the other target and just restricts checking to some of the asm files.)

> I'm more than happy to leave the heavy verification of all files on the one CI Coq job, and skip entirely on the others. I mean the files wont change frequently, And I believe its rather unlikely that a change in the Fiat-Code makes a file unverifiable only on non-linux machines.

I think we should leave at least a couple asm checks on all platforms, to make sure that we don't mess up line endings, etc.  But I agree that the core functionality shouldn't differ across platforms

dderjoel:
> What dependencies?

the dependencies to run the binary. On windows that seems to be Cygwin (which is created by the opam package)

Okay lets to it this way. I'll revert most of the last commit to check the lite target on all platforms but the coq workflow and try to get it running on Win with the cygwin-setup step

.github/workflows/docker-coq.yml
```diff
@@ -98,3 +98,26 @@ jobs:
...
+      uses: actions/checkout@v3
+      with:
+        submodules: recursive
+    - name: Download a Build Artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: ExtractionOCaml-master
+        path: src/ExtractionOCaml
+    - name: make binaries executable
+      run: chmod +x src/ExtractionOCaml/solinas_reduction src/ExtractionOCaml/word_by_word_montgomery src/ExtractionOCaml/unsaturated_solinas src/ExtractionOCaml/dettman_multiplication
```
JasonGross, line 121: Might be more future proof to do
```suggestion
      run: chmod +x src/ExtractionOCaml/*
```
?
(Also, do you need `sudo`? And/or to say `chmod a+x`? Or `chown`?)

dderjoel:
I was so confident the CI would just run through this time... 
So the Assembly checks pass, see here: https://github.com/mit-plv/fiat-crypto/actions/runs/4759329181/jobs/8460673296?pr=1587#step:5:94

But it does not like that there is a diff https://github.com/mit-plv/fiat-crypto/actions/runs/4759329181/jobs/8460673296?pr=1587#step:5:600

I'm unsure where this comes from. Assuming it domes from [here](https://github.com/mit-plv/fiat-crypto/blob/master/etc/ci/github-actions-make.sh#L58), Should we add a step that we remove the `src/ExtractionOcaml` files? Or can we disable that check with `ALLOW_DIFF=1`? Or should I just run `make` instead of `etc/ci/github-actions-make.sh -j2 only-test-amd64-files SLOWEST_FIRST=1`

dderjoel:
Is there anything else to do here or can we merge?

.github/workflows/coq-windows.yml
```diff
@@ -78,7 +78,7 @@ jobs:
       shell: cmd
     - name: only-test-amd64-files-lite
       run: |
-        %CYGWIN_ROOT%\bin\bash.exe -l -c 'cd "%cd%"; opam exec -- etc/ci/github-actions-make.sh -j${NJOBS} only-test-amd64-files-lite SLOWEST_FIRST=1'
+        %CYGWIN_ROOT%\bin\bash.exe -l -c 'cd "%cd%"; etc/ci/github-actions-make.sh -j${NJOBS} only-test-amd64-files-lite SLOWEST_FIRST=1'
```
JasonGross, line 81: Why this change?  Is it necessary?

JasonGross:
Can you revert the change to .github/workflows/coq-windows.yml?, and then I'll merge (squash and merge, unless you want me to preserve history)

dderjoel:
I don't really care if the history is preserved, but I think it would be a bit useful no have delete / add amd64 files separated, and the adjustments to the workflows.

I've just rebased and cleaned up the history in https://github.com/dderjoel/fiat-crypto/tree/cleanmaster.
I could force push this to my fork at master, or change this PR to originate from cleanmaster, or I could close this PR and open a new one.
I'd let you decide whats best in terms of documentation.

JasonGross:
I'd suggest just force pushing to this PR (your master), though I'd also be fine with a separate PR if you'd prefer

PR #1596, opened 2023-04-28
Cmov using sbb when possible

andres-erbsen:
This is required to get decent code for saturated subtraction on clang with intrinsics


src/Rewriter/RulesProofs.v
```diff
@@ -571,6 +571,13 @@ Proof using Type.
   all: repeat interp_good_t_step_arith.
   all: remove_casts; try fin_with_nia.
   all: try (reflect_hyps; lia).
+
+  match goal with H5 : _ |- _ => apply unfold_is_bounded_by_bool in H5; cbn in H5 end.
+  replace c with 1 in * by lia.
+  rewrite Z.mod_opp_l_nz; rewrite ?Z.mod_small.
+  replace (M+1-1) with M by lia.
+  remove_casts.
+  all : try lia.
```
JasonGross, line 580: Is there any way this can be made more automated, and integrated with the above automation?  Like `M+1-1` should be folded into `interp_good_t_step_arith` if possible, etc

andres-erbsen: I don't see anything useful to do here. I could wrap the thing in a  match clause but I worry doing so might make maintenance harder, not easier.

PR #1597, opened 2023-04-28
Add a constant-time binary extended GCD algorithm.

jadephilipoom:
This is a lightly updated version of #333, as discussed on that PR. The only changes I made were to fix things that broke because lemmas got renamed, moved, etc. in the last few years. Given the amount of time that's passed, there were surprisingly few of these!


andres-erbsen:
So the older PR has some suggestions but I think evaluating and integrating these would require paging in the context for actually developing the work, which might delay it further. I am fine merging on the basis that  we'd rather have this than not and none of the suggested improvements are both obviously fine and likely to result in a significant maintenance benefit.

Issue #1582, opened 2023-04-01
Possible optimization for secp256k1


sipa:
The current C code for the 5x52 field implementation in libsecp256k1 has (at least) this optimization which seems not present in the fiat-crypto output (it's also absent in libsecp256k1's asm code, which probably explains how it was missed): https://github.com/bitcoin-core/secp256k1/pull/810

I'm really happy with how close the fiat-crypto generated code seems to be getting, and we're seriously contemplating incorporating it in the library, but it'd be an easier sell if it has at least all the same high-level optimizations.

dderjoel:
Hi,
just wanted to gauge how complex this would be to implement in fiat-crypto.

As far as I can see from the [changes in C-code](https://github.com/bitcoin-core/secp256k1/pull/810/files#diff-35bbd3490af4fba588fa0e1cabfb6642d198e639170a6efc0b8dd5ee44d214b2L180-L197), it replaces an 
```
c & M * R  // producing AND and MULX
c>>52        // producing SHR 
...
c * R          // 128*64 -> 2 MULX + ADDs (or maybe 2 ADDs)
```
by 
```
R*c                  // MULX
c>>64              // free
(R<<12) * c     // SHL and MULX
```

This happens twice in each function.

I think this is genius!

I've dug a little through the fiat-code and only found [this Dettman-related file](https://github.com/mit-plv/fiat-crypto/blob/master/src/Arithmetic/DettmanMultiplication.v#L82-L99) and in it I could only identify lines 82 -- 99, which seem like they specify the multiplication, but I am too unfamiliar with what is actually happening. 
I'm more than happy to help with this wherever I can, but I'd need some help from the experts @OwenConoly, @JasonGross, @andres-erbsen .

OwenConoly:
> it's also absent in libsecp256k1's asm code, which probably explains how it was missed

Yeah, I referenced the asm code when I wrote fiat-crypto's template, so that would explain it.

My thought is that this change would not be very difficult to implement in fiat-crypto.  It looks just like the same sort of operations that we already have.  I haven't yet had time to look at this in detail, though.

dderjoel:
Hi @OwenConoly , did you have time already?
Would it help, if I would rewrite the assembly incorporating the optimization?

OwenConoly:
[...]

PR #1601, opened 2023-05-03
Optimize dettman algorithm

OwenConoly:
This implements the optimization described in #1582.

This should resolve almost all of #1582; the only thing missing is the aesthetic improvement mentioned [here](https://github.com/mit-plv/fiat-crypto/commit/b7798cb03a6ecd597d2fb5f55c2b67cba6347d93#r111587543) about removing an unnecessary & operation, which I don't know how to fix.

Incidentally, I've realized that the dettman template, in its state before this PR, did unreasonable things when given a fractional limbwidth.  I've rewritten it so that it will hopefully---I haven't tested this or anything---now behave reasonably given a fractional limbwidth.  This happened to require the same sort of change that I was making to implement the optimization.  This change should make the algorithm a bit easier to read, too.


dderjoel:
FWIW, I am unsure if the current files in [fiat-amd64/fiat_secp256k1_dettman_mul ](https://github.com/mit-plv/fiat-crypto/tree/master/fiat-amd64/fiat_secp256k1_dettman_mul) (and square respectively) would still be proven correct.
I've updated the files in [CryptOpt](https://github.com/0xADE1A1DE/CryptOpt/tree/main/generated/secp256k1) already. Feel free to replace them, in case the validation does not succeed. I can then later provide the 3-best-per-architechture files just like for the other curves.

OwenConoly:
Ah!  I forgot about updating those.  Indeed, they [failed](https://github.com/mit-plv/fiat-crypto/actions/runs/4876486620/jobs/8702345705?pr=1601).  I'll replace them with the CryptOpt files that you linked.

OwenConoly:
I'm trying to add [this](https://github.com/mit-plv/fiat-crypto/commit/b7798cb03a6ecd597d2fb5f55c2b67cba6347d93#r111587543) enhancement, via a new and/or generalized rewriting rule, before merging this PR.
I've added some commits to this branch where I make some attempts at progress on that, so this branch is no longer in a good state to merge.

OwenConoly:
The dettman code generation actually works now (with the redundant `&` removed!), but the new rewriting pass that I've added has broken some other things.

Code generation for `fiat-c/src/p521_32.c` (among others, I assume) doesn't work---the error is with "converting to C", as you can see in [the build of my most recent commit](https://github.com/mit-plv/fiat-crypto/actions/runs/5006935804/jobs/8972875948?pr=1601).

@andres-erbsen, @JasonGross - do you have any idea what would cause this?  Might it be the same issue as in https://github.com/mit-plv/fiat-crypto/issues/1604?

OwenConoly:
Oops, never mind my previous comment.  I've inferred (correctly, I think?) from #1604 that my problem has something to do with dead code elimination, so I've removed the dead code elimination thing that I added in `BoundsPipeline.v`.  This seems to have fixed the issue.

OwenConoly:
Looks like the redundant ands, which are removed by the new rewrite pass that I added, actually appeared in a lot of places other than the dettman code.  Adding the new rewrite pass caused the total length of generated code to decrease by about 27000 lines.  I was not expecting this, but I think it's a good thing.

andres-erbsen:
This cleanup is indeed more impressive than anticipated :D. Great job!

JasonGross (approved):
Other than needing to update one more 3 into 4 for accurate error messages, this looks good to me!

PR #1608, opened 2023-06-03
Generalize dettman algorithm

OwenConoly:
Now we can use the same template to generate code for both the 5x52 and 10x26 implementations of the secp256k1 mul/square algorithm.


src/Arithmetic/DettmanMultiplication.v
```diff
@@ -14,26 +14,30 @@ Module DettmanMultiplication.
...
+        (wprops : @weight_properties weight)
+        
         (weight_0 := weight_0 wprops)
         (weight_positive := weight_positive wprops)
         (weight_multiples := weight_multiples wprops)
         (weight_divides := weight_divides wprops).
-    
-    Let c := Associational.eval c_.
 
+    Definition c := Positional.to_associational weight limbs
```
andres-erbsen, line 38: Does this algorithm have no useful flexibility from choosing different encodings of the same value for c?

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -93,132 +97,56 @@ Module DettmanMultiplication.
...
-      let b_assoc := Positional.to_associational weight limbs b in
-      let r0 := Associational.mul a_assoc b_assoc in
-      reduce_carry_borrow r0.
+    Hint Rewrite eval_reduce_one Associational.eval_carry eval_dedup_weights eval_borrow : push_eval.
 
-    Definition squaremod a :=
-      let a_assoc := Positional.to_associational weight limbs a in
-      let r0 := Associational.square a_assoc in
-      reduce_carry_borrow r0.
+    Definition carry_from_position i x :=
```
andres-erbsen, line 125: Did we really not have this one before? :D Might be good to share with other unsaturated code.

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -263,37 +191,113 @@ Module DettmanMultiplication.
...
+      - reflexivity.
+      - simpl. autorewrite with push_eval. assumption.
+    Qed.
+    Hint Rewrite eval_carry_reduce_chain : push_eval.
+
+    Local Open Scope nat_scope.
+
+    Definition seq_from_to a b := seq a (Z.to_nat (b - a + 1)).
+
+    Definition reduce_carry_borrow x :=
```
andres-erbsen, line 235: This code is clearly improved, and I still don't quite grasp the algorithm. Perhaps a couple of sentences of comment would do the trick now?

fiat-c/src/secp256k1_dettman_64.c
```diff
@@ -43,71 +44,75 @@ FIAT_SECP256K1_DETTMAN_FIAT_EXTENSION typedef unsigned __int128 fiat_secp256k1_d
 static FIAT_SECP256K1_DETTMAN_FIAT_INLINE void fiat_secp256k1_dettman_mul(uint64_t out1[5], const uint64_t arg1[5], const uint64_t arg2[5]) {
   fiat_secp256k1_dettman_uint128 x1;
   uint64_t x2;
-  fiat_secp256k1_dettman_uint128 x3;
-  uint64_t x4;
+  uint64_t x3;
```
andres-erbsen, line 47: Did you confirm that changes here are just due to redundant &, or are there actual changes as well?

andres-erbsen (approved):

src/Arithmetic/DettmanMultiplication.v
OwenConoly: We had [something](https://github.com/mit-plv/fiat-crypto/blob/fd26eec818c7533ac11530fedfb7fd5ccd979646/src/Arithmetic/Core.v#LL964C5-L968C57) very similar to this before.  The main difference is that the thing we had before uses a positional representation, and it relies on converting from associational to positional to combine terms, instead of using `dedup_weights`.
I didn't want to bother with converting from associational to positional and back here, since we have [some values](https://github.com/mit-plv/fiat-crypto/blob/964bce197be266ca3ed08f52b6d684529453d54b/src/Arithmetic/DettmanMultiplication.v#L244) that land at the weight `s`, and don't seem natural to encode in the positional representation.
So I don't think the other unsaturated code would be useful here, but do you think that this `carry_from_position` is general enough to be useful elsewhere?  Should I move it to `Core.v` perhaps?

src/Arithmetic/DettmanMultiplication.v
OwenConoly: I can't think of when that would be useful.  I believe we want to split `c` into pieces so that each piece fits in a limb (e.g., 26 bits for the 10x26 implementation).  But splitting it into more pieces than necessary would just make the reductions less efficient, right?

fiat-c/src/secp256k1_dettman_64.c
OwenConoly: I had not confirmed that, no.  Just now, I temporarily turned on the redundant-and removal on this branch, regenerated the dettman 64 code, and compared it with https://github.com/mit-plv/fiat-crypto/commit/fd26eec818c7533ac11530fedfb7fd5ccd979646.

There is only one thing causing the changes in secp256k1_dettman_64.c, other than the redundant & stuff.  For some reason, there are some changes of the form a + b --> b + a.  This seems to have something to do with the carries?  I'm not sure exactly what I did that led to that.  But I think it's not a bad thing.

src/Arithmetic/DettmanMultiplication.v
OwenConoly: I've added some comments!  Hopefully it's mostly graspable now.

src/Arithmetic/DettmanMultiplication.v
OwenConoly: I suppose we don't necessarily need each piece to fit into a limb; what I meant to say is that it's important to encode `c` like `[(2^(limbwidth*0), value_0); (2^(limbwidth*1), value_1); ... (2^(limbwidth*k), value_k)]`.

Otherwise, I believe that [`reduce_one`](https://github.com/mit-plv/fiat-crypto/blob/964bce197be266ca3ed08f52b6d684529453d54b/src/Arithmetic/Core.v#L222), when it's called in the multiplication algorithm, will return an associational representation with values at weird weights (i.e., weights that aren't of the form `weight i` for some `i`).  And I don't think that would be good.

And I don't think it's very likely that we'd could want, for instance, `value_0` to have some value that's greater than `2^(limbwidth*1)`.  So using the `simple_encode` function to get an encoding for `c` seems to be the natural choice.

OwenConoly:
The Generated C tests are currently failing with 
`cannot stat '/home/runner/work/fiat-crypto/fiat-crypto/fiat-c/src/curve25519_64_adx.c': No such file or directory`.
I don't know why it's expecting this file to exist.

JasonGross:
@davidben @andres-erbsen did BoringSSL recently add more variants?  We may need to extend the blacklist at https://github.com/mit-plv/fiat-crypto/blob/fd26eec818c7533ac11530fedfb7fd5ccd979646/etc/ci/test-fiat-c-boringssl.sh#L30-L32

JasonGross:
(maybe we should also add a better error message when the file doesn't exist?)

src/Arithmetic/DettmanMultiplication.v
andres-erbsen: Ok. Perhaps just add a comment that this is a heurisitc and not important for proofs.

src/Arithmetic/DettmanMultiplication.v
andres-erbsen: Wouldn't the positional representation also have weight `s`? In case you care about keeping all weights as-is, I agree the new `codem akes sense. I wonder whether `carry_from_position` could replace the old code.

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -232,26 +241,59 @@ Module DettmanMultiplication.
...
+    (* The main idea here is to have something like this:
+           Definition overly_simple_reduce_carry_borrow x :=
+                      carry_reduce_chain (seq_from_to 0 (l - 1)) x.
+       Below, we have something like this, with a few extra things added in.
+     *)
     Definition reduce_carry_borrow x :=
       let l := limbs in
       let x := dedup_weights x in
 
+      (* In the 'overly_simple_reduce_carry_borrow', the following code block
```
andres-erbsen, line 253: Do you think it would be helpful to generate and put in a comment the carry chains this code corresponds to on 32-bit and 64-bit instantiations as lists of indices, highlighting specially optimized carries with comments that can be matched to the source code?

andres-erbsen (commented):
Yep this is helpful. I am getting the flavor of the optimizations now. Feel free to address the comments I left or not.

src/Arithmetic/DettmanMultiplication.v
OwenConoly: Note that, for example, in the 5x52 algorithm, we have `weight i = (2^52)^i`, and `s = 2^256`, so there is no `i` s.t. `weight i = s`.
So, if we wanted to make this work with a positional representation, we would have to define a new function `weight'`, where
```
weight' i := if i <=? limbs then
  if i =? limbs then s else weight i
else
  weight (i - 1)
```
This just seems a bit awkward and unnecessarily confusing, though; I think it's nicer to just stay in the associational representation.

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -232,26 +241,59 @@ Module DettmanMultiplication.
...
       let x := borrow' (weight l) s x in
       let x := reduce' s (weight 0) x in
       let x := carry_from_position 0 x in
-      
+
+      (* This code block is what remains (after we've finished adding in all the
+         extra stuff) of the 'overly_simple_reduce_carry_borrow' referred to above.
+
+         The only clever part is that the last two carry_reduces have been replaced
+         with a carry_reduce_rw.
```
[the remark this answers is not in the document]
OwenConoly: I considered this earlier, but I don't think there is really a good reason to expect this in general?  The requirement for `carry_reduce_rw i x` to be sensible is that the weight `weight (limbs + i + 1)` should have value `0` in `x`.  And this wouldn't necessarily happen, in general, at the end of a `carry_reduce_chain`.  So I think it's more conceptually appealing to keep `carry_reduce_rw` separate from `carry_reduce_chain`.

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -232,26 +241,59 @@ Module DettmanMultiplication.
...
       let l := limbs in
       let x := dedup_weights x in
 
+      (* In the 'overly_simple_reduce_carry_borrow', the following code block
+         paragraph would belong in the position marked with an 'A' down below.
+         It turns out that doing that might make the output bounds too loose.
+
+         In particular, we need last_reduction to be small enough so that the final
+         carry chain (right below the 'A') does not overflow position (l - 1).
+         We solve this issue by putting the code block here (instad of at 'A').
```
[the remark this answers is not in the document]
OwenConoly: I don't think that's quite right, if I understand what you mean by pre-reduce.  The reductions in the first chain are the ones that land in the weights greater than the weight where the last reduction lands.  I just added a comment in the code, which I think should be more helpful than what I've written here.

src/Arithmetic/DettmanMultiplication.v
```diff
@@ -232,26 +241,59 @@ Module DettmanMultiplication.
...
+         last_reduction parameter to be as large as possible, since when it gets
+         smaller, the last carry chain gets longer (and everything else takes
+         the same amount of time).
+       *)
       let x := carry_reduce_chain (seq_from_to (last_reduction + 1) (l - 3)) x in
-      
       let x := carry_reduce_rw (l - 2) x in
 
-      let x := carry_from_position (l - 1) x in
+      (* This next code block is a reduction from position l to position 0---
```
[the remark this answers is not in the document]
OwenConoly: Yeah, that sounds right.  And factoring it into a function would be very nice; I'll do that.

src/Arithmetic/DettmanMultiplication.v
OwenConoly: That might be nice; not sure I'll get around to doing it in this PR though

src/Arithmetic/DettmanMultiplication.v
andres-erbsen: Ok I'm now on board with all this.

OwenConoly:
> Yes, I added some fiat-crypto code to Boringssl. Please change that test to explicitly list all files that should be patched and tested, or to be not sensitive to BoringSSL additions in some other way.

Is this good?
https://github.com/mit-plv/fiat-crypto/pull/1608/commits/5ef196152bada47ccb17167baa4530a4e77a49ad

JasonGross:
> > Yes, I added some fiat-crypto code to Boringssl. Please change that test to explicitly list all files that should be patched and tested, or to be not sensitive to BoringSSL additions in some other way.
> 
> Is this good? [5ef1961](https://github.com/mit-plv/fiat-crypto/commit/5ef196152bada47ccb17167baa4530a4e77a49ad)

Looks good to me

PR #1611, opened 2023-06-25
Update for Zig 0.11.0

jedisct1:
`@truncate()` and `@bitCast()` now do type inference, so the code can be simplified a bit.

I still really wish there was a way to make the distinction between casts that truncate, and casts that extend the types (and thus are always guaranteed to be safe).

### DIFF
--- a/fiat-zig/src/curve25519_32.zig
+++ b/fiat-zig/src/curve25519_32.zig
@@ -21,12 +21,11 @@ inline fn cast(comptime DestType: type, target: anytype) DestType {
         const dest = @typeInfo(DestType).Int;
         const source = @typeInfo(@TypeOf(target)).Int;
         if (dest.bits < source.bits) {
-            return @bitCast(DestType, @truncate(std.meta.Int(source.signedness, dest.bits), target));
-        } else {
-            return @bitCast(DestType, @as(std.meta.Int(source.signedness, dest.bits), target));
+            const T = std.meta.Int(source.signedness, dest.bits);
+            return @bitCast(@as(T, @truncate(target)));
         }
     }
-    return @as(DestType, target);
+    return target;
 }
 
 // The type LooseFieldElement is a field element with loose bounds.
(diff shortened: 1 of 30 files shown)


JasonGross:
Errors with
```
zig test Debug native: error: the following command failed with 10 compilation errors:
/opt/hostedtoolcache/zig/master/x64/zig test /home/runner/work/fiat-crypto/fiat-crypto/fiat-zig/src/main.zig --cache-dir /home/runner/work/fiat-crypto/fiat-crypto/fiat-zig/zig-cache --global-cache-dir /home/runner/.cache/zig --name test --listen=- 
Build Summary: 0/3 steps succeeded; 1 failed (disable with --summary none)
test transitive failure
+- run test transitive failure
   +- zig test Debug native 10 errors
src/secp256k1_montgomery_64.zig:30:20: error: expected 2 arguments, found 1
src/poly1305_64.zig:25:20: error: expected 2 arguments, found 1
src/poly1305_32.zig:25:20: error: expected 2 arguments, found 1
src/p384_64.zig:30:20: error: expected 2 arguments, found 1
src/p521_64.zig:25:20: error: expected 2 arguments, found 1
src/p448_solinas_64.zig:25:20: error: expected 2 arguments, found 1
src/p256_64.zig:30:20: error: expected 2 arguments, found 1
src/curve25519_64.zig:25:20: error: expected 2 arguments, found 1
src/curve25519_32.zig:25:20: error: expected 2 arguments, found 1
src/p256_32.zig:30:20: error: expected 2 arguments, found 1
```

> I still really wish there was a way to make the distinction between casts that truncate, and casts that extend the types (and thus are always guaranteed to be safe).

Yeah, really the problem here is that we're using casts both to track bounds info and for truncation, and ideally we'd split off the bounds analysis tracking and have the language-specific code insert whatever casts it needs based on bounds analysis, with the language-generic code only inserting truncating casts.  In theory we could already track this now by hand, if we use `Z.modulo` for truncating casts and `ident.cast` for non-truncating casts.

jedisct1:
Hi Jason!

> Errors with [...]

I didn't expect such a fast response :) This is to prepare for the next Zig snapshot, that will be uploaded in a couple hours.

PR #1623, opened 2023-07-22
Rust: Create newtype structs for typedefs

### LINKED ISSUE #1620: Types for `strict` and `loose` field elements should prevent misuse
Context is related to this issue in another Rust crate that uses fiat-crypto: https://github.com/crate-crypto/Ed448-Goldilocks/issues/28

In a nutshell: some code is using the p448 field (for implementing computations in the edwards448 elliptic curve). The `p448_solinas_64` fiat-crypto backend is used; in that backend, there are two internal representations of integers, called "tight" and "loose", that differ in the allowed range for the value limbs. In the generated Rust code, two distinct *type names* are defined:
https://github.com/mit-plv/fiat-crypto/blob/63a6ba42f418d2031342cf8e63d3e4428c88473d/fiat-rust/src/p448_solinas_64.rs#L23-L29

However, the two names are defined as type aliases to `[u64; 8]`, i.e. from the Rust compiler point of view, these are the same type. In particular, this means that the following code compiles without any error or warning:
```rust
pub fn bad_code(z: &mut fiat_p448_tight_field_element,
                x: &fiat_p448_tight_field_element,
                y: &fiat_p448_tight_field_element)
{
    fiat_p448_add(z, x, y);
}
```
`fiat_p448_add()` expects the output (first parameter) to be a "loose" value, and here I pass a reference to a variable with the "tight" type, and the compiler sees nothing wrong since these are the same type. The result is a "loose" value disguised as a "tight" value.

In practice,  a lot of the routines accept a larger range than what is formally proven correct, so much misuse can be absorbed with no exploitable consequence; this is the case with the Ed448-Goldilocks crate referenced above. But it still means that the formal proof of correctness is lost.

Such misuse would be easier to catch if the "tight" and "loose" types were actually distinct, and not two aliases on the same type. The Rust definition would look like this:
```rust
pub struct fiat_p448_loose_field_element(pub [u64; 8]);
pub struct fiat_p448_tight_field_element(pub [u64; 8]);
```
i.e. simple wrappers which have no actual extra runtime overhead. This would have some source-level impact, though:

  - References to limbs would have to be modified, e.g. `x[3]` becoming `x.0[3]`. I suppose this would be something to change in the code generator that produced the `p448_solinas_64.rs` file.
  - Callers who instantiate values directly (by providing the limb values) would need to use the appropriate syntax (e.g. `fiat_p448_tight_field_element([0u64; 8])` instead of `[0u64; 8]`). I expect such direct instantiation to be used for tables of precomputed constants (in an elliptic curve context, you would do that for precomputed small multiples of the base point, to speed up point multiplication by a scalar).

The second point means that the suggested change would break source compatibility with some external code that uses this crate.

Here I use the p448 field as example, but a cursory look shows that the same applies to other fields with tight/loose representation duality, and, arguably, also to fields with a Montgomery/non-Montgomery duality.

divergentdave:
This PR fixes #1620 by replacing type aliases for arrays with newtypes. As the types for Montgomery/non-Montgomery or tight/loose field elements are now distinct, this will prevent misuse by consumers. Since this is a breaking change, I bumped the crate version to 0.2.0.

I couldn't figure out how to make the suggested changes to stringification of `IR.List_nth` expressions, so I solved that part of the problem in Rust instead. Bare arrays, without typedefs, are still used in function arguments. Thus, any `IR.List_nth` operation would either need to be lowered to `({v}[{n}])` or `({v}.0[{n}])`, depending on the type of the operation's argument. I couldn't figure out how to access type information starting from these IR nodes, so I instead implemented `Index` and `IndexMut` on the new Rust structs, in order to forward indexing operations to the wrapped arrays.

FYI, I test-ran the Rust CI scripts locally, and I noted that in the curve25519-dalek and Ed448-Goldilocks tests, Cargo is ignoring the patched dependencies because of version number mismatches. (path = "../fiat-rust" has version 0.2.0, while the two tested crates depend on 0.1, so Cargo still uses the original crate from `crates.io`)

### DIFF
--- a/fiat-rust/Cargo.toml
+++ b/fiat-rust/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fiat-crypto"
-version = "0.1.21"
+version = "0.2.0"
 authors = ["Fiat Crypto library authors <jgross@mit.edu>"]
 edition = "2018"
 description = "Fiat-crypto generated Rust"
--- a/fiat-rust/src/curve25519_32.rs
+++ b/fiat-rust/src/curve25519_32.rs
@@ -22,11 +22,43 @@ pub type fiat_25519_i2 = i8;
 
 /* The type fiat_25519_loose_field_element is a field element with loose bounds. */
 /* Bounds: [[0x0 ~> 0xc000000], [0x0 ~> 0x6000000], [0x0 ~> 0xc000000], [0x0 ~> 0x6000000], [0x0 ~> 0xc000000], [0x0 ~> 0x6000000], [0x0 ~> 0xc000000], [0x0 ~> 0x6000000], [0x0 ~> 0xc000000], [0x0 ~> 0x6000000]] */
-pub type fiat_25519_loose_field_element = [u32; 10];
+#[derive(Clone, Copy)]
+pub struct fiat_25519_loose_field_element(pub [u32; 10]);
+
+impl std::ops::Index<usize> for fiat_25519_loose_field_element {
+    type Output = u32;
+    #[inline]
+    fn index(&self, index: usize) -> &Self::Output {
+        &self.0[index]
+    }
+}
+
+impl std::ops::IndexMut<usize> for fiat_25519_loose_field_element {
+    #[inline]
+    fn index_mut(&mut self, index: usize) -> &mut Self::Output {
+        &mut self.0[index]
+    }
+}
 
 /* The type fiat_25519_tight_field_element is a field element with tight bounds. */
 /* Bounds: [[0x0 ~> 0x4000000], [0x0 ~> 0x2000000], [0x0 ~> 0x4000000], [0x0 ~> 0x2000000], [0x0 ~> 0x4000000], [0x0 ~> 0x2000000], [0x0 ~> 0x4000000], [0x0 ~> 0x2000000], [0x0 ~> 0x4000000], [0x0 ~> 0x2000000]] */
-pub type fiat_25519_tight_field_element = [u32; 10];
+#[derive(Clone, Copy)]
+pub struct fiat_25519_tight_field_element(pub [u32; 10]);
+
+impl std::ops::Index<usize> for fiat_25519_tight_field_element {
+    type Output = u32;
+    #[inline]
+    fn index(&self, index: usize) -> &Self::Output {
+        &self.0[index]
+    }
+}
+
+impl std::ops::IndexMut<usize> for fiat_25519_tight_field_element {
+    #[inline]
+    fn index_mut(&mut self, index: usize) -> &mut Self::Output {
+        &mut self.0[index]
+    }
+}
 
 
 /// The function fiat_25519_addcarryx_u26 is an addition with carry.
--- a/fiat-rust/src/curve25519_64.rs
+++ b/fiat-rust/src/curve25519_64.rs
@@ -22,11 +22,43 @@ pub type fiat_25519_i2 = i8;
 
 /* The type fiat_25519_loose_field_element is a field element with loose bounds. */
 /* Bounds: [[0x0 ~> 0x18000000000000], [0x0 ~> 0x18000000000000], [0x0 ~> 0x18000000000000], [0x0 ~> 0x18000000000000], [0x0 ~> 0x18000000000000]] */
-pub type fiat_25519_loose_field_element = [u64; 5];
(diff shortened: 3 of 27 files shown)


andres-erbsen:
Thank you for the contribution. I will do my best to help this get merged, working with my limited knowledge of rust best practices.

The version and implementation changes look good. Indeed, the stringification IR is unfortunately untyped. Defining indexing on the wrapper struct seems fine even without the need for a workaround here, so I am happy to go with that. (If you'd prefer not to, I don't think the unknown-length case is actually used anywhere, so a quick alternative may be to make the type declaration for that error and change the pretty-printing of indexing in a way that only works for wrapped fixed-length arrays.)

Thanks for the heads-up about tests. I am not sure what the right thing to do for integration-testing breaking changes in Cargo packages is, but I think it would make sense to attempt to update the relevant crates and not leave fiat-rust untested. I guess the ideal scenario here would be to open PRs against the crates used in tests, perhaps following "[Prepublishing a breaking change](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#prepublishing-a-breaking-change)" with your feature branch. Then once the tests pass (on their CI or ours), this change could be merged with confidence, and downstream PRs could be reverted to refer to the usual fiat-crypto repository before merging. In the lucky case that downstreams don't need source-code changes, all this would be a little bit redundant, but probably still fine.

Do you think you could take a stab at this and expect success without too much effort? I am overall agnostic about whether fiat-crypto should be testing against reverse dependencies, so adding some appropriate tests in this repo could be an alternative.

divergentdave:
I've got the Ed448-Goldilocks changes up in a PR, and made a first draft of changes for curve25519-dalek, but the latter results in significant benchmark regressions. I'll have to look at profiling and assembly listings closely, and maybe change how I'm initializing and copying things.

divergentdave:
Good news: adding a `#[inline]` on `index()` and `index_mut()` took care of my benchmark regressions. I've updated the test scripts to run against PR branches of downstream dependencies for now, and they're passing locally.

etc/ci/test-fiat-rust-crate-crypto-Ed448-Goldilocks.sh
```diff
@@ -5,10 +5,13 @@ set -ex
...
 # Tests for crate-crypto/Ed448-Goldilocks
 ################################################################################
-git clone https://github.com/crate-crypto/Ed448-Goldilocks.git --branch=master Ed448-Goldilocks || exit $?
+git clone https://github.com/divergentdave/Ed448-Goldilocks.git --branch=fiat-crypto-0.2 Ed448-Goldilocks || exit $?
 
 pushd Ed448-Goldilocks >/dev/null || exit $?
+git checkout 671c75e9377b9cb64b14ce9a3749fde716f97d95 || exit $?
 
+mv Cargo.toml Cargo.toml.old
+head -n -3 Cargo.toml.old > Cargo.toml
```
JasonGross, line 14: These are temporary and should go away before this is merged, right?

divergentdave: That depends on what we want to do with these tests against reverse dependencies. If we make no change to these scripts, then they would only test the head of Ed448-Goldilocks and curve25519-dalek against fiat-crypto 0.1.20 from crates.io (see discussion above). With the changes I have here currently, they would test pinned commits of Ed448-Goldilocks and curve25519-dalek against the head of this repository. Once each downstream crate upgrades, its test script could be rolled back, in order to test the head of that repository against the head of this repository. The third option would be to rip out tests against the other repositories and add an independent set of tests here. (see @andres-erbsen's comment)

divergentdave: I removed the pinned commits, let me know what you think.

andres-erbsen:
@JasonGross can you bump https://crates.io/crates/fiat-crypto/? (How does this work?)

JasonGross:
There's a CI job that bumps it whenever we tag a release

JasonGross:
I'll tag one now

JasonGross:
I've tagged https://github.com/mit-plv/fiat-crypto/releases/tag/v0.0.21, https://github.com/mit-plv/fiat-crypto/actions/runs/6032879535/job/16368724510 is bumping the rust crate

PR #1625, opened 2023-07-27
add Alpine CI

andres-erbsen:


andres-erbsen:
[`ocamlfind: Package `coq-core.plugins.ltac' not found`](https://github.com/mit-plv/fiat-crypto/actions/runs/5675163948/job/15379985109?pr=1625#step:4:357) :(

andres-erbsen:
It's not feature-complete wrt the Ubuntu LTS CI, but we could say we want it anyway.

JasonGross:
@andres-erbsen I'm looking into what it would take to automatically publish standalone binaries on releases, and it seems like we'd probably want alpine for this.  Should we merge this?

andres-erbsen:
I don't have a vision here, i'd be happy if you adopted this one.

PR #1632, opened 2023-08-07
Debian CI

andres-erbsen:
https://github.com/mit-plv/fiat-crypto/pull/1631 will not support 8.15 but we want CI with maintained packages

### DIFF
--- a/.github/workflows/coq-debian.yml
+++ b/.github/workflows/coq-debian.yml
@@ -30,39 +30,22 @@ jobs:
     - uses: actions/checkout@v3
       with:
         submodules: recursive
-    - name: bootstrap
-      run: |
-        sudo mkdir /chroot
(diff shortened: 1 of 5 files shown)


.github/workflows/coq-debian.yml
```diff
@@ -0,0 +1,136 @@
...
+      group: ${{ github.workflow }}-${{ matrix.env.DEBIAN }}-${{ github.head_ref || github.run_id }}
+      cancel-in-progress: true
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: recursive
+    - name: bootstrap
+      run: |
+        sudo mkdir /chroot
```
JasonGross, line 35: Might be worth sticking this in a script in etc/ci/ or whatever so it can be reused in both this job and the test-amd64 one

PR #1633, opened 2023-08-12
update CI to Coq 8.16, and remove Ubuntu LTS because we now have Debian

samuelgruetter:


andres-erbsen:
:+1: please also bump the version requirement in README when you merge

JasonGross:
The motivation here is to drop 8.15 / Ubuntu LTS so that we can incur less compat overhead in bedrock2/rupicola?

samuelgruetter:
> The motivation here is to drop 8.15 / Ubuntu LTS so that we can incur less compat overhead in bedrock2/rupicola?

Yes, exactly

PR #1635, opened 2023-08-16
Pipe through carry_add function

bMacSwigg:
`carry_add` has tighter bounds on the results than the basic `add` function. The source already exists in PushButtonSynthesis. This defines & specs the function in Field.v, proves the correctness of the existing source relative to that spec, and then derives the Field25519 and Field1305 versions.

A later PR will replace the mul-by-2 in #1622 with `carry_add` instead.

### DIFF
--- a/src/Bedrock/Field/Synthesis/New/WordByWordMontgomery.v
+++ b/src/Bedrock/Field/Synthesis/New/WordByWordMontgomery.v
@@ -750,6 +750,7 @@ Definition field_parameters_prefixed
     M_pos a24
     (prefix ++ "mul")
     (prefix ++ "add")
+    (prefix ++ "carry_add_dontuse")
     (prefix ++ "sub")
     (prefix ++ "opp")
     (prefix ++ "square")
--- a/src/Bedrock/End2End/Poly1305/Field1305.v
+++ b/src/Bedrock/End2End/Poly1305/Field1305.v
@@ -78,6 +78,14 @@ Section Field.
          As fe1305_add_correct.
   Proof. Time derive_bedrock2_func add_op. Qed.
 
+  Derive fe1305_carry_add
+         SuchThat (forall functions,
+                      spec_of_BinOp bin_carry_add
+                        (field_representation:=field_representation n s c)
+                        (fe1305_carry_add :: functions))
+         As fe1305_carry_add_correct.
+  Proof. Time derive_bedrock2_func carry_add_op. Qed.
+
   Derive fe1305_sub
          SuchThat (forall functions,
                       spec_of_BinOp bin_sub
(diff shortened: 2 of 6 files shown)


bMacSwigg:
Hm. Tried to reproduce the failed CI checks locally by running `make src/Bedrock/Field/Synthesis/New/UnsaturatedSolinas.vo`. Hit a different error:

```
File "./src/Bedrock/Field/Common/Util.v", line 1038, characters 6-37:
Error: Expected a single focused goal but 3 goals are focused.
```

which I could reproduce on my main branch without these changes (see issue #1636).

JasonGross:
> Hm. Tried to reproduce the failed CI checks locally by running `make src/Bedrock/Field/Synthesis/New/UnsaturatedSolinas.vo`. Hit a different error:

Try `git submodule update --recursive`

src/Bedrock/Field/Synthesis/New/WordByWordMontgomery.v
```diff
@@ -758,6 +758,7 @@ Definition field_parameters_prefixed
     M_pos a24
     (prefix ++ "mul")
     (prefix ++ "add")
+    (prefix ++ "carry_add_dontuse")
```
JasonGross, line 761: Why "dontuse"?

andres-erbsen: I think this is because there isn't a particularly satisfying implementation of this function in montgomery domain. Having it be an alias for add would be fine, though. (We could consider calling this function "add", and the other one "add_limbwise" or "unreduced_add" or "unsafe_add"...)

bMacSwigg: Correct, there's no existing implementation; and in fact, the existing `add` function in Montgomery already guarantees `tight_bounds`, so it doesn't look like there would be any benefit to the `carry_add` function.

However, as an unfortunate consequence of the way FieldParameters is defined right now, any function in the spec must be defined (or "defined" by just providing an arbitrary string) everywhere that uses the spec, even if it's not needed.

(See for example `fe1305_scmul1_dontuse` in [RupicolaCrypto/Low.v](https://github.com/mit-plv/fiat-crypto/blob/63a6ba42f418d2031342cf8e63d3e4428c88473d/src/Bedrock/End2End/RupicolaCrypto/Low.v#L1333).)

bMacSwigg: What I said about Montgomery `add` guaranteeing `tight_bounds` isn't really accurate; I was misremembering. But it guarantees `montgomery_domain_bounds`, i.e. that the result is in the Montgomery domain, which I think is analogous?

https://github.com/mit-plv/fiat-crypto/blob/63a6ba42f418d2031342cf8e63d3e4428c88473d/src/PushButtonSynthesis/WordByWordMontgomery.v#L319-L326

bMacSwigg:
> > Hm. Tried to reproduce the failed CI checks locally by running `make src/Bedrock/Field/Synthesis/New/UnsaturatedSolinas.vo`. Hit a different error:
> 
> Try `git submodule update --recursive`?

That worked :) I was able to reproduce the failed CI and pushed a couple fixes. Re-running CI now :crossed_fingers:

PR #1642, opened 2023-08-25
Edwards point doubling for X25519

bMacSwigg:


src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -0,0 +1,519 @@
...
+  fe25519_mul(ox, cX, cT);
+  fe25519_mul(oy, cY, cZ);
+  fe25519_mul(oz, cZ, cT);
+  fe25519_mul(ot, cX, cY)
+}.
+
+Section WithParameters.
+  Context {two_lt_M: 2 < M_pos}.
+  (* TODO: Can we provide actual values/proofs for these, rather than just sticking them in the context? *)
+  Context {char_ge_3 : (@Ring.char_ge (F M_pos) Logic.eq F.zero F.one F.opp F.add F.sub F.mul (BinNat.N.succ_pos BinNat.N.two))}.
```
bMacSwigg, line 108: These are all things that are required/referenced by the context of Edwards.XYZT.Basic. It seems like these should be easily available from the FieldParameters etc (e.g. `a` and `d` seem to be parameters of the field, but FieldParameters only has a value for `(a-2)/4` right now).

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -0,0 +1,519 @@
...
+  fe25519_add(oz, D, C);
+  fe25519_sub(ot, D, C);
+  fe25519_mul(ox, ox, ot);
+  fe25519_mul(oy, oy, oz);
+  fe25519_mul(oz, ot, oz);
+  fe25519_mul(ot, ox, oy)
+}.
+
+(* Equivalent of m1double in src/Curves/Edwards/XYZT/Basic.v *)
+(* TODO: T is unused..? *)
```
bMacSwigg, line 78: Could probably rip this out (and change the spec to be `exists T,`) but I wasn't sure if we should, since the Gallina `m1double` takes a `point`.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -0,0 +1,519 @@
...
+  | x := _ : BinNums.Z |- _ => clear x
+  | x := _ : unit |- _ => clear x
+  | x := _ : bool |- _ => clear x
+  | x := _ : list _ |- _ => clear x
+  | x := _ : nat |- _ => clear x
+  | |- forall _, _ => intros
+  | |- let _ := _ in _ => intros
+  | |- dlet.dlet ?v (fun x => ?P) => change (let x := v in P); intros
+  | _ => progress (cbn [Semantics.interp_binop] in * )
+  | H: exists _, _ |- _ => destruct H; assert_fails (idtac; let __ := constr:(H) in idtac)
```
bMacSwigg, line 223: How do I properly make this change in the submodule? (For now I just made these local versions but obviously that's not the end state we want)

src/Bedrock/End2End/X25519/EdwardsXYZT.v
andres-erbsen: T is indeed unused in doubling. I don't think it's worth ripping it out yet as we might need to use a struct for the parameters anyway.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
andres-erbsen: I'm not sure what to do here, so let's keep them as is.

`d` and `a` and such are the parameters of the curve. however, the field implementation here is somewhat specialized to the curve being constructed, so I can see how it was expedient to have `(a+2)/4)` in the field parameters. And perhaps one day not that far in the future we'll blast FieldParameters into Context statements so we can mix and match as we please.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
bMacSwigg: This is now done in #1655 (and associated PRs to rupicola and bedrock2 submodules)

PR #1646, opened 2023-09-01
Rust: Reference newtype operators from `core` instead of `std`

### LINKED ISSUE #1645: Rust: `no_std` is broken due to use of `std::ops`
#1623 introduced newtype structs to prevent misuse, which was a great addition. But because it references `ops` from `std` instead of `core`, compiling with `--no-default-features` no longer works.

brycx:
closes #1645 

I have also added a Actions test to prevent this from happening in the future.


andres-erbsen:
Good catch, thank you!

PR #1648, opened 2023-09-04
Rust: Add no_std CI

pinkforest:
Adds CI for no_std on a target that does not have it

Prevents https://github.com/mit-plv/fiat-crypto/issues/1645

(The current one doesn't use no_std target and may miss it - this is the [same workflow we use in curve25519-dalek](https://github.com/dalek-cryptography/curve25519-dalek/blob/main/.github/workflows/curve25519-dalek.yml))

EDIT: The workflow curently in curve25519-dalek is [no_std.yml](https://github.com/dalek-cryptography/curve25519-dalek/blob/main/.github/workflows/no_std.yml)

(I will send further PR's to make the workflows compact + faster separately after this - I see some low hanging fruit :strawberry: )


.github/workflows/no_std.yml
```diff
@@ -0,0 +1,40 @@
...
+          - crate: fiat-crypto
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          targets: thumbv7em-none-eabi
+      - uses: taiki-e/install-action@cargo-hack
+      # No default features build
+      - name: no_std / no feat ${{ matrix.crate }}
```
JasonGross, line 36: Where does `matrix.crate` come from?

JasonGross (changes requested):
Could you add this as a separate job in the `rust.yml` file?  I'd rather have just one file for all the Rust tests

.github/workflows/no_std.yml
pinkforest: Line 27 - I can take matrix out given it's only one crate unlike in dalek we have multiple crates

pinkforest: I was hoping to isolate this until I renovate the workflow more but I will hack it into rust.yml

.github/workflows/rust.yml
```diff
@@ -18,12 +22,31 @@ jobs:
...
     - name: crate-crypto/Ed448-Goldilocks test
-      run: etc/ci/test-fiat-rust-crate-crypto-Ed448-Goldilocks.sh
+      run: ../etc/ci/test-fiat-rust-crate-crypto-Ed448-Goldilocks.sh
+
+  build-nostd:
+    name: Build on no_std target (thumbv7em-none-eabi)
+    runs-on: ubuntu-latest
+
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
JasonGross, line 38: I think we want
```suggestion
      group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}-${{ github.job }}
```
on both this and the one above, or else the jobs are going to cancel each other

.github/workflows/no_std.yml
JasonGross, line 1: This file should be removed now, right?

pinkforest: Yeah I took it out

.github/workflows/rust.yml
pinkforest: Fixed

.github/workflows/rust.yml
```diff
@@ -6,24 +6,47 @@ on:
...
-      run: (cd fiat-rust; cargo build --verbose)
+      run: cargo build --verbose
     - name: cargo test
-      run: (cd fiat-rust; cargo test --verbose)
-    - name: cargo test (no_std)
-      run: (cd fiat-rust; cargo test --no-default-features --verbose)
+      run: cargo test --verbose
     - name: curve25519-dalek test
-      run: etc/ci/test-fiat-rust-curve25519-dalek.sh
+      run: (cd .. ; etc/ci/test-fiat-rust-curve25519-dalek.sh)
```
JasonGross, line 29: Could you instead make the scripts robust to cwd changes by sticking in a line like
https://github.com/mit-plv/fiat-crypto/blob/dba0b22ee3f7f8dac8050ae04a727d6e085a1c92/etc/ci/github-actions-make.sh#L4
in the script itself?

pinkforest: I can do that in a follow-up PR where I will remove those shellscripts together and it will be natively github action

JasonGross:
> this is the [same workflow we use in curve25519-dalek](https://github.com/dalek-cryptography/curve25519-dalek/blob/main/.github/workflows/curve25519-dalek.yml)

I don't see the `taiki-e/install-action@cargo-hack` in that link, what's up with this?

pinkforest:
Here: https://github.com/dalek-cryptography/curve25519-dalek/blob/main/.github/workflows/no_std.yml

I forgot I had a PR which isn't yet in main: https://github.com/dalek-cryptography/curve25519-dalek/pull/572/files#diff-92d698a5ce32a269344d647e18fb6501bf6a58b69ea93cf4a80fd5c9c12efc21 :sweat_smile:

.github/workflows/rust.yml
JasonGross: Okay, sounds good.  In that case I think you'll need to fix the `cd ..; ../etc/...` that's a couple lines down though

.github/workflows/rust.yml
```diff
@@ -6,24 +6,47 @@ on:
...
     - name: crate-crypto/Ed448-Goldilocks test
-      run: etc/ci/test-fiat-rust-crate-crypto-Ed448-Goldilocks.sh
+      run: (cd .. ; ../etc/ci/test-fiat-rust-crate-crypto-Ed448-Goldilocks.sh)
```
JasonGross, line 31:
```suggestion
      run: (cd .. ; etc/ci/test-fiat-rust-crate-crypto-Ed448-Goldilocks.sh)
```

.github/workflows/rust.yml
pinkforest: Oh thanks fixed :purple_heart:

.github/workflows/rust.yml
pinkforest: Oops fixed

.github/workflows/rust.yml
```diff
@@ -6,24 +6,47 @@ on:
...
+    working-directory: fiat-rust
+
 jobs:
   test-rust:
 
     runs-on: ubuntu-latest
 
     concurrency:
-      group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+      group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}-${{ github.job }}
```
JasonGross, line 19: Hmm, this doesn't work the way I hoped, and still [cancels the job](https://github.com/mit-plv/fiat-crypto/actions/runs/6076805923/job/16485443157?pr=1648)
Let's try
```suggestion
      group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}-${{ job.name }}
```
Or else just manually disambiguate them?

.github/workflows/rust.yml
```diff
@@ -6,24 +6,47 @@ on:
...
     - name: crate-crypto/Ed448-Goldilocks test
-      run: etc/ci/test-fiat-rust-crate-crypto-Ed448-Goldilocks.sh
+      run: (cd .. ; ../etc/ci/test-fiat-rust-crate-crypto-Ed448-Goldilocks.sh)
+
+  build-nostd:
+    name: Build on no_std target (thumbv7em-none-eabi)
+    runs-on: ubuntu-latest
+
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}-${{ github.job }}
```
JasonGross:
```suggestion
      group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}-${{ job.name }}
```
I can try adding this change from the web interface

.github/workflows/rust.yml
pinkforest: I think it needs ot be at root level

.github/workflows/rust.yml
```diff
@@ -6,24 +6,43 @@ on:
   schedule:
     - cron: '0 0 1 * *'
 
+defaults:
+  run:
+    working-directory: fiat-rust
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}-${{ github.job }}
+    cancel-in-progress: true
```
JasonGross, line 15: "Invalid workflow file: .github/workflows/rust.yml#L15
You have an error in your yaml syntax on line 15"
I think the indentation is inconsistent

JasonGross:
```suggestion
  cancel-in-progress: true
```

.github/workflows/rust.yml
```diff
@@ -6,24 +6,43 @@ on:
...
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          targets: thumbv7em-none-eabi
+      - uses: taiki-e/install-action@cargo-hack
+      # No default features build
+      - name: no_std / no feat fiat-crypto
+        run: cargo build -p fiat-crypto --target thumbv7em-none-eabi --release --no-default-features
+      - name: no_std / cargo hack fiat-crypto
+        run: cargo hack build -p fiat-crypto --target thumbv7em-none-eabi --release --each-feature --exclude-features default,std
```
JasonGross, line 48: Is this supposed to complete in 0.00s?
Run cargo build -p fiat-crypto --target thumbv7em-none-eabi --release --no-default-features
   Compiling fiat-crypto v0.2.2 (/home/runner/work/fiat-crypto/fiat-crypto/fiat-rust)
    Finished release [optimized] target(s) in 6.75s
0s
Run cargo hack build -p fiat-crypto --target thumbv7em-none-eabi --release --each-feature --exclude-features default,std
  cargo hack build -p fiat-crypto --target thumbv7em-none-eabi --release --each-feature --exclude-features default,std
  shell: /usr/bin/bash -e {0}
  env:
    CARGO_INCREMENTAL: 0
    CARGO_TERM_COLOR: always
info: running `cargo build --release --target thumbv7em-none-eabi --no-default-features` on fiat-crypto (1/1)
    Finished release [optimized] target(s) in 0.00s

JasonGross (approved):
Thank you!

.github/workflows/rust.yml
pinkforest: Yes because it only builds it and it uses incremental compilation

PR #1671, opened 2023-09-23
rust: Annotates literals with their type.

armfazh:
Cargo clippy reports unnecessary literal casting, instead do annotate every literal with its type. This change does not affect any static cast.

[1] https://rust-lang.github.io/rust-clippy/master/index.html#/unnecessary_cast


JasonGross:
Please update the autogenerated files (you should be able to download them from CI on the summary page [here for generated-files-bookworm](https://github.com/mit-plv/fiat-crypto/suites/16475947928/artifacts/941821710), or you can run `make`)

JasonGross (changes requested):
Please update the autogenerated files (you should be able to download them from CI on the summary page [here for generated-files-bookworm](https://github.com/mit-plv/fiat-crypto/suites/16475947928/artifacts/941821710), or you can run `make`)

armfazh:
It looks like one additonal cast must be added in the intermediate representation:

For example:
```rust
fiat_p256_subborrowx_u64(&mut x104, &mut x105, x103, x96, 0xffffffff_u32);
```

With the current patch, every literal has its own type annotation (which corresponds to the bitsize of the value). However, there should be another cast to `u64` since it will be used to invoke the `fiat_p256_subborrowx_u64` function.

Previously, there wasn't an issue as the rust compiler automatically converts these constants.

I couldn't find where to include this additional cast (note that for C, all constants are casted to u64, when they used as parameters of a function). It seems that this cast must come from the intermediate representation tree. @JasonGross , could you please look into this issue?

JasonGross:
The cast for the third argument should be inserted at
https://github.com/mit-plv/fiat-crypto/blob/37dd5afc21a11fd476ba0822169c4892c00e408b/src/Stringification/IR.v#L1353
https://github.com/mit-plv/fiat-crypto/blob/37dd5afc21a11fd476ba0822169c4892c00e408b/src/Stringification/IR.v#L1040-L1044
C says that this is `false`, though:
https://github.com/mit-plv/fiat-crypto/blob/37dd5afc21a11fd476ba0822169c4892c00e408b/src/Stringification/C.v#L557-L558
while Rust says `true`
https://github.com/mit-plv/fiat-crypto/blob/37dd5afc21a11fd476ba0822169c4892c00e408b/src/Stringification/Rust.v#L382-L383
so I don't see how the casts are getting there in the C code.  But continuing on, we have
https://github.com/mit-plv/fiat-crypto/blob/37dd5afc21a11fd476ba0822169c4892c00e408b/src/Stringification/IR.v#L566-L590
https://github.com/mit-plv/fiat-crypto/blob/37dd5afc21a11fd476ba0822169c4892c00e408b/src/Stringification/IR.v#L462-L478
I expect `Zcast` to give us
https://github.com/mit-plv/fiat-crypto/blob/37dd5afc21a11fd476ba0822169c4892c00e408b/src/Stringification/IR.v#L471
so we should be getting a cast at
https://github.com/mit-plv/fiat-crypto/blob/37dd5afc21a11fd476ba0822169c4892c00e408b/src/Stringification/IR.v#L473

I don't have time to dig into this more at the moment, but if you're down to try a bit more adventurous debugging, there are a bunch of debugging examples in `SlowPrimeSynthesisExamples`, if you pick one that's reasonably close to what you want, change a line like
https://github.com/mit-plv/fiat-crypto/blob/37dd5afc21a11fd476ba0822169c4892c00e408b/src/SlowPrimeSynthesisExamples.v#L422
to use Rust rather than C, then you can investigate what's going on.  You can either try to do controlled reduction like many of the examples in there do, or you can install https://github.com/coq-community/reduction-effects and do printf-debugging with `lazy` or `cbv` by inserting printing effects.

PR #1711, opened 2023-11-09
Upload standalone binaries to release pages

JasonGross:
There's also some minimal testing to see that the binaries work without installing all the extra Coq infrastructure.

Presumably we should get #1625 working and upload only statically-linked elf binaries (+ mac & windows), not separate binaries for docker, debian, etc.  Also TODO: figuring out if we need to do extra things to get Windows binaries working (probably?).

On top of #1710


JasonGross:
Only dumpbin fails, I'm going to merge this and then fix it immediately after, I don't want to bother with another round of CI.

PR #1712, opened 2023-11-10
Model coq-alpine after coq-debian

JasonGross:


JasonGross:
@andres-erbsen alpine is missing bash?  What should we do about
```
etc/ci/describe-system-config.sh
  shell: /home/runner/rootfs/alpine-edge-x86_64/abin/alpine.sh {0}
env: can't execute 'bash': No such file or directory
```
?

PR #1713, opened 2023-11-10
Adapt to coq/coq#18280 (case relevance outside case info)

SkySkimmer:


JasonGross (approved):

JasonGross:
This will be backwards-compatible, right?

SkySkimmer:
should be

PR #1718, opened 2023-11-12
Statically link alpine binaries: `-ccopt -static`

JasonGross:


JasonGross:
This builds except on Windows, so I'm going to merge it without re-running the CI

PR #1719, opened 2023-11-12
remove RupicolaCrypto.Low

andres-erbsen:
This file has admits and hasn't had a non-maintenance change since almost a year ago.


DIJamner (approved):
I checked my branch and can confirm this file is safe to delete.

PR #1721, opened 2023-11-12
Only upload one copy of linux binaries to releases

JasonGross:
We upload the statically-linked Alpine edge binaries, as these are
probably the most portable.

Note that the statically-linked Alpine binaries are actually smaller
than the Debian executables (~35 MB archive vs ~41 MB archive), though
the Debian ones are pie (position-independent executables, supporting
Address Space Layout Randomization).

Current names file names:
fiat-crypto-windows.zip
fiat-crypto-macos.tar.gz
fiat-crypto-linux-x86_64.tar.gz

Suggestions welcome cc @andres-erbsen

### DIFF
--- a/.github/workflows/coq-macos.yml
+++ b/.github/workflows/coq-macos.yml
@@ -147,6 +147,9 @@ jobs:
       contents: write  # IMPORTANT: mandatory for making GitHub Releases
     steps:
     - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0 # Fetch all history for all tags and branches
+        tags: true     # Fetch all tags as well, `fetch-depth: 0` might be sufficient depending on Git version
     - name: Download standalone MacOS
       uses: actions/download-artifact@v3
       with:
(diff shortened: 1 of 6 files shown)


.github/workflows/coq-macos.yml
```diff
@@ -157,7 +160,12 @@ jobs:
...
         mkdir dist
-        mv dist-macos/standalone.tar.gz dist/fiat-crypto-macos.tar.gz
+        ( cd dist-macos && tar -xzvf standalone.tar.gz )
+        arch="$(etc/ci/find-arch.sh dist-macos/word_by_word_montgomery)"
+        tag="$(git describe --tags HEAD)"
+        fname="Fiat-Cryptography_${tag}_macOS_${arch}.tar.gz"
+        echo "$fname"
+        mv dist-macos/standalone.tar.gz "dist/$fname"
         find dist
         tar -tvf dist/fiat-crypto-macos.tar.gz
```
JasonGross, line 170: Mac fails because I forgot to update this line, everything else passes

PR #1733, opened 2023-11-15
Test install target on CI (docker only)

### LINKED ISSUE #1732: Test install target on CI
We should test the install target on CI to avoid issues like https://github.com/coq/coq/pull/18191#issuecomment-1811982115

JasonGross:
We match the various opam install targets.

Non-docker targets in #1734 

Fixes #1732


JasonGross:
Now blocked only on https://github.com/mit-plv/bedrock2/issues/388 I believe

PR #1738, opened 2023-11-19
Adapt wrt Coq/Coq#18164

Villetaneuse:
These changes should be mostly backwards compatible (at least down to 8.16).
PS: the sp2019 target is intended, I'm trying to pass "fiat_crypto_legacy" in Coq's CI.


JasonGross (approved):
Thank you!

JasonGross:
@Villetaneuse is there a similar patch that applies to fiat-crypto master?

PR #1743, opened 2023-11-20
bump rupicola

samuelgruetter:
Two changes are needed because coqutil's Bitwidth instances are now `#[export]` instead of `#[global]`.


samuelgruetter:
(this PR is deliberately out-of-date with the master branch because there have been CI failures on the master branch, so I based it on the [last commit with a green tick](https://github.com/mit-plv/fiat-crypto/commit/2f79c57219f3417137a80e55419e0f079d180478))

PR #1755, opened 2023-11-27
Use objects instead of arrays in coq-opam-package.yml

dra27:
Addresses the specific complaint in https://github.com/ocaml/setup-ocaml/issues/720. A run of this workflow can be seen [on my fork](https://github.com/dra27/fiat-crypto/actions/runs/7003071299). In particular, the "Set up OCaml" step has this for the three Windows jobs:

```yaml
Run ocaml/setup-ocaml@v2
  with:
    ocaml-compiler: 4.11.1
    opam-repositories: opam-repository-mingw: https://github.com/ocaml-opam/opam-repository-mingw.git#sunset
  default: https://github.com/ocaml/opam-repository.git
```
(both `opam-repository-mingw` and `default` for `opam-repositories`) but the 6 macOS and Ubuntu jobs have:

```yaml
Run ocaml/setup-ocaml@v2
  with:
    ocaml-compiler: 4.11.1
    opam-repositories: 
  default: https://github.com/ocaml/opam-repository.git
```

Likewise, the step "Run `opam install coq-fiat-crypto --with-test` visibly still has `-j 1` appended on the Windows jobs and similarly the `COQEXTRAFLAGS` environment variable is only set on Windows.

### DIFF
--- a/.github/workflows/coq-opam-package.yml
+++ b/.github/workflows/coq-opam-package.yml
@@ -18,34 +18,23 @@ jobs:
       matrix:
         coq-version: ['dev', '8.18.0', '8.17.0']
         ocaml-compiler: ['4.11.1']
-        os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
-        extra-ocaml-repositories: ['', 'opam-repository-mingw: https://github.com/ocaml-opam/opam-repository-mingw.git#sunset']
-        coq-extra-flags: ['-async-proofs-j 1', '']
-        opam-jobs-flag: ['-j 1', '']
-        exclude:
-        - os: 'ubuntu-latest'
-          coq-extra-flags: '-async-proofs-j 1'
-        - os: 'macos-latest'
-          coq-extra-flags: '-async-proofs-j 1'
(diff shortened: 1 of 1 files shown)


.github/workflows/coq-opam-package.yml
```diff
@@ -55,8 +44,8 @@ jobs:
       uses: ocaml/setup-ocaml@v2
       with:
         ocaml-compiler: ${{ matrix.ocaml-compiler }}
-        opam-repositories: |
-          ${{ matrix.extra-ocaml-repositories }}
+        opam-repositories: |-
```
JasonGross, line 47: What does using `|-` rather then `|` do?

JasonGross (approved):
I did not realize this was a possibility, thank you!

.github/workflows/coq-opam-package.yml
dra27: It’s cosmetic - it chomps the unnecessary new line at the end of the field (which was showing up in the logs when I checked the parameter was propagating correctly)

PR #1793, opened 2023-12-23
[ci] [alpine] Use --root instead of sudo

JasonGross:


JasonGross:
Does not work due to path issues

PR #1798, opened 2023-12-25
Unify Coq CI into a single .yml file

JasonGross:
This shares structure a lot more, and will also enable us to (eventually) just check if the generated OCaml code is identical across Coq versions and platforms.

Note that we no longer check generated files on most platforms, and we only validate on Coq master.


JasonGross:
Hm, probably it would be better to use [reusable workflows](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputs) rather than blocking future steps across platforms

andres-erbsen:
I think branching on the platform inside a larger script isn't great. Reusable workflows might work, but I am not familiar with them.

If we didn't have CI-specific workarounds, I think I would like to have a deps script for each platform and then source the right one. Unfortunately it seems that most of the files in this diff are full of stuff that is really specific to oddities of the proprietary CI infrastructure and I am not sure what to make of it.

PR #1804, opened 2023-12-31
Fix nested unordered lists in README.md

divergentdave:
This changes unordered lists at the third nesting level to use `*` instead of `.` as bullets, so that they are properly formatted by Markdown.

(same idea as mit-plv/rewriter#147)


JasonGross (approved):
Thank you!

PR #1806, opened 2024-01-01
[CI] [ArchLinux] Update keys

JasonGross:
To avoid "Signature is unknown trust"


JasonGross:
I think the packages just need updating

PR #1807, opened 2024-01-09
Adapt to Coq's PR #18445 fixing inheritance of multiple signatures of implicit arguments in notations

herbelin:
With coq/coq#18445, notations hiding constants with multiple signatures of implicit arguments inherit as expected the underlying implicit arguments.

File `src/Compilers/Named/ContextProperties/Proper.v` was depending on the "bug". We adapt it by changing the notation so that it becomes indifferent to potential insertion of maximal implicit arguments.

Note that this is one possible fix among others (for instance, it is probably also possible to change the signatures of the underlying constant).

It is (a priori) backwards compatible and can be merged as soon as now.


JasonGross:
Thank you!

PR #1823, opened 2024-02-27
Joye double-add ladder for short Weierstrass curves in co-Z arithmetic

atrieu:
Not sure if this is of interest for inclusion, feel free to close if not. 
I have proved correctness of an implementation of Joye's double-add ladder for short Weierstrass curves in co-Z arithmetic (basically algorithm 14 from ["Scalar Multiplication on Weierstraß Elliptic Curves from Co-Z Arithmetic", Goundar et al., Journal of Cryptographic Engineering 2011](https://marcjoye.github.io/papers/GJMRV11regpm.pdf)) to study on Fiat-Crypto a bit.

Just for my own interest, what would be the easiest way to synthesize some C code out of it?

### DIFF
--- a/src/Curves/Weierstrass/Jacobian/CoZ.v
+++ b/src/Curves/Weierstrass/Jacobian/CoZ.v
@@ -467,8 +467,8 @@ Module Jacobian.
       co_z P (opp P).
     Proof. unfold co_z; faster_t. Qed.
 
-    Definition make_co_z_inner (P Q : F*F*F) : (F*F*F) * (F*F*F) :=
-      match P, Q with
+    Program Definition make_co_z (P Q : point) (HQaff : z_of Q = 1) : point*point :=
+      match proj1_sig P, proj1_sig Q return (F*F*F)*(F*F*F) with
       | (x1, y1, z1), (x2, y2, z2) =>
           let t1 := x2 in
           let t2 := y2 in
@@ -479,29 +479,25 @@ Module Jacobian.
           let t2 := t3 * t2 in
           (P, (t1, t2, t3))
       end.
-
-    Hint Unfold is_point co_z make_co_z_inner : points_as_coordinates.
-
-    Program Definition make_co_z (P Q : point) (HQaff : z_of Q = 1) : (point * point) :=
-      make_co_z_inner (proj1_sig P) (proj1_sig Q).
     Next Obligation. Proof. faster_t. Qed.
     Next Obligation. Proof. faster_t. Qed.
 
-    Hint Unfold make_co_z : points_as_coordinates.
+    Hint Unfold is_point co_z make_co_z : points_as_coordinates.
 
     Lemma make_co_z_correct (P Q : point) (HQaff : z_of Q = 1)
       (HPnz : z_of P <> 0) :
-      eq P (fst (make_co_z P Q HQaff))
-      /\ eq Q (snd (make_co_z P Q HQaff))
-      /\ co_z (fst (make_co_z P Q HQaff)) (snd (make_co_z P Q HQaff)).
+      let '(A, B) := make_co_z P Q HQaff in
+      eq P A
+      /\ eq Q B
+      /\ co_z A B.
     Proof. faster_t. Qed.
 
     (* Scalar Multiplication on Weierstraß Elliptic Curves from Co-Z Arithmetic *)
     (* Goundar, Joye, Miyaji, Rivain, Vanelli *)
     (* Algorithm 19 Co-Z addition with update (register allocation) *)
     (* 6 field registers, 5M, 2S *)
-    Definition zaddu_inner (P Q : F*F*F) : (F*F*F)*(F*F*F) :=
-      match P, Q with
+    Program Definition zaddu (P Q : point) (H : co_z P Q) : point * point :=
+      match proj1_sig P, proj1_sig Q return (F*F*F)*(F*F*F) with
       | (x1, y1, z1), (x2, y2, z2) =>
           let t1 := x1 in
           let t2 := y1 in
@@ -524,11 +520,6 @@ Module Jacobian.
           let t5 := t5 - t2 in
           ((t4, t5, t3), (t1, t2, t3))
       end.
-
-    Hint Unfold zaddu_inner : points_as_coordinates.
-
-    Program Definition zaddu (P Q : point) (H : co_z P Q) : point * point :=
-      zaddu_inner (proj1_sig P) (proj1_sig Q).
     Next Obligation. Proof. faster_t_noclear. Qed.
     Next Obligation. Proof. faster_t. Qed.
 
@@ -561,8 +552,8 @@ Module Jacobian.
     (* Goundar, Joye, Miyaji, Rivain, Vanelli *)
     (* Algorithm 20 Conjugate co-Z addition (register allocation) *)
     (* 7 field registers, 6M, 3S*)
-    Definition zaddc_inner (P Q : F*F*F) : (F*F*F) * (F*F*F) :=
-      match P, Q with
+    Program Definition zaddc (P Q : point) (H : co_z P Q) : point * point :=
+      match proj1_sig P, proj1_sig Q return (F*F*F)*(F*F*F) with
       | (x1, y1, z1), (x2, y2, z2) =>
           let t1 := x1 in
           let t2 := y1 in
@@ -596,12 +587,6 @@ Module Jacobian.
           let t2 := t2 + t6 in
           ((t1, t2, t3), (t4, t5, t3))
       end.
-
-    Hint Unfold zaddc_inner : points_as_coordinates.
-
-    Program Definition zaddc (P Q : point)
-      (H : co_z P Q) : point * point :=
-      zaddc_inner (proj1_sig P) (proj1_sig Q).
     Next Obligation. Proof. faster_t_noclear. Qed.
     Next Obligation. Proof. faster_t_noclear. Qed.
 
@@ -633,8 +618,8 @@ Module Jacobian.
     (* Goundar, Joye, Miyaji, Rivain, Vanelli *)
     (* Algorithm 21 Co-Z doubling with update (register allocation) *)
     (* 6 field registers, 1M, 5S *)
-    Definition dblu_inner (P : F*F*F) : (F*F*F) * (F*F*F) :=
-      match P with
+    Program Definition dblu (P : point) (H: z_of P = 1) : point * point :=
+      match proj1_sig P return (F*F*F)*(F*F*F) with
       | (x1, y1, _) =>
           let t0 :=  a in
           let t1 := x1 in
@@ -660,12 +645,6 @@ Module Jacobian.
           let t5 := t5 - t2 in
           ((t4, t5, t3), (t1, t2, t3))
       end.
-
-    Hint Unfold dblu_inner : points_as_coordinates.
-
-    Program Definition dblu (P : point)
-      (H: z_of P = 1) : point * point :=
-      dblu_inner (proj1_sig P).
     Next Obligation. Proof. faster_t. Qed.
     Next Obligation. Proof. faster_t. Qed.
 
@@ -699,25 +678,11 @@ Module Jacobian.
     (* Goundar, Joye, Miyaji, Rivain, Vanelli *)
     (* Algorithm 22 Co-Z tripling with update (register allocation) *)
     (* 6M, 7S *)
-    Definition tplu_inner (P : F*F*F) : (F*F*F)*(F*F*F) :=
-      zaddu_inner (snd (dblu_inner P)) (fst (dblu_inner P)).
-
-    Hint Unfold tplu_inner : points_as_coordinates.
-
     Program Definition tplu (P : point) (H : z_of P = 1) : point * point :=
-      tplu_inner (proj1_sig P).
-    Next Obligation. Proof. faster_t. Qed.
-    Next Obligation. Proof. faster_t. Qed.
-
-    Program Definition tplu2 (P : point) (H : z_of P = 1) : point * point :=
       zaddu (snd (dblu P H)) (fst (dblu P H)) _.
     Next Obligation. faster_t. Qed.
 
-    Hint Unfold tplu tplu2 : points_as_coordinates.
-
-    Lemma tplu_tplu2 (P : point) (H : z_of P = 1) :
-      eq (fst (tplu P H)) (fst (tplu2 P H)) /\ eq (snd (tplu P H)) (snd (tplu2 P H)).
-    Proof. faster_t. Qed.
+    Hint Unfold tplu : points_as_coordinates.
 
     Lemma tplu_correct0 (P : point) (H : z_of P = 1) (Hyz : y_of P = 0) :
       let '(R1, R2) := tplu P H in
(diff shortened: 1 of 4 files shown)


andres-erbsen:
Yes this looks like something we'd be very happy to have; I'll try to find time to review in detail this week or maybe next.

src/Curves/Weierstrass/Jacobian/ScalarMult.v
```diff
@@ -0,0 +1,965 @@
...
+              {Hn : (2 <= n < 2^scalarbitsz)%Z}
+              {Hscalarbitsz : (2 <= scalarbitsz)%Z}
+              {P : Wpoint} {HPnz : P <> ∞ :> Wpoint}
+              {ordP : Z} {HordPpos : (2 < ordP)%Z}
+              {HordPodd : Z.odd ordP = true :> bool}
+              {HordP : forall l, (l <> 0 :> Z)%Z -> ((Weq (scalarmult l P) ∞) <-> exists k, (l = k * ordP :> Z)%Z)}
+              {HordPn : (n + 2 < ordP)%Z}.
+
+      Local Notation testbitn := (Z.testbit n).
+      Local Notation n' := (if testbitn 0 then n else (n + 1)%Z).
```
andres-erbsen, line 481: `Z.setbit`?

src/Curves/Weierstrass/Jacobian/ScalarMult.v
```diff
@@ -0,0 +1,965 @@
...
+      SS (Z.to_nat scalarbitsz - 1) = n :> Z.
+    Proof.
+      rewrite SS_decomposition.
+      replace (Z.of_nat (S (Z.to_nat scalarbitsz - 1)))%Z with scalarbitsz by lia.
+      apply Ones.Z.land_ones_low_alt. split; lia.
+    Qed.
+
+    Lemma SS_plus_TT (k: nat) :
+      (SS k + TT k)%Z = (2^(Z.of_nat (S k)))%Z :> Z.
+    Proof.
```
andres-erbsen, line 146: Might these two properties be desirable as definitions of SS and TT, maybe using modulo instead of land?

src/Curves/Weierstrass/Jacobian/CoZ.v
```diff
@@ -0,0 +1,915 @@
...
+
+    Definition co_z (P Q : point) : Prop :=
+      z_of P = z_of Q.
+
+    Lemma opp_co_z (P : point) :
+      co_z P (opp P).
+    Proof. unfold co_z; faster_t. Qed.
+
+    Definition make_co_z_inner (P Q : F*F*F) : (F*F*F) * (F*F*F) :=
+      match P, Q with
```
andres-erbsen, line 471: Was there a coq-technical issue with using sigma types for points? I see you have the non-dependently-typed bodies as separate definitions. Previously, I had found specifying `return F*F*F` or such sufficient for matches within `Program Definition` so that the proof gets attached on the very outside. Or do you think having the types propagate on-curve invariants through the loop is just not helpful enough to be worth it?

src/Curves/Weierstrass/Jacobian/CoZ.v
```diff
@@ -0,0 +1,915 @@
...
+      try solve [ lazymatch goal with
+                  | [ |- _ = _ ] => clear_neq
+                  | _ => idtac
+                  end;
+                  fsatz ].
+
+    Hint Unfold Jacobian.double negb andb : points_as_coordinates.
+    Hint Unfold W.eq W.add Jacobian.to_affine Jacobian.of_affine Jacobian.add Jacobian.add_impl Jacobian.opp : points_as_coordinates.
+
+    (* These could go into Jacobian.v *)
```
andres-erbsen, line 411: Might it be preferable to prove that `point` is a commutative group by isomorphism to affine and then use lemmas about that?

src/Curves/Weierstrass/Jacobian/CoZ.v
```diff
@@ -0,0 +1,915 @@
...
+    Next Obligation. Proof. faster_t. Qed.
+    Next Obligation. Proof. faster_t. Qed.
+
+    Hint Unfold make_co_z : points_as_coordinates.
+
+    Lemma make_co_z_correct (P Q : point) (HQaff : z_of Q = 1)
+      (HPnz : z_of P <> 0) :
+      eq P (fst (make_co_z P Q HQaff))
+      /\ eq Q (snd (make_co_z P Q HQaff))
+      /\ co_z (fst (make_co_z P Q HQaff)) (snd (make_co_z P Q HQaff)).
```
andres-erbsen, line 496: `fst` and `snd` here are totally in line with what we have been doing, and yet the larger scale of this file highlights that they are not great for readability. If I was hacking on this, I'd try out `let (a,b,c) := ... in ..` instead, but feel free to ignore this suggestion.

src/Curves/Weierstrass/Jacobian/ScalarMult.v
```diff
@@ -0,0 +1,965 @@
...
+      Lemma Pynz :
+        y_of (of_affine P) <> 0.
+      Proof.
+        intro Hy. assert (HA : eq (of_affine P) (opp (of_affine P))).
+        - apply eq_proof_irr. destruct P as [ [ [X Y] | u] HP]; simpl; cbv in Hy; repeat split; fsatz.
+        - apply (mult_two_power 1%Z ltac:(lia)).
+          replace Wzero with (scalarmult 0%Z P) by reflexivity.
+          eapply scalarmult_eq_weq_conversion.
+          replace (2 ^ 1)%Z with (1 - -1)%Z by lia.
+          rewrite (@scalarmult_sub_l point eq add zero opp Pgroup _ (@scalarmult_ref_is_scalarmult _ _ _ _ _ Pgroup)).
```
andres-erbsen, line 612: maybe specifying `(...:=Pgroup)` should be enough to get the right implicit arguments here?

src/Curves/Weierstrass/Jacobian/ScalarMult.v
```diff
@@ -0,0 +1,965 @@
...
+        replace (1 + 2)%Z with 3%Z by lia. reflexivity.
+      Qed.
+
+      Lemma joye_inner_correct (Hjoye : is_point (joye_ladder_inner scalarbitsz testbitn (proj1_sig (of_affine P)))) :
+        eq
+          (exist _
+             (joye_ladder_inner scalarbitsz testbitn (proj1_sig (of_affine P)))
+             Hjoye)
+          (scalarmult' n (of_affine P)).
+      Proof.
```
andres-erbsen, line 698: This lemma is long and complicated enough that it would be valuable to split out one or another conceptually coherent part as a lemma, if practical. If not, some comments on opening braces would go a long way.

src/Curves/Weierstrass/Jacobian/ScalarMult.v
```diff
@@ -0,0 +1,965 @@
...
+        set (R0 := if testbitn 1 then proj1_sig (fst TPLU) else proj1_sig (snd TPLU)).
+        replace (if testbitn 1 then (proj1_sig (snd TPLU), proj1_sig (fst TPLU)) else (proj1_sig (fst TPLU), proj1_sig (snd TPLU))) with (R1, R0) by (destruct (testbitn 1); reflexivity).
+        set (WW := while (fun '(_, i) => (i <? scalarbitsz)%Z)
+        (fun '(R2, R3, i) =>
+         let
+         '(R4, R5) := if testbitn i then (R3, R2) else (R2, R3) in
+          let
+          '(R6, R7) :=
+           let '(P0, Q) := zdau_inner R4 R5 in if testbitn i then (Q, P0) else (P0, Q)
+          in (R6, R7, Z.succ i)) (Z.to_nat scalarbitsz) (R1, R0, 2%Z)).
```
andres-erbsen, line 713:
```suggestion
        set (WW := while _ _ _ _)
```

src/Curves/Weierstrass/Jacobian/ScalarMult.v
```diff
@@ -0,0 +1,965 @@
...
+      Proof.
+        clear -Hxeq Hzeq. prept; [tauto|fsatz|fsatz|].
+        assert (f4 = f1 \/ f4 = Fopp f1) by (destruct (dec (f4 = f1)); [left; assumption|right; fsatz]).
+        destruct H; [left; repeat split; fsatz|right; repeat split; fsatz].
+      Qed.
+
+      Lemma tplu_scalarmult' (Hz1 : z_of (of_affine P) = 1) :
+        eq (fst (tplu (of_affine P) Hz1)) (scalarmult' 3 (of_affine P))
+        /\ eq (snd (tplu (of_affine P) Hz1)) (scalarmult' 1 (of_affine P))
+        /\ co_z (fst (tplu (of_affine P) Hz1)) (snd (tplu (of_affine P) Hz1)).
```
andres-erbsen, line 657: Is `of_affine P` rather than general argument actually a valuable specialization here for a deep reason (if so, please comment).

src/Curves/Weierstrass/Jacobian/ScalarMult.v
atrieu: Specifying `(groupG:=Pgroup)` wasn't sufficient, I had to also specify `(mul_is_scalarmult:=scalarmult_ref_is_scalarmult (groupG:=Pgroup))` which wasn't that much shorter, so I did not apply the changes everywhere, just the parts I went over.

src/Curves/Weierstrass/Jacobian/ScalarMult.v
atrieu: I added comments.

src/Curves/Weierstrass/Jacobian/ScalarMult.v
atrieu: This lemma is not correct in general as the co-Z formulas are incomplete. If P is of order 2 or 3, you would have `tplu P = (0, 0)` instead of `(3P, P)`.

src/Curves/Weierstrass/Jacobian/CoZ.v
atrieu: Affine is a commutative group assumes `char_ge_12` and `discriminant_non_zero` which these lemmas do not require, so I guess it's either move those lemmas into `Jacobian.v` which does not have the assumptions, or these could indeed be proved using isomorphism with affine group commutativity. Any preference ?

atrieu:
Thanks for the review!

> Though, in particular, the details of SS/TT are opaque me and yet I have an intuition that perhaps the relevant definitions and theory could be simplified. 

SS/TT are the intermediate values taken by a Joye ladder state. The recurrence relation is what is happening during a loop iteration: `(P, Q) <- (if bit i then 2P + Q else P, if bit i then Q else 2Q + P)`.

Though since co-Z formulas are incomplete, the neutral point must be avoided, which is why there are the `HSS` and `HTT` assumptions using SS/TT. I guess I could just have used the equivalent `(n mod (2 ^ (Z.of_nat (S i))))` and `(2^(Z.of_nat (S i)) - (n mod (2 ^ (Z.of_nat (S i)))))` formulations instead of using SS/TT.

andres-erbsen:
Thank you! I'll try my hand at the SS/TT situation for an hour or so and probably just merge regardless of how it goes.

andres-erbsen:
I refactored the SS/TT part that I had in mind (please chery-pick [here](https://github.com/mit-plv/fiat-crypto/commit/cad12cb717900fb39a30fac69a213f59889beeb9). I will comment some more inline, but I think we are close to landing this.

src/Curves/Weierstrass/Jacobian/ScalarMult.v
andres-erbsen: Do you think it would be possible to factor this proof in the following ways:

- state in terms of SS and TT and integer inequalities the fact that invalid points are never encountered, and prove it separately?
- consider separately the part that treats n' as opaque, perhaps even factoring the code as well as the proof, and then wrap that argument and implementation with the parts about how n' is derived and related back to n?

src/Curves/Weierstrass/Jacobian/ScalarMult.v
```diff
@@ -0,0 +1,854 @@
...
+        assert (A3 = (if testbitn 1 then A2 else A1) :> point) as HA3 by (destruct (testbitn 1); inversion HA1; auto).
+        assert (A4 = (if testbitn 1 then A1 else A2) :> point) as HA4 by (destruct (testbitn 1); inversion HA1; auto).
+        clear HA1. destruct (tplu_scalarmult' (tplu_co_z_points_obligation_1 P HPnz)) as (HeqA1 & HeqA2 & _).
+        rewrite Htplu in HeqA1, HeqA2. cbn [fst snd] in HeqA1, HeqA2.
+        (* While loop *)
+        set (WW := while _ _ _ _).
+        (* Invariant is:
+           - loop counter i is such that 2 ≤ i ≤ scalarbitsz
+           - R1 = [TT (i - 1)]P
+           - R0 = [SS (i - 1)]P
```
andres-erbsen, line 617: TT and SS add 1 to their argument, and here we are calling them with an explicit -1. Might it be that a more natural definition would let us skip both steps?

src/Curves/Weierstrass/Jacobian/ScalarMult.v
```diff
@@ -0,0 +1,854 @@
...
+                    rewrite Q; reflexivity
+                  |
+                  | eapply (mult_two_power (Z.succ i) ltac:(lia));
+                    replace Wzero with (scalarmult 0 P) by reflexivity;
+                    apply scalarmult_eq_weq_conversion;
+                    rewrite Q; reflexivity
+                  ];
+                  replace zero with (scalarmult' 0 (of_affine P)) in Q by reflexivity;
+                  apply scalarmult_eq_weq_conversion in Q;
+                  generalize (SS_monotone0 n' scalarbitsz ltac:(generalize Hn'; lia) ltac:(lia) (Z.to_nat (i - 1)%Z)); rewrite SS0; intro QS;
```
andres-erbsen, line 774: My changes affected the arguments of `SS_monotone0` and I was luckily able to adapt this proof based on the error message alone.  Yet I worry that if the proof breaks here later for whatever reason, the current structure makes it quite challenging to debug. So, I would really appreciate if you could take another look at conceptual factoring. ALternatively, if the cases come in groups, a tactic that I sometimes get mileage out of is to match-goal-with and shelve all but one group, then use `all:` to make progress on that group, and Unshelve to pick the next group.

src/Curves/Weierstrass/Jacobian/ScalarMult.v
```diff
@@ -0,0 +1,854 @@
...
+      Qed.
+
+      (* We compute [n]P where P ≠ ∞ and n < ord(P) *)
+      Context {n scalarbitsz : Z}
+              {Hn : (2 <= n < 2^scalarbitsz)%Z}
+              {Hscalarbitsz : (2 <= scalarbitsz)%Z}
+              {P : Wpoint} {HPnz : P <> ∞ :> Wpoint}
+              {ordP : Z} {HordPpos : (2 < ordP)%Z}
+              {HordPodd : Z.odd ordP = true :> bool}
+              {HordP : forall l, (l <> 0 :> Z)%Z -> ((Weq (scalarmult l P) ∞) <-> exists k, (l = k * ordP :> Z)%Z)}
```
andres-erbsen, line 369: Is `(l <> 0 :> Z)` doing something meaningful here?

atrieu: It's indeed useless, good catch ! :)

src/Curves/Weierstrass/Jacobian/ScalarMult.v
atrieu: Indeed! I fixed it.

src/Curves/Weierstrass/Jacobian/ScalarMult.v
atrieu: I did some changes and used `match goal with` which should hopefully make the proof more readable.

src/Curves/Weierstrass/Jacobian/ScalarMult.v
atrieu: Yes, this should be possible, I will try to get around to it later this week.

atrieu:
I refactored the code so that there is an "inner" ladder manipulating Jacobian coordinates, and assuming the scalar input is odd, and a wrapper to convert affine <-> Jacobian coordinates and adjusting the result of the inner loop in case the actual input isn't odd.
Proofs have also been refactored in consequence.

andres-erbsen:
Thank you! This is an awesome contribution.

PR #1838, opened 2024-03-19
Use etc/ci/github-actions-make.sh on alpine

JasonGross:
To display more errors nicely


JasonGross:
Let's try merging.  Might have to revert if we get non-deterministic timeouts on alpine though

PR #1841, opened 2024-03-29
merge MontgomeryEquivalence into MontgomeryLadder

andres-erbsen:


andres-erbsen:
@JasonGross is this CI failure indicative of real issues?

JasonGross:
It looks like a real issue, but unrelated to this PR.  Should be fixed by https://github.com/mit-plv/fiat-crypto/pull/1842, if the error message is accurate.

PR #1876, opened 2024-04-16
Do not use Ndigits in NUtil (coq/coq #18936)

andres-erbsen:


JasonGross:
Is there some lemma that we use here that has no adequate replacement?  Or is it just easier to remove the lemmas than to port them?

JasonGross:
Oh, this is sp2019latest.  Do we not use this anymore in master?

andres-erbsen:
Yes, all of these.

PR #1878, opened 2024-04-17
[CI] use container instead of chroot for Debian

JasonGross:
Passing `GITHUB_STEP_SUMMARY="$GITHUB_STEP_SUMMARY"` through to the chroot was not working, so let's try using a container instead

### DIFF
--- a/.github/workflows/coq-debian.yml
+++ b/.github/workflows/coq-debian.yml
@@ -18,96 +18,84 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - env: { DEBIAN: "sid" }
-        #- env: { DEBIAN: "bookworm" }# restore once 8.17 lands in Debian stable
+        - debian: "sid"
+        #- debian: "bookworm" # restore once 8.17 lands in Debian stable
 
     runs-on: 'ubuntu-22.04'
-    env: ${{ matrix.env }}
-    name: debian-${{ matrix.env.DEBIAN }}
+    name: debian-${{ matrix.debian }}
+    container: debian:${{ matrix.debian }}
 
     concurrency:
-      group: ${{ github.workflow }}-${{ matrix.env.DEBIAN }}-${{ github.head_ref || github.run_id }}
+      group: ${{ github.workflow }}-${{ matrix.debian }}-${{ github.head_ref || github.run_id }}
       cancel-in-progress: true
 
     steps:
     - uses: actions/checkout@v4
       with:
         submodules: recursive
-    - name: setup Debian chroot
-      run: etc/ci/setup-debian-chroot.sh "$DEBIAN"
-    - name: host build params
+    - name: install system dependencies
+      run: |
+        apt-get -o Acquire::Retries=30 update -y
+        apt-get -q -y --allow-unauthenticated -o Acquire::Retries=30 install sudo git make time jq python3 python-is-python3 ocaml coq libcoq-core-ocaml-dev libfindlib-ocaml-dev ocaml-findlib cabal-install js-of-ocaml
+    - name: container build params
       run: etc/ci/describe-system-config.sh
-    - name: chroot build params
-      shell: in-debian-chroot.sh {0}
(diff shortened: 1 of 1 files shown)


andres-erbsen:
Can you point me to what was not working?

JasonGross:
Look at the failing debian CI before https://github.com/mit-plv/fiat-crypto/commit/6bce0ff7b316874b4706cc3de066a46364791ce6, where I reverted the change that broke things.  GitHub Actions defines GITHUB_STEP_SUMMARY but the chroot doesn't seem to know about it

JasonGross:
> Look at the failing debian CI before [6bce0ff](https://github.com/mit-plv/fiat-crypto/commit/6bce0ff7b316874b4706cc3de066a46364791ce6), where I reverted the change that broke things. GitHub Actions defines GITHUB_STEP_SUMMARY (per step, I think) but the chroot doesn't seem to know about it

[Here it says] (https://github.com/mit-plv/fiat-crypto/actions/runs/8710498721/job/23892538556)
```
id -u
+ id -g
+ pwd
+ exec sudo chroot /chroot setpriv --reuid 1001 --regid 127 --init-groups /bin/sh -e -u -x -c cd "$1"; shift; exec sh -e -u -x "$@" -- /home/runner/work/fiat-crypto/fiat-crypto /home/runner/work/_temp/6cba05fe-d649-48be-a360-34765f85caa5
+ cd /home/runner/work/fiat-crypto/fiat-crypto
+ shift
+ exec sh -e -u -x /home/runner/work/_temp/6cba05fe-d649-48be-a360-34765f85caa5
/home/runner/work/_temp/6cba05fe-d649-48be-a360-34765f85caa5: 1: GITHUB_STEP_SUMMARY: parameter not set
```

andres-erbsen:
Oh, is https://github.com/mit-plv/fiat-crypto/commit/6bce0ff7b316874b4706cc3de066a46364791ce6#diff-66eee238a3c2480cc0fb962e08ff6a6b91570e9944ab4058a7ed9fe5cff62360L42 the code? That does not work because the rhs is evaluated inside chroot as well, failing because the variable is unset there. See [here](https://github.com/jirutka/setup-alpine/blob/359b91e094d3dc6292e4b627abb84d8856ab7ef2/alpine.sh#L12-L16) for how it works in the alpine script -- I am not sure passing through all the variables is great, which is why I didn't imitate it, but if we could pass just the ci-script-derived ones that would be good. (Does github actions stash them in some file already?) Or we could just imitate that.

andres-erbsen:
(As I have just confirmed, it's not chroot that messes with the env anyway, it's [sudo](https://github.com/mit-plv/fiat-crypto/blob/master/etc/ci/in-debian-chroot.sh#L6))

JasonGross:
> Oh, is [6bce0ff#diff-66eee238a3c2480cc0fb962e08ff6a6b91570e9944ab4058a7ed9fe5cff62360L42](https://github.com/mit-plv/fiat-crypto/commit/6bce0ff7b316874b4706cc3de066a46364791ce6#diff-66eee238a3c2480cc0fb962e08ff6a6b91570e9944ab4058a7ed9fe5cff62360L42) the code?

Yes

> That does not work because the rhs is evaluated inside chroot as well, failing because the variable is unset there. See [here](https://github.com/jirutka/setup-alpine/blob/359b91e094d3dc6292e4b627abb84d8856ab7ef2/alpine.sh#L12-L16) for how it works in the alpine script 

Ah, I see

> I am not sure passing through all the variables is great, which is why I didn't imitate it, but if we could pass just the ci-script-derived ones that would be good. 

Sure, I'm happy to revert this in favor of the chroot solution.  Do you have a preference for chroot over docker container?

> Does github actions stash them in some file already?

Yes, the file name is stored in `$GITHUB_ENV`

andres-erbsen:
> Do you have a preference for chroot over docker container?

Yes. When chroot goes wrong, I have some idea of how to troubleshoot it.

PR #1901, opened 2024-05-09
Reuse docker images for validate job

### LINKED ISSUE #1900: Use the same docker image in validate and build docker jobs on master
[Here](https://github.com/mit-plv/fiat-crypto/actions/runs/9015124748/job/24784247494) we have a failure due to the Coq image being updated between jobs.  We can figure out the tag either with https://github.com/coq-community/docker-coq-action/pull/89 or by parsing `docker images --format "{{.Repository}}:{{.Tag}}"`.  We can then emit output with something like `echo "::set-output name=coq_image_tag::$coq_image_tag"` and something like
```yaml
name: Example Workflow

on: [push]

jobs:
  set-output-job:
    runs-on: ubuntu-latest
    outputs:
      coq_image_tag: ${{ steps.set-output-step.outputs.coq_image_tag }}
    steps:
      - name: Set coq_image_tag Output
        id: set-output-step
        run: |
          # Simulate obtaining some value; in this case, we use a hard-coded string
          echo "COQ_IMAGE_TAG=dev" > result.txt

          # Extract the value and set it as output for the job
          coq_image_tag=$(cat result.txt | cut -d= -f2)
          echo "::set-output name=coq_image_tag::$coq_image_tag"

  fetch-output-job:
    needs: set-output-job
    runs-on: ubuntu-latest
    steps:
      - name: Fetch coq_image_tag from Previous Job
        run: echo "COQ_IMAGE_TAG is ${{ needs.set-output-job.outputs.coq_image_tag }}"
```

JasonGross:
Fixes #1900
(hopefully; the CI doesn't test the validate job)

### DIFF
--- a/.github/workflows/coq-docker.yml
+++ b/.github/workflows/coq-docker.yml
@@ -24,6 +24,9 @@ jobs:
     env: ${{ matrix.env }}
     name: docker-${{ matrix.env.COQ_VERSION }}
 
+    outputs:
+      coq_image_name: ${{ steps.set-output-step.outputs.coq_image_name }}
+
     concurrency:
       group: ${{ github.workflow }}-${{ matrix.env.COQ_VERSION }}-${{ github.head_ref || github.run_id }}
       cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


.github/workflows/coq-docker.yml
```diff
@@ -122,6 +125,41 @@ jobs:
...
+        if [ -z "${COQ_IMAGE}" ]; then
+          echo "::warning::.coqimage is empty, parsing docker images for coqorg/coq:.*${{ matrix.env.DOCKER_COQ_VERSION }}.*"
+          echo "::group::docker images"
+          docker images
+          echo "::endgroup::"
+          echo "::group::docker images --format '{{.Repository}}:{{.Tag}}'"
+          docker images --filter "reference=coqorg/coq*" --format "{{.Repository}}:{{.Tag}}"
+          echo "::endgroup::"
+          COQ_IMAGE="$(docker images --filter "reference=coqorg/coq*" --format "{{.Repository}}:{{.Tag}}" | grep -o 'coqorg/coq:.*${{ matrix.env.DOCKER_COQ_VERSION }}.*')"
+        fi
```
JasonGross, line 151: Note that this block can be removed once https://github.com/coq-community/docker-coq-action/pull/89 is merged

PR #1946, opened 2024-08-15
fix Debian CI

JasonGross:


JasonGross:
Does not fix things, but seems to help a bit anyway, let's merge

PR #1947, opened 2024-08-15
[CI] [Debian] use -j1 on js-of-ocaml

JasonGross:
Maybe this will fix the CI


JasonGross:
Seems to help, but is not sufficient

PR #1958, opened 2024-09-18
Attempt to fix wasm_of_ocaml

JasonGross:
As per https://github.com/ocaml-wasm/wasm_of_ocaml?tab=readme-ov-file#installation, to deal with the CI failure in https://github.com/mit-plv/fiat-crypto/pull/1957 / https://github.com/mit-plv/fiat-crypto/actions/runs/10906100009/job/30271149044?pr=1957


JasonGross:
wasm is not building without this change, so let's merge, even though this doesn't completely fix things

PR #1959, opened 2024-09-19
Revert "Also install .wasm.map files"

JasonGross:
Reverts mit-plv/fiat-crypto#1908

The build time of the WASM job somehow jumped from 14m to 2h with mit-plv/fiat-crypto#1908, though I don't see how.  Investigating again.


JasonGross:
Tracking at https://github.com/ocaml-wasm/wasm_of_ocaml/issues/81

PR #1961, opened 2024-09-19
Don't sed wasm files, instead use subfolders to ensure the binary names are correct from the start

JasonGross:
This should also pave the way to more easily installing files in assets folders when we land #1958


JasonGross:
The CI passed, but it seems to have somehow not output anything on the install step?!  I will merge anyway, so we can add the next followup PR

PR #2009, opened 2025-01-31
switch GarageDoor to LeakageSemantics

### LINKED ISSUE #2007: GarageDoor does not compile with latest bedrock2 (Leakage traces)
GarageDoor.v calls memequal. memequal has a proof in LeakageSemantics. Garagedoor uses Semantics. It should be changed to use LeakageSemantics or MetricLeakageSemantics, but

- I couldn't find a weakening lemma to prove a degenerate LeakageSemantics call using Semantics
- I couldn't find programlogic & etc for MetricLeakageSemantics

The first one is probably easier to fix. Do you agree, @OwenConoly ? Do you then know which semantics we'd want to call the compiler-correctness theorem with?

As for who would do this work: I was thinking I'd do it when I merged the bedrock2 PR but now I have questions for you (and I am at a conference anyway) so if you want to give it a stab it would be very welcome.

(As a last-resort fallback, e.g. if this ends up blocking a coq-compat PR, we could also duplicate memequal.)

OwenConoly:
Resolves #2007.

This needs mit-plv/bedrock2#445 to be merged first.

### DIFF
--- a/src/Bedrock/End2End/X25519/GarageDoor.v
+++ b/src/Bedrock/End2End/X25519/GarageDoor.v
@@ -119,16 +119,9 @@ Definition BootSeq : list OP -> Prop :=
                ||| lan9250_boot_timeout _
                ||| (any+++spi_timeout _)).
 
-Import WeakestPrecondition ProgramLogic SeparationLogic.
+Import LeakageSemantics LeakageWeakestPrecondition LeakageProgramLogic SeparationLogic.
 Local Notation "m =* P" := ((P%sep) m) (at level 70, only parsing) (* experiment*).
 Local Notation "xs $@ a" := (Array.array ptsto (word.of_Z 1) a xs) (at level 10, format "xs $@ a").
-Global Instance spec_of_initfn : spec_of "initfn" :=
-  fnspec! "initfn" / bs R,
-  { requires t m := m =* bs $@(word.of_Z PK) * R /\ length bs = 32%nat;
-    ensures t' m' := m' =* garageowner$@(word.of_Z PK) * R /\
-      exists iol, t' = iol ++ t /\
-      exists ioh, mmio_trace_abstraction_relation ioh iol /\
-      BootSeq ioh }.
 
 Ltac fwd :=
   repeat match goal with
@@ -138,6 +131,17 @@ Ltac fwd :=
          end; trivial.
 Ltac slv := try solve [ trivial | ecancel_assumption | SepAutoArray.listZnWords | intuition idtac | reflexivity | eassumption].
 
+Section WithPickSp.
+Context {pick_sp: PickSp}.
+
+Global Instance spec_of_initfn : spec_of "initfn" :=
+  fnspec! "initfn" / bs R,
+  { requires k t m := m =* bs $@(word.of_Z PK) * R /\ length bs = 32%nat;
+    ensures k' t' m' := m' =* garageowner$@(word.of_Z PK) * R /\
+      exists iol, t' = iol ++ t /\
+      exists ioh, mmio_trace_abstraction_relation ioh iol /\
+      BootSeq ioh }.
+
 Local Instance spec_of_memconst_pk : spec_of "memconst_pk" := spec_of_memconst "memconst_pk" garageowner.
 Local Instance WHY_spec_of_lan9250_init : spec_of "lan9250_init" := spec_of_lan9250_init.
 Lemma initfn_ok : program_logic_goal_for_function! initfn.
@@ -146,13 +150,13 @@ Proof.
   straightline_call.
   { ssplit. eassumption. rewrite H2. all : vm_compute; trivial. }
   repeat straightline.
-  eapply WeakestPreconditionProperties.interact_nomem; repeat straightline.
+  eapply LeakageWeakestPreconditionProperties.interact_nomem; repeat straightline.
   eexists; fwd; slv.
   { vm_compute. intuition congruence. }
-  eapply WeakestPreconditionProperties.interact_nomem; repeat straightline.
+  eapply LeakageWeakestPreconditionProperties.interact_nomem; repeat straightline.
   eexists; fwd; slv.
   { vm_compute. intuition congruence. }
-  eapply WeakestPreconditionProperties.interact_nomem; repeat straightline.
+  eapply LeakageWeakestPreconditionProperties.interact_nomem; repeat straightline.
   eexists; fwd; slv.
   { vm_compute. intuition congruence. }
   straightline_call; repeat straightline; fwd.
@@ -251,8 +255,8 @@ Definition memrep bs R : state -> map.rep(map:=SortedListWord.map _ _) -> Prop :
 
 Global Instance spec_of_loopfn : spec_of "loopfn" :=
   fnspec! "loopfn" / seed sk bs R,
-  { requires t m := memrep bs R (Build_state seed sk) m;
-    ensures T M := exists SEED SK BS, memrep BS R (Build_state SEED SK) M /\
+  { requires k t m := memrep bs R (Build_state seed sk) m;
+    ensures K T M := exists SEED SK BS, memrep BS R (Build_state SEED SK) M /\
     exists iol, T = iol ++ t /\
     exists ioh, SPI.mmio_trace_abstraction_relation ioh iol /\
     protocol_step (Build_state seed sk) ioh (Build_state SEED SK) }.
@@ -272,20 +276,20 @@ Proof.
   rename H11 into Lseed. rename H12 into Lsk. rename H13 into H11.
   straightline_call; try ecancel_assumption; trivial; repeat straightline.
   intuition idtac; repeat straightline;
-  eexists; split; repeat straightline; split; intros; try contradiction; [|]; repeat straightline.
+  eexists; eexists; split; repeat straightline; split; intros; try contradiction; [|]; repeat straightline.
   2: fwd; slv.
 
   pose proof H12 as Hbuf.
   seprewrite_in @bytearray_index_merge Hbuf. { ZnWords. }
 
-  eexists; split; repeat straightline.
+  eexists; eexists; split; repeat straightline.
   rewrite word.unsigned_ltu, ?word.unsigned_of_Z_nowrap by ZnWords.ZnWords;
   destr Z.ltb; rewrite ?word.unsigned_of_Z_0, ?word.unsigned_of_Z_1; intuition try discriminate;
   autoforward with typeclass_instances in E.
   2: { fwd; slv. right. left. fwd; slv; intuition try ZnWords. }
 
   repeat straightline.
-  eapply WeakestPreconditionProperties.dexpr_expr.
+  eapply LeakageWeakestPreconditionProperties.dexpr_expr.
   eexists; split; repeat straightline.
 
   case (SepAutoArray.list_expose_nth x3 12 ltac:(ZnWords)) as (Hpp&Lpp).
@@ -318,7 +322,7 @@ Proof.
     cbv [word.wrap]; (rewrite_strat (bottomup (terms (Zmod_small)))); try ZnWords.ZnWords.
     { rewrite Z.lor_comm; reflexivity. } }
 
-  eexists; split; repeat straightline.
+  eexists; eexists; split; repeat straightline.
   rewrite word.unsigned_ltu, ?word.unsigned_of_Z_nowrap by ZnWords.ZnWords;
   destr Z.ltb; rewrite ?word.unsigned_of_Z_0, ?word.unsigned_of_Z_1; intuition try discriminate;
   autoforward with typeclass_instances in E0.
@@ -341,7 +345,7 @@ Proof.
   repeat eplace (word.add (word.add buf _) _) with (word.add buf _) in H12 by (ring_simplify; trivial).
 
   repeat straightline.
-  eexists; split; repeat straightline.
+  eexists; eexists; split; repeat straightline.
   subst protocol.
   pose proof byte.unsigned_range ipproto.
   rewrite word.unsigned_eqb, ?word.unsigned_and_nowrap, ?word.unsigned_of_Z_nowrap by ZnWords.ZnWords.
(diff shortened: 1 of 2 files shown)


andres-erbsen (approved):
To be merged after dependencies and CI.

andres-erbsen:
For merging this, I think a simple plan would be to wait for the bedrock2 change to get into rupicola, then, bump rupicola synchronously with the GarageDoor changes in this PR. IIUC either wouldn't build without the other.

andres-erbsen:
@JasonGross any guesses for what's up with windows? I wouldn't have expected the file in question to be heavy.

JasonGross:
Almost certainly this is an issue with building deps with `-j2`.

https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources says Windows runners have 16 GB RAM.

The [last successful Windows run claims](https://github.com/mit-plv/fiat-crypto/actions/runs/13018569905/job/36313691507#step:11:5786) a peak memory usage on deps of 19666352 ko for Rewriter/Rewriter/Reify.vo, which seems bogus, probably `time` can't measure memory on Windows accurately?

MacOS 13 and Windows both claim to be using Coq 8.18, and [MacOS on this PR claims](https://github.com/mit-plv/fiat-crypto/actions/runs/13073618980/job/36480407593?pr=2009#step:10:7710) a peak memory usage on deps of 2282440 ko for rupicola/bedrock2/compiler/src/compiler/FlatToRiscvFunctions.vo which is a lot more reasonable. 

For kyberslash MacOS claims
```
 /Users/runner/work/fiat-crypto/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/kyberslash.vo (real: 20.35, user: 19.86, sys: 0.38, mem: 687276 ko)
```
which is indeed not that heavy.

It's hard to see what's running at the same time as kyberslash, so let's rebase on top of https://github.com/mit-plv/fiat-crypto/pull/2010 to remove `--output-sync` so we can see things a bit better.

andres-erbsen:
I can't find any concurrency in the log.

```File "D:/a/fiat-crypto/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/lightbulb.v", line 94, characters 2-40:
D:/a/fiat-crypto/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/lightbulb.vo (real: 128.19, user: 0.12, sys: 0.42, mem: 289064 ko)
make[4]: *** [D:/a/fiat-crypto/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/kyberslash.vo] Deleting file 'D:/a/fiat-crypto/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/kyberslash.glob'
COQC D:/a/fiat-crypto/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/kyberslash.v
While loading initial state:
Warning: Native compiler is disabled, -native-compiler ondemand option
ignored. [native-compiler-disabled,native-compiler,default]
Fatal error: out of memory
Command exited with non-zero status 3
D:/a/fiat-crypto/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/kyberslash.vo (real: 1074.22, user: 0.04, sys: 0.27, mem: 128372 ko)
Error: make[4]: *** [Makefile.coq.ex:838: D:/a/fiat-crypto/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/kyberslash.vo] Error 3
Error: make[3]: *** [Makefile.coq.ex:409: all] Error 2
Error: make[2]: *** [Makefile:71: ex] Error 2
Error: make[1]: *** [Makefile:90: bedrock2_ex] Error 2
+ python ./etc/coq-scripts/timing/make-one-time-file.py --real time-of-build.log time-of-build-pretty.log
make: *** No rule to make target 'bedrock2', needed by 'deps'.  Stop.
+ git status
HEAD detached at pull/2009/merge
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
  (commit or discard the untracked or modified content in submodules)
	modified:   rupicola (untracked content)

no changes added to commit (use "git add" and/or "git commit -a")
+ git ls-files --others --exclude-standard
+ git diff
+ cat time-of-build-pretty.log
     Time |    Peak Mem | File Name                                                                                                                   
------------------------------------------------------------------------------------------------------------------------------------------------------
43m52.00s | 19715484 ko | Total Time / Peak Mem                                                                                                       
------------------------------------------------------------------------------------------------------------------------------------------------------
17m54.22s |   128372 ko | D:/a/fiat-crypto/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/kyberslash.vo                                  
 2m08.19s |   289064 ko | D:/a/fiat-crypto/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/lightbulb.vo                                   
```

JasonGross:
Can you tell the CI to upload files on failure?  If we can look at D:/a/fiat-crypto/fiat-crypto/rupicola/bedrock2/bedrock2/src/bedrock2Examples/kyberslash.v.timing, we can see how far it got before failing.

JasonGross:
> https://github.com/mit-plv/fiat-crypto/actions/runs/13168644056
> 
> ```
> windows
> No files were found with the provided path: timing-files.zip. No artifacts will be uploaded.
> ```

Sorry, fixed by https://github.com/mit-plv/fiat-crypto/commit/66b7bf559153134a28bfa44e8f621a35ef36d493, let's try again

JasonGross:
> Does either of you have windows with coq set up to repro locally?

I don't have time to investigate right now (and am now on Mac), but you can [get a Windows 11 VM on kvm](https://getlabsdone.com/how-to-install-windows-11-on-kvm/).

> I am puzzled by this failure and if it only happens on GithHub actions I think we should not block the merge on it.

I'd like to at least get a separate bedrock2 target, etc, that excludes these files from the build, and modify the windows CI to skip the files that don't build on Windows.

OwenConoly:
> Does either of you have windows with coq set up to repro locally?

I have a laptop with Windows 10 --- setting up Coq there would probably (?) be easy.  Or I could try the Windows 11 VM.  Do you think one of these things is worth doing?

JasonGross:
> Do you think one of these things is worth doing?

Yes, please!

There is something weird going on with kyberslash.  The .timing file is empty, suggesting that it fails to even run the first `Require Import`, if that can be believed?  The log claims it takes 784.50s (771.68s the second time through).  This is obscenely longer than the 14.89s it takes on alpine edge (compare with lightbulb, which takes 134.28s on Windows and 100.40s on alpine).

If you can match the [version & install config we use on CI](https://github.com/mit-plv/fiat-crypto/blob/master/.github/workflows/coq-windows.yml) (Coq 8.18.0, OCaml 4.13.1, install opam and then `opam install conf-time conf-gcc 'ocamlfind>=1.9.7' && opam pin add --kind=version coq 8.18.0` and then `opam exec -- make COQEXTRAFLAGS="-async-proofs-j 1" NJOBS="2" ...`), then that makes it more likely you'll see similar behavior.  It might be the case that simply moving to a different version of Coq or OCaml will fix the issue (in which case I'll record the incompatible version in the opam file and we can just update the CI).  But if it's still present on newer versions of Coq, maybe there's a Coq or OCaml bug lurking here (maybe for some reason the comment at the top of the file triggers some weird issue?)

OwenConoly:
Windows makes everything difficult, apparently.  I've tried a couple ways to install Coq 8.18.0 and have given up for now.

Do you know off the top of your head what the attached error message means?
[[ERROR] The compilation of coq-stdlib.txt](https://github.com/user-attachments/files/18713914/ERROR.The.compilation.of.coq-stdlib.txt)

The error code "-1073741515" is meaningless, so far as I can tell from googling it...

OwenConoly:
I get the same error code with a fresh install of windows in a VM, so it is probably not due to "corrupted system files or damaged operating system files", as a vague answer on the impressively useless website https://answers.microsoft.com suggests.

andres-erbsen:
Let's drop windows altogether? We're at least as short on maintainer time as ever, and if we're not using it actively, I don't think it's wise to spend time trying to learn it to fix it.

JasonGross:
Let's at least try updating first, see if that fixes anything

JasonGross:
Seems a bit of a shame to lose one of the few intensive CI tests of Coq on Windows, and I imagine there will be some issues when the next Coq platform is released (cc @MSoegtropIMC), but you're right, not worth spending time on given limited resources.

MSoegtropIMC:
P.S.: for sure you can't compile fiat-crypto on Windows without Coq Platform - you need the large stack size patches.

JasonGross:
@MSoegtropIMC The concern is that the latest version of bedrock2 fails to compile on our Windows CI, and we are planning on removing Windows from our CI here to deal with that because none of us have time to debug Windows.  I'm concerned about whether this will mean that fiat-crypto gets excluded from the Platform when the next release rolls around.  The issue seems to be a bug in Coq, because as far as I can tell, `coqc` dies with an out of memory error while processing a beginning-of-file comment followed by `Require Import Coq.Lists.List.` (evidence: passing `-timing-file` to `coqc` results in an empty timing file) (cf https://github.com/coq/coq/issues/20212), and `time coqc` does not record excessive memory usage.  If you happen to have the time and inclination to test whether or not the latest dev version of coq-bedrock2 successfully compiles `src/bedrock2Examples/kyberslash.v` (I think `From bedrock2Examples Require Import kyberslash.` should work as a smoke test) on Windows, and potentially look at what the actual error is / post a reproducer on https://github.com/coq/coq/issues/20212, that could help prevent accumulating divergences that might have to be fixed at platform-release-time.

MSoegtropIMC:
@JasonGross : a stack overflow can also give an out of memory (depends on how stacks are organised). Afaik this changed in OCaml 5 from OCaml 4. In OCaml 5 (afaik) the initial process stack is not really used and a new thread with dynamic stack created (just hearsay - I have no solid knowledge about this). I have no idea how large these stacks are and if they can grow on Windows. With OCaml 4.X I can safely set the main thread stack size to arbitrary sizes by patching the binaries. This stack does grow. I currently do this for ocamlc, but not for coqc.

This is done here: https://github.com/coq/platform/blob/c64cb9a6d4282f03c39b0b6171cb87ef411a5021/shell_scripts/install_ocaml_stacksize.sh#L15

If one wants to do it for coqc, one would have to do it after coqc has been compiled. But if you have a Coq Platform cygwin the patch program is compiled and you can just call it for coqc after installing coq and before compiling bedrock.

I can have a look later this week. Is your Windows CI is Coq Platform based as the Coq CI?

JasonGross:
I don't see how stack overflow could cause a failure without emitting anything to the timing file, though.  And note that this is bedrock2, not fiat-crypto, and shouldn't require excessive stack.

> I can have a look later this week. Is your Windows CI is Coq Platform based as the Coq CI?

Thank you!  The setup is based on the [setup-ocaml action](https://github.com/ocaml/setup-ocaml/) + `opam install coq`, not based on the platform script.  Happy to move to the platform script if that fixes things though.

MSoegtropIMC:
OK, I will have a look - it will take a few days, though (busy with the platform release).

OwenConoly:
> If you can't compile Coq itself, you are doing something wrong.

Well yes, I figured that was the case!  Could you help me see what I am doing wrong?  I basically just installed opam in the way recommended on the opam website, and then I tried to install coq (and the other packages Jason said I should install) using opam, and it failed.  Is the actual process more complicated than that?

Here is a complete transcript of everything I did, starting from installing windows 11 in a VM.  Maybe you can point out the problem?

[coq_installation_attempt.txt](https://github.com/user-attachments/files/18744783/coq_installation_attempt.txt)




Of course, it is quite understandable if you'd rather just look at the thing yourself rather than try to fix whatever I am doing wrong :)

MSoegtropIMC:
@OwenConoly : I would strongly recommend to follow the instructions here:

https://github.com/coq/platform/blob/main/doc/README_Windows.md#installation-by-compiling-from-sources-using-opam-on-cygwin

This is very well tested in Coq Platform CI, Coq CI and by many users. This is also the officially recommended way to install Coq (also on other platforms btw.). See:

https://coq.inria.fr/download

The Windows support in opam is quite new (last year) and not yet mature enough for bigger things. The Coq Platform Windows support dates back to Coq 8.6 and did constantly work since then. Bedrock and fiat crypto are in Coq Platform and if they don't work I fix it.

I hope I find some time this year to fix the issues in plain opam - there shouldn't be many. Basically since 2.2 (or so) opam follows a similar approach as Coq Platform, that is to do a MinGW cross build on cygwin, so the gaps should be small.

OwenConoly:
Thanks for the link!  The instructions there worked smoothly.

OwenConoly:
I found the problem; I think it should be considered a bug in Coq.  I opened coq/coq#20228 to discuss it.

As a workaround, merging mit-plv/bedrock2#447 (and then bumping bedrock2 here) should make the Windows CI work.

.github/workflows/coq-windows.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
     env:
       NJOBS: "2"
-      COQ_VERSION: "8.18.0" # minimal major version required for bedrock2 components
+      COQ_VERSION: "8.20.0"
```
JasonGross, line 29: Can you revert / drop this CI change when rebasing/bumping bedrock2?

.github/workflows/coq-windows.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     - name: Set up OCaml
       uses: ocaml/setup-ocaml@v3
       with:
-        ocaml-compiler: 4.13.1
+        ocaml-compiler: 4.14
```
JasonGross, line 46: And this one too

OwenConoly:
> @OwenConoly : so I guess there is no need that I look into it?

Right!  Thanks for the offer though.

andres-erbsen:
:tada:

PR #2013, opened 2025-02-06
Refined/Signed Barrett Reduction

atrieu:
This proves the correctness of the Barrett reduction strategy that is used for instance in Kyber/ML-KEM with regards to a signed modulo operation.

I haven't proved that this signed modulo operation gives rise to a field structure isomorphic to the usual one used in Fiat-Crypto, though it shouldn't be too hard if one needs it.

### DIFF
--- a/src/Arithmetic/BarrettReduction/Refined.v
+++ b/src/Arithmetic/BarrettReduction/Refined.v
@@ -1,0 +1,335 @@
+From Coq Require Import ZArith QArith Qabs Qround.
+From Coq Require Import Lia.
+Require Import Util.QUtil.
+
+(*
+  This is a formalization of (a part of) section 2.4 of the following paper
+  Efficient Multiplication of Somewhat Small Integers Using Number-Theoretic Transforms
+  Hanno Becker, Vincent Hwang, Matthias J. Kannwischer, Lorenz Panny, Bo-Yin Yang
+  IWSEC 2022
+*)
+
+Local Open Scope Q_scope.
+
+Section Qround_half_up.
+  Local Coercion inject_Z : Z >-> Q.
+  Local Coercion Zpos : positive >-> Z.
+
+  Definition Qround_half_up (x: Q): Z := Qfloor (x + (1#2))%Q.
+
+  Lemma Qfloor_approx (x: Q):
+    Qabs (x - Qfloor x) < 1%Z.
+  Proof.
+    rewrite Qabs_Qlt_condition.
+    pose proof (Qfloor_le x) as H1.
+    pose proof (Qlt_floor x) as H2.
+    destruct x as (x_num & x_den).
+    cbv [Qlt Qle] in *. simpl in *. lia.
+  Qed.
+  Lemma Qround_half_up_approx (x: Q):
+    Qabs (x - Qround_half_up x) <= 1#2.
+  Proof.
+    rewrite Qabs_Qle_condition.
+    pose proof (Qfloor_le (x + (1#2))) as H1.
+    pose proof (Qlt_floor (x + (1#2))) as H2.
+    destruct x as (x_num & x_den).
+    cbv [Qlt Qle Qplus Qround_half_up] in *. simpl in *. lia.
+  Qed.
+
+  Lemma Qfloor_diff (x1 x2: Q) (Hlt: (Qfloor x1 < Qfloor x2)%Z):
+    exists (n: Z), x1 < n <= x2.
+  Proof.
+    exists (Qfloor x2).
+    pose proof (Qlt_floor x1) as H1.
+    pose proof (Qfloor_le x2) as H2.
+    split; [|assumption].
+    apply (Qlt_le_trans _ _ _ H1).
+    cbv [Qle]; simpl. lia.
+  Qed.
+
+  Lemma Qround_half_up_diff (x1 x2: Q)
+    (Hlt: (Qround_half_up x1 < Qround_half_up x2)%Z):
+    exists (n: Z), x1 < n + (1#2) <= x2.
+  Proof.
+    unfold Qround_half_up in Hlt.
+    apply Qfloor_diff in Hlt.
+    destruct Hlt as (n & Hlt & Hle).
+    exists (n - 1)%Z.
+    destruct x1 as (x_num1 & x_den1).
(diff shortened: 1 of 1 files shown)


src/Arithmetic/BarrettReduction/Refined.v
```diff
@@ -0,0 +1,335 @@
...
+  Qed.
+End SignedMod.
+
+Section RefinedBarrettReduction.
+  Local Coercion inject_Z : Z >-> Q.
+  Local Coercion Zpos : positive >-> Z.
+
+  (* Barrett reduction is the special case with b = 1 *)
+  Definition barrett_multiplication_approx
+    (approx: Q -> Z) (R a b: Z) (q: positive): Z :=
```
andres-erbsen, line 141: `approx` seems to be always called with freshly constructed fractions, maybe it would look better taking two arguments?

src/Arithmetic/BarrettReduction/Refined.v
```diff
@@ -0,0 +1,335 @@
...
+    (a * b - q * Qround_half_up ((a * (approx ((b * R)#q)))%Z / R))%Z.
+
+  (* Not the same bounds on |a| and |b| as in the paper, as theirs cannot be proved *)
```
andres-erbsen, line 144: This sounds like it could be a fun story. Did formal verification help identify a gap here?

src/Arithmetic/BarrettReduction/Refined.v
```diff
@@ -0,0 +1,335 @@
...
+      rewrite Z.sub_0_r. reflexivity.
+    - assert (Z.abs 1 = Z.pow 2 0)%Z as -> by reflexivity.
+      apply Z.pow_le_mono_r; lia.
+    - assert (Z.log2_up _ = 0)%Z as -> by reflexivity.
+      assert (M - 2 - (0 - (k - 1)) = M - 2 + (k - 1))%Z as -> by lia.
+      exact Ha'.
+  Qed.
+End RefinedBarrettReduction.
+
+(* Section MLKEM_Example. *)
```
andres-erbsen, line 295: This code would be welcome too. Perhaps `Module ExampleMLKEM`

src/Arithmetic/BarrettReduction/Refined.v
```diff
@@ -0,0 +1,335 @@
...
+  Definition barrett_multiplication_approx
+    (approx: Q -> Z) (R a b: Z) (q: positive): Z :=
+    (a * b - q * Qround_half_up ((a * (approx ((b * R)#q)))%Z / R))%Z.
+
+  (* Not the same bounds on |a| and |b| as in the paper, as theirs cannot be proved *)
+  Lemma barrett_multiplication_approx_correct
+    (approx: Q -> Z) (M R a b: Z) (k q: positive)
+    (Hk: Qabs (((b * R)#q) - approx ((b * R)#q)) <= 1#(Pos.pow 2 k))
+    (HOddq: Z.Odd q) (HR: R = Z.pow 2 (M - 1 + Z.log2 q - Z.log2_up (Z.abs b))%Z)
+    (HM: (2 <= M)%Z) (Ha: (Z.abs a <= Z.pow 2 (M - 1))%Z)
```
andres-erbsen, line 149: I understand there is some value in following the paper here, but I do wonder whether `Z.pow 2 (M - 1)` could be quantified over instead of `M and so on.

andres-erbsen (approved):
:+1: , I'd be happy to have this with or without whatever changes my comments inspire you to undertake

cc @jadephilipoom who looked into Barrett reduction variants earlier

src/Arithmetic/BarrettReduction/Refined.v
```diff
@@ -0,0 +1,335 @@
...
+  Lemma Zmod_is_mod_approx_floor:
+    forall (a: Z) (b: positive),
+      Z.modulo a b = mod_approx Qfloor a b.
+  Proof.
+    intros. unfold mod_approx.
+    rewrite Qmake_Qdiv, <- Zdiv_Qdiv.
+    apply Z.mod_eq. discriminate.
+  Qed.
+
+  Definition signed_mod: Z -> positive -> Z := mod_approx Qround_half_up.
```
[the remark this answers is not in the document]
atrieu: Yes, here is a proof below. I don't mind using your `smodulo` as the specification instead, do you want to add your file to Fiat-Crypto first and I rebase over it? Or the other way around, merge this first and you can fix it in another PR later?

```Coq
Lemma smodulo_eq_signed_mod (a: Z) (b: positive):
  smodulo a b = signed_mod a b.
Proof.
  cbv [smodulo]. rewrite (omod_inj_mod _ a (signed_mod a b)) by (apply Zmod_signed_mod).
  apply omod_small_iff. left.
  rewrite Z.quot_div_nonneg by lia.
  rewrite signed_mod_eq_Zmod. pose proof (Zlt_cases (2 * (a mod b)) b) as Hlt.
  destruct (_ <? _); Z.to_euclidean_division_equations; lia.
Qed.
```

src/Arithmetic/BarrettReduction/Refined.v
atrieu: I don't think formal verification was necessary to identify the issue here. I believe it's mainly an off-by-1 calculation error, unless I'm missing something in the reasoning.

This is Fact 2 on page 7 https://eprint.iacr.org/2022/439.pdf

One difference is that they consider `|a| < 2 ^ (M - 1)` which would fail to account for `a = INT16_MIN = - 2 ^ 15` when M = 16. Though, this doesn't really change the proof.

The main issue is at the end of their proof on page 7, where one needs to prove that `|a| < (2^(k + w - 1)) / q`, and since `q < 2 ^ (1 + log2 q)`, then it should suffice to show that `|a| <= (2^(k + w - 1)) / (2 ^ (1 + log2 q)) = 2 ^ (k + w - 2 - log2 q)`, whereas the paper concludes with `2 ^ (k + w - 1 - log2 q)`.

src/Arithmetic/BarrettReduction/Refined.v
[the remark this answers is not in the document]
andres-erbsen: Sounds like you'd still keep your version either way, so I guess I'll keep procrastinating on that even longer.

PR #2016, opened 2025-03-01
Add support for more assembly

JasonGross:
I want to be able to handle the output of gcc/clang on our C code.

Right now, I have added support for parsing the assembly output. Equivalence checking is still to come.


src/Assembly/Syntax.v
```diff
@@ -3,42 +3,74 @@ From Coq Require Import NArith.
...
+(* Special registers *)
+(*| st0 | st1 | st2 | st3 | st4 | st5 | st6 | st7  (* FPU stack registers *)*)
+| k0 | k1 | k2 | k3 | k4 | k5 | k6 | k7          (* AVX-512 mask registers *)
+| gdtr | idtr | ldtr | tr
+| cw | sw | tw | fp_cs | fp_opc | fp_ds
+(* Flags registers *)
+| rflags
+| eflags
+| flags
 .
```
JasonGross, line 47: @andres-erbsen Do you have any objection to extending the registers like this?

src/Assembly/Syntax.v
```diff
@@ -71,21 +105,51 @@ Inductive OpCode :=
...
 | setc
+| sete      (* Set byte if equal *)
+| setne     (* Set byte if not equal *)
 | seto
 | shl
 | shlx
+| shld
 | shr
 | shrx
 | shrd
```
JasonGross, line 155: @andres-erbsen Do you have any opinion on / objection to extending the list of opcodes?

PR #2020, opened 2025-03-03
Use Listable to prove equality of registers and opcodes

JasonGross:
This is much faster, especially as we add more registers and opcodes


JasonGross:
I want to run timing numbers on this before merging

JasonGross:
Attempting to run timing at https://github.com/JasonGross/fiat-crypto/actions/runs/13647667831

JasonGross:
Previous build succeeded except for https://github.com/ocaml/setup-ocaml/issues/938, going to merge

PR #2027, opened 2025-03-04
Add support for parsing more assembly

JasonGross:
I want to be able to handle the output of gcc/clang on our C code.

Right now, I have added support for parsing the assembly output.
Equivalence checking is still to come.

Also still to come is parsing data declarations


JasonGross:
Built before rebase, so I will merge

PR #2029, opened 2025-03-05
Handle rip-relative addressing

JasonGross:
The idea to parse `rip` specially rather than as a register is due to @andres-erbsen 

Supersedes #2023
Closes #2023


JasonGross:
Build succeeds, will rebase and merge

PR #2039, opened 2025-03-07
Move signed to Z.signed in ZUtil.Definitions

JasonGross:


JasonGross:
@andres-erbsen Do you object to moving `Definition signed` like this?

PR #2058, opened 2025-03-12
Adapt to coq/coq#19690 (Hint Extern follows proof mode)

SkySkimmer:


SkySkimmer:
Should be backwards compatible

PR #2068, opened 2025-03-31
Readdition in bedrock2

miriampolzer:
Implement the readdition function in bedrock2 and prove it correct


andres-erbsen:
Alpine failures seem unrelated, I intend to override them. ghc depends on llvm15 which depends on an unavailable library.

PR #2069, opened 2025-04-02
Enable `+fragile-hint-constr`

JasonGross:
For compatibility with https://github.com/rocq-prover/rocq/pull/20373

Is someone willing to fix the issues that this causes so that fiat-crypto doesn't get removed from Coq's CI?  cf https://github.com/rocq-prover/rocq/pull/20373#issuecomment-2768550004


JasonGross:
The first one that needs to be fixed is https://github.com/mit-plv/fiat-crypto/blob/ec7316df6a73b3b350ee26793801b16e5e94dbba/src/Util/ZUtil/Pow.v#L48-L49, the solution is to write something like
```coq
Local Definition f_equal_pow_b b := f_equal (fun e => b ^ e).
Local Definition f_equal_pow_e e := f_equal (fun b => b ^ e).
#[global]
  Hint Resolve f_equal_pow_b f_equal_pow_e : zarith.
```
And I expect all other instances can be fixed in similar ways.

Issue #1394, opened 2022-10-01
Github CI Anomaly Zoo


andres-erbsen:
- https://github.com/mit-plv/fiat-crypto/actions/runs/3163210890/jobs/5150550477 was terminated after 320 minutes, which is a round enough number to be a timeout, but the [documented limit](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration) is 6 hours (360 minutes).
- https://github.com/mit-plv/bedrock2/actions/runs/3161224197/jobs/5146573405 shows that it was canceled, but none of us remember doing it nor can think of a reason why we would have done it.
- https://github.com/mit-plv/bedrock2/commit/7654f3fdaaa5208902a4d4de90e0b56e17d13cc2 shows :x:  even though all builds of it show :heavy_check_mark:
- An intermediate head on the Garagedoor PR showed :heavy_check_mark: even though listing its builds showed some :x:s

JasonGross:
> * [mit-plv/bedrock2@7654f3f](https://github.com/mit-plv/bedrock2/commit/7654f3fdaaa5208902a4d4de90e0b56e17d13cc2) shows ❌  even though all builds of it show ✔️

To shed a bit more light on this: If you go to the PR at https://github.com/mit-plv/bedrock2/pull/280, 	we see
![image](https://user-images.githubusercontent.com/396076/193449960-c6d52f5f-dbf8-488c-abd3-ad56abbd203f.png)
and the [actions log](https://github.com/mit-plv/bedrock2/actions/workflows/coq.yml) shows
![image](https://user-images.githubusercontent.com/396076/193450049-a4da5e95-adbe-4ee1-a585-330c764b5be7.png)
so it looks like the overall status of the commit includes runs of that commit across all branches, but the displayed checks are only for master or something?

JasonGross:
> * [mit-plv/fiat-crypto/actions/runs/3163210890/jobs/5150550477](https://github.com/mit-plv/fiat-crypto/actions/runs/3163210890/jobs/5150550477) was terminated after 320 minutes, which is a round enough number to be a timeout, but the [documented limit](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration) is 6 hours (360 minutes).

I'm thinking this might actually be an OOM kill, especially given that we're running the checker while replacing `vm_compute` with the `lazy` reduction machine.

JasonGross:
[...]

PR #2078, opened 2025-04-24
Downgrade setup-ocaml to v3.2.16

JasonGross:
This might fix the issue where Coq is failing to set printing width big enough?

cf https://github.com/mit-plv/fiat-crypto/issues/1394#issuecomment-2825962067


JasonGross:
This build gets further

PR #2080, opened 2025-04-24
feat: make rust `const`

Daniel-Aaron-Bloom:
`&mut` is `const`-stable as of rust `1.83`. As this is a pretty big jump in MSRV, it should probably be a minor version bump.

Unfortunately using `Index` and `IndexMut` in `const` functions isn't `const`, except on built-in arrays. However you can get around this with an internal helper type (`IndexConst` in this PR) doing the array indexing. This does make the `Index` and `IndexMut` impls internally redundant, but they're harmless and removing them would break downstream crates, which would be a major version bump.

At some point in the distant future when `const trait`s are stabilized, `IndexConst` can go away and the original code will just work with essentially no change to public API. But that's probably a few years away.


fiat-rust/Cargo.toml
```diff
@@ -2,12 +2,13 @@
...
 version = "0.2.10"
 authors = ["Fiat Crypto library authors <jgross@mit.edu>"]
-edition = "2018"
+edition = "2021"
 description = "Fiat-crypto generated Rust"
 homepage = "https://github.com/mit-plv/fiat-crypto"
 repository = "https://github.com/mit-plv/fiat-crypto"
 readme = "README.md"
 license = "MIT OR Apache-2.0 OR BSD-1-Clause"
+rust-version = "1.83.0"
```
JasonGross, line 11: Does this set a minimum version or pin to this particular version?  What's the standard best practice in specifying Rust version constraints?

Daniel-Aaron-Bloom: This is the standard best practice for setting the minimum version.

Per [the spec](https://doc.rust-lang.org/cargo/reference/manifest.html) this field is: `The minimal supported Rust version.`

fiat-rust/Cargo.toml
```diff
@@ -2,12 +2,13 @@
 name = "fiat-crypto"
 version = "0.2.10"
```
JasonGross, line 3: We should bump the version here, right?
```suggestion
version = "0.3.0"
```

fiat-rust/src/curve25519_64.rs
```diff
@@ -134,32 +183,32 @@ pub fn fiat_25519_cmovznz_u64(out1: &mut u64, arg1: fiat_25519_u1, arg2: u64, ar
...
-  let x17: u128 = (((arg1[1]) as u128) * ((arg2[3]) as u128));
-  let x18: u128 = (((arg1[1]) as u128) * ((arg2[2]) as u128));
-  let x19: u128 = (((arg1[1]) as u128) * ((arg2[1]) as u128));
-  let x20: u128 = (((arg1[1]) as u128) * ((arg2[0]) as u128));
-  let x21: u128 = (((arg1[0]) as u128) * ((arg2[4]) as u128));
-  let x22: u128 = (((arg1[0]) as u128) * ((arg2[3]) as u128));
-  let x23: u128 = (((arg1[0]) as u128) * ((arg2[2]) as u128));
-  let x24: u128 = (((arg1[0]) as u128) * ((arg2[1]) as u128));
-  let x25: u128 = (((arg1[0]) as u128) * ((arg2[0]) as u128));
+pub const fn fiat_25519_carry_mul(mut out1: &mut fiat_25519_tight_field_element, arg1: &fiat_25519_loose_field_element, arg2: &fiat_25519_loose_field_element) {
```
JasonGross, line 186: Why do we need `mut out1` now?  And will this break any downstream crates?

JasonGross (approved):
Thank you!

src/Stringification/Rust.v
```diff
@@ -87,6 +101,23 @@ Module Rust.
        (["";
         "#![allow(unused_parens)]";
         "#![allow(non_camel_case_types)]";
+        "";
+        "struct IndexConst<T: ?Sized>(T);";
```
JasonGross, line 105: It might be worth adding a Rust comment here about why this type exists (basically just the contents of the PR message)

fiat-rust/src/curve25519_64.rs
Daniel-Aaron-Bloom: The short answer is "no, everything is fine". `mut` in only part of the internal signature of `out1` not the signature of `fiat_25519_carry_mul`. 

The reason why `out1` need to be `mut` is a quirk of how rust mutable borrowing works. There are multiple indexing operations into `out1`, which means multiple `IndexConst`s are constructed. `IndexConst` take ownership of whatever is passed in, which is fine when dealing with `&` which are `Copy`, but `&mut` is not `Copy`. To get around this we borrow `out1` and pass that in instead. The issue is if we used `&` to borrow `out1`, then we couldn't get `&mut` out because once you're non-`mut` you're stuck that way. So instead we use `&mut`, but for that to work the local variable `out1` must be mutable.

As state above though, this doesn't impact callers because it's only the local variable `out1` which is being changed to be mutable. It's the same as putting `let mut out1 = out1;` as the first line and shadowing the variable for the whole function.

PR #2117, opened 2025-07-06
[CI] [MacOS] Enable JSOO again

JasonGross:
As per https://github.com/mit-plv/fiat-crypto/commit/4986ba94a6edc20fb09f82d9a1144c922ebd5fa5#r161506462, jsoo.6.1.0 includes significant memory usage improvements, and will hopefully work even work even with M1 GH Action runners which have only 7GB RAM.

cc @hhugo

### DIFF
--- a/.github/workflows/coq-macos.yml
+++ b/.github/workflows/coq-macos.yml
@@ -20,7 +20,7 @@ jobs:
         # macOS 12, 13 are x86_64; macOS 14 is arm64
         include:
         - os: { name: 'macOS 13', arch: 'x86_64', runs-on: 'macos-13' }
-          ocaml-compiler: '4.11.1'
+          ocaml-compiler: '4.13.0'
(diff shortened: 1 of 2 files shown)


JasonGross:
@hhugo do you know what's going on with the `wasm_of_ocaml` failure?  Does `wasm_of_ocaml --source-map --no-inline  src/ExtractionJsOfOCaml/WithBedrock/fiat_crypto.byte` no longer produce `src/ExtractionJsOfOCaml/fiat_crypto.wat` (and instead produce files in `src/ExtractionJsOfOCaml/fiat_crypto.assets` or something?)

hhugo:
- we not longer emit textual wat files unless `--debug wat` is provided. Emitting it slow and binaryen tools are slow reading it.
- the latest jsoo release 6.1 requires OCaml 4.13 at least. You should bump ocaml-compiler version for the jsoo job.
- you should experiment removing `--no-inline`

PR #1778, opened 2023-12-08
Add remaining rewrite rules for saturated arithmetic

JasonGross:
For #1609

PR #1777, opened 2023-12-08
Unobjectionable rewrite rules for saturated arithmetic

JasonGross:
For https://github.com/mit-plv/fiat-crypto/pull/1609

We don't yet include the ones that are higher-order-than 3, those need more debugging.

PR #1761, opened 2023-12-02
Don't replace abstract state of higher order functions with `bottom` when we can avoid it (now guarded by `--fancy-and-powerful-but-exponentially-slow-bounds-analysis`)

JasonGross:
Partial progress towards #1609.

Reworked value type to more closely match the possible spec.

Deduplicated `interp_ident` / `abstract_interp_ident` overlapped state.

The specification of `reify` and `reflect` has been simplified to more
closely match the intuition: `reify`'s spec no longer talks about output
bounds, only interpretation, and a companion lemma
`abstraction_relation_of_related_bounded_value'` shows that the spec of
`reflect` is effectively one direction of an isomorphism between
interpreted-values bounded by abstract state, and `value`s which are in
relationship with the given interpreted-value.  Not really sure how best
to describe this in words yet, probably needs some digesting work.

There's some interesting lemmas where we have an equivalency between two
ways of representing things only for up-to-second-order types, and an
implication for up-to-third-order types, which is currently all the
identifiers.  But this means that if we add fourth-order identifiers,
we'll have to deal with two different sorts of `Proper` abstraction
relations (one used by `Assembly/Symbolic` and one used by bounds
analysis, though it's possible the `Assembly/Symbolic` one can be
adapted).

Timing info coming soon

Remaining admits:
- [x] `wf_interp` (this is a big one, but hopefully nothing too surprising)
- [x] `always_strip_annotation_related`
- [x] `interp_eval_with_bound` (requires fixing a mismatch between `abstraction_relation` and `type.related_hetero abstraction_relation'`, maybe some other stuff)
- [x] `interp_eta_expand_with_bound`
- [x] `Interp_EvalWithBound` (requires dealing with newly introduced `ToFlat` / `FromFlat` round-trip)
- [x] `interp_strip_annotations`
- [x] `Interp_PartialEvaluateWithBounds_bounded`
- [x] `CheckedPartialEvaluateWithBounds_Correct`

Note that, other than the `wf_interp` lemma, I expect most of these to just be figuring out some gluing.

cc @andres-erbsen

PR #2120, opened 2025-07-15
Saturated experiment revived

andres-erbsen:
Perhaps there is a way forward here after all.

- Instead of #1761, a less ambitious alternative would be to add a special `if` to the pipeline IR for cases where a specific branch is expected to be taken, and the other branch is only there for proof purposes. Only the expected-dead branch should be thunked. Thus there wouldn't be a need to do bounds analysis inside a thunk.
- #1778 perf impact could be avoided (or at least reduced) by uncurrying list_rect in templates
- #1777 issue is probably a rewrite-rule conflict and it looks like perhaps it could be debugged to satisfaction. EDIT: I posted a hypothesis/proposal there.
- #2121 would need to be finished up to actually get appropriately efficient Bedrock2 code out after template instantiation

@JasonGross did I get this right?


JasonGross:
> @JasonGross did I get this right?

Yes, that sounds right to me

andres-erbsen:
The first step listed above seems like the most uncertain still, so if we want to move forward here, that would be the one to try first. Following

- 31c76f820862e902f19f043fb428ab4521fe92f8 is a commit that adds new operations to rewrite
- https://github.com/mit-plv/fiat-crypto/blame/master/src/Util/Bool.v#L135 has full thunked bool operators; we want one that is only thunked in one argument
- This new operation would be used to replace all `if` in synthesizable definitions in [src/Arithmetic/SaturatedPseudoMersenne.v](https://github.com/mit-plv/fiat-crypto/pull/2120/files/b78221c5e7a6428012a1a984b1ea07076bbac0ad#diff-5764d7d974de9ef1636ef7c8ef4b4cc86b1a1fc46fddd96efe424e4eebff1915)

PR #2122, opened 2025-07-15
Ed25519: Point-based separation logic for doubling

miriampolzer:
Convert the doubling function to a point based format, the rest of the functions will follow.

### DIFF
--- a/src/Bedrock/End2End/X25519/EdwardsXYZT.v
+++ b/src/Bedrock/End2End/X25519/EdwardsXYZT.v
@@ -21,6 +21,7 @@ From coqutil.Tactics Require Import Tactics letexists eabstract rdelta reference
 Require Import coqutil.Word.Bitwidth32.
 Require Import coqutil.Word.Bitwidth.
 Require Import coqutil.Word.Interface.
+Require Import coqutil.Word.Naive.
 From Coq Require Import Init.Byte.
 From Coq Require Import List.
 From Coq Require Import String.
@@ -76,29 +77,37 @@ Definition add_precomputed := func! (ox, oy, oz, ota, otb, X1, Y1, Z1, Ta1, Tb1,
   fe25519_mul(oz, F, G)
 }.
 
+Definition felem_size := Eval lazy in felem_size_in_bytes.
+
+(* Member access notation for projective points: (x, y, z, ta, tb). *)
+Local Notation "A .x" := (expr.op Syntax.bopname.add A (0)) (in custom bedrock_expr at level 2, left associativity).
+Local Notation "A .y" := (expr.op Syntax.bopname.add A (felem_size)) (in custom bedrock_expr at level 2, left associativity).
+Local Notation "A .z" := (expr.op Syntax.bopname.add A (felem_size + felem_size)) (in custom bedrock_expr at level 2, left associativity).
+Local Notation "A .ta" := (expr.op Syntax.bopname.add A (felem_size + felem_size + felem_size)) (in custom bedrock_expr at level 2, left associativity).
+Local Notation "A .tb" := (expr.op Syntax.bopname.add A (felem_size + felem_size + felem_size + felem_size)) (in custom bedrock_expr at level 2, left associativity).
+
 (* Equivalent of m1double in src/Curves/Edwards/XYZT/Basic.v *)
-(* Note: Ta/Tb are unused, but leaving in place in case we want to switch to a point struct in the future *)
-Definition double := func! (ox, oy, oz, ota, otb, X, Y, Z, Ta, Tb) {
-  stackalloc 40 as trX;
-  fe25519_square(trX, X);
-  stackalloc 40 as trZ;
-  fe25519_square(trZ, Y);
-  stackalloc 40 as t0;
-  fe25519_square(t0, Z);
-  stackalloc 40 as trT;
+Definition double := func! (p_out, p_a) {
+  stackalloc felem_size as trX;
+  fe25519_square(trX, p_a.x);
+  stackalloc felem_size as trZ;
+  fe25519_square(trZ, p_a.y);
+  stackalloc felem_size as t0;
+  fe25519_square(t0, p_a.z);
+  stackalloc felem_size as trT;
   fe25519_carry_add(trT, t0, t0);
-  stackalloc 40 as rY;
-  fe25519_add(rY, X, Y);
+  stackalloc felem_size as rY;
+  fe25519_add(rY, p_a.x, p_a.y);
   fe25519_square(t0, rY);
-  fe25519_carry_add(otb, trZ, trX);
-  stackalloc 40 as cZ;
+  fe25519_carry_add(p_out.tb, trZ, trX);
+  stackalloc felem_size as cZ;
   fe25519_carry_sub(cZ, trZ, trX);
-  fe25519_sub(ota, t0, otb);
-  stackalloc 40 as cT;
+  fe25519_sub(p_out.ta, t0, p_out.tb);
+  stackalloc felem_size as cT;
   fe25519_sub(cT, trT, cZ);
-  fe25519_mul(ox, ota, cT);
-  fe25519_mul(oy, otb, cZ);
-  fe25519_mul(oz, cZ, cT)
+  fe25519_mul(p_out.x, p_out.ta, cT);
+  fe25519_mul(p_out.y, p_out.tb, cZ);
+  fe25519_mul(p_out.z, cZ, cT)
 }.
 
 (* Converting a normal point to a cached point to prepare it for readdition. *)
@@ -148,13 +157,12 @@ Section WithParameters.
   Context {a_eq_minus1:a = F.opp F.one} {twice_d} {k_eq_2d:twice_d = (F.add d d)} {nonzero_d: d<>F.zero}.
 
 Local Notation "m =* P" := ((P%sep) m) (at level 70, only parsing).
-Local Notation "xs $@ a" := (Array.array ptsto (word.of_Z 1) a xs) (at level 10, format "xs $@ a").
 
 Local Notation FElem := (FElem(FieldRepresentation:=frep25519)).
 Local Notation bounded_by := (bounded_by(FieldRepresentation:=frep25519)).
 Local Notation word := (Naive.word 32).
 Local Notation felem := (felem(FieldRepresentation:=frep25519)).
-Local Notation point := (point(Fzero:=F.zero)(Fadd:=F.add)(Fmul:=F.mul)(a:=a)(d:=d)).
+Local Notation point := (point(Feq:=Logic.eq)(Fzero:=F.zero)(Fadd:=F.add)(Fmul:=F.mul)(a:=a)(d:=d)).
 Local Notation cached := (cached(Fzero:=F.zero)(Fadd:=F.add)(Fmul:=F.mul)(a:=a)(d:=d)).
 Local Notation coordinates := (coordinates(Feq:=Logic.eq)).
 Local Notation m1double :=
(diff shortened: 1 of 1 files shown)


src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -76,29 +77,40 @@ Definition add_precomputed := func! (ox, oy, oz, ota, otb, X1, Y1, Z1, Ta1, Tb1,
   fe25519_mul(oz, F, G)
 }.
 
+(* TODO can I pull those together somehow? *)
+Definition felem_size_term := Eval hnf in felem_size_in_bytes.
+(* Felem size in bytes in computed form. *)
+Definition felem_size := Eval compute in felem_size_term.
+
+(* TODO how do we call these points? I call them normal points in my head :D *)
```
andres-erbsen, line 85: Maybe projective points? I've also seen the 4-component representation called xyzt and Niels coordinates.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -76,29 +77,40 @@ Definition add_precomputed := func! (ox, oy, oz, ota, otb, X1, Y1, Z1, Ta1, Tb1,
...
+Notation "A .x" := (expr.op Syntax.bopname.add A (0*felem_size)) (in custom bedrock_expr at level 3, left associativity).
```
andres-erbsen, line 86: level 3 seems like it would bind less tightly than [negation](https://github.com/mit-plv/bedrock2/blob/5ec3fc97bb4ee29a1d6b7013e7f417ec7cb3311b/bedrock2/src/bedrock2/NotationsCustomEntry.v#L18). Maybe it was a mistake to put negation at level 1. But perhaps 0 works here?

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -76,29 +77,40 @@ Definition add_precomputed := func! (ox, oy, oz, ota, otb, X1, Y1, Z1, Ta1, Tb1,
...
+Notation "A .y" := (expr.op Syntax.bopname.add A (1*felem_size)) (in custom bedrock_expr at level 3, left associativity).
+Notation "A .z" := (expr.op Syntax.bopname.add A (2*felem_size)) (in custom bedrock_expr at level 3, left associativity).
+Notation "A .ta" := (expr.op Syntax.bopname.add A (3*felem_size)) (in custom bedrock_expr at level 3, left associativity).
+Notation "A .tb" := (expr.op Syntax.bopname.add A (4*felem_size)) (in custom bedrock_expr at level 3, left associativity).
```
andres-erbsen, line 90: In P256 i only used addition for field offseting, not multiplication, following what a code generator adding up field sizes would do. It's possible it didn't matter, and it's possible it did :shrug:

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -216,28 +227,34 @@ Global Instance spec_of_add_precomputed : spec_of "add_precomputed" :=
         bounded_by loose_bounds otb' /\
         m' =* (FElem X1K X1) * (FElem Y1K Y1) * (FElem Z1K Z1) * (FElem Ta1K Ta1) * (FElem Tb1K Tb1) * (FElem half_ypxK half_ypx) * (FElem half_ymxK half_ymx) * (FElem xydK xyd) * (FElem oxK ox') * (FElem oyK oy') * (FElem ozK oz') * (FElem otaK ota') * (FElem otbK otb') * R }.
 
+Definition coords_bounds := (Field.bounds * Field.bounds * Field.bounds * Field.bounds * Field.bounds)%type.
+(* TODO find a good name for this construct. *)
```
andres-erbsen, line 231: I think it's a representation of (projective) points. So maybe point_repr, or point.repr`...

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -216,28 +227,34 @@ Global Instance spec_of_add_precomputed : spec_of "add_precomputed" :=
...
+Definition bounded_coords := (felem * felem * felem * felem * felem)%type.
+Definition feval_coords '(x, y, z, ta, tb) := (feval x, feval y, feval z, feval ta, feval tb).
+
+Definition coords_bounded_by (bounds : coords_bounds) '(x,y,z,ta,tb) : Prop :=
+  let '(bounds_x, bounds_y, bounds_z, bounds_ta, bounds_tb) := bounds in
+  bounded_by bounds_x x /\ bounded_by bounds_y y /\ bounded_by bounds_z z /\ bounded_by bounds_ta ta /\ bounded_by bounds_tb tb.
+
+Local Notation "p .+ n" := (word.add p (word.of_Z n)) (at level 50, format "p .+ n", left associativity).
+Local Notation "c $@ p" := (let '(x,y,z,ta,tb) := c in sep (sep (sep (sep (FElem (p .+ 0) x) (FElem (p .+ 40) y)) (FElem (p .+ 80) z)) (FElem (p .+ 120) ta)) (FElem (p .+ 160) tb)) (at level 10, format "c $@ p").
```
andres-erbsen, line 240: So far `$@` has been for `list byte` only (already confusingly using two definitions with different out-of-memory corner cases). When I was suggesting using `$@` here, I meant to suggest writing to_bytes function for point representations and then passing that to a separation-logic predicate that takes bytes. This suggestion was in anticipation of deallocating and (conditional) memcpying of points. Here you don't really need it; and using `*` of `FElem` is fine -- but if you stick with it, consider a different name/symbol than `$@`.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -216,28 +227,34 @@ Global Instance spec_of_add_precomputed : spec_of "add_precomputed" :=
...
-      bounded_by loose_bounds Z /\
-      bounded_by loose_bounds Ta /\
-      bounded_by loose_bounds Tb /\
-      m =* (FElem XK X) * (FElem YK Y) * (FElem ZK Z) * (FElem TaK Ta) * (FElem TbK Tb) * (FElem oxK ox) * (FElem oyK oy) * (FElem ozK oz) * (FElem otaK ota) * (FElem otbK otb) * R;
-    ensures t' m' :=
+    (p_out p_a : word) /
+    (p : point) (a anything: bounded_coords) (R : _ -> Prop), {
+    requires t m :=
+      coords_bounded_by (tight_bounds, tight_bounds, loose_bounds, loose_bounds, loose_bounds) a /\
+      feval_coords a = coordinates p /\
```
andres-erbsen, line 249: This precondition is fine; an alternative to consider would be to make bounded_coords a sigma type that only contains points on curve. This is not a recommendation given what I have seen so far, but it is worth keeping in mind.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -216,28 +227,34 @@ Global Instance spec_of_add_precomputed : spec_of "add_precomputed" :=
...
+      m =* anything $@ p_out * a $@ p_a * R;
```
andres-erbsen, line 250: The caller of this function will need to construct a `*`-of-`FElem` representation. This helps callers that are overwriting an existing point representation, but is work for callers that are passing in a freshly allocated memory. (If the separation-logic relation for projective points was defined in terms of to_bytes for projective points, requiring `anything` to be a list of bytes of the appropriate length would work for both kinds of callers)

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -216,28 +227,34 @@ Global Instance spec_of_add_precomputed : spec_of "add_precomputed" :=
...
-        ((feval ox'), (feval oy'), (feval oz'), (feval ota'), (feval otb')) = coordinates (@m1double p) /\
-        bounded_by tight_bounds ox' /\
-        bounded_by tight_bounds oy' /\
-        bounded_by tight_bounds oz' /\
-        bounded_by loose_bounds ota' /\ (* could be tight_bounds if we need it, but I don't think we do *)
-        bounded_by tight_bounds otb' /\
-        m' =* (FElem XK X) * (FElem YK Y) * (FElem ZK Z) * (FElem TaK Ta) * (FElem TbK Tb) * (FElem oxK ox') * (FElem oyK oy') * (FElem ozK oz') * (FElem otaK ota') * (FElem otbK otb') * R }.
+      exists a_double : bounded_coords,
+      coordinates (@m1double p) = feval_coords a_double /\
+      coords_bounded_by (tight_bounds, tight_bounds, tight_bounds, loose_bounds, tight_bounds) a_double /\
```
andres-erbsen, line 255: This bounds postcondition is stronger than the precondition, but perhaps it would be convenient to weaken it to match the precondition? Really depends on what the caller needs though.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -446,53 +450,61 @@ Proof.
...
-    remember (Bignum.Bignum felem_size_in_words ozK _) as Pz in H87.
-    remember (Bignum.Bignum felem_size_in_words oyK _) as Py in H87.
-    remember (Bignum.Bignum felem_size_in_words oxK _) as Px in H87.
-    remember (Bignum.Bignum felem_size_in_words otaK _) as Pta in H87.
-    remember (Bignum.Bignum felem_size_in_words otbK _) as Ptb in H87.
-    do 7 (seprewrite_in @Bignum.Bignum_to_bytes H87).
-    subst Pz Py Px Pta Ptb.
-    extract_ex1_and_emp_in H87.
+
+  (* Now we need to convert scalar to ptsto, because straightline_dealloc can't handle scalar yet. *)
```
andres-erbsen, line 466: The code below looks as expected, but are there actually scalar-s being deallocated, or only Bignums? either way we'd want to automate this, but slightly different steps may be needed for each.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -446,53 +450,61 @@ Proof.
...
+  cbv [FElem] in *.
+  (* Prevent output from being rewritten by seprewrite_in *)
+  remember (Bignum.Bignum felem_size_in_words (p_out.+80) x12) as Pz in H92.
+  remember (Bignum.Bignum felem_size_in_words (p_out.+40) x11) as Py in H92.
+  remember (Bignum.Bignum felem_size_in_words (p_out.+0) x10) as Px in H92.
+  remember (Bignum.Bignum felem_size_in_words (p_out.+120) x8) as Pta in H92.
+  remember (Bignum.Bignum felem_size_in_words (p_out.+160) x6) as Ptb in H92.
+  do 7 (seprewrite_in @Bignum.Bignum_to_bytes H92).
```
andres-erbsen, line 474: does `repeat` run too long here or otherwise mess up?

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -446,53 +450,61 @@ Proof.
...
+  subst Pz Py Px Pta Ptb.
```
andres-erbsen, line 475: Does `subst` without arguments do something wrong here?

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -216,28 +227,34 @@ Global Instance spec_of_add_precomputed : spec_of "add_precomputed" :=
...
-        bounded_by tight_bounds ox' /\
-        bounded_by tight_bounds oy' /\
-        bounded_by tight_bounds oz' /\
-        bounded_by loose_bounds ota' /\ (* could be tight_bounds if we need it, but I don't think we do *)
-        bounded_by tight_bounds otb' /\
-        m' =* (FElem XK X) * (FElem YK Y) * (FElem ZK Z) * (FElem TaK Ta) * (FElem TbK Tb) * (FElem oxK ox') * (FElem oyK oy') * (FElem ozK oz') * (FElem otaK ota') * (FElem otbK otb') * R }.
+      exists a_double : bounded_coords,
+      coordinates (@m1double p) = feval_coords a_double /\
+      coords_bounded_by (tight_bounds, tight_bounds, tight_bounds, loose_bounds, tight_bounds) a_double /\
+      m' =* a_double $@ p_out * a $@ p_a * R
```
andres-erbsen, line 256: Consider putting the separation-logic postcondition first so that solving it by cancellation can instantiate the evar for `a_double`. Ditto for precondition in the caller's proof.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -446,53 +450,61 @@ Proof.
...
+  do 7 (seprewrite_in @Bignum.Bignum_to_bytes H92).
+  subst Pz Py Px Pta Ptb.
+  extract_ex1_and_emp_in H92.
 
   (* Solve stack/memory stuff *)
   repeat straightline.
 
   (* Post-conditions *)
-  exists x9,x10,x11,x7,x5; ssplit. 2,3,4,5,6:solve_bounds.
+  exists (x10,x11,x12,x8,x6); ssplit. 2,3,4,5,6:solve_bounds.
```
andres-erbsen, line 482: After moving the separation-logic assertion first in the postcondition, I hope running `eexists` without arguments (or perhaps with a tuple of the right number of underscores) would be enough here for `ecancel_assumption` to solve the separation-logic assertion and instantiate evars for other subgoals.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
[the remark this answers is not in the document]
miriampolzer: Oh yes, I overlooked adding that.
I still need to learn a bit about sections, modules, and notation scopes, and how they interact with imports.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
miriampolzer: Probably faster, right? I'll change it.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
miriampolzer: It might make the spec easier to use, I'll keep it in mind for later and try what works best. Thank!

src/Bedrock/End2End/X25519/EdwardsXYZT.v
miriampolzer: Good point, I'll use a different symbol. Using * of FElem becomes tedious quickly, that's why I wanted to have a more concise format for now (and as we discussed probably switch to byte format at a later point).

src/Bedrock/End2End/X25519/EdwardsXYZT.v
miriampolzer: I guess it doesn't hurt to be stronger in terms of bounds, one can always weaken. Let's see how it works in practice.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
miriampolzer: Called them projective points and gave the format.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
```diff
@@ -76,29 +77,40 @@ Definition add_precomputed := func! (ox, oy, oz, ota, otb, X1, Y1, Z1, Ta1, Tb1,
...
-  stackalloc 40 as trX;
-  fe25519_square(trX, X);
-  stackalloc 40 as trZ;
-  fe25519_square(trZ, Y);
-  stackalloc 40 as t0;
-  fe25519_square(t0, Z);
-  stackalloc 40 as trT;
+Definition double := func! (p_out, p_a) {
+  stackalloc felem_size as trX;
+  fe25519_square(trX, p_a.x);
```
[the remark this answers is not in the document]
miriampolzer: \o/

src/Bedrock/End2End/X25519/EdwardsXYZT.v
miriampolzer: It does, but only because we weren't remembering enough values. I think I can fix this and create a generic tactic.

miriampolzer: I turned it around and created a tactic that finds the addresses that need to be converted into byteformat. Much easier than telling the tactic to remember a certain set of addresses we still need.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
miriampolzer: I think I found a tactic that will work for the other proofs too.

src/Bedrock/End2End/X25519/EdwardsXYZT.v
andres-erbsen: which leaves us with the mystery of *why* level 0 is right-associative.  But okay, make it level 2 and then we can move - to 3 later, perhaps roughly following [c++](https://en.cppreference.com/w/cpp/language/operator_precedence.html).

PR #2142, opened 2025-08-22
Do not warn for deprecations in 8.20, 9.0

andres-erbsen:


andres-erbsen:
Cc @miriampolzer all remaining warnings should be fixable now

PR #2144, opened 2025-08-25
Placeholder for field operation preconditions

miriampolzer:
Use Placeholder instead of FElem in field operation preconditions. Placeholder for output parameters is more reasonable then FElem, because it states "there is some memory" instead of requiring a specific field element to be present in that memory.


miriampolzer:
I've decided to abandon this in favor of https://github.com/mit-plv/fiat-crypto/pull/2154 as using existential quantifiers inside the predicate turned out to be inconvenient.

PR #2154, opened 2025-09-10
Field ops with map.of_list_word_at precondition

miriampolzer:
Using map.of_list_word_at instead of FElem/Bignum/array will make it a bit easier to handle allocation of bigger chunks of memory that are then partially fed into field operations to fill them with values.

map.of_list_word_at requires a lenght assumption, which is wrapped into Placeholder.

The main difference between of_list_word_at and FElem is that one allows for equational reasoning (below Placeholder) while the other one does not need an explicit length assumption, as separation logic has it inherently.

As far as I understand it, both FElem and Placeholder have their advantages and disadvantages, this change hopefull allows to switch a bit more freely between them.

I've also removed some duplicated list byte <-> list word translations and removed unnecessary imports when I saw them. Furthermore, I tried to reduce the number of typeclass instances for fields and words, to avoid unintended
non-equality because of some instance parameter being different.

### DIFF
--- a/src/Bedrock/End2End/Poly1305/Field1305.v
+++ b/src/Bedrock/End2End/Poly1305/Field1305.v
@@ -1,6 +1,7 @@
 From Coq Require Import String.
 From Coq Require Import List.
 From Coq Require Import ZArith.
+Require Import Lia.
 Require Import Crypto.Arithmetic.PrimeFieldTheorems.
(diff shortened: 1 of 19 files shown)


miriampolzer:
@andres-erbsen could you review this, please? Feel free to leave any comments you have, I'm happy to learn.
It's not ready to merged, https://github.com/mit-plv/rupicola/pull/179 will need to be present.

PR #2239, opened 2026-01-20
Pull field element length into felem

miriampolzer:
The main changes are in Field.v. Extract the length of the word list representing a field element in memory from the Bignum hidden inside FElem over to the felem type. That way, one does not need to fiddle with separation logic everytime the length of that list is needed.

There are a few helper lemmas for conversions between the different ways to represent field elements in memory in Field.v, and two new tactics to automatically rewrite between the types.

Use those new tactics throughout existing proofs, and get rid of some of the convolution the Bignum type was causing.


miriampolzer:
Not sure if those failures are related to my change. It worked fine locally, but I haven't synced my branch in a while. I can check that tomorrow.

PR #2256, opened 2026-02-12
Add P256 curve file

miriampolzer:
- Move all p256 curve related logic to a separate file, adding commutative group and scalarmult.
- Make point and coord in P256 specs a notation and pull the notation inside the module to ensure it is only active when needed. I had a few issues when rewriting with the definition instead of notation, that's why I tried changing it. Let me know if that is not a good idea, I don't think it is necessary.
- My auto formatter removed some trailing spaces.

### DIFF
--- a/src/Curves/Weierstrass/P256.v
+++ b/src/Curves/Weierstrass/P256.v
@@ -57,6 +57,27 @@ End W.
 
 Definition p256_group_order := 0xffffffff00000000ffffffffffffffffbce6faada7179e84f3b9cac2fc632551.
 
+#[export] Instance prime_order : prime p256_group_order.
+Proof.
+  simple refine (Pocklington_refl
+    (Pock_certif _ 7 [(187019741,1);(38189,1);(17449,1);(3407,1);(131,1);(71,1);(2,4)] 0xf72444b78fa654e632d4b)
+    [ Pock_certif 187019741 2 [(9350987,1);(2,2)] 1;
+      Pock_certif 9350987 2 [(229,1);(2,1)] 264;
+      Pock_certif 38189 2 [(9547,1);(2,2)] 1;
+      Pock_certif 17449 14 [(3,1);(2,3)] 1;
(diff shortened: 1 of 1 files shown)


src/Curves/Weierstrass/P256.v
```diff
@@ -0,0 +1,44 @@
...
+From Crypto Require Import PrimeFieldTheorems ModInv.
+Require Import Crypto.Spec.WeierstrassCurve.
+Require Import Crypto.Curves.Weierstrass.Affine.
+Require Import Crypto.Curves.Weierstrass.AffineProofs.
+
+Local Open Scope positive_scope.
+
+Definition p256 := 2^256 - 2^224 + 2^192 + 2^96 - 1.
+
+#[export] Instance prime_p256 : prime p256. Admitted.
```
andres-erbsen:
```suggestion
From Coq Require Import List. Import ListNotations.
From Coqprime Require Import PocklingtonCertificat(Pocklington_refl,singleCertif(..)).

#[export] Instance prime_p256 : prime p256.
Proof.
  remember prime; vm_compute; subst.
  simple refine (Pocklington_refl
    (Pock_certif _ 6 [(6700417,1);(490463,1);(65537,1);(641,1);(257,1);(17,1);(5,2);(3,1);(2,1)] 0x6f199ff1e192ee086c186e)
    [ Pock_certif 6700417 5 [(3,1);(2,7)] 552;
      Pock_certif 490463 7 [(53,1);(2,1)] 174;
      Pock_certif 65537 3 [(2,16)] 1;
      Pock_certif 641 3 [(2,7)] 1;
      Pock_certif 257 3 [(2,8)] 1;
      Pock_certif 53 2 [(2,2)] 4;
      Pock_certif 17 3 [(2,4)] 1;
      Pock_certif 5 2 [(2,2)] 1;
      Proof_certif 3 prime_3;
      Proof_certif 2 prime_2] _)%positive.
  native_cast_no_check (@eq_refl bool true).
Qed.
```

PR #2258, opened 2026-02-16
P256: set zero, select, coord opp

miriampolzer:


miriampolzer:
Still need to check that those are exactly what will be needed, but I wanted to show you this draft to ensure this is the right way to spec and use functions generated by the pipeline branch.

PR #2276, opened 2026-03-02
Update to min 9.0.0 in the CI

miriampolzer:
Trying out if this works. If I understand correctly, CI workflows should be triggered once I create a draft pull request.


miriampolzer:
As we've run into this multiple times now with several changes, I'm inclined to merge it. Any objections?

PR #2310, opened 2026-04-20
Fix incorrect Proof commands

SkySkimmer:


andres-erbsen:
In what sense are these incorrect? `Proof` is [documented](https://rocq-prover.org/doc/master/refman/proofs/writing-proofs/proof-mode.html#rocq:cmd.Proof) as a no-op

miriampolzer:
Thanks for fixing! The Proof command was at the wrong position.

PR #2313, opened 2026-04-28
Adapt spec_of_p256_get_multiple

miriampolzer:
The old spec was expecting a point as output, but we want to allow raw bytes to be able to call with just allocated memory.

To that end, add a zeroing at the beginning of the function that selects a point from the table. This is how it is done in boringssl today and probably better because then the result is zero for out of bounds calls.


miriampolzer:
Me and @lukaszobernig figured this may be the best approach and it mirrors existing boringssl implementation in p256. @andres-erbsen what do you think? There are alternatives such as using constant_time_conditional_memcpy.

andres-erbsen:
This is a good fix, we are hoping the compiler will ripup this logic anyway and simplify resulting vpternlog instructions.

PR #2334, opened 2026-05-22
Adapt to rocq-prover/rocq#21987 (secvar status)

SkySkimmer:


src/Util/FSets/FMapTrie.v
```diff
@@ -1119,11 +1119,11 @@ Module Import ListWSfun_gen_proofs.
...
         clear.
         cbv [MapsTo find].
-        destruct m as [m' pf]; cbn [proj1_sig]; clear m pf.
+        destruct (m) as [m' pf]; cbn [proj1_sig]; clear m pf.
         revert x y m'.
-        induction x as [|x' xs IH], y as [|y' ys], m' as [m|]; clear x y m';
+        induction (x) as [|x' xs IH], (y) as [|y' ys], (m') as [m|]; clear x y m';
           try specialize (IH ys);
-          try (induction m as [d m'] using (t_case (elt:=elt)); clear m).
+          try (induction (m) as [d m'] using (t_case (elt:=elt)); clear m).
```
SkySkimmer: this fix is incomplete but I don't understand what's going on enough to finish it
the proof gets stuck with 1 remaining goal (using https://github.com/rocq-prover/rocq/pull/21987)
please help @JasonGross

SkySkimmer: I got it to work, for some reason `induction (m)`  below was still clearing m so I used  `try clear` instead

SkySkimmer:
@JasonGross what's the CI issue?

JasonGross:
CI issue is that Rocq is ≈1.5x--2x slower on Mac x86_64 than Mac ARM and we haven't trimmed our CI target for Mac yet.  cc @andres-erbsen

PR #2347, opened 2026-06-17
P256 use xor for conditional point select

miriampolzer:
The previous approach used three chained coord selects that were slower than an xor.


andres-erbsen:
(I think the failures are https://github.com/rocq-prover/rocq/issues/22227)

miriampolzer:
Failures are unrelated to this change, and there is sufficient coverage now by the CI runners that passed, I will bypass and merge this.

PR #2354, opened 2026-07-06
Make redc.v robust to the Rocq ≥ 9.3 shelf-restriction change (rocq-prover/rocq#22182)

JasonGross:
rocq-prover/rocq#22182 ("clear do not unshelve shelved evars") changes `Evd.restrict` so that a restricted shelved evar keeps its position on the shelf instead of being re-declared as a future goal (which effectively moved it to the end of the shelf, or surfaced it). `src/Bedrock/Field/Synthesis/Examples/redc.v` currently fails on that PR's head at line 210 with `Expected a single focused goal but 2 goals are focused.` — it is the only fiat-crypto failure in that PR's CI.

Two knock-on effects on this proof script (full analysis with probes in [this comment on the Rocq PR](https://github.com/rocq-prover/rocq/pull/22182#issuecomment-4894750927)):

- bedrock2's `straightline` loop case (`Markers.unique (Markers.left _)`, implemented with `unshelve (...); []`) now succeeds where it previously failed, so `repeat straightline` advances much further into the `tailrec` obligations, and the loop-condition witness becomes a named local definition `br := if word.ltu ...` (via `letexists`) instead of an inlined evar instantiation;
- the goal order after the final `Unshelve` changes.

The fix (4 hunks, one file, commented in the source):

- guard the manual loop-invariant steps behind a goal-shape check — a no-op on Rocq ≥ 9.3, removable once Rocq < 9.3 support is dropped;
- inline the `br := if word.ltu ...` local definition (matched by shape, not by autogenerated hypothesis name) where the script needs the conditional inlined — a no-op on Rocq < 9.3;
- decouple `repeat (repeat eexists; straightline)` into `repeat eexists. all: repeat straightline.` — `t1; t2` discards everything when `t2` fails on any goal produced by `t1`, and on Rocq ≥ 9.3 this point is reached with `straightline` already at a fixpoint, so the original form was a silent no-op;
- solve the final `Unshelve` goals order-independently (`all: eauto.` instead of positional bullets).

**Verification**: compiled this exact file against both the current Rocq master tip (`2274c5dc56`) and rocq-prover/rocq#22182's head (`769b5325e4`), each with its own GitLab-CI-built dependency stack (coqutil/bedrock2/rewriter/stdlib artifacts from the respective pipelines): unpatched it passes on master and fails on the PR head at line 210; patched it passes on both.

Intended to serve as the fiat-crypto overlay for rocq-prover/rocq#22182 (cc @SkySkimmer).

https://claude.ai/code/session_019ttctspSoVoquHLQtbPVZw

### DIFF
--- a/src/Bedrock/Field/Synthesis/Examples/redc.v
+++ b/src/Bedrock/Field/Synthesis/Examples/redc.v
@@ -207,7 +207,28 @@ Section WithParameters.
         - ecancel_assumption.
         - rewrite word.unsigned_of_Z_0. Lia.lia. }
 
-      { repeat straightline. eexists. split. { repeat straightline. } repeat straightline. split.
+      { repeat straightline.
+        (* Backwards compatibility: on Rocq < 9.3 (before rocq-prover/rocq#22182),
+           [repeat straightline] cannot make progress on the loop-invariant goal
+           [Markers.unique (Markers.left _)] (its [unshelve]-based case fails when
+           an evar on the shelf gets restricted), so we do the corresponding steps
+           manually.  On Rocq >= 9.3 the automation already went past this goal,
+           making this a no-op; it can be removed once Rocq < 9.3 support is
+           dropped. *)
+        all: try lazymatch goal with
+             | |- Markers.unique (Markers.left _) =>
+               eexists; split; [ repeat straightline | repeat straightline; split ]
+             end.
+        (* On Rocq >= 9.3 (rocq-prover/rocq#22182), the loop-condition witness is
+           introduced by [letexists] (inside [straightline]) as a context-local
+           definition [br := if word.ltu ... then ... else ...] which hypotheses
+           mention by name, whereas on Rocq < 9.3 it is a plain evar whose
+           instantiation leaves the conditional inlined in the hypotheses.  The
+           [destruct (word.ltu ...)]/[rewrite ... in H*] steps below need the
+           inlined form, so inline the definition; no-op on Rocq < 9.3. *)
+        all: try match goal with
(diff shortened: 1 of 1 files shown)


ppedrot:
This is allegedly backwards compatible, could you merge quickly so that we can get the Rocq PR in before the branch?

JasonGross:
Yes. I tasked Fable with investigating the docker failure, it said:
> docker-master fails at src/Bedrock/End2End/X25519/MontgomeryLadder.v:34 with Unable to unify "?ladderstep" with (["X1"; "X2"; "Z2"; "X3"; "Z3"], ...) (unresolved ?field_representaton/?width evars in the
  candidate term). I pulled the log of fiat-crypto master's most recent run of the same workflow (July 1, five days before your PR) — it fails at the byte-identical location with the byte-identical error. This job
  builds against rocq/rocq-prover:dev, and fiat-crypto master has been red against rocq dev for weeks (master runs of coq-docker.yml failing on Jul 1, Jun 15, Jun 13). Meanwhile your redc.v compiled fine in this
  very run (redc.vo in 6.07s) before the unrelated failure aborted the build. Note also that since rocq PR #22182 isn't merged into dev yet, this job runs old-shelf-semantics Rocq — and the patched redc.v passing
  there is exactly the backwards-compatibility half of what the PR promises.

I guess the rocq CI target does not quite test enough of fiat-crypto to catch all the bugs, since I don't see why
https://github.com/mit-plv/fiat-crypto/blob/f6267279e9608234e190ace7e10f364bc14af14d/src/Bedrock/End2End/X25519/MontgomeryLadder.v#L33-L34

should be failing.  I have tasked Fable with investigating this too.

PR #2359, opened 2026-07-08
add etc/coqccache.sh and Makefile.cached

andres-erbsen:


Makefile.cached
```diff
@@ -0,0 +1,2 @@
+COQC := $(CURDIR)/etc/coqccache.sh $(CURDIR)/.coqdep.mk coqc
```
JasonGross:
```suggestion
COQC := $(CURDIR)/etc/coqccache.sh $(CURDIR)/.coqdep.mk $(COQBIN)coqc
```

PR #2365, opened 2026-07-24
Depend on libcoq-stdlib in debian.yml

andres-erbsen:


andres-erbsen:
```
File "src/Rewriter/Util/plugins/strategy_tactic.ml.v93", line 6, characters 102-103:
Error: The value q has type Names.GlobRef.t list
       but an expression was expected of type Environ.env list
       Type Names.GlobRef.t is not compatible with type Environ.env
```

@JasonGross ?

JasonGross:
Sounds like a version issue.  If I had to guess, I'd guess that rewriter is incorrectly inferring the version of Rocq, either due to a Debian packaging bug or a bug in the rewriter version guessing code.  (Maybe Debian is reporting the version as dev even though it is a released version? (It is a released version, right?)). I will get AI to investigate

PR #2368, opened 2026-07-28
Implement and verify `u320_muladd`.

klinashka:
This PR verifies and implements the `u320_mulladd` function which performs the following: 

Input: a 320-bit number `p_v`, a 256-bit number `p_m`, a word `c_prime. The specification imposes that the variables do not alias. 
Output: `p_v` is updated as follows: `p_v += p_m * c_prime`. The carry limb `c` is returned.

### DIFF
--- a/src/Bedrock/P256/modinv/u320_muladd.v
+++ b/src/Bedrock/P256/modinv/u320_muladd.v
@@ -11,7 +11,7 @@ Local Open Scope string_scope. Local Open Scope Z_scope.
 Local Notation eval := (fold_right (fun (a : word) (s : Z) => a + 2^64*s) 0).
 Local Notation array := (array scalar (word.of_Z 8)).
 
-Local Instance spec_of_u320_muladd : spec_of "u320_muladd" := 
+#[export] Instance spec_of_u320_muladd : spec_of "u320_muladd" := 
     fnspec! "u320_muladd" (p_v p_m c_prime : word) / (v m : list word) R ~> c,
     {
         requires t m' := 
@@ -67,11 +67,6 @@ Local Existing Instance spec_of_full_mul.
 Local Ltac lists_into_elements := repeat match goal with
   | H : length ?l = ?n |- _ =>  constr_eq true ltac:(isnatcst n);
   let x := fresh l "0" in destruct l as [(*nil*)|x l]; inversion H; clear H end.
-
-Lemma lt_word_prod (w1 w2 : word) : w1 * w2 <= (2^64 - 1) * (2^64 - 1).
-Proof.
-    eapply Zorder.Zmult_le_compat; ZnWords.
-Qed.
 
 Lemma u320_muladd_correct : program_logic_goal_for_function! u320_muladd.
 Proof. repeat straightline. lists_into_elements. unfold array in *.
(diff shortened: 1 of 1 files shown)


src/Bedrock/P256/modinv/u320_muladd.v
```diff
@@ -0,0 +1,94 @@
...
+Import coqutil.Tactics.Tactics.
+From coqutil Require Import WithBaseName.
+Local Open Scope string_scope. Local Open Scope Z_scope.
+
+(** * Specification *)
+
+Local Notation eval := (fold_right (fun (a : word) (s : Z) => a + 2^64*s) 0).
+Local Notation array := (array scalar (word.of_Z 8)).
+
+Local Instance spec_of_u320_muladd : spec_of "u320_muladd" :=
```
lukaszobernig: Will you use this spec in another file? If so, let's mark it as `#[export]` instead of `Local`.

src/Bedrock/P256/modinv/u320_muladd.v
```diff
@@ -0,0 +1,94 @@
...
+    store(p_v + $8 + $8 + $8, s3);
+    store(p_v + $8 + $8 + $8 + $8, s4);
+
+    c = carry_t + carry_add
+}.
+
+(** * Proof *)
+
+Local Existing Instance spec_of_full_add.
+Local Existing Instance spec_of_full_mul.
```
lukaszobernig: Might be worth exporting them explicitly in the files that they are defined, but that will require a change to Bedrock2.

src/Bedrock/P256/modinv/u320_muladd.v
```diff
@@ -0,0 +1,94 @@
...
+
+Local Ltac lists_into_elements := repeat match goal with
+  | H : length ?l = ?n |- _ =>  constr_eq true ltac:(isnatcst n);
+  let x := fresh l "0" in destruct l as [(*nil*)|x l]; inversion H; clear H end.
+
+Lemma lt_word_prod (w1 w2 : word) : w1 * w2 <= (2^64 - 1) * (2^64 - 1).
+Proof.
+    eapply Zorder.Zmult_le_compat; ZnWords.
+Qed.
```
lukaszobernig: Looks like this is unused?

lukaszobernig (commented):
Looks good, thanks for the PR!

src/Bedrock/P256/modinv/u320_muladd.v
klinashka: Yes, this will be used in another file. I will mark it as `#[export]`

src/Bedrock/P256/modinv/u320_muladd.v
klinashka: Yes, indeed! Thanks for pointing this out. I removed it.

src/Bedrock/P256/modinv/u320_muladd.v
klinashka: That is a good suggestion! The change to bedrock2 has already been merged. Once that propagates to `fiat-crypto` I will make the necessary adjustments.

PR #2377, opened 2026-08-06
Verify `u320_shr` function.

klinashka:
This PR implements and verifies the `u320_shr` function, which performs a right shift operation on unsigned 320-bit integers. 

The function assumes the following:
- The `p_x` parameter is a 320-bit unsigned integer. 
- The `carry` parameter contains only one bit of information (I.e. `carry < 2`)
- The `shift` parameter is exclusively between 0 and 64. The `shift` being less than 64 ensures that the shift is only performed across one limb. The `shift` being greater than zero together with the carry being exactly one bit ensures that the carry will be entirely shifted into the number. This simplifies the postcondition expression. 

If the preconditions are satisfied, the function ensures that `p_x` will hold the result of a right shift of the number `p_x + 2^320 * carry`.

### DIFF
--- a/src/Bedrock/P256/modinv/u320_shr.v
+++ b/src/Bedrock/P256/modinv/u320_shr.v
@@ -1,5 +1,6 @@
 From Coq Require Import BinInt String List InitialRing.
 From bedrock2 Require Import BasicC64Semantics WeakestPrecondition ProgramLogic.
+Require Import bedrock2.NotationsCustomEntry bedrock2.ZnWords Coq.ZArith.ZArith Lia.
 From coqutil Require Import WithBaseName.
 Import ListNotations ProgramLogic.Coercions SeparationLogic Array Scalars.
 Local Open Scope string_scope. Local Open Scope Z_scope.
@@ -10,20 +11,19 @@ Local Notation eval := (fold_right (fun (a : word) (s : Z) => a + 2^64*s) 0).
 Local Notation eval_bool := (fold_right (fun (a : word) (s : Z) => Z.lor a (Z.shiftl s 64)) 0).
 Local Notation array := (array scalar (word.of_Z 8)).
 
-Local Instance spec_of_u320_shr : spec_of "u320_shr" := 
+#[export] Instance spec_of_u320_shr : spec_of "u320_shr" := 
 fnspec! "u320_shr" (p_x carry shift : word) / (x r : list word) (s : word) R,
 {
     requires t m := 
         m =* array p_x x ⋆ R /\
             length x = 5%nat /\ length r = 5%nat /\ 
-            shift < 64%nat /\ 0%nat < shift /\ carry < 2^1;
+            0%nat < shift < 64%nat /\ carry < 2;
     ensures T M := 
         T = t /\ exists (r : list word) , M =* array p_x r ⋆ R /\
             length r = 5%nat /\ eval r = Z.shiftr (eval (x ++ [carry])) shift
 }.
 
 (** * Implementation *)
-Require Import bedrock2.NotationsCustomEntry bedrock2.ZnWords Coq.ZArith.ZArith Lia.
 
 Definition u320_shr := func! (p_x, carry, shift) 
 {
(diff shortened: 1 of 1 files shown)


src/Bedrock/P256/modinv/u320_shr.v
```diff
@@ -0,0 +1,108 @@
...
+Import ListNotations ProgramLogic.Coercions SeparationLogic Array Scalars.
+Local Open Scope string_scope. Local Open Scope Z_scope.
+
+(** * Specification *)
+
+Local Notation eval := (fold_right (fun (a : word) (s : Z) => a + 2^64*s) 0).
+Local Notation eval_bool := (fold_right (fun (a : word) (s : Z) => Z.lor a (Z.shiftl s 64)) 0).
+Local Notation array := (array scalar (word.of_Z 8)).
+
+Local Instance spec_of_u320_shr : spec_of "u320_shr" :=
```
lukaszobernig: I think this should be `#[export]` here.

src/Bedrock/P256/modinv/u320_shr.v
```diff
@@ -0,0 +1,108 @@
...
+Local Notation eval_bool := (fold_right (fun (a : word) (s : Z) => Z.lor a (Z.shiftl s 64)) 0).
+Local Notation array := (array scalar (word.of_Z 8)).
+
+Local Instance spec_of_u320_shr : spec_of "u320_shr" := 
+fnspec! "u320_shr" (p_x carry shift : word) / (x r : list word) (s : word) R,
+{
+    requires t m := 
+        m =* array p_x x ⋆ R /\
+            length x = 5%nat /\ length r = 5%nat /\ 
+            shift < 64%nat /\ 0%nat < shift /\ carry < 2^1;
```
lukaszobernig: My stylistic choice would be to write `0%nat < shift < 64%nat`, but I have no strong opinion.

src/Bedrock/P256/modinv/u320_shr.v
```diff
@@ -0,0 +1,108 @@
...
+    ensures T M := 
+        T = t /\ exists (r : list word) , M =* array p_x r ⋆ R /\
+            length r = 5%nat /\ eval r = Z.shiftr (eval (x ++ [carry])) shift
+}.
+
+(** * Implementation *)
+Require Import bedrock2.NotationsCustomEntry bedrock2.ZnWords Coq.ZArith.ZArith Lia.
```
lukaszobernig: Let's collect the imports at the top of the file.

src/Bedrock/P256/modinv/u320_shr.v
```diff
@@ -0,0 +1,108 @@
...
+    repeat match goal with |- context [Z.testbit ?a ?b] => rewrite (Z.testbit_neg_r a b) by ZnWords end;
+    repeat match goal with |- context [Z.testbit ?a ?b] => rewrite (@prove_Zeq_bitwise.testbit_above 64 a) by ZnWords end;
+    repeat match goal with |- context [Z.testbit ?a ?b] => rewrite (@prove_Zeq_bitwise.testbit_above 1 a) by ZnWords end).
+Qed.
+
+(** * Linking Proof *)
+Definition u320_shr_funcs := &[, u320_shr].
+
+Lemma link_u320_shr : spec_of_u320_shr (Interface.map.of_list u320_shr_funcs).
+Proof. apply u320_shr_correct; trivial. Qed.
```
lukaszobernig: We probably don't need this here since the function does not use any other functions with specs.

src/Bedrock/P256/modinv/u320_shr.v
klinashka: Good catch! I will make the change.

src/Bedrock/P256/modinv/u320_shr.v
klinashka: I agree that it is a better choice! I have also changed the last conjunct to be `carry < 2`

src/Bedrock/P256/modinv/u320_shr.v
klinashka: I have removed it from the file.

lukaszobernig:
Looks good, thanks for adding this!

PR #2383, opened 2026-08-17
Verify optimized brctz

klinashka:
This PR implements an implementation of `br_ctz`, which takes as input a 64-bit unsigned word and returns the number of trailing Zeros.   

The implementation repeatedly shifts the input left by 1, until the input becomes 0. We have opted for this implementation, rather than the more intuitive alternative, because the current implementation targets optimizations performed in the `LoopIdiomRecognize` pass of the LLVM compiler. This results in efficient compilations by both `clang` and `gcc` (even though we did not specifically target `gcc` initially). More details on this can be found [here](https://godbolt.org/z/aqa8x488z). 

The PR is currently a draft, as it should be merged after https://github.com/mit-plv/coqutil/pull/177 propagates to fiat-crypto.


src/Bedrock/P256/modinv/br_ctz.v
```diff
@@ -0,0 +1,108 @@
...
+    }
+Qed.
+
+
+(** * Specification *)
+
+#[export] Instance spec_of_br_ctz : spec_of "br_ctz" := 
+    fnspec! "br_ctz" (value : word) ~> count,
+    {
+        requires t m := value > 0;
```
miriampolzer: Could this be value >= 0?

src/Bedrock/P256/modinv/br_ctz.v
```diff
@@ -0,0 +1,108 @@
...
+        ensures T M := T = t /\ M = m /\ count = word.of_Z (lctz 64 value)
```
miriampolzer: Not sure if it plays a role here, but we usually have an easier time preferring non-truncating conversions. e.g. here you would write `word.unsigned count = lctz 64 value`.

src/Bedrock/P256/modinv/br_ctz.v
```diff
@@ -0,0 +1,108 @@
...
+            match goal with 
+            | [H1 : ?a = (?b * 2^?c) mod ?d |- _] => rewrite H1 by ZnWords
+            end.
+            rewrite Z.mul_mod_idemp_l by ZnWords.
+            f_equal. rewrite_strat bottomup Z.mod_small. 
+            all: idtac + rewrite ?Z.pow_add_r; ZnWords.
+        }
+    }
+    { repeat straightline. ZnWords. }
+Qed.
```
miriampolzer: style nit: Newline at end of file.

src/Bedrock/P256/modinv/br_ctz.v
```diff
@@ -0,0 +1,108 @@
...
+(** * Specification Proof *)
+Lemma br_ctz_ok : program_logic_goal_for_function! br_ctz.
+Proof.
+    repeat straightline.
+    refine ((Loops.tailrec
+    (* types of ghost variables*) 
+        (HList.polymorphic_list.nil)
+    (* program variables *) (["value";"count";"tmp"] : list String.string))
+    (fun v t m value_ count tmp => PrimitivePair.pair.mk (* precondition *)
+      (v = word.unsigned count /\ word.unsigned tmp = (word.unsigned value * 2^count) mod 2^64 /\ lctz 64 (word.unsigned tmp) = lctz 64 (word.unsigned value) + (word.unsigned count) /\ value = value_)
```
miriampolzer: style nit: fiat crypto doesn't have any guideline, but I try to avoid lines above 100 chars, and make a hard cut at 120.

src/Bedrock/P256/modinv/br_ctz.v
```diff
@@ -0,0 +1,108 @@
...
+From coqutil Require Import Tactics.Tactics WithBaseName Z.CountTrailingZeros.
+Local Open Scope string_scope. Local Open Scope Z_scope.
+
+Local Notation eval := (fold_right (fun (a : word) (s : Z) => a + 2^64*s) 0).
+Local Notation array := (array scalar (word.of_Z 8)).
+
+From Coq Require Import ZArith Lia.
+
+Lemma lctz_word_slu (w : word) : (w > 0) -> lctz 64 (word.slu w (word.of_Z 1)) = 1 + lctz 64 w.
+Proof.
```
miriampolzer, line 14: This one can be simplified a bit to reduce explicit replace statements and ZnWords_pre at the start, e.g. 
```
  intros H.
  rewrite Properties.word.unsigned_slu_shamtZ by lia.
  cbv [word.wrap].
  rewrite Z.shiftl_mul_pow2, Z.pow_1_r by lia.
  match goal with |- context [lctz _ ?x] => destr (x =? 0) end.
  { pose proof (Properties.word.unsigned_range w).
    apply Z.mod_divide in E; [|lia].
    destruct E.
    replace (word.unsigned w) with (2^63) by lia.
    reflexivity. }
  { rewrite lctz_eq_mod_pow2 with (n := 64%nat) by lia.
    rewrite Z.mul_comm, lctz_double by lia.
    reflexivity. }
```
This is just one option, feel free to alter it yourself as you prefer to write it.

miriampolzer (approved):
Thanks! Looks good overall, just some detail comments. Let me know when you rebased and updated so that I can trigger the CI and merge.

src/Bedrock/P256/modinv/br_ctz.v
klinashka: Yes, I believe it could be `>=`, since `lctz 64 0 = 64`, which is what the function also returns.

src/Bedrock/P256/modinv/br_ctz.v
klinashka: Thank you for the suggestion! I will take a look at it.

src/Bedrock/P256/modinv/br_ctz.v
klinashka: > Yes, I believe it could be `>=`, since `lctz 64 0 = 64`, which is what the function also returns.

Should I change it?

src/Bedrock/P256/modinv/br_ctz.v
klinashka: Thanks for the suggestion, I will change this.

src/Bedrock/P256/modinv/br_ctz.v
miriampolzer: I would say yes. It certainly doesn't hurt.

klinashka: I have removed the assertion `value > 0`. Since `value` is a word now, the function doesn't have a precondition anymore.

miriampolzer (approved):

PR #2385, opened 2026-08-21
Verify beeu normalize

klinashka:
This PR proves the specification of the function `beeu_normalize`, with the following specifications: 

Precondition: two pointers `p_y` and `p_mod`, which point to 320-bit unsigned integers `y` and `MOD`, where `MOD > 0`. The two pointers do not alias. 

Postcondition: `p_y` will point to a 320-bit unsigned integer `y'`, s.t. `y' = y mod MOD`. 

Additionally, the PR modifies the signature and specification of `u320_sub` such that the result of the difference is stored in the input. The only usage of `u320_sub` occurs in this function, which does not need the more general version, where the difference is stored in an arbitrary pointer.

### DIFF
--- a/src/Bedrock/P256/modinv/beeu_normalize.v
+++ b/src/Bedrock/P256/modinv/beeu_normalize.v
@@ -7,7 +7,6 @@ Require Import Util.ZRange.
 Require Import P256.modinv.u320_sub.
 From Coq Require Import Zmod ZArith.
 Local Open Scope string_scope. Local Open Scope Z_scope.
-Require Import Setoid.
 
 Local Notation eval := (fold_right (fun (a : word) (s : Z) => a + 2^64*s) 0).
 Local Notation array := (array scalar (word.of_Z 8)).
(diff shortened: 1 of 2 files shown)


src/Bedrock/P256/modinv/beeu_normalize.v
```diff
@@ -0,0 +1,205 @@
+From Coq Require Import BinInt String List InitialRing.
+From bedrock2 Require Import BasicC64Semantics WeakestPrecondition ProgramLogic NotationsCustomEntry ZnWords ArrayCasts.
+Import ListNotations ProgramLogic.Coercions SeparationLogic Array Scalars.
+From coqutil Require Import Tactics.Tactics WithBaseName Map.SeparationLogic.
+Require Import bedrock2Examples.full_sub.
+Require Import Util.ZRange.
+Require Import P256.modinv.u320_sub.
+From Coq Require Import Zmod ZArith.
+Local Open Scope string_scope. Local Open Scope Z_scope.
+Require Import Setoid.
```
lukaszobernig: Are you using `setoid_rewrite`?

src/Bedrock/P256/modinv/beeu_normalize.v
```diff
@@ -0,0 +1,205 @@
...
+            m =* array p_y y ⋆ array p_m MOD ⋆ R /\ 
+            length y = 5%nat /\ length MOD = 5%nat /\ eval MOD > 0;
+        ensures T M := exists (r : list word), 
+            M =* array p_y r ⋆ array p_m MOD ⋆ R /\ length r = 5%nat /\ 
+                Zmod.of_Z (eval MOD) (eval y) = Zmod.of_Z (eval MOD) (eval r) 
+                /\ 0%nat <= (eval r) < (eval MOD)
+    }
+.
+
+#[export] Instance spec_of_u320_set : spec_of "u320_set" := fnspec! "u320_set" (p_x p_y : word) / (x y : list word) R,
```
lukaszobernig: Do you think that you will need this helper somewhere else?

lukaszobernig (commented):
Thank you for submitting this! Generally it looks good to me, I just left two small questions.

src/Bedrock/P256/modinv/beeu_normalize.v
klinashka: I don't think I will use this anywhere else. I can mark it as local.

src/Bedrock/P256/modinv/beeu_normalize.v
klinashka: No, I used `rewrite_strat` in a previous version. I will remove the import since it is not required.

PR #2434, opened 2026-09-12
Consolidate F onto stdlib Zmod; coqutil Word.Naive onto stdlib bits

JasonGross:
`F` is now stdlib `Zmod` (rocq-stdlib 9.1, which master already requires), and coqutil's `Word.Naive` is stdlib `bits` via the rupicola bump (mit-plv/rupicola#225).

What changes for users of `F`:
- `F : Z -> Type` instead of `positive -> Type`; the `Z.pos` coercion keeps `F p` working for `p : positive`. Lemmas that were stated over a `positive` modulus now take `m : Z` with `0 < m` (or `prime m`) where the proof needs it.
- `F.pow` takes a `Z` exponent (it is `Zmod.pow`), so the stdlib power lemmas with their `0 <= n` premises apply. `Fq_inv_fermat` is `x ^ (q - 2)`, `euler_criterion` uses `q / 2`.
- `F.of_Z`, `F.to_Z`, `F.add`, ..., `F.inv`, `F.div` are notations for the `Zmod` operations, so every stdlib `Zmod` lemma applies directly; the `F_scope` notations are unchanged. `Arithmetic/ModularArithmeticPre.v` is gone.
- `F`'s own restatements of stdlib lemmas are gone (`to_Z_of_Z`, `of_Z_to_Z`, `eq_to_Z_iff`, `to_Z_add`, `of_Z_mul`, `to_Z_range`, `pow_succ_r`, `pow_add_r`, `pow_pow_l`, `to_Z_pow`, `inv_0`, `to_Z_1`, ... 33 in total); use sites name the stdlib lemma (`Zmod.unsigned_of_Z`, `Zmod.of_Z_unsigned`, `Zmod.unsigned_inj_iff`, `Zmod.unsigned_add`, `Zmod.unsigned_pos_bound`, `Zmod.pow_succ_nonneg_r`, `Zmod.pow_add_r_nonneg`, `Zmod.pow_mul_r_nonneg`, ...). What stays in `F` are lemmas whose statement differs from stdlib's (listed in the ModularArithmeticTheorems commit message).
- Removed names: `F.F` (only the `F` notation remains), `F.inv_with_spec`, `F.pow_with_spec`, `F.pow_spec`, `F.inv_spec`, the `F.unwrap_F` tactic, the (already commented-out) `F.Iso` module. Changed shapes: `F.char_gt` is `Ring.char_ge (Z.to_pos m)`; `F.power_theory` injects exponents through `Z.of_N` and `F.is_pow_constant` returns `N` constants, so custom `Add Ring`/`Add Field` declarations with `power_tac` must use `F.is_pow_constant`; `F.quotrem` is floor-based (`Z.div`/`Z.modulo`); `F.to_nat_of_nat`, `F.of_nat_to_nat`, `F.of_nat_mod` take `0 < m`. `Curve25519.p`, `l`, `order`, `P256.p256` are `Z`; `Spec.Test.X25519.monty` takes a `Z` scalar.
- Two global effects on stdlib's type, both restoring what the old `F` had: `Global Coercion Zmod.unsigned : Zmod >-> Z` (stdlib declares none) and `Bind Scope F_scope with Zmod` (`F_scope` maps `+ - * ^ 0 1` to the same constants as stdlib's `Zmod_scope` and adds `/` for `Zmod.mdiv`).

The rest of the diff is the fallout: `COperationSpecifications`, the curve specs, and the bedrock backend (`src/Bedrock/**`, 22 files). In the bedrock backend a few proofs that let tactics search or compute through the transparent field or word representation were restructured to stay at the interface (`word.unsigned_*` lemmas, goal-first sep-logic matching, `rewrite` instead of `replace`), because unification through `Zmod`'s definitions is slow (in `Secp256k1/Addchain.v` the addition-chain equation is rewritten with the stdlib pow lemmas after generalizing the base term: with the concrete `feval x` as base, `rewrite` compares mismatching closed subterms up to conversion and computes modular powers before failing, over 100 s; with the base abstracted the same rewrites take 0.37 s, numbers in that commit's message); per-file compile times of the slowest files on this machine are within about 1.1-1.25x of master's times on the GitHub runners, so not directly comparable but no sign of a regression.

The macOS and Windows CI jobs now pin Rocq 9.2.0 (matching #2432): the `coq 9.1.0` opam metapackage requires `coq-stdlib = 9.0.0`, which has no `Zmod`, so those opam-based jobs cannot test 9.1 with `Zmod` (the docker-based upstream repos install rocq-stdlib 9.1 in their 9.1 image instead).

Verified locally (Rocq 9.2 + rocq-stdlib 9.1.0, the supported minimum): `make coq` and `make all-except-generated-and-js-of-ocaml` pass with generated outputs unchanged; the touched core files also compile against stdlib master (no removed `to_Z_*` aliases are used). No new `Admitted`.

Follow-ups: #2435 flips `Bedrock/Specs/Field.v`'s `M_pos : positive` to a `Z` field; mit-plv/kami#51 moves Kami's own word type.

Depends on mit-plv/rupicola#225 (which depends on mit-plv/bedrock2#561 and mit-plv/coqutil#186); CI cannot fetch the pinned rupicola commit until it merges, so this stays a draft until then.

Written by Claude Fable 5.1 from Andres Erbsen's feasibility sketch (https://gist.github.com/andres-erbsen/17b583447d29dc085714bb8e9c7cdc89).


JasonGross:
Stacked follow-up: #2435 (FieldParameters modulus as a Z field, drop M_pos).

---

src/Bedrock/Secp256k1/Addchain.v
```diff
@@ -288,6 +297,28 @@ Section WithParameters.
     eexists; ssplit; eassumption.
   Qed.
 
+  (* [chain = x ^ e] for an addition chain [chain] built from [x] with [*] and
```
andres-erbsen: Could we use rewrite_strat with Hint Opaque?

src/Arithmetic/ModularArithmeticTheorems.v
```diff
@@ -14,127 +13,98 @@ Require Import Crypto.Util.ZUtil.Tactics.PullPush.Modulo.
...
+    Proof using Type. symmetry; apply Zmod.unsigned_inj_iff. Qed.
 
     Lemma eq_of_Z_iff : forall x y : Z, x mod m = y mod m <-> F.of_Z m x = F.of_Z m y.
-    Proof using Type. split; unwrap_F; congruence. Qed.
-
+    Proof using Type. intros; symmetry; apply Zmod.of_Z_inj. Qed.
 
     Lemma to_Z_of_Z : forall z, F.to_Z (F.of_Z m z) = z mod m.
-    Proof using Type. unwrap_F; trivial. Qed.
+    Proof using Type. apply Zmod.unsigned_of_Z. Qed.
```
andres-erbsen: for lemmas that stdlib already has a version of, we should not keep our duplicate version. instead uses of our lemma should be replaced with uses of the stdlib lemma.

JasonGross: Done repo-wide: every F lemma that restates a stdlib Zmod lemma is gone (27 in ModularArithmeticTheorems, inv_0 and to_Z_1 in PrimeFieldTheorems, three FLia helpers, zero_iff_to_Z in Bedrock/P256/Specs), and the use sites across src/ name the stdlib lemma (names checked against stdlib master too). What stays are lemmas whose statement differs, listed in the commit message of the ModularArithmeticTheorems commit; one borderline case is F.of_Z_pow, which is stdlib's of_Z_pow composed with of_Z_mod and unused, kept only because of the mod placement. Happy to drop it as well.

---

src/Bedrock/Secp256k1/Addchain.v
JasonGross: It works once the field operations are opaque for the rewrite unifier (`#[local] Hint Constants Opaque : rewrite`; the `rewrite` hint database only makes variables opaque by default, so without it rewrite_strat unfolds Zmod.pow/mul and does not finish), but it is much slower than the structural tactic here: top-down 51 s and bottom-up over 100 s in this proof against 1.4 s (standalone on the extracted chain: 6 s vs 0.7 s). So pow_chain stays; its history-narrating comment is cut to one line and the measurements are in the commit message.

---

JasonGross: Follow-up after profiling with `Set Typeclasses Debug` and `Set Keyed Unification`: the typeclass side is not the cost (about five trivial queries per step, no backtracking; adding `Proper` instances up front or `Typeclasses Opaque` changes nothing). Two things were slow. With transparent constants, unification unfolds `Zmod.pow`, whose cost grows with the exponent value, so it never finishes on this chain. With the concrete `feval x` as the base, `rewrite` compares mismatching closed subterms up to conversion and computes big modular powers before each failed match; that is why plain `rewrite` took over 100 s here. Keyed unification does not change either (it keeps full conversion on arguments and does not reach rewrite_strat).

With the base generalized to a variable, plain `rewrite` with the three stdlib pow lemmas takes 0.37 s end to end, against 1.3 s for the structural tactic and 47 s for the best rewrite_strat setup (opaque constants, generalized base, cleared context gets it to 6.5 s). So the custom lemmas and tactic are gone: the step is now `generalize (feval x)` followed by `repeat rewrite <-Zmod.pow_mul_r_nonneg, <-Zmod.pow_succ_nonneg_r, <-Zmod.pow_add_r_nonneg`. Numbers are in the commit message.

---

src/Bedrock/End2End/X25519/GarageDoor.v
```diff
@@ -36,11 +36,17 @@ Import Tuple LittleEndianList.
 Definition garageowner : list byte :=
   [x7b; x06; x18; x0c; x54; x0c; xca; x9f; xa3; x16; x0b; x2f; x2b; x69; x89; x63; x77; x4c; xc1; xef; xdc; x04; x91; x46; x76; x8b; xb2; xbf; x43; x0e; x34; x34].
 
+(* [Decidable.vm_decide] below refers to [Rewriter.Util.Decidable] (loaded
+   through [Crypto.Language.API]), a copy of [Crypto.Util.Decidable] without
+   an instance for equality on [Zmod]. *)
+#[local] Instance Rewriter_Decidable_eq_F {m} : Rewriter.Util.Decidable.DecidableRel (@eq (F m))
+  := fun x y => F.eq_dec x y.
```
andres-erbsen: Instead change the call to `Rewriter.Util.Decidable.vm_decide` to instead call Crypto.Util.Decidable.vm_decide`.

src/Bedrock/P256/Coord.v
```diff
@@ -258,22 +265,22 @@ Proof.
...
 
   cbv [Semantics.interp_op1] in *.
   assert (x9 = word.of_Z 0 /\ F.to_Z y <= F.to_Z x
         \/x9 = word.of_Z 1 /\ F.to_Z x < F.to_Z y) as [ [-> ?]|[-> ?]] by
-      (rewrite <-!word.unsigned_inj_iff; cbv [p256] in *; ZnWords.ZnWords).
-  { rewrite ?Z.add_0_r in *; cbv [p256] in *. ZnWords.ZnWords. }
-  rewrite <-(Z.mod_add _ 1), Z.mod_small by (cbv [p256] in *; ZnWords.ZnWords).
+      (rewrite <-!word.unsigned_inj_iff; unfold_p256; ZnWords.ZnWords).
+  { rewrite ?Z.add_0_r in *; unfold_p256. rewrite Z.mod_small by lia. ZnWords.ZnWords. }
+  rewrite <-(Z.mod_add _ 1), Z.mod_small by (unfold_p256; ZnWords.ZnWords).
```
andres-erbsen: Do not replace calls to cbv with calls to unfold -- cbv is more predictable.

src/Bedrock/Field/Synthesis/New/WordByWordMontgomery.v
```diff
@@ -826,7 +841,7 @@ Require Import bedrock2.ProgramLogic.
   Instance field_parameters : FieldParameters.
   Proof using Type.
     let M := (eval vm_compute in (Z.to_pos (m))) in
-    (* Curve25519 "A" parameter (see section 4.1 of RFC 7748) *)
+    (* dummy 'A' parameter since we don't care about scmula24 here *)
```
andres-erbsen: Please do not make changes unrelated to the Word/Zmod/bits refactoring.

src/Arithmetic/ModularArithmeticTheorems.v
```diff
@@ -329,60 +252,9 @@ Module F.
...
-
-    Lemma pow_2_r (x:F m) : x^2 = x*x.
-    Proof using Type.
-      change 1%N with (N.succ (N.succ 0)); repeat rewrite ?pow_succ_r, ?pow_0_r; ring.
-    Qed.
-
     Lemma pow_3_r (x:F m) : x^3 = x*x*x.
     Proof using Type.
-      change 1%N with (N.succ (N.succ (N.succ 0))); repeat rewrite ?pow_succ_r, ?pow_0_r; ring.
+      change 3%Z with (Z.succ 2); rewrite Zmod.pow_succ_nonneg_r, Zmod.pow_2_r by lia; ring.
```
andres-erbsen: With the ring power changes above, is `ring` sufficient to prove this lemma?

src/Bedrock/End2End/X25519/GarageDoor.v
JasonGross: Done: the local instance is gone and both `vm_decide` calls in this file (this one and the pre-existing one at line 378, which also resolved to the Rewriter copy) now call `Crypto.Util.Decidable.vm_decide`. No other place in the diff referenced the Rewriter copy.

---

src/Bedrock/P256/Coord.v
JasonGross: Reverted across the PR: `p256` is now defined as `Eval cbv in (2^256 - 2^224 + 2^192 + 2^96 - 1)`, so `cbv [p256] in *` produces the literal modulus as it did with the `positive` definition, and Coord.v and Coord32.v are back to master's `cbv [p256] in *` at all 14 sites (the `unfold_p256` tactic is gone). The reason for the earlier replacement was that `cbv [p256]` on the unevaluated `Z` expression left arithmetic that lia's certificate check did not finish on. No other cbv-to-unfold replacement was in the diff.

---

src/Bedrock/Field/Synthesis/New/WordByWordMontgomery.v
JasonGross: Reverted: this commented-out test block is back to master's text, as are a removed blank line in Spec/ModularArithmetic.v and an added example in FLia.v. The rest of the diff was checked hunk by hunk; what remains is either needed to compile or sits inside a rewritten proof.

---

src/Arithmetic/ModularArithmeticTheorems.v
JasonGross: Yes: with the power theory registered, `ring` alone proves it (also `x^2 = x*x`, `(x*y)^2 = x^2*y^2`, `(x^2)^3 = x^6`), so the proof is now just `ring`. The lemma stays because the P256 Jacobian proofs and the AdditionChains rewrite database use it by name.

---

andres-erbsen:
Please investigate this failure and propose a workaround: https://github.com/mit-plv/fiat-crypto/actions/runs/34926821354/job/104246511064

andres-erbsen:
As a follow-up, let's get rid of `Definition F` altogether -- F.foo should be replaced with Zmod.foo.

JasonGross:
Root cause: an upstream Rocq bug in the new `rocq makefile --rocq-package` install scheme, Windows only, unrelated to this PR or to the merge commit. On Windows-native OCaml, `rocq makefile -destination-of` joins paths with a backslash (`Stdlib\\Arith`), and the generated install rule compares that against the shell glob `Stdlib/*`; nothing matches, so `make install` for rocq-stdlib.dev prints `SKIP ... since it is outside rocqpath Stdlib` for all 1746 files and installs only META (the `--legacy-support` copy is skipped by the same `continue`). Every dependent then fails with "library ZArith ... has not been found in the loadpath". Introduced by rocq-prover/rocq#21564 plus the stdlib's move to that scheme (coq/stdlib#286); this cell has failed the same way on every run since 2026-07-13 (last green 2026-06-15). Linux/macOS use `/` and pass; the 9.2.0 Windows cell passes because stdlib 9.2 still uses the user-contrib install. The mechanism reproduces on Linux by making `-destination-of` emit backslashes. Filed as rocq-prover/rocq#22478 with a one-line fix suggestion.

Workaround: #2440 runs the `opam pin add coq dev` step with `ROCQPACKAGEOPTIONS=''` for the Windows dev cell only, so the stdlib Makefile (`?=`) falls back to the classic user-contrib install. Not verified on a Windows runner here, so that PR's own run is the test; the fallback would be `continue-on-error` for the cell until upstream fixes it.

---

JasonGross:
Done as #2441 (on master): `F` is gone, everything names `Zmod.foo` directly in `Zmod_scope` (plus `/` for `Zmod.mdiv`), and the fiat-crypto additions sit in `Module Zmod.` extension blocks with only new names (checked with `Locate`), following the quartz#2 pattern. The only definitional change is that the P256 `point_iszero` spec now uses the stdlib `Zmod.eqb` instead of the local `if eq_dec ...` version; the extracted synthesis code is byte-identical.

---

PR #2435, opened 2026-09-12
Bedrock: FieldParameters modulus is a Z field (M), drop M_pos

JasonGross:
Stacked on #2434 (base branch `fable/zmod`); the diff shown here is only the follow-up.

With `F : Z -> Type`, the bedrock backend's `FieldParameters` class no longer needs a `positive` modulus. `M : Z` becomes a real field and `M_pos` is gone: `a24 : F M`, `fe_copy := @id (F M)`. Fallout in 16 other files under `src/Bedrock`: `F M_pos` is `F M`, the section hypotheses `F_M_pos : Z.pos M_pos = m` are `F_M : M = m`, `positive` facts about the modulus become the same facts in `Z` (for example `28 <= M`), and the field instantiations pass the modulus directly instead of `Z.to_pos`. Positivity comes from `M_prime` where a proof needs it; the only lemma that gained a hypothesis is the unused `M_nonzero` in `Specs/Field.v`, which now takes `FieldParameters_ok`. No theorem statement changes meaning. One consequence of the flip: the class itself no longer rules out a zero or negative modulus; every theorem that needs positivity states it (via `FieldParameters_ok`'s `M_prime`), so nothing is silently weakened.

Verified locally on Rocq 9.2 + rocq-stdlib 9.1.0: `make coq` (the bedrock backend recompiles in about 23 minutes at -j96) and `make all-except-generated-and-js-of-ocaml` pass, generated outputs unchanged, no new `Admitted`.

Written by Claude Fable 5.1.


andres-erbsen:
Please create yet another PR on top of this one to remove F altogether in favor of stdlib Zmod.

PR #2440, opened 2026-09-15
CI (opam, Windows dev): install rocq-stdlib with the classic user-contrib scheme

JasonGross:
Works around rocq-prover/rocq#22478 in the "CI (Coq, opam)" workflow: on Windows-native OCaml, rocq master's `rocq makefile --rocq-package` install scheme installs no .vo files (its `-destination-of` prints backslash-joined paths that the install rule never matches), so rocq-stdlib.dev installs only META and every dependent fails with "library ZArith ... has not been found in the loadpath". The Windows dev cell has failed this way on every run since 2026-07-13; Linux/macOS and the 9.2.0 Windows cell are unaffected.

For the Windows dev cell only, the `opam pin add coq dev` step now runs with `ROCQPACKAGEOPTIONS=''`, which makes the stdlib Makefile fall back to the classic user-contrib install that Rocq 9.2 and earlier use everywhere. Investigated for @andres-erbsen's question on #2434; the mechanism was reproduced on Linux by emulating the backslash output.

This workflow only runs on pushes to master, so the second commit adds a `workflow_dispatch` trigger; the manual run on this branch is the test: https://github.com/mit-plv/fiat-crypto/actions/runs/35016400443

Written by Claude Fable 5.1.


JasonGross:
The manual run of the opam workflow on this branch is green in all six cells, including the Windows dev cell that has failed on master since July: https://github.com/mit-plv/fiat-crypto/actions/runs/35016400443 (Windows dev: 102 min, `opam install coq-fiat-crypto --with-test` and the `fiat_crypto` checks pass). So the workaround does what the upstream issue predicts; it can be dropped once rocq-prover/rocq#22478 is fixed.

---

PR #2444, opened 2026-09-15
Verify modinv.

klinashka:
This PR contains the verification of `beeu_modinv` which computes the modular inverse of a 256-bit unsigned integer modulo an odd 256-bit unsigned integer, which is greater than 2. 

Input: 
`p_a` : 256-bit unsigned integer. The number whose modular inverse is computed. 
`p_m`: 256-bit unsigned integer. The modulus wrt. which the modular inverse is computed. The number must be odd and greater than 1.  
`inv_m` : 64-bit unsigned integer which is equal to `(- p_m^-1) mod p_m`. 

Output: 
`c`: 0 if `p_a` and `p_m` are not coprime, indicating the inverse does not exist. Otherwise, 1. 
If `c` is 1, then the value of the modular inverse of `a` modulo `m` will be stored in `p_out`.


JasonGross:
Branch with this PR's changes on top of current master (after #2445): https://github.com/mit-plv/fiat-crypto/tree/fable/bits-modinv (tip 741b62d0c = master a5ee22098 + the two commits here, cherry-picked verbatim, + one commit porting `beeu_modinv.v` to stdlib `bits`, the same way the other `P256/modinv` files were ported).

Admits are unchanged (the six in the x,y-range block and `helper_loop_ok`); `beeu_modinv_ok` otherwise depends only on `helper_loop_ok`. Two things beyond the mechanical port were needed to make the file compile at all: `beeu_normalize`'s spec now has the `T = t` conjunct that the proof of `beeu_modinv_ok` relies on (the trace after that call was unconstrained before), and three hypothesis references were made robust to the extra `0 <= c < 2` conjunct that #2424 added to `u320_muladd`'s postcondition. The deleted coqutil helpers (`mod2_cases`, `Z_mod_mult'`) are file-local lemmas. Happy to open it as a PR or push it here, whichever you prefer.

---

src/Bedrock/P256/modinv/beeu_modinv.v
```diff
@@ -0,0 +1,1877 @@
...
+Local Open Scope string_scope. Local Open Scope Z_scope.
+
+Local Lemma mod2_cases (n : Z) : n mod 2 = 0 \/ n mod 2 = 1.
+Proof. pose proof (Z.mod_pos_bound n 2); lia. Qed.
+
+Local Lemma Z_mod_mult' (a b : Z) : (a * b) mod a = 0.
+Proof. rewrite Z.mul_comm; apply Z_mod_mult. Qed.
+
+Local Notation eval := (fold_right (fun (a : word) (s : Z) => a + 2^64*s) 0).
+Local Notation eval_bool := (fold_right (fun (a : word) (s : Z) => Z.lor a (Z.shiftl s 64)) 0).
```
lukaszobernig: This seems unused?

src/Bedrock/P256/modinv/beeu_modinv.v
```diff
@@ -0,0 +1,1877 @@
...
+        cbv [c] in *. rewrite Z.gcd_comm, <- H62.
+        destruct (eval x2 =? 1) eqn: Hx2; eauto.
+        (* Contradiction *)
+        rewrite Z.eqb_eq, Hx2, bits.unsigned_1 in * by lia.
+        eapply Zmod.eqb_eq in H65.
+        rewrite H65 in *. discriminate.
+    }
+Qed.
+
+From bedrock2Examples Require Import full_mul.
```
lukaszobernig: `full_mul` is already imported (see line 4).

src/Bedrock/P256/modinv/beeu_modinv.v
```diff
@@ -0,0 +1,1877 @@
...
+        apply u256_shr_correct ||
+        apply helper_subtract_ok ||
+        apply u320_shr_correct ||
+        apply u256_sub_ok ||
+        apply u256_dec_ok ||
+        apply u320_add_correct ||
+        apply br_ctz_ok ||
+        apply beeu_normalize_ok ||
+        apply u320_sub_correct ||
+        apply u320_set_ok ||
```
lukaszobernig: You already call that in line 1855.

src/Bedrock/P256/modinv/beeu_modinv.v
```diff
@@ -0,0 +1,1877 @@
...
+                all: try solve [
+                    eapply mod_pow2_inv with (n := Z.min (lctz 64 (eval x2 - eval x1)) 63); try lia;
+                    rewrite <- Z.mul_assoc, (Z.mul_comm (eval b)), Z.mul_assoc;
+                    rewrite ?H47; Z.push_mod_step; Z.push_mod_step; rewrite ?H32, ?H33, ?H49; Z.push_pull_mod; f_equal; lia
+                ].
+                all: try solve [
+                    eapply mod_pow2_inv with (n := Z.min (lctz 64 (eval x2 - eval x1)) 63); try lia;
+                    rewrite <- Z.mul_assoc, (Z.mul_comm (eval b)), Z.mul_assoc;
+                    rewrite ?H47; Z.push_mod_step; Z.push_mod_step; rewrite ?H32, ?H33, ?H49; Z.push_pull_mod; f_equal; lia
+                ].
```
lukaszobernig: These seem duplicated?

src/Bedrock/P256/modinv/beeu_modinv.v
```diff
@@ -0,0 +1,1877 @@
...
+    unpack! d0, b = br_full_sub(load(p_x), load(p_y), b);
+    unpack! d1, b = br_full_sub(load(p_x + $8), load(p_y + $8), b);
+    unpack! d2, b = br_full_sub(load(p_x + $8 + $8), load(p_y + $8 + $8), b);
+    unpack! d3, b = br_full_sub(load(p_x + $8 + $8 + $8), load(p_y + $8 + $8 + $8), b);
+
+    store(p_out, d0);
+    store(p_out + $8, d1);
+    store(p_out + $8 + $8, d2);
+    store(p_out + $8 + $8 + $8, d3)
+}.
```
lukaszobernig: If I understand this correctly, the whole function is an exact duplicate of `u256_sub`. Instead of doing that, can't you prove an alternative spec `spec_of_u256_sub'` of `u256_sub` with the longer `y`? Or destruct the input as `array y [y0; y1; y2; y3] ⋆ scalar (y + $32) y4`?

src/Bedrock/P256/modinv/beeu_modinv.v
```diff
@@ -0,0 +1,1877 @@
...
+            length out = 4%nat /\ length a = 4%nat /\ length MOD = 4%nat /\
+            eval MOD > 1 /\ eval MOD mod 2 = 1 /\
+            0 <= eval a < eval MOD /\
+            inv_m * (eval MOD) mod (2^64) = 2^64 - 1;
+        ensures T M := T = t /\ exists OUT,
+            M =* array p_out OUT ⋆ array p_a a ⋆ array p_m MOD ⋆ R /\
+            length OUT = 4%nat /\
+                if (Z.gcd (eval a) (eval MOD) =? 1)
+                then
+                    c = bits.of_Z _ 1 /\ ((eval OUT) * (eval a)) mod (eval MOD) = 1
```
lukaszobernig: I think it would make sense to add `0 <= eval OUT < eval MOD` to the list of postconditions in the success (c = 1) case, to explicitly express that it has been reduced. What do you think?

src/Bedrock/P256/modinv/beeu_modinv.v
klinashka: Good catch! I will remove it.

src/Bedrock/P256/modinv/beeu_modinv.v
klinashka: I agree that this would strengthen the postcondition. Also, I don't believe this is a difficult condition to add to the proof. I will make this change.

src/Bedrock/P256/modinv/beeu_modinv.v
klinashka: I do like the option where `u256_sub` is called with two different specifications. I will try to make this change.

src/Bedrock/P256/modinv/beeu_modinv.v
klinashka: `eval_bool` is indeed unused. I had it defined in `u256_shr` and `u320_shr` as it simplified the respective proofs. I will remove it from this file.

